PR #7, opened 2026-02-04
App Builder - Unit and Integration tests

eshurakov:
This pull request introduces improvements to the Cloudflare App Builder project focusing on enhanced test coverage.

**Testing & CI Improvements**
- Added integration tests for the App Builder. 

**API Schema Additions**
- Introduced new API schemas for the `/tree` and `/blob` endpoints, defining types and validation for tree and blob responses.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@ jobs:
     outputs:
       cloud_agent: ${{ steps.filter.outputs.cloud_agent }}
       webhook_agent: ${{ steps.filter.outputs.webhook_agent }}
+      app_builder: ${{ steps.filter.outputs.app_builder }}
     steps:
       - uses: actions/checkout@v4
         with:
@@ -30,6 +31,8 @@ jobs:
               - 'cloud-agent/**'
             webhook_agent:
               - 'cloudflare-webhook-agent-ingest/**'
+            app_builder:
+              - 'cloudflare-app-builder/**'
 
   test:
     runs-on: ubuntu-24.04-8core
@@ -250,6 +253,37 @@ jobs:
       - name: Run webhook-agent-ingest tests
         run: pnpm --filter cloudflare-webhook-agent-ingest test
 
+  app-builder:
+    needs: changes
+    if: needs.changes.outputs.app_builder == 'true'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          lfs: true
+          ref: ${{ github.head_ref }}
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v2
+        with:
+          version: latest
+          run_install: false
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Typecheck (app-builder)
+        run: pnpm --filter app-builder typecheck
+
+      - name: Run app-builder tests
+        run: pnpm --filter app-builder test
+
   cloudflare:
     strategy:
       fail-fast: false
@@ -259,8 +293,6 @@ jobs:
             filter: kilo-deploy-builder
           - name: deploy-dispatcher
             filter: deploy-dispatcher
-          - name: app-builder
-            filter: app-builder
           - name: code-review
             filter: kilo-code-review-worker
     name: cloudflare-${{ matrix.name }}
--- a/cloudflare-app-builder/package.json
+++ b/cloudflare-app-builder/package.json
@@ -8,7 +8,9 @@
     "deploy": "wrangler deploy",
     "cf-typegen": "wrangler types --env-interface CloudflareEnv worker-configuration.d.ts",
     "typecheck": "tsc --noEmit",
-    "test:git": "tsx src/_integration_tests/test-git-integration.ts"
+    "test": "vitest run",
+    "test:watch": "vitest",
+    "test:git:all": "./src/_integration_tests/run-all-tests.sh"
   },
   "dependencies": {
     "@ashishkumar472/cf-git": "1.0.5",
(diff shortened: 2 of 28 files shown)


cloudflare-app-builder/src/git-repository-do.ts
```diff
@@ -263,4 +263,144 @@ export class GitRepositoryDO extends DurableObject<Env> {
...
     });
   }
+
+  /**
+   * Get directory tree contents at a specific path (RPC method)
+   * @param ref - Branch name (e.g., "main", "HEAD") or commit SHA
+   * @param path - Optional path to a subdirectory (defaults to root)
+   * @returns Array of tree entries with name, type, oid, and mode
+   */
+  async getTree(
```
marius-kilocode, line 278: This code is duplicated, can we reuse the git implementation from here?

cloudflare-app-builder/src/git-repository-do.ts
```diff
@@ -263,4 +263,144 @@ export class GitRepositoryDO extends DurableObject<Env> {
...
+    });
+  }
+
+  /**
+   * Get blob (file) contents at a specific path (RPC method)
+   * @param ref - Branch name (e.g., "main", "HEAD") or commit SHA
+   * @param path - Path to the file
+   * @returns File content (utf-8 or base64 encoded), encoding type, size, and blob SHA
+   */
+  async getBlob(
```
marius-kilocode, line 306: Also duplicated and the return types are mismatching?

cloudflare-app-builder/src/handlers/files.ts
```diff
@@ -0,0 +1,133 @@
...
+      path,
+      error: error instanceof Error ? error.message : String(error),
+    });
+    return new Response(
+      JSON.stringify({
+        error: 'not_found',
+        message: error instanceof Error ? error.message : 'Path not found',
+      }),
+      {
+        status: 404,
```
marius-kilocode, line 74: This will return 404 also for 'Error(`Path not found: ${normalizedPath}`)' and 'Error(`Path is not a directory: ${normalizedPath}`)'. It depends on the quality we want for this API, but an optional improvement would be to handle not found cases explicitly by cascading the error from `getTree`

cloudflare-app-builder/src/handlers/files.ts
```diff
@@ -0,0 +1,133 @@
...
+      path,
+      error: error instanceof Error ? error.message : String(error),
+    });
+    return new Response(
+      JSON.stringify({
+        error: 'not_found',
+        message: error instanceof Error ? error.message : 'File not found',
+      }),
+      {
+        status: 404,
```
marius-kilocode, line 133: See comment above. Same thing here.

cloudflare-app-builder/package.json
```diff
@@ -18,13 +20,16 @@
...
   },
   "devDependencies": {
-    "@types/jsonwebtoken": "^9.0.9",
     "@cloudflare/containers": "^0.0.30",
     "@cloudflare/sandbox": "0.6.7",
     "@cloudflare/workers-types": "^4.20251014.0",
+    "@types/better-sqlite3": "^7.6.13",
+    "@types/jsonwebtoken": "^9.0.9",
     "@types/node": "^22.10.1",
+    "better-sqlite3": "^12.6.0",
```
marius-kilocode, line 29: We can already use: `12.6.2`, same applies for the other new deps and their patch releases.

cloudflare-app-builder/src/git/test-utils.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Close all databases opened during tests
+ */
+export function closeAllDatabases(): void {
+  for (const db of databases) {
+    try {
+      db.close();
+    } catch {
+      // Ignore - might already be closed
```
marius-kilocode, line 68: Should we not log this?

cloudflare-app-builder/src/git/test-utils.ts
```diff
@@ -0,0 +1,72 @@
+/**
+ * Test utilities for GitVersionControl unit tests
```
marius-kilocode, line 2: The code does't appears really domain specific to git. It rather looks like common code that should be sharable and reusable?

.github/workflows/ci.yml
```diff
@@ -250,6 +253,37 @@ jobs:
...
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Typecheck (app-builder)
+        run: pnpm --filter app-builder typecheck
+
+      - name: Run app-builder tests
+        run: pnpm --filter app-builder test
```
marius-kilocode, line 285: I am a bit confused here should we not run the integration tests?

eshurakov:
thank you @marius-kilocode, PR is ready for another pass

cloudflare-app-builder/src/_integration_tests/run-all-tests.sh
```diff
@@ -0,0 +1,78 @@
...
+  local test_file="$2"
+
+  echo ""
+  echo "========================================"
+  echo "Running: $test_name"
+  echo "========================================"
+  echo ""
+
+  if pnpm exec tsx "$test_file"; then
+    TOTAL_PASSED=$((TOTAL_PASSED + 1))
```
marius-kilocode, line 42: Is there a reason why we do all of this manually? Should we not use a proper test runner instead?

cloudflare-app-builder/src/_integration_tests/test-init-edge-cases.ts
```diff
@@ -0,0 +1,371 @@
...
+
+  // Test various special characters - most should work since pattern is [a-z0-9_-]
+  const validTestId = `test-init_special-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
+
+  try {
+    const result = await initProjectRaw(validTestId);
+
+    if (result.status === 201 && result.body.success === true) {
+      logSuccess('Init with valid special chars (underscore, hyphen) succeeds');
+      return true;
```
marius-kilocode, line 281: I don't get why we are returning true or false here? We need assertions here or not and a proper test framework to detect failures.

cloudflare-app-builder/src/handlers/files.ts
```diff
@@ -0,0 +1,143 @@
...
+ * File tree and blob endpoint handlers
+ * GET /apps/{app_id}/tree/{ref} - Get directory contents
+ * GET /apps/{app_id}/blob/{ref}/{path} - Get file content
+ */
+
+import { verifyBearerToken } from '../utils/auth';
+import { logger } from '../utils/logger';
+import type { Env } from '../types';
+
+const notFoundPatterns = [
```
marius-kilocode, line 11: This is a really hacky model reward hack, once any error string changes the logic changes. We should introduce either proper error codes or error implementations.

cloudflare-app-builder/package.json
marius-kilocode: Still the old version?

PR #9, opened 2026-02-04
feat(share): implement shared session page with database integration

### LINKED ISSUE #105: `/share` does not work
### Description

I ran `/share` and got this link for my session: https://app.kilo.ai/s/84343b3d-69aa-4cee-8f36-085fb2dd9258
All that page shows is this:

Image

### Plugins

N/A

### OpenCode version

@kilocode/cli@1.0.13

### Steps to reproduce

1. Open a new session and do a simple task
2. Run `/share`
3. Open the URL from your clipboard in a browser

### Screenshot and/or share link

https://app.kilo.ai/s/84343b3d-69aa-4cee-8f36-085fb2dd9258

### Operating System

macOS 26.2

### Terminal

Terminal.app

iscekic:
Closes https://github.com/Kilo-Org/kilo/issues/105

image


markijbema (commented):
Don't we want to show a bit more of what the session is about? If i'd get linked to this I'd be quite wary of it

src/app/s/[sessionId]/page.tsx
```diff
@@ -1,10 +1,109 @@
...
+  }
+
+  const sessionResult = await db
+    .select({
+      ownerName: kilocode_users.google_user_name,
+      title: cli_sessions_v2.title,
+    })
+    .from(cli_sessions_v2)
+    .leftJoin(kilocode_users, eq(cli_sessions_v2.kilo_user_id, kilocode_users.id))
+    .where(eq(cli_sessions_v2.public_id, sessionId))
```
Comment (markijbema) [inline src/app/s/[sessionId]/page.tsx:33]: having a public_id is the security? or is every session shared if the id is known? I'd expect there to be some check so the user has to actively mark a session as public, but i might be missing context (other prs) here

iscekic:
> Don't we want to show a bit more of what the session is about? If i'd get linked to this I'd be quite wary of it

We are going to show a preview, but it's blocked on work currently in progress by @eshurakov and @pandemicsyn - making the components which display the new message format. Once those are done, integrating them here will be simple. Otherwise, we're just doing the same work twice. 😄

src/app/s/[sessionId]/page.tsx
Comment (iscekic) [reply src/app/s/[sessionId]/page.tsx:33]: `public_id` is `null` if the session isn't shared - the user generates and sets a public id when executing `/share`. The code for this is in `cloudflare-session-ingest`.

markijbema (approved):

PR #14, opened 2026-02-04
feat(cloud-agent): new worker for new kilo cli based on focus-week branch

pandemicsyn:
This is the same as the https://github.com/Kilo-Org/kilocode-backend/tree/kilo branch we've been hacking on since focus-week, just with the cloud-agent v2 routes changes broken out into a new dedicated worker cloud-agent-next.

The only difference is that this does also have the userId encoding fix that landed in main ~today~ yesterday, and that this worker has none of the legacy routes.

https://www.loom.com/share/9ee9ace2883f4dc9ae8266e3c3d339e8


pandemicsyn:
Same caveat about kilo review items as on the original PR. They're valid, but trying to avoid introducing new changes in this pr since this pr is mostly existing code. I can idle a cloud agent to fix items kilo review flagged after this lands.

cloud-agent-next/wrangler.jsonc
pandemicsyn, line 1: This is probably the most important bit of this whole PR and the only thing that really warrants close 👁️ by reviewers. 

Big thing to look out for is accidental references to the same resources as the original `cloud-agent` worker.

name is unique/new
dev.port is unique across the repo
workers using a new subdomain 
callback queue is a new one

cloud-agent-next/wrangler.jsonc
```diff
@@ -0,0 +1,268 @@
...
+  "$schema": "node_modules/wrangler/config-schema.json",
+  "name": "cloud-agent-next",
+  "account_id": "e115e769bcdd4c3d66af59d3332cb394",
+  "main": "src/index.ts",
+  "compatibility_date": "2025-09-15",
+  "compatibility_flags": ["nodejs_compat"],
+  "preview_urls": false,
+  "workers_dev": true,
+  "dev": {
+    "port": 8794,
```
dagadbm, line 15: i actually found an issue in the current repo: `"port": 8792` exists in 2 wrangler.jsonc files.

cloud-agent-next/wrangler.jsonc
```diff
@@ -0,0 +1,268 @@
...
+      "r2_buckets": [
+        {
+          "binding": "R2_BUCKET",
+          "bucket_name": "kilocode-sessions-dev",
+        },
+      ],
+      "containers": [
+        {
+          "class_name": "Sandbox",
+          "image": "./Dockerfile.dev",
```
dagadbm, line 229: maybe this is too nitpicky but these dockerfile names can be confusing.

Dockerfile => old CLI
Dockerfile.next => new CLI

dagadbm: else then it should be more obvious between the wrangler.jsonc of `cloud-agent` in the KILOCODE_CLI_VERSION should be the old version and in here the next (as it is done)

dagadbm (approved):
i left some non blocking comments in wrangler.jsonc file.
I  approved to not block but I dont have enough context to decide if there are other things that should be looked at more carefully.

cloud-agent-next/wrangler.jsonc
pandemicsyn: But this is using 8794 ?

cloud-agent-next/wrangler.jsonc
pandemicsyn: Theres no distinction between old and new cli in this worker. Its only new cli.

cloud-agent-next/wrangler.jsonc
dagadbm: i mean its not a problem for this PR. I just found this as an existing issue.

PR #19, opened 2026-02-04
Cloudflare Git Token Service - Add new worker for GitHub token management

eshurakov:
## Summary

- Add `cloudflare-git-token-service` Cloudflare Worker for centralized GitHub App token management
- Implement installation lookup with org membership verification to ensure users can only access installations they're authorized for
- Support both standard (read/write) and lite (read-only) GitHub App types with per-app credential configuration
- Add KV-based token caching with automatic TTL to reduce GitHub API calls
- Include GitHub Actions workflow for automated deployment


cloudflare-git-token-service/src/github-token-service.ts
```diff
@@ -24,16 +23,6 @@ type GeneratedToken = {
...
-  GITHUB_TOKEN_CACHE?: KVNamespace;
-  // Standard app credentials
-  GITHUB_APP_ID?: string;
-  GITHUB_APP_PRIVATE_KEY?: string;
-  // Lite app credentials
-  GITHUB_LITE_APP_ID?: string;
-  GITHUB_LITE_APP_PRIVATE_KEY?: string;
-};
-
 const TokenCacheEntrySchema = z.object({
```
jrf0110: Seems like `TokenCacheEntry`, `GeneratedToken`, and `TokenCacheEntrySchema` are all the same. Should probably consolidate into a single type inferred from the zod schema:

```tsx
type Token = z.infer<typeof Token>
const Token = z.object({
  token: z.string(),
  expiresAt: z.number(),
});
```

cloudflare-git-token-service/src/installation-lookup-service.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    const rows = await db.query(
+      /* sql */ `
+        SELECT
+          ${platform_integrations.platform_installation_id},
+          ${platform_integrations.platform_account_login},
+          ${platform_integrations.github_app_type}
+        FROM ${platform_integrations}
+        -- For org installations, verify user is a member of the org
+        LEFT JOIN ${organization_memberships}
```
jrf0110, line 99: do we also need to join on `kilocode_users` and ensure that `blocked_reason is null`?

jrf0110 (approved):

PR #27, opened 2026-02-05
feat: add free Claude Opus 4.5 for Kilo for Slack integration

RSO:
## Summary
- Add free Claude Opus 4.5 model (`anthropic/claude-opus-4.5:slackbot`) for Kilo for Slack integration
- Introduce `slackbot_only` flag to `KiloFreeModel` type to support models only available through internal API
- Hide slackbot-only models from the public model list for better UX
- Use generic `isSlackbotOnlyModel()` helper for access control instead of model-specific checks

## Testing
- All 131 test suites pass (1867 tests)
- Updated model list snapshot to reflect slackbot models being hidden

## Notes
- The Opus 4.5 slackbot model uses the existing Anthropic API key through Vercel AI Gateway
- No dedicated promotional API key needed (unlike MiniMax which uses `MINIMAX_FREE_SLACKBOT_PROMOTION_API_KEY`)

### DIFF
--- a/src/lib/providers/kilo-free-model.ts
+++ b/src/lib/providers/kilo-free-model.ts
@@ -14,6 +14,8 @@ export type KiloFreeModel = {
   gateway: ProviderId;
   internal_id: string;
   inference_providers: [OpenRouterInferenceProviderId, ...OpenRouterInferenceProviderId[]];
+  /** If true, this model is only available through Kilo for Slack (internalApiUse) and hidden from public model list */
+  slackbot_only?: boolean;
 };
 
 export function convertFromKiloModel(model: KiloFreeModel) {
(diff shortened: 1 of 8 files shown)


src/lib/providers/kilo-free-model.ts
```diff
@@ -14,6 +14,8 @@ export type KiloFreeModel = {
   gateway: ProviderId;
   internal_id: string;
   inference_providers: [OpenRouterInferenceProviderId, ...OpenRouterInferenceProviderId[]];
+  /** If true, this model is only available through Kilo for Slack (internalApiUse) and hidden from public model list */
+  slackbot_only?: boolean;
```
markijbema, line 18: Consider:
allowed_uses: ['slackbot']

that way we can reuse for reviewer, ...

markijbema (approved):
small suggestion

src/lib/providers/kilo-free-model.ts
RSO: While I like the suggestion, I won't have time to address it, so let's cross that bridge when we cross it 😄

PR #29, opened 2026-02-05
Add inactive user re-engagement rollout script

marius-kilocode:
.


src/scripts/d2025-02-04_cli-v1-rollout.ts
```diff
@@ -0,0 +1,246 @@
...
+import pLimit from 'p-limit';
+import * as fs from 'node:fs/promises';
+import * as path from 'node:path';
+
+const isDryRun = !process.argv.includes('--apply');
+
+// Resume support: If script crashes, you can resume from the last processed user ID
+// Example: pnpm script src/scripts/d2025-02-04_cli-v1-rollout.ts --resume=user-abc-123
+const resumeFromArg = process.argv.find(arg => arg.startsWith('--resume='));
+const RESUME_FROM_USER_ID = resumeFromArg ? resumeFromArg.split('=')[1] : null;
```
markijbema: I would instead just let it only process users who dont have credits yet, this seems like nonsense

src/scripts/d2025-02-04_cli-v1-rollout.ts
```diff
@@ -0,0 +1,250 @@
...
+  const scriptStartTime = Date.now();
+
+  // Step 1: Pre-compute recently active users (last 30 days)
+  console.log(`Fetching recently active user IDs (last ${INACTIVE_DAYS} days)...`);
+  const cutoff = new Date(Date.now() - INACTIVE_DAYS * 24 * 60 * 60 * 1000);
+
+  const recentlyActiveRows = await db
+    .selectDistinct({ userId: microdollar_usage.kilo_user_id })
+    .from(microdollar_usage)
+    .where(gte(microdollar_usage.created_at, cutoff.toISOString()));
```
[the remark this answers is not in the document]
markijbema: don't know about this one, please test/try

markijbema (commented):
I think i would make the script get the X recently active rows without these credits, and then give them all credits (maybe do a few batches, but not as small as opus thinks)

Then you have a bit more control, and it's easier to resume; 

Looking at who we are trying to target, I'd even say it would be better in two phases; first add them to a cohort (@olearycrew added this last week), then give everyone in that cohort the $1. That way you have complex code defining the cohort, but that can do real cheap sql writing. 

And then the actualy iteration code is easier. And it would make it easier to also query them for mailing them or something

src/scripts/d2025-02-04_cli-v1-rollout.ts
RSO: This was just copied from the previous rollout script

src/scripts/d2025-02-04_cli-v1-rollout.ts
```diff
@@ -0,0 +1,246 @@
...
+
+const isDryRun = !process.argv.includes('--apply');
+
+// Resume support: If script crashes, you can resume from the last processed user ID
+// Example: pnpm script src/scripts/d2025-02-04_cli-v1-rollout.ts --resume=user-abc-123
+const resumeFromArg = process.argv.find(arg => arg.startsWith('--resume='));
+const RESUME_FROM_USER_ID = resumeFromArg ? resumeFromArg.split('=')[1] : null;
+
+const BATCH_SIZE = 10;
+const SLEEP_AFTER_BATCH_MS = 1000;
```
RSO: I'd try to have this much lower. I think we needed it this high because of Orb, but there's no Orb anymore. Lowering should make the script run much faster

RSO (approved):
Small comments that I hope will make this script run faster. Let's try a dry run?

src/scripts/d2025-02-04_cli-v1-rollout.ts
markijbema: I'm not saying the previous script wasnt vibed :P

PR #41, opened 2026-02-05
Swap recommended model to Opus 4.6

lambertjosh:


lambertjosh:
This should indeed be fine, here is a change to add some models: https://github.com/Kilo-Org/OLD-UNUSED-kilocode-dnekcab/pull/4478/changes

The defaults need to get updated since tests specifically look for that.

iscekic (approved):

PR #43, opened 2026-02-05
chore(cloudflare-webhook-agent-ingest): try to distinguish errors

pandemicsyn:
Trying to track down a timeout issue some users are see'ing.


cloudflare-webhook-agent-ingest/src/db/node-postgres.ts
```diff
@@ -13,7 +13,7 @@ types.setTypeParser(types.builtins.INT8, val => parseInt(val, 10));
 export const createNodePostgresConnection: CreateDatabaseConnection = connectionString => {
   const pool = new Pool({
     connectionString,
-    max: 5,
+    max: 10,
```
pandemicsyn, line 16: Maybe 5 was to conservative ?

jrf0110 (approved):

PR #47, opened 2026-02-05
chore(cloudflare-webhook-agent-ingest): Use client instead of pool

pandemicsyn:
Switch to pg client per use.

### DIFF
--- a/cloudflare-webhook-agent-ingest/src/services/profile-resolution-service.ts
+++ b/cloudflare-webhook-agent-ingest/src/services/profile-resolution-service.ts
@@ -15,18 +15,8 @@ type ResolveProfileParams = {
   orgId?: string | null;
 };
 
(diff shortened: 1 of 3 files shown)


cloudflare-webhook-agent-ingest/src/services/profile-resolution-service.ts
```diff
@@ -15,18 +15,8 @@ type ResolveProfileParams = {
   orgId?: string | null;
 };
 
-// Singleton instance for connection pooling
-let singleton: ProfileResolutionService | null = null;
```
pandemicsyn, line 19: I don't _think_ this would really cause issues, but i definitely didn't have this singleton in my spec - i think the robot snuck this one past me at some point.

eshurakov (approved):

PR #54, opened 2026-02-06
fix(ci): add test run to autofix workflow for snapshot updates

lambertjosh:
The autofix workflow was missing test runs that update approval/snapshot files. This was present in the old repo's ci.yml but got lost during the repo migration.

- Old PR that seems to have added this: https://github.com/Kilo-Org/OLD-UNUSED-kilocode-dnekcab/pull/4295
- History which shows this that the snapshot updates were largely automated: https://github.com/Kilo-Org/OLD-UNUSED-kilocode-dnekcab/commits/main/src/tests/openrouter-models-sorting.approved.json

Changes:
- Add postgres service for tests that need it
- Run tests (with || true to not fail on test failures)
- Add guard against infinite commit loops
- Update commit message to reflect snapshot updates

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -11,6 +11,26 @@ jobs:
   autofix:
     runs-on: ubuntu-24.04-8core
 
+    services:
+      postgres:
+        image: pgvector/pgvector:pg16
+        env:
+          POSTGRES_HOST_AUTH_METHOD: trust
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+
+    env:
+      NODE_ENV: test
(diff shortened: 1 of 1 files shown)


chrarnoldus (approved):

lambertjosh:
@RSO - do you want to look at this one too?

markijbema (changes requested):
Previously this ran on pr, it makes no sense to silently update approvals on the master branch. I think it's fine to re-add on pr, but then you'd need to think about how this should work for external contributors (that is, prs should still work, which is why linting etc was moved)

lambertjosh:
I'm just going to close this for now, it's unclear the value as opposed to just failing the test and requiring the user to update the files manually. Doing it automatically seems like a risk for an unintentional breakage of ordering or similar. (Although it should be evident in other diffs, but perhaps not as clearly)

PR #64, opened 2026-02-09
Log API requests made by kilocode employees

chrarnoldus:
To help us diagnose issue we encounter or reproduce ourselves.

### DIFF
--- a/src/lib/handleRequestLogging.ts
+++ b/src/lib/handleRequestLogging.ts
@@ -1,0 +1,46 @@
+import { api_request_log, type User } from '@/db/schema';
+import { db } from '@/lib/drizzle';
+import { KILO_ORGANIZATION_ID } from '@/lib/organizations/constants';
+import type { OpenRouterChatCompletionRequest } from '@/lib/providers/openrouter/types';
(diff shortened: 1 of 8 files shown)


src/lib/handleRequestLogging.ts
```diff
@@ -0,0 +1,46 @@
...
+  organization_id: string | null;
+  provider: string;
+  model: string;
+  request: OpenRouterChatCompletionRequest;
+}) {
+  const { clonedResponse, user, organization_id, provider, model, request } = params;
+  if (
+    !user?.google_user_email.endsWith('@kilo.ai') &&
+    !user?.google_user_email.endsWith('@kilocode.ai') &&
+    organization_id !== KILO_ORGANIZATION_ID
```
markijbema, line 20: I'd really appreciate an variable isKiloEmployee instead, and have a negated check here. My in-head demorgan is always a bit slow

markijbema (approved):

src/lib/handleRequestLogging.ts
chrarnoldus: ok

PR #72, opened 2026-02-09
AppBuilder - Add GitHub migration feature

eshurakov:
## Summary

- Add migrate-to-GitHub endpoint (`POST /apps/{app_id}/migrate-to-github`) to the Cloudflare App Builder worker that pushes the internal git repository to a GitHub remote and switches the preview to clone from GitHub.
- Extend `GitRepositoryDO` with `pushToRemote` (exports git objects to an in-memory FS and pushes all branches) and `scheduleDelete` (alarm-based self-deletion after a grace period).
- Extend `PreviewDO` with `setGitHubSource` and `migrateToGithub` methods; after migration, fresh GitHub tokens are fetched from `git-token-service` via RPC on every clone/fetch.
- Add `MigrateToGitHubDialog` React component with repo creation, installation linking, and migration flow UI.
- Add `github-migration-service.ts` on the Next.js side orchestrating repo creation, installation token retrieval, and calling the worker migration endpoint.
- Add GitHub adapter helpers for creating repositories and generating installation tokens.
- Add new DB migration (`0004`) with `app_builder_github_repos` and `app_builder_github_repo_commits` tables.
- Wire up new tRPC routes (`app-builder-router`, `organization-app-builder-router`) for triggering migration and listing GitHub installations.
- Update push webhook handler to recognize app-builder repos and record commits.


src/db/migrations/0009_minor_leo.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE "app_builder_project_sessions" (
+	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+	"project_id" uuid NOT NULL,
+	"cloud_agent_session_id" text NOT NULL,
```
markijbema, line 4: wouldnt this imply the project_id, that is, is project_id here redundant? If so I think I'd prefer not saving it, as the speed gain isnt worth the change of data inconsistency, or the cost of an update later

src/lib/app-builder/app-builder-service.ts
```diff
@@ -70,6 +88,53 @@ async function getProjectWithOwnershipCheck(
   return project;
 }
 
+/**
+ * Lazily backfill session tracking for existing projects.
+ * Uses ON CONFLICT DO NOTHING so it's safe to call repeatedly.
```
markijbema: is this better? It sounds like this means the sessions could get out of date? and/or why would we expect conflicts? Why is it better to keep the value we have instead of what we would calculate in the backfill?

src/lib/app-builder/app-builder-service.ts
```diff
@@ -70,6 +88,53 @@ async function getProjectWithOwnershipCheck(
...
+  return db
+    .select()
+    .from(app_builder_project_sessions)
+    .where(
+      and(
+        eq(app_builder_project_sessions.project_id, projectId),
+        isNull(app_builder_project_sessions.ended_at)
+      )
+    )
+    .limit(1);
```
markijbema: in these cases i'd prefer a limit 2, and an exception if it is more than one

markijbema: (unless there is a guarantee by an index of course)

cloudflare-app-builder/src/handlers/migrate-to-github.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    logger.info({ source: 'MigrateToGithubHandler', appId }, 'Successfully migrated to GitHub');
+
+    return new Response(JSON.stringify({ success: true }), {
+      status: 200,
+      headers: { 'Content-Type': 'application/json' },
+    });
+  } catch (error) {
+    logger.error({ source: 'MigrateToGithubHandler' }, 'Migrate to GitHub handler error', error);
+    return new Response(
```
markijbema, line 139: I think an error response method with status/error/message as params would make this file a lot more readable, sa now a big part is error handling which makes the code itself harder to follow

cloudflare-app-builder/src/handlers/migrate-to-github.ts
```diff
@@ -0,0 +1,128 @@
...
+          success: false,
+          error: 'internal_error',
+          message: 'Repository not found',
+        }),
+        { status: 404, headers: { 'Content-Type': 'application/json' } }
+      );
+    }
+
+    logger.info({ source: 'MigrateToGithubHandler', appId }, 'Pushing repository to remote', {
+      remoteUrl: remoteUrl.replace(/\/\/[^@]*@/, '//***@'),
```
markijbema: could you add a comment here? showing what is substituted here? In general I typically prefer regexes which match either begin, end or both, as without any anchoring you have a bigger change of changing something else than you intend

cloudflare-app-builder/src/git-repository-do.ts
```diff
@@ -263,4 +265,136 @@ export class GitRepositoryDO extends DurableObject<Env> {
       logger.info('Repository deleted successfully');
     });
   }
+
+  /**
+   * Schedule self-deletion after a delay.
+   * Used after GitHub migration to clean up the internal git repo
+   * while keeping a grace period for rollback.
+   */
+  async scheduleDelete(delayMs: number): Promise<void> {
```
markijbema, line 275: Do we need this? Can't we remove all internal git repo's april first or something like that?

cloudflare-app-builder/src/git-repository-do.ts
```diff
@@ -263,4 +265,136 @@ export class GitRepositoryDO extends DurableObject<Env> {
...
+      await this.initializeFS();
+    }
+
+    if (!this._initialized || !this.fs) {
+      return { success: false, error: 'Repository not initialized' };
+    }
+
+    logger.info('Pushing repository to remote', {
+      id: this.ctx.id.toString(),
+      remoteUrl: remoteUrl.replace(/\/\/[^@]*@/, '//***@'), // Redact token in logs
```
markijbema: or maybe extract this to a method censorToken?

cloudflare-app-builder/src/git-repository-do.ts
```diff
@@ -263,4 +265,136 @@ export class GitRepositoryDO extends DurableObject<Env> {
...
+            dir: '/',
+            url: remoteUrl,
+            ref: branch,
+            remoteRef: branch,
+            onAuth: () => ({ username: 'x-access-token', password: authToken }),
+            force: false, // Don't force push
+          });
+
+          logger.info('Successfully pushed branch', { branch });
+          if (branch === 'main') {
```
markijbema, line 369: why is main hardcoded; don't we want to use githubs default branch field for this?

cloudflare-app-builder/src/index.ts
```diff
@@ -27,6 +28,7 @@ const COMMIT_PATTERN = new RegExp(`^/apps/(${APP_ID_PATTERN_STR})/commit$`);
 const PREVIEW_STATUS_PATTERN = new RegExp(`^/apps/(${APP_ID_PATTERN_STR})/preview$`);
 const BUILD_TRIGGER_PATTERN = new RegExp(`^/apps/(${APP_ID_PATTERN_STR})/build$`);
 const BUILD_LOGS_PATTERN = new RegExp(`^/apps/(${APP_ID_PATTERN_STR})/build/logs$`);
+const MIGRATE_TO_GITHUB_PATTERN = new RegExp(`^/apps/(${APP_ID_PATTERN_STR})/migrate-to-github$`);
```
markijbema, line 31: are we sure APP_ID_PATTERN_STR contains no characters which can mean something in a regex like `.` ?

src/lib/integrations/platforms/github/adapter.ts
```diff
@@ -481,6 +481,181 @@ function isHttpError(error: unknown): error is { status: number; message: string
...
+  cloneUrl: string;
+  htmlUrl: string;
+  isEmpty: boolean;
+  isPrivate: boolean;
+} | null> {
+  const tokenData = await generateGitHubInstallationToken(installationId, appType);
+  const octokit = new Octokit({ auth: tokenData.token });
+
+  const [owner, repo] = repoFullName.split('/');
+  if (!owner || !repo) {
```
markijbema, line 510: do we also want to check for the failure case that there is more than 1 slash? (I don't know the source of this field, and how trustworthy it is)

src/lib/integrations/platforms/github/adapter.ts
```diff
@@ -481,6 +481,181 @@ function isHttpError(error: unknown): error is { status: number; message: string
...
+  const repositories: Array<{
+    fullName: string;
+    createdAt: string;
+    isPrivate: boolean;
+  }> = [];
+  let page = 1;
+  const perPage = 100;
+  const maxRepos = 200;
+
+  while (repositories.length < maxRepos) {
```
markijbema: this feels like a lot of infrastructure to do 1 or 2 requests. Can't we just request 200 at once?

src/lib/app-builder/github-migration-service.ts
```diff
@@ -0,0 +1,320 @@
...
+  MigrateToGitHubInput,
+  MigrateToGitHubResult,
+  CanMigrateToGitHubResult,
+} from '@/lib/app-builder/types';
+
+/**
+ * Sanitize a string for use as a GitHub repository name.
+ * GitHub repo names can only contain alphanumeric characters, hyphens, underscores, and periods.
+ */
+function sanitizeRepoName(name: string): string {
```
markijbema: is there no method for this in octokit?

src/lib/app-builder/github-migration-service.ts
```diff
@@ -0,0 +1,320 @@
...
+  return name
+    .toLowerCase()
+    .replace(/[^a-z0-9._-]/g, '-') // Replace invalid chars with hyphens
+    .replace(/-+/g, '-') // Collapse multiple hyphens
+    .replace(/^[-.]|[-.]$/g, '') // Remove leading/trailing hyphens and dots
+    .substring(0, 100); // Truncate to 100 chars
```
markijbema: i think we want to truncate first, and then remove trailing stuff, otherwise you can still have trailng hyphens and dots

markijbema (approved):
Added some questions/sanity checks, but none blocking I'd say

cloudflare-app-builder/src/handlers/migrate-to-github.ts
eshurakov: good point, we already have a method to strip credentials from git urls, I will use it here

cloudflare-app-builder/src/handlers/migrate-to-github.ts
eshurakov: good point, I will extract it later!

cloudflare-app-builder/src/git-repository-do.ts
eshurakov: It's up to users to use GitHub, they start with our internal repo and may use it as long as they want. So it is possible that users will never migrate to GitHub.

cloudflare-app-builder/src/git-repository-do.ts
eshurakov: App Builder is a single branch platform right now, so it only supports main

cloudflare-app-builder/src/index.ts
eshurakov: Looks fine, but I will migrate to Hono later, this was fine with a few routes, now it gets out of hands

src/db/migrations/0009_minor_leo.sql
eshurakov: we can't figure out app builder project id only from cloud agent session id, so both fields are needed here

src/lib/app-builder/app-builder-service.ts
eshurakov: hmm yeah, I think I'll just run one time migration

src/lib/app-builder/app-builder-service.ts
eshurakov: actually unused method, removed it

src/lib/app-builder/github-migration-service.ts
eshurakov: didn't find anything like this

src/lib/integrations/platforms/github/adapter.ts
eshurakov: yeah, I've removed this method completely :-)

cloudflare-app-builder/src/git-repository-do.ts
markijbema: no i get that, but why not let the single branch be the default branch, for instance it is 'dev' on our kilo repo, and 'master' on some legacy git repos, and this is a field in the github config

markijbema (approved):

PR #79, opened 2026-02-09
Deployments - Add custom slug support

eshurakov:
## Summary

- Decouple the public-facing subdomain (slug) from the internal Cloudflare worker name by adding an `internal_worker_name` column (`dpl-<uuid>` format) to the deployments table
- Add deployment rename functionality (`renameDeployment` tRPC endpoint) that updates the public slug/URL while preserving the internal CF worker name
- Consolidate `DEPLOY_AUTH_KV` and `SLUG_MAPPINGS_KV` into a single `DEPLOY_KV` namespace in the dispatcher, with key prefixes: `password:<worker>`, `slug2worker:<slug>`, `worker2slug:<worker>`
- Add slug mapping CRUD API routes to the dispatcher and a unified backend client (`dispatcher-client.ts`, replacing the old `password-client.ts`)
- Update dispatcher middleware to resolve `slug -> worker name` via KV lookup, with backwards compatibility (falls back to using slug directly as worker name)
- Key password protection by internal worker name so auth persists across renames
- Add `checkSlugAvailability` endpoint (validates format, reserved words, and uniqueness against both `deployment_slug` and `internal_worker_name`)
- Slugs are auto-generated at creation time; custom subdomains are only available via rename after the fact
- Clean up orphaned CF workers best-effort when a slug uniqueness race causes the DB insert to fail

## Notes

- Migration `0005_strong_rachel_grey.sql` adds `internal_worker_name` as a text column, backfills it from `deployment_slug`, then sets `NOT NULL`
- `DEPLOY_KV` namespace IDs in `wrangler.jsonc` reuse the existing KV namespace (same IDs as the old `DEPLOY_AUTH_KV`)


src/lib/user-deployments/deployments-service.ts
```diff
@@ -649,39 +709,127 @@ export async function createDeployment(params: {
...
+  const newUrl = `https://${newSlug}.${DEFAULT_DEPLOYMENT_DOMAIN}`;
+
+  // DB update first — the unique constraint is the authoritative guard against races.
+  // Use optimistic locking: only update if the slug hasn't changed since we read it.
+  const [updated] = await db
+    .update(deployments)
+    .set({
+      deployment_slug: newSlug,
+      deployment_url: newUrl,
+      internal_worker_name: internalWorkerName,
```
marius-kilocode: Why is this read and written just after?

eshurakov: Good question. Though it doesn't do any harm, it is not also doing anything good. Removing it

marius-kilocode (approved):

PR #133, opened 2026-02-11
Add SlackBot models to model list

chrarnoldus:
otherwise you can't go back after switching


jrf0110:
image

This make it such that the free Opus shows up in the model selector. Pretty sure we don't want that!

chrarnoldus:
That's exactly what I want, we don't have different model selectors currently.

(being able to select it doesn't mean you're able to use it)

jrf0110:
Fair enough. If that's intended, then it LGTM

jrf0110 (approved):

jrf0110:
I do wonder if there's another way we can plumb through slackbot only models without exposing junk models to the cli/extension

chrarnoldus:
I'm sure we can with enough effort, on the other hand the list already contains a load of junk realistically.

PR #169, opened 2026-02-13
Replace Cloudflare Containers with Fly.io Machines

pandemicsyn:
## Summary

Migrate KiloClaw's compute backend from Cloudflare Containers (`@cloudflare/sandbox`) to Fly.io Machines. The CF Worker + Durable Objects remain as the orchestration and auth layer.

The KiloClawInstance Durable Object is the coordination primitive for each user's instance. Every mutation to the instance's lifecycle (provision, start, stop, destroy) and every read of its state (status, config, machine ID) is serialized through a single DO.

This gives us:

1. Single-writer semantics — No two concurrent requests can race to create a machine or double-destroy a volume. The DO's single-threaded execution model prevents it.
2. Authoritative state — The DO is the source of truth for flyMachineId, flyVolumeId, status, pending-destroy IDs, health check counters. Fly is the source of truth for whether the machine is actually running, but the DO decides what the system should do about it.
3. Self-healing via alarm — The reconciliation loop is co-located with the state it repairs. There's no external cron or central reconciler — each DO independently detects and fixes drift for its own instance.
4. Proxy routing — The catch-all in index.ts resolves the flyMachineId from the DO before forwarding to Fly Proxy. The DO is the lookup index.

The only state that lives outside the DO is:
   - Postgres — same as before this PR - registry/backup (Next.js is sole writer, worker reads only for restore) this is where custom config stuff related to OpenClaw will  be stored.
   - Fly metadata — recovery index (written at machine creation, queried only when DO state is lost)
  - Fly itself — the actual compute and volume resources

All three are fallback/recovery paths. The DO is the atom.

## Architecture

```
Browser → CF Worker (claw.kilo.ai)
            │ JWT auth, look up flyMachineId from DO
            │ add fly-force-instance-id header
            ↓
         Fly Proxy ({FLY_APP_NAME}.fly.dev, TLS)
            │ routes to pinned machine
            ↓
         Fly Machine (openclaw gateway on port 18789)
            │ Fly Volume mounted at /root
```

### Compute Backend
- **New**: Fly Machines API client (`src/fly/client.ts`, `src/fly/types.ts`) — create/start/stop/destroy/wait/list machines + volumes
- **New**: Fly metadata tagging (`kiloclaw_user_id`, `kiloclaw_sandbox_id`) for recovery
- **Removed**: `@cloudflare/sandbox` dependency, `KiloClawSandbox` DO, R2 sync, process management

### Reliability
- **Two-phase destroy**: Pending resource IDs persisted before deletion; DO state only cleared when both machine + volume confirmed deleted; alarm retries failures
- **Reconciliation loop**: Runs for all instance statuses (provisioned/running/stopped/destroying); fixes drift, missing volumes, stale machine IDs, wrong mounts
- **Metadata recovery**: When DO state is lost, recovers machine/volume IDs from Fly metadata tags with deterministic selection (prefer started > stopped, newest first)
- **No duplicate machines**: Only recreates on 404 (confirmed gone); transient Fly errors re-thrown, not masked
- **Machine ID persisted before wait**: `createNewMachine()` writes ID to durable storage immediately after `fly.createMachine()`, before `waitForState()`
- **Proxy blocks during destroy**: Returns 409 when instance status is `destroying`
- **Crash recovery refreshes routing**: After `attemptCrashRecovery`, re-resolves machine ID before retry

### Env Var Transport
- `KILOCODE_MODELS_JSON` — user model list serialized as JSON (was silently dropped before)
- `TELEGRAM_DM_POLICY` / `DISCORD_DM_POLICY` — worker env passthrough to machine

### App-Side/nextjs Updates
- Added `destroying` status handling to types, router, hooks, and UI
- Removed `deleteData` from destroy flow (always deletes volume)
- Removed storage/sync endpoints and UI (410 Gone stubs on worker)


pandemicsyn:
Do still need to sort proper secrets injection - just using plain env right now.

jrf0110 (approved):

PR #170, opened 2026-02-13
Add fireworks as provider for glm-5:free and update display name

chrarnoldus:


iscekic (approved):
Nit: might be overkill to generate sha256, a faster hashing algorithm might be a slight improvement (this is a hot path after all).

chrarnoldus:
>Nit: might be overkill to generate sha256, a faster hashing algorithm might be a slight improvement (this is a hot path after all).

I removed that code because OR says they already do it

PR #171, opened 2026-02-13
Add survey to Onboarding Users

jfcsantos:
## Summary
- Add welcome survey modal using Typeform that displays for new users on the profile page
- Survey shows after user signup only. Once submitted or dismissed we store it in db and never show it again.
- Track survey completion status in the database to only show the modal once
- Integrate with tRPC to persist completion state

## Changes
- New component: WelcomeTypeformModal - embedded Typeform widget in a responsive modal dialog
- Database: Added completed_welcome_form boolean column to kilocode_users table
- API: New markWelcomeFormCompleted tRPC mutation to record when users complete/skip the survey
- Profile page: Conditionally render the modal for users who haven't completed the form

## Implementation Details
- Modal prevents accidental dismissal (blocks outside clicks and escape key)
- Handles both form submission and skip/close actions
- Includes error handling with retry capability if the API call fails
- Passes user email as a hidden field to Typeform for tracking
- Typeform link: https://admin.typeform.com/form/xNTrQO4E/create?block=18842f7b-22e5-489b-955a-e7d8483656bd

Screenshot 2026-02-13 at 12 15 38


src/db/schema.ts
```diff
@@ -175,6 +175,7 @@ export const kilocode_users = pgTable(
     is_bot: boolean().default(false).notNull(),
     default_model: text(),
     cohorts: jsonb().$type<Record<string, number>>().default({}).notNull(),
+    completed_welcome_form: boolean().default(false).notNull(),
```
iscekic, line 178: Should we default to false? This way every user is going to see it, not just new users. 🤔

src/components/profile/WelcomeTypeformModal.tsx
iscekic, line 1: I think the bot may have a good point about theming

src/components/profile/WelcomeTypeformModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <DialogContent
+        className="h-[50vh] max-w-[90vw] overflow-hidden p-0 sm:h-[400px] sm:max-w-xl [&>button]:rounded-md [&>button]:p-2 [&>button]:text-black [&>button]:transition-colors [&>button]:hover:bg-black/10 [&>button]:hover:text-black/60"
+        onPointerDownOutside={e => e.preventDefault()}
+        onEscapeKeyDown={e => e.preventDefault()}
+      >
+        <DialogTitle className="sr-only">Welcome to the app</DialogTitle>
+        <Widget
+          id={WELCOME_FORM_ID}
+          style={{ width: '100%', height: '100%', minHeight: '400px' }}
+          className="my-form"
```
iscekic: This class seems redundant?

src/components/profile/WelcomeTypeformModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+        if (!open) handleSkip();
+        setIsOpen(open);
+      }}
+    >
+      <DialogContent
+        className="h-[50vh] max-w-[90vw] overflow-hidden p-0 sm:h-[400px] sm:max-w-xl [&>button]:rounded-md [&>button]:p-2 [&>button]:text-black [&>button]:transition-colors [&>button]:hover:bg-black/10 [&>button]:hover:text-black/60"
+        onPointerDownOutside={e => e.preventDefault()}
+        onEscapeKeyDown={e => e.preventDefault()}
+      >
+        <DialogTitle className="sr-only">Welcome to the app</DialogTitle>
```
iscekic:
```suggestion
        <DialogTitle className="sr-only">Welcome to Kilo</DialogTitle>
```

src/components/profile/WelcomeTypeformModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <Widget
+          id={WELCOME_FORM_ID}
+          style={{ width: '100%', height: '100%', minHeight: '400px' }}
```
iscekic: Styles here and above are a bit weird, but I guess that's what we need to use because of the library?

jfcsantos: it actually works fine without this.

src/db/schema.ts
jfcsantos: hmm, could be useful info but I'm guessing we don't want that for existing users @emilieschario? 
if we don't we can just update the migration to set it `true` for existing users, what do you think? 

```
ALTER TABLE "kilocode_users" ADD COLUMN "completed_welcome_form" boolean DEFAULT false NOT NULL;
UPDATE "kilocode_users" SET "completed_welcome_form" = true;
```

src/components/profile/WelcomeTypeformModal.tsx
jfcsantos: It's weird because dark mode on the modal wasn't triggering for me until I forced it in the browser console, regardless it should look good now

src/db/schema.ts
emilieschario: let's make sure we only show to new users

jfcsantos: I'll push the migration update to set that field to `true` for all existing users.

src/db/migrations/0012_square_marvel_apes.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "kilocode_users" ADD COLUMN "completed_welcome_form" boolean DEFAULT false NOT NULL;
+UPDATE "kilocode_users" SET "completed_welcome_form" = true;
```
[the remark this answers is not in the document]
jfcsantos: This is such an edge case and the only fix (I believe) would be to add a timestamp to make sure we capture users created in-between. I'm okay with this current approach

jfcsantos:
@iscekic might giving this another look?

src/db/migrations/0013_opposite_groot.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "kilocode_users" ADD COLUMN "completed_welcome_form" boolean DEFAULT false NOT NULL;
```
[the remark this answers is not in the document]
jfcsantos: sorry @iscekic I lost the update on the merge conflict with main...fixing it now

iscekic (approved):

PR #239, opened 2026-02-16
Organization Switcher skeleton loader

jfcsantos:
## Summary
The OrganizationSwitcher in the sidebar was taking randomly a long time to load. Navigating to an organization page also briefly flashed "Personal Workspace" before showing the correct org name.

## Changes
- Show skeleton UI while organizations list is loading
- Prevents layout shift by matching skeleton dimensions to actual button
- Uses existing Skeleton component from UI library
- Server-side prefetch (PrefetchedOrganizations.tsx): New server component that prefetches the org list during SSR via tRPC server caller + React Query's HydrationBoundary. The OrganizationSwitcher now renders with data immediately instead of showing a skeleton.
- Synchronous URL parsing (useUrlOrganizationId.ts): Replaced useState/useEffect with useMemo so the org ID is available on the first render, eliminating the "Personal Workspace" flash.
- Layout integration (layout.tsx): Wrapped the app sidebar content with PrefetchedOrganizations.

## Root Causes
- Client-side waterfall: The organizations.list query only fired after the client-side React tree mounted, causing a visible loading state on every page navigation.
- Deferred URL parsing: useUrlOrganizationId used useState + useEffect to extract the org ID from the URL, so organizationId was null on the first render — causing a flash of "Personal Workspace".

### DIFF
--- a/src/app/(app)/components/OrganizationSwitcher.tsx
+++ b/src/app/(app)/components/OrganizationSwitcher.tsx
@@ -8,6 +8,7 @@ import {
   DropdownMenuSeparator,
   DropdownMenuTrigger,
 } from '@/components/ui/dropdown-menu';
+import { Skeleton } from '@/components/ui/skeleton';
 import { useTRPC } from '@/lib/trpc/utils';
 import { Check, ChevronDown } from 'lucide-react';
 import { useQuery } from '@tanstack/react-query';
@@ -51,7 +52,17 @@ export default function OrganizationSwitcher({ organizationId = null }: Organiza
   if (isLoading) {
     return (
       <div className="mt-1">
-        <div className="h-[62px] w-full animate-pulse rounded-lg border border-gray-700 bg-gray-800" />
+        <Button
+          variant="ghost"
+          disabled
+          className="h-auto w-full justify-between rounded-lg border border-gray-700 p-3 text-left"
+        >
+          <div className="flex flex-col items-start">
+            <Skeleton className="h-5 w-24" />
+            <Skeleton className="h-4 w-16" />
+          </div>
+          <ChevronDown className="h-4 w-4 text-gray-500" />
+        </Button>
       </div>
     );
   }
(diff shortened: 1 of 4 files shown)


jfcsantos:
ccing @emilieschario since we were testing this.

markijbema (approved):
I mean, this is fine, but would we actually expect this to be slow to load? It's just a simple sql query right?

jfcsantos:
> I mean, this is fine, but would we actually expect this to be slow to load? It's just a simple sql query right?

There may be some query improvements we can make here - adding that to my list, but it sometimes takes close to 1 min+ to load the org switcher.

https://github.com/user-attachments/assets/f4f420fa-ff2d-43fb-8cb1-8cc67e7ac832

emilieschario:
> I mean, this is fine, but would we actually expect this to be slow to load? It's just a simple sql query right?

I lose at least 10 minutes per day waiting for this to load. 🫠

markijbema:
But that's just insane loading time right? We shouldnt build a loader because something which should take maybe 100ms takes a minute right? We should fix that. Personally I thought it was just broken. If you go to /profile it loads fine, and fast, so there's something weird happening here

jfcsantos:
@markijbema If you could have a look, this should be much better now.

jfcsantos:
> But that's just insane loading time right? We shouldnt build a loader because something which should take maybe 100ms takes a minute right? We should fix that. Personally I thought it was just broken. If you go to /profile it loads fine, and fast, so there's something weird happening here

I believe there's probably some further optimizations that could be done on the queries and caching so I'll put it on my list to look at that...the goal here was mainly to load the org switcher faster and reduce "janky" ui

src/app/(app)/components/OrganizationSwitcher.tsx
```diff
@@ -22,7 +23,7 @@ export default function OrganizationSwitcher({ organizationId = null }: Organiza
   const router = useRouter();
 
   // Fetch user organizations
-  const { data: organizations } = useQuery(trpc.organizations.list.queryOptions());
+  const { data: organizations, isPending } = useQuery(trpc.organizations.list.queryOptions());
```
jrf0110: I _believe_ the reason this is slow a lot of times is due to batching queries with trpc. You can use:


```suggestion
  const { data: organizations, isPending } = useQuery(trpc.organizations.list.queryOptions(
    {},
    {
      trpc: {
        context: {
          skipBatch: true,
        },
      },
    }
  ));
```

To make this query skip being part of a batch

markijbema (approved):

PR #242, opened 2026-02-16
CloudAgentNext - Move cli_sessions_v2 creation to session-ingest RPC

eshurakov:
## Summary

Update Cloud Agent Next to use the Ingest Service to pre-create CLI sessions and add a link between Cloud Agent Session and CLI session.

- Separate session-ingest RPC entrypoint from Hono app


cloud-agent-next/src/persistence/CloudAgentSession.ts
```diff
@@ -574,41 +573,6 @@ export class CloudAgentSession extends DurableObject {
...
-    if (!metadata?.kilocodeToken) {
-      throw new Error('Cannot link session: missing kilocodeToken');
-    }
-
-    const backendUrl = (this.env as unknown as WorkerEnv).KILOCODE_BACKEND_BASE_URL;
-    if (!backendUrl) {
-      throw new Error('Cannot link session: KILOCODE_BACKEND_BASE_URL not configured');
-    }
-
-    const response = await fetch(`${backendUrl}/api/cloud-sessions/linkSessions`, {
```
eshurakov, line 594: this endpoint doesn't even exist :-)

cloudflare-session-ingest/src/session-ingest-rpc.ts
iscekic, line 1: I wonder if we could extract some functions for reuse 🤔

iscekic (approved):

cloudflare-session-ingest/src/session-ingest-rpc.ts
eshurakov: definitely, can do a service layer

PR #251, opened 2026-02-16
feat(kiloclaw): per-user Fly apps with network isolation

pandemicsyn:
Create a dedicated Fly App per user instead of sharing a single app. Each user's app gets its own isolated private network, preventing cross-user lateral movement over Fly's internal DNS, and also means we can start using App Secrets. (app per user is also recommended by Fly for this scenario - https://fly.io/docs/machines/guides-examples/one-app-per-user-why/)

New KiloClawApp Durable Object manages per-user app lifecycle:
- Idempotent app creation with retry alarm on partial failure
- IPv4 (shared) + IPv6 allocation via Machines REST API (i think we probably don't need to keep provisioning ipv4 addrs , so i'll probably remove in a follow up)
- Network isolation via app-scoped private network
- Environment-aware naming: dev-{hash} / acct-{hash}

Key changes:
- Apps REST API client (create, get, delete, IP assignment)
- KiloClawInstance calls App DO on first provision, caches flyAppName
- Proxy routes to per-user {flyAppName}.fly.dev with legacy fallback
- Docker images stay in shared FLY_REGISTRY_APP, referenced by all apps (which is great, means we just ship one image to `fly-machines`)
- Postgres restore recovers flyAppName from App DO (persists independently)

### DIFF
--- a/kiloclaw/.dev.vars.example
+++ b/kiloclaw/.dev.vars.example
@@ -11,15 +11,12 @@ WORKER_ENV=development
 # KILOCODE_API_BASE_URL=https://api.kilo.ai/api/openrouter/
 
 # Fly.io Machines API
-# Get an org-scoped token: fly tokens create org
+# Get a token (run: `fly tokens create some-name-for-this-dev-token`)
 FLY_API_TOKEN=fo1_...
 # Your Fly org slug (run: `fly orgs list`)
 FLY_ORG_SLUG=personal
 # Shared Docker image registry app (images are pushed here, referenced by all per-user apps)
 FLY_REGISTRY_APP=kiloclaw-machines-dev
-# Legacy fallback for existing instances without per-user apps.
-# New instances get per-user apps (acct-{hash}) created automatically.
-FLY_APP_NAME=kiloclaw-machines
 # Default region for new users. Comma-separated priority list of region codes or aliases.
 # Fly tries the first region, falls back to the next if unavailable.
 # Examples: "us,eu" (try US first, then Europe), "lhr,fra" (London first, then Frankfurt)
@@ -29,6 +26,10 @@ FLY_REGION=us,eu
 # In dev, use "latest" or run scripts/push-dev.sh which sets a timestamped tag.
 # In production, pin to a version (e.g., v40, sha-abc1234).
 FLY_IMAGE_TAG=latest
+
+# Legacy fallback for existing instances without per-user apps.
+# New instances get per-user apps (acct-{hash}) created automatically.
+FLY_APP_NAME=kiloclaw-machines
 
 # Allowed origins for the OpenClaw Control UI WebSocket (comma-separated).
 # Must include the origins that the browser connects from.
--- a/kiloclaw/.dev.vars.example
+++ b/kiloclaw/.dev.vars.example
@@ -13,7 +13,7 @@ WORKER_ENV=development
 # Fly.io Machines API
 # Get an org-scoped token: fly tokens create org
 FLY_API_TOKEN=fo1_...
-# Your Fly org slug (run: fly orgs list)
+# Your Fly org slug (run: `fly orgs list`)
 FLY_ORG_SLUG=personal
 # Shared Docker image registry app (images are pushed here, referenced by all per-user apps)
 FLY_REGISTRY_APP=kiloclaw-machines-dev
--- a/kiloclaw/.dev.vars.example
+++ b/kiloclaw/.dev.vars.example
@@ -11,15 +11,21 @@ WORKER_ENV=development
 # KILOCODE_API_BASE_URL=https://api.kilo.ai/api/openrouter/
 
 # Fly.io Machines API
-# Get your API token from: https://fly.io/docs/flyctl/tokens/
+# Get an org-scoped token: fly tokens create org
 FLY_API_TOKEN=fo1_...
+# Your Fly org slug (run: fly orgs list)
+FLY_ORG_SLUG=personal
+# Shared Docker image registry app (images are pushed here, referenced by all per-user apps)
+FLY_REGISTRY_APP=kiloclaw-machines-dev
+# Legacy fallback for existing instances without per-user apps.
+# New instances get per-user apps (acct-{hash}) created automatically.
 FLY_APP_NAME=kiloclaw-machines
 # Default region for new users. Comma-separated priority list of region codes or aliases.
 # Fly tries the first region, falls back to the next if unavailable.
 # Examples: "us,eu" (try US first, then Europe), "lhr,fra" (London first, then Frankfurt)
 # The platform API can override per-user at provision time.
 FLY_REGION=us,eu
-# Docker image tag on registry.fly.io/{FLY_APP_NAME}:{tag}.
+# Docker image tag on registry.fly.io/{FLY_REGISTRY_APP}:{tag}.
 # In dev, use "latest" or run scripts/push-dev.sh which sets a timestamped tag.
 # In production, pin to a version (e.g., v40, sha-abc1234).
 FLY_IMAGE_TAG=latest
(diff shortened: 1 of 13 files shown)


kiloclaw/src/routes/debug.ts
```diff
@@ -2,17 +2,38 @@ import { Hono } from 'hono';
...
+  userId: string | undefined
+): Promise<FlyClientConfig | null> {
+  if (!env.FLY_API_TOKEN) return null;
+
+  if (userId) {
+    // Look up from Instance DO first (may have a cached flyAppName)
+    const stub = env.KILOCLAW_INSTANCE.get(env.KILOCLAW_INSTANCE.idFromName(userId));
+    const status = await stub.getStatus();
+    const prefix = env.WORKER_ENV === 'development' ? 'dev' : undefined;
+    const appName = status.flyAppName ?? (await appNameFromUserId(userId, prefix));
```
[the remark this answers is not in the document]
pandemicsyn: its fine, debug routes getting nuked is my next pr

kiloclaw/src/durable-objects/kiloclaw-instance.ts
```diff
@@ -1212,6 +1241,18 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
...
+      // If the user has a per-user app, the App DO still knows about it.
+      // If both DOs were wiped, derive the app name deterministically so
+      // restore remains self-healing (the Fly app still exists on Fly's side).
+      // For legacy users who never had a per-user app, derivation produces a
+      // name that won't exist on Fly, but getFlyConfig() falls back to
+      // env.FLY_APP_NAME in that case since the derived app won't route.
+      const appStub = this.env.KILOCLAW_APP.get(this.env.KILOCLAW_APP.idFromName(userId));
+      const prefix = this.env.WORKER_ENV === 'development' ? 'dev' : undefined;
+      const recoveredAppName =
+        (await appStub.getAppName()) ?? (await appNameFromUserId(userId, prefix));
```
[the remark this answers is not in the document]
pandemicsyn: There's no legacy instances other than a few of our own.

kiloclaw/.dev.vars.example
```diff
@@ -11,15 +11,21 @@ WORKER_ENV=development
 # KILOCODE_API_BASE_URL=https://api.kilo.ai/api/openrouter/
 
 # Fly.io Machines API
-# Get your API token from: https://fly.io/docs/flyctl/tokens/
+# Get an org-scoped token: fly tokens create org
```
jeanduplessis: This reads weird: "fly tokens create org"?

jeanduplessis: Is this a CLI command? Maybe we can wrap it in code ticks then.

kiloclaw/.dev.vars.example
```diff
@@ -11,15 +11,21 @@ WORKER_ENV=development
...
 FLY_API_TOKEN=fo1_...
+# Your Fly org slug (run: fly orgs list)
```
jeanduplessis:
```suggestion
# Your Fly org slug (run: `fly orgs list`)
```

pandemicsyn:
Screenshot of this in action in my dev env

SCR-20260216-lply

got a little blur happy, but - thats a per user app.

jeanduplessis (approved):
Approving. Would be good to make the small improvements to the docs I flagged but can be in a follow-up as well.

eshurakov (approved):
looks good to me!

kiloclaw/.dev.vars.example
pandemicsyn: Yep, its a cli call, ill clean it up - also technically for the dev setup doesn't even have to be an org token. And probably actually shouldn't be one.

kiloclaw/src/fly/apps.ts
```diff
@@ -0,0 +1,147 @@
+/**
+ * Fly.io Apps + IP allocation REST API.
+ *
+ * Manages per-user Fly Apps: creation, existence checks, deletion,
+ * and IP address allocation (IPv4 shared + IPv6).
+ * All calls use the Machines REST API (https://api.machines.dev).
```
iscekic, line 6: I wonder if we could generate a ts client using their open api spec - would provide great dx.

kiloclaw/src/fly/apps.ts
```diff
@@ -0,0 +1,147 @@
...
+ *
+ * The hash portion is the same regardless of prefix, so you can compare
+ * across environments by stripping the prefix.
+ *
+ * @param prefix - Environment prefix (e.g. "dev" for WORKER_ENV=development). Omit for production.
+ */
+export async function appNameFromUserId(userId: string, prefix?: string): Promise<string> {
+  const encoder = new TextEncoder();
+  const data = encoder.encode(userId);
+  const hashBuffer = await crypto.subtle.digest('SHA-256', data);
```
iscekic, line 31: nit: depending on how often we call this, a faster hashing algorithm may be better?

iscekic (approved):

kiloclaw/src/fly/apps.ts
pandemicsyn: If things go ok, this will only ever be called once during initial provisioning. Only other spot is the "oh crap we accidentally deleted all the DO's so i need to try and rebuild" spot.

kiloclaw/src/index.ts
```diff
@@ -46,7 +47,8 @@ function validateRequiredEnv(env: KiloClawEnv): string[] {
   if (!env.NEXTAUTH_SECRET) missing.push('NEXTAUTH_SECRET');
   if (!env.GATEWAY_TOKEN_SECRET) missing.push('GATEWAY_TOKEN_SECRET');
   if (!env.FLY_API_TOKEN) missing.push('FLY_API_TOKEN');
-  if (!env.FLY_APP_NAME) missing.push('FLY_APP_NAME');
+  if (!env.FLY_ORG_SLUG) missing.push('FLY_ORG_SLUG');
```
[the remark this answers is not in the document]
pandemicsyn: > when FLY_ORG_SLUG/FLY_REGISTRY_APP are unset


fine/desirable - because that implies things have gone wrong.

kiloclaw/src/routes/debug.ts
```diff
@@ -2,17 +2,38 @@ import { Hono } from 'hono';
...
+async function resolveDebugFlyConfig(
+  env: AppEnv['Bindings'],
+  userId: string | undefined
+): Promise<FlyClientConfig | null> {
+  if (!env.FLY_API_TOKEN) return null;
+
+  if (userId) {
+    // Look up from Instance DO first (may have a cached flyAppName)
+    const stub = env.KILOCLAW_INSTANCE.get(env.KILOCLAW_INSTANCE.idFromName(userId));
+    const status = await stub.getStatus();
```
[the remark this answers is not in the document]
pandemicsyn: I keep forgetting that review bot has no memory, but - nuking debug routes in my pr

kiloclaw/src/durable-objects/kiloclaw-app.ts
```diff
@@ -0,0 +1,181 @@
...
+      // Partial state persisted above — arm a retry alarm so the DO self-heals
+      // even if the caller doesn't retry.
+      if (!this.ipv4Allocated || !this.ipv6Allocated) {
+        await this.ctx.storage.setAlarm(Date.now() + RETRY_ALARM_MS);
+        console.error('[AppDO] Partial failure, retry alarm armed for:', appName, err);
+      }
+      throw err;
+    }
+
+    return { appName };
```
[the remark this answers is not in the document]
pandemicsyn: id rather the recovery/alarm handler clear itself that have this accidentally proactively clear it.,

PR #252, opened 2026-02-16
Randomize the order of the churn reason array every time it's displayed

alexkgold:


src/components/profile/kilo-pass/KiloPassSubscriptionSettingsModal.tsx
```diff
@@ -261,19 +261,22 @@ export function KiloPassSubscriptionSettingsModal(props: SettingsModalProps) {
...
+      'Performance issues',
+      'Using another tool',
+    ];
+    // Fisher-Yates shuffle
+    for (let i = reasons.length - 1; i > 0; i--) {
+      const j = Math.floor(Math.random() * (i + 1));
+      [reasons[i], reasons[j]] = [reasons[j], reasons[i]];
+    }
+    return [...reasons, 'Other'];
+  }, [isOpen]);
```
iscekic: Ignore the bot, remove `isOpen` from the dep array and I think this'll be good to go. Reasons will be randomized on each modal open, but we likely don't care. 😄

iscekic (changes requested):

src/components/profile/kilo-pass/KiloPassSubscriptionSettingsModal.tsx
alexkgold:
```suggestion
  });
```

iscekic: `}, []);` is what you want.

iscekic (approved):

PR #244, opened 2026-02-16
feat: Add linkedin and githubto user profile and domain to organizations

alex-alecu:
### User profile: LinkedIn & GitHub URLs

Adds `linkedin_url` and `github_url` columns to users. The profile page now shows these links (or "Not set" placeholders) and has an edit button that opens a dialog to update them.

- New [`UserProfileCard`](src/components/profile/UserProfileCard.tsx) component replaces the inline avatar/email markup on the profile page
- New [`EditProfileDialog`](src/components/profile/EditProfileDialog.tsx) with client-side URL validation
- New [`user.updateProfile`](src/routers/user-router.ts:282) tRPC mutation — validates URLs are `https?://`, rejects `javascript:` etc.
- Tests in [`user-router.test.ts`](src/routers/user-router.test.ts)

### Organization: company domain

Adds a `company_domain` column to organizations. Can be set during org creation or updated afterward by org owners.

- [`CompanyDomainSchema`](src/lib/organizations/company-domain.ts:55) normalizes input (accepts bare domains or full URLs, extracts hostname) and validates domain format
- [`organizations.updateCompanyDomain`](src/routers/organizations/organization-router.ts:174) — owner-only mutation, supports clearing with `null` or `""`
- Domain field added to [`CreateOrganizationPage`](src/components/organizations/new/CreateOrganizationPage.tsx) and the [`OrganizationInfoCard`](src/components/organizations/OrganizationInfoCard.tsx) admin view
- Tests for the domain normalization logic and the router mutations in [`company-domain.test.ts`](src/lib/organizations/company-domain.test.ts) and [`organization-router.test.ts`](src/routers/organizations/organization-router.test.ts)
### Migration

[`0013_yielding_whiplash.sql`](src/db/migrations/0013_yielding_whiplash.sql) — adds `linkedin_url`, `github_url` to `kilocode_users` and `company_domain` to `organizations` (all nullable text columns, no data backfill needed).

### Demo

https://github.com/user-attachments/assets/df358974-3487-4671-8894-c979446f9695

PR #273, opened 2026-02-17
Add unicode/IDN domain support

alex-alecu:
This is a followup to https://github.com/Kilo-Org/cloud/pull/244#discussion_r2813324801

Added support for domains like `münchen.de`

https://github.com/user-attachments/assets/36534a8d-8391-4799-b51a-dbce8ef9d585


src/lib/organizations/company-domain.ts
```diff
@@ -4,43 +4,51 @@ import * as z from 'zod';
...
+ *   "münchen.de" → "münchen.de"
  *   "  " → null
  *   "" → null
  */
 export function normalizeCompanyDomain(input: string): string | null {
   const trimmed = input.trim();
   if (!trimmed) return null;
 
-  // If it looks like a URL (has ://), try to parse it
+  let hostname = trimmed;
```
marius-kilocode: What are we storing now punycode or unicode? This and the previous pr just makes this human readable right?

alex-alecu: We are now storing `unicode` in the db

src/lib/organizations/company-domain.ts
```diff
@@ -4,47 +4,46 @@ import * as z from 'zod';
...
+ * and domainToUnicode() from node:url isn't available in the browser.
  */
 export function normalizeCompanyDomain(input: string): string | null {
   const trimmed = input.trim();
   if (!trimmed) return null;
 
-  // If it looks like a URL (has ://), try to parse it
-  if (trimmed.includes('://')) {
-    try {
-      const url = new URL(trimmed);
```
marius-kilocode, line 22: I don't think this is a good tradeoff. Why would we replace `URL`? It probably makes much more sense to parse with URL and then decode punycode back instead?

marius-kilocode (changes requested):

src/lib/organizations/company-domain.ts
alex-alecu: @marius-kilocode the simple short solution is to use [punycode](https://www.npmjs.com/package/punycode) like so:

```tsx
import { toUnicode } from 'punycode/';

export function normalizeCompanyDomain(input: string): string | null {
  const trimmed = input.trim();
  if (!trimmed) return null;

  const urlString = trimmed.includes('://') ? trimmed : `https://${trimmed}`;

  try {
    const url = new URL(urlString);
    // URL punycode-encodes IDN hostnames; decode back to unicode
    const hostname = toUnicode(url.hostname);
    return hostname || null;
  } catch {
    return null;
  }
}
```

Is it worth adding a new lib to be used a single time?

marius-kilocode (approved):

src/lib/organizations/company-domain.ts
marius-kilocode: As you prefer, I guess it's internal anyway so there are not too many risks

PR #275, opened 2026-02-17
Improved support for non OpenRouter providers

chrarnoldus:
* CoreThink: delete incompatible parameters, rewrite response to be compatible with OpenCode
* Grok Code Optimized: delete incompatible parameters, implement preserved reasoning

### DIFF
--- a/src/lib/providers/index.ts
+++ b/src/lib/providers/index.ts
@@ -90,7 +90,7 @@ export const PROVIDERS = {
     apiUrl: 'https://api.mistral.ai/v1',
     apiKey: getEnvVariable('MISTRAL_API_KEY'),
     hasGenerationEndpoint: false,
-    requiresResponseRewrite: true,
+    requiresResponseRewrite: false,
   },
   MINIMAX: {
     id: 'minimax',
--- a/src/lib/providers/index.ts
+++ b/src/lib/providers/index.ts
@@ -46,6 +46,7 @@ export type Provider = {
   apiUrl: string;
   apiKey: string;
   hasGenerationEndpoint: boolean;
+  requiresResponseRewrite: boolean;
 };
 
 export const PROVIDERS = {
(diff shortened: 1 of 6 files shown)


src/lib/rewriteModelResponse.ts
```diff
@@ -1,17 +1,42 @@
...
 import { NextResponse } from 'next/server';
 import type OpenAI from 'openai';
 
-export async function redactedModelResponse(response: Response, model: string) {
+function convertReasoningToOpenRouterFormat(message: MessageWithReasoning) {
+  if (!message.reasoning_content) {
+    return;
+  }
+  if (!message.reasoning) {
+    message.reasoning = message.reasoning_content;
```
markijbema, line 14: Why not do this with `||=` or something instead of ifs?

markijbema (approved):

src/lib/rewriteModelResponse.ts
```diff
@@ -1,17 +1,42 @@
...
+  }
+  if (!message.reasoning_details) {
```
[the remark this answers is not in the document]
chrarnoldus: I don't think an upstream will do this.

src/lib/rewriteModelResponse.ts
chrarnoldus: I just didn't think of that

src/lib/providers/index.ts
```diff
@@ -54,60 +55,70 @@ export const PROVIDERS = {
...
     apiKey: getEnvVariable('MARTIAN_API_KEY'),
     hasGenerationEndpoint: false,
+    requiresResponseRewrite: true,
   },
   MISTRAL: {
     id: 'mistral',
     apiUrl: 'https://api.mistral.ai/v1',
     apiKey: getEnvVariable('MISTRAL_API_KEY'),
     hasGenerationEndpoint: false,
+    requiresResponseRewrite: true,
```
chrarnoldus:
```suggestion
    requiresResponseRewrite: false,
```

PR #299, opened 2026-02-17
fix(kiloclaw): detect app name hash collisions on Fly app creation

St0rmz1:
On 409 (app already exists), verify ownership by listing machines and checking userId metadata. 
Throws AppNameCollisionError if the existing app belongs to a different user, preventing silent network namespace sharing from SHA-256 truncation collisions.

Fails open if machine listing errors to preserve existing retry behavior. Legacy machines without metadata are treated as safe (no false positives).


pandemicsyn (approved):
🚢

PR #307, opened 2026-02-17
feat(usage): add feature attribution to microdollar usage

pedroheyerdahl:
Adds a nullable `feature` column to `microdollar_usage` so we can track which product feature generated each token usage record. This is the backend portion of a cross-repo change.

**What this does:**
- Adds `feature` column to `microdollar_usage` table and `microdollar_usage_view`
- Creates `src/lib/feature-detection.ts` with an allow-list of 14 feature values and a `validateFeatureHeader()` function
- Both gateway endpoints (`/api/openrouter` and `/api/fim/completions`) read the new `X-KILOCODE-FEATURE` header and store the validated value
- `sendProxiedChatCompletion` accepts an optional `feature` field and forwards it as a header
- Security agent and slack bot callers send their respective feature values
- `cloud-agent-next` server manager accepts an optional `feature` parameter to set `KILOCODE_FEATURE` env var when launching `kilo serve`

**How it works:**
Callers send `X-KILOCODE-FEATURE: <value>`. The gateway validates against the allow-list and stores it. No header = NULL. No fallback heuristics in the write path.

**Companion changes needed (separate PRs):**
- `Kilo-Org/kilo`: Add `HEADER_FEATURE` constant to kilo-gateway, read `KILOCODE_FEATURE` env var
- `Kilo-Org/kilocode`: Add `X-KiloCode-Feature` header to `customRequestOptions()` and `streamFim()`

See `plans/microdollar-feature-tracking.md` for the full implementation plan.

---


cloud-agent-next/src/kilo/server-manager.ts
```diff
@@ -209,10 +209,11 @@ export async function findAvailablePort(
  * @param port - The port to listen on
  * @returns The full command string
  */
-export function buildKiloServeCommand(sessionId: string, port: number): string {
+export function buildKiloServeCommand(sessionId: string, port: number, feature?: string): string {
```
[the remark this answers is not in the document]
pedroheyerdahl: cloud-agent-nextcan't import from src/lib/feature-detection.ts. Keeping as string here. The value is validated at gateway before storage, no injection risk in practice

src/lib/llm-proxy-helpers.ts
```diff
@@ -625,6 +625,8 @@ export type ProxiedChatCompletionRequest = {
   userAgent: string;
   body: OpenRouterChatCompletionRequest;
   organizationId?: string;
+  /** Feature attribution value for microdollar usage tracking. */
+  feature?: string;
```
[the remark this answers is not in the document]
pedroheyerdahl: applied

cloud-agent-next/src/kilo/server-manager.ts
```diff
@@ -235,7 +236,8 @@ export async function ensureKiloServer(
   sandbox: SandboxInstance,
   session: ExecutionSession,
   sessionId: string,
-  workspacePath: string
+  workspacePath: string,
+  feature?: string
```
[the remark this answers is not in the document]
pedroheyerdahl: cloud-agent-nextcan't import from src/lib/feature-detection.ts. Keeping as string here. The value is validated at gateway before storage, no injection risk in practice

src/lib/llm-proxy-helpers.ts
```diff
@@ -649,6 +651,10 @@ export async function sendProxiedChatCompletion<T>(
     headers.set('X-KiloCode-OrganizationId', request.organizationId);
   }
 
+  if (request.feature) {
+    headers.set('X-KiloCode-Feature', request.feature);
```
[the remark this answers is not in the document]
pedroheyerdahl: applied

cloud-agent-next/src/kilo/server-manager.ts
```diff
@@ -278,7 +280,7 @@ export async function ensureKiloServer(
 
   for (let attempt = 0; attempt < MAX_PORT_RETRY_ATTEMPTS; attempt++) {
     const port = await findAvailablePort(sandbox, sessionId);
-    const command = buildKiloServeCommand(sessionId, port);
+    const command = buildKiloServeCommand(sessionId, port, feature);
```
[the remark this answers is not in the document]
pedroheyerdahl: Intentional. The callers need to pass the feature value based on what initiated the session (cloud-agent, code-review, auto-triage, etc.), which comes from the upstream Cloudflare workers. Updating the callers is tracked in the implementation plan (Step 8) and will be done as a follow-up once the feature identity propagation through the session creation flow is worked out.

src/app/api/gateway/[...path]/route.ts
```diff
@@ -1 +1,8 @@
-export { POST } from '@/app/api/openrouter/[...path]/route';
+import type { NextRequest } from 'next/server';
+import { POST as openrouterPOST } from '@/app/api/openrouter/[...path]/route';
+import { FEATURE_HEADER } from '@/lib/feature-detection';
+
+export function POST(request: NextRequest) {
+  request.headers.set(FEATURE_HEADER, 'direct-gateway');
```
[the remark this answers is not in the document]
Comment (pedroheyerdahl) [reply src/app/api/gateway/[...path]/route.ts]: applied

src/app/api/gateway/[...path]/route.ts
```diff
@@ -1 +1,15 @@
-export { POST } from '@/app/api/openrouter/[...path]/route';
+import { NextRequest } from 'next/server';
```
Comment (pedroheyerdahl) [inline src/app/api/gateway/[...path]/route.ts]: @jrf0110 I changed the gateway route from a re-export to a thin wrapper that clones the request with X-KILOCODE-FEATURE: direct-gateway before forwarding to the openrouter handler. This lets us positively identify external gateway consumers in microdollar_usage without depending on them to send any headers. Tested locally and confirmed the feature value is stored correctly. Any concerns with this approach or the request cloning pattern?

Image

Comment (jrf0110) [reply src/app/api/gateway/[...path]/route.ts]: I'm not so sure about relying on that the route. There are currently direct-gateway users that use the openrouter endpoint

Comment (pedroheyerdahl) [reply src/app/api/gateway/[...path]/route.ts]: @jrf0110 Thanks for flagging. Is it actually intended behavior or is that just where we ended up?
If external consumers can hit both /api/gateway/ and /api/openrouter/, we lose a clean way to identify direct gateway traffic. Ideally, each feature should have its own positive signal so we don’t end up with misattribution or silent failure modes.
Can we enforce separation moving forward? If not, what’s the right way to make direct gateway traffic reliably identifiable without relying on a catch all fallback from excluding all other known internal features signals?

Comment (jrf0110) [reply src/app/api/gateway/[...path]/route.ts]: > Is it actually intended behavior or is that just where we ended up?

I suppose it's both. I think we never should have had our gateway route at /api/openrouter, but that's the way it was. Clients (Kilo or otherwise) hit the gateway all the same. IMO kilo feature clients should hit the gateway with some identifying header (e.g. x-kilocode-feature). Of course this is spoofable, so it should only be used for reporting purposes

Comment (chrarnoldus) [reply src/app/api/gateway/[...path]/route.ts]: I agree we should not introduce behavioral differences between the endpoints (and we should get rid of the poorly named openrouter endpoint eventually).

src/lib/feature-detection.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const featureSet = new Set<string>(FEATURE_VALUES);
+
+function isFeatureValue(value: string): value is FeatureValue {
+  return featureSet.has(value);
+}
+
+export const FEATURE_HEADER = 'x-kilocode-feature';
+
+export function validateFeatureHeader(headerValue: string | null): FeatureValue | null {
```
chrarnoldus, line 39: using zod would be the typical way of parsing structured data

src/app/api/gateway/[...path]/route.ts
Comment (pedroheyerdahl) [reply src/app/api/gateway/[...path]/route.ts]: > kilo feature clients should hit the gateway with some identifying header 

That's exactly what this PR along with others in kilo and kilocode repo is doing. The open question is specifically about direct-gateway. It's a feature that needs a positive signal just like the others. Since we can't rely on external consumers to send the header, how should we identify their traffic? I'd appreciate your help defining that so we don't end up using "no header" as a catch-all, which would mask tracking issues.

src/lib/feature-detection.ts
pedroheyerdahl: accepted

src/app/api/gateway/[...path]/route.ts
Comment (pedroheyerdahl) [reply src/app/api/gateway/[...path]/route.ts]: reverted the wrapper. will create an issue for direct-gateway usage attribution to unblock the current work

chrarnoldus (approved):

PR #311, opened 2026-02-18
feat(kiloclaw): Telegram bot token config + channel pairing approval

pandemicsyn:
## Summary

- Add Telegram bot token input to the CreateInstanceCard (at provision time) and SettingsTab (post-provision update with save/remove)
- Add channel pairing approval UI — a "Pairing Requests" card on the instance page that lists pending requests and lets users approve them
- Wire the full stack: frontend → tRPC → internal client → worker platform routes → DO methods → Fly Machines exec API → `openclaw pairing list/approve` CLI

### Telegram Token Configuration
- Masked input with eye toggle, `data-1p-ignore` + `autoComplete=off` to prevent 1Password interference
- `updateChannels` DO method for post-provision channel token updates (PATCH /api/platform/channels)
- Supports add, replace, and remove (pass null) for any channel token

### Channel Pairing Approval
- Uses Fly Machines exec API to run `openclaw pairing list` (via wrapper) and `openclaw pairing approve` on the running machine
- `openclaw-pairing-list.js` helper script on the Docker image aggregates pairing requests across all configured channels in a single exec call
- Results cached in KV (`KV_CLAW_CACHE`) to minimize Fly exec calls
-120s background polling + manual refresh button (bypasses KV cache) for the frontend.
- Command injection prevention: channel/code validated against strict regex before shell interpolation

### Startup Script Changes
- **Right now we do a full channel object overwrite** with `dmPolicy` + `allowFrom`, in the future we probably need to try and just patch entries we know about. 
- Set `plugins.entries.telegram.enabled = true` when configuring the channel (fixes `openclaw onboard --skip-channels` disabling the plugin)
- `HOME=/root` prefix on exec commands (Fly exec sets `HOME=/`)

### Infrastructure
- New `KV_CLAW_CACHE` KV namespace for pairing list cache
- New `execCommand()` in fly/client.ts for the Fly Machines exec API
- New `ChannelsPatchSchema` for channel update validation


pandemicsyn:
A "known defect" i wanna flag. when you delete a telegram token , we don't go and remove it from the deployed config quite yet. Trying to figure out how we're going to safely do that or if we should even support removal only the fly at this stage of the EA.

kiloclaw/openclaw-pairing-list.js
pandemicsyn, line 1: At least in this version, openclaw doesn't let you list all pairing requests, and so that we don't have to call fly machine exec a bunch of times, we just call this script instead. 

Guessing this sort of pattern will become more common until we cave and do a C&C sidecar process.

kiloclaw/start-openclaw.sh
```diff
@@ -253,6 +253,8 @@ config.tools.exec.security = 'deny';
 config.tools.exec.ask = 'on-miss';
 
 // Telegram configuration
+// Overwrite entire channel object to drop stale keys that would fail
+// OpenClaw's strict config validation (matches moltworker behavior)
```
pandemicsyn, line 257: I spent some time trying to do a patch pattern instead - but in this early version, if a user is hacking up the openclaw config manually - we're just gonna clobber their changes.

kiloclaw/src/durable-objects/kiloclaw-instance.ts
```diff
@@ -438,6 +438,210 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
...
+      return { success: false, message: 'Invalid channel name' };
+    }
+    if (!/^[A-Za-z0-9]{1,32}$/.test(code)) {
+      return { success: false, message: 'Invalid pairing code' };
+    }
+
+    const result = await fly.execCommand(
+      flyConfig,
+      flyMachineId,
+      ['/bin/bash', '-c', `HOME=/root openclaw pairing approve ${channel} ${code} --notify`],
```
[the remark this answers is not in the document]
pandemicsyn: kinda fixed in the follow up, kinda not true.

St0rmz1:
A few suggestions to consider:

1. openclaw-pairing-list.js: Use execFileSync instead of execSync

```
  // Current — shell interpolation
  const output = execSync(`openclaw pairing list ${channel} --json`, { ... });
```

  ```
// Recommended — no shell involved
  const output = execFileSync('openclaw', ['pairing', 'list', channel, '--json'], { ... });
```

The channel values currently come from a hardcoded set so risk is low, but execFileSync eliminates the class of risk entirely for free.

2. kiloclaw-instance.ts — Possibly greedy regex in JSON parsing

`  const jsonMatch = result.stdout.match(/\{[\s\S]*\}/);
`

 The greedy [\s\S]* matches from the first { to the last } in stdout. If openclaw ever prints a log line containing braces before the JSON output, this produces invalid JSON. Use non-greedy (/\{[\s\S]*?\}/) or anchor to the expected shape (/\{"requests"[\s\S]*\}/). Same issue exists in openclaw-pairing-list.js.

 3. kiloclaw-instance.ts — Potential broad success condition in approvePairingRequest

`  const success = result.exit_code === 0 || result.stdout.toLowerCase().includes('approved');
`
If the command fails but output contains "approved" (e.g., "Error: already approved", "could not be approved"), this is incorrectly treated as success and busts the cache. Can we use `exit_code === 0`.

St0rmz1 (approved):

PR #319, opened 2026-02-18
Make Sonnet 4.6 default reviewer and make it free for one week in review mode

alex-alecu:
## What

1. Make sonnet 4.6 default reviewer - is already in `https://app.kilo.ai/organizations/.../code-reviews`. Has same costs as sonnet 4.5 [per docs](https://platform.claude.com/docs/en/about-claude/pricing)
2. Time-boxed promotion (2026-02-18T14:00Z → 2026-02-25T14:00Z) making `sonnet 4.6` free for automated code reviews. After the window closes, the promo silently stops — no cleanup needed.

### How it works

A single predicate `isActiveReviewPromo(botId, model)` is the sole source of truth. It returns `true` only when all three conditions hold:

1. `botId === 'reviewer'` (set in the JWT by `prepare-review-payload.ts`)
2. `model === 'anthropic/claude-sonnet-4.6'`
3. Current time is before the promo end date

This predicate is checked in two places:
- **Balance gate** (`route.ts`): skips the 402 "insufficient credits" rejection
- **Cost recording** (`processUsage.ts`): zeros `cost_mUsd` so the user isn't charged

### Changes

**Auth stack — thread `botId` from JWT to usage context:**
- `tokens.ts`: return `botId` from `validateAuthorizationHeader` (was missing despite being in `JWTTokenExtraPayload`)
- `user.server.ts`: add `botId` to `GetAuthResponse` type and thread through `getUserFromAuth` → `validateUserAuthorization`
- `route.ts`: destructure `botId` from auth, add to `usageContext`, add `isActiveReviewPromo` to balance gate condition

**Promo logic:**
- `constants.ts`: add `REVIEW_PROMO_MODEL`, `REVIEW_PROMO_START`, `REVIEW_PROMO_END`, and `isActiveReviewPromo()` predicate. Update `DEFAULT_CODE_REVIEW_MODEL` to `anthropic/claude-sonnet-4.6`.
- `processUsage.ts`: add `botId?: string` to `MicrodollarUsageContext`; extend cost-zeroing condition with `isActiveReviewPromo`
- `prepare-review-payload.ts`: add promo logging when `isActiveReviewPromo` returns true

**Admin visibility:**
- `admin-code-reviews-router.ts`: add `getReviewPromotionStats` query — aggregate + daily breakdown from `microdollar_usage` joined with `microdollar_usage_metadata`, filtered to zero-cost non-BYOK sonnet-4.6 requests within the promo window

**Integration tests:**
- `test-sonnet-46-review-promo.ts`: 8 tests covering promo accepted, cost zeroed, non-promo model not free, wrong botId, no botId, balance gate bypass with zero balance, `isActiveReviewPromo` unit test, admin stats query shape validation

### What this does NOT change

- No new model slugs, no `KiloFreeModel` entries, no DB migrations
- No changes to `isFreeModel`/`kiloFreeModels`/free model rate limiter
- No changes to model picker UI or FIM route
- `prepare-review-payload.ts` JWT creation already sets `botId: 'reviewer'` — unchanged

### Testing

```
pnpm script src/scripts/test-sonnet-46-review-promo.ts
```

### DIFF
--- a/src/lib/code-reviews/core/constants.ts
+++ b/src/lib/code-reviews/core/constants.ts
@@ -8,16 +8,29 @@
 // Review Configuration
 // ============================================================================
 
-/**
- * Default model for code reviews
- * Using Claude Sonnet 4.5 as specified in the plan
- */
-export const DEFAULT_CODE_REVIEW_MODEL = 'anthropic/claude-sonnet-4.5';
+/** Default model for code reviews */
+export const DEFAULT_CODE_REVIEW_MODEL = 'anthropic/claude-sonnet-4.6';
 
 /**
  * Default mode for cloud agent sessions
  */
 export const DEFAULT_CODE_REVIEW_MODE = 'code' as const;
+
+// ============================================================================
+// Sonnet 4.6 Review Promotion
+// ============================================================================
+
+export const REVIEW_PROMO_MODEL = 'anthropic/claude-sonnet-4.6';
+export const REVIEW_PROMO_START = '2026-02-18T14:00:00Z'; // used only for admin logging
+export const REVIEW_PROMO_END = '2026-02-25T14:00:00Z';
+
+/** Single source of truth: is the free-review promo active for this request? */
+export function isActiveReviewPromo(botId: string | undefined, model: string): boolean {
+  if (botId !== 'reviewer') return false;
+  if (model !== REVIEW_PROMO_MODEL) return false;
+
+  return Date.now() < Date.parse(REVIEW_PROMO_END);
+}
 
 // ============================================================================
 // Pagination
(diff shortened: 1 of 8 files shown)


src/lib/code-reviews/core/constants.ts
```diff
@@ -8,17 +8,30 @@
 // Review Configuration
 // ============================================================================
 
-/**
- * Default model for code reviews
- * Using Claude Sonnet 4.5 as specified in the plan
- */
-export const DEFAULT_CODE_REVIEW_MODEL = 'anthropic/claude-sonnet-4.5';
+/** Default model for code reviews */
+export const DEFAULT_CODE_REVIEW_MODEL = 'anthropic/claude-sonnet-4.6';
```
chrarnoldus, line 12: you could replace this with the current sonnet constant that's defined somewhere

chrarnoldus (approved):

src/lib/code-reviews/core/constants.ts
alex-alecu: I though about that, but I think it's safer this way and follows the previous pattern. What do you think?

chrarnoldus: I think we want it to stay in sync with the latest version

PR #324, opened 2026-02-18
Sonnet 4.6 is free, but users still see the upstream cost

alex-alecu:
Followup to https://github.com/Kilo-Org/cloud/pull/319

Sonnet 4.6 is free for code review, but users still see the upstream cost.


chrarnoldus:
due to the potential performance impact and potential for bugs of rewriting the response stream I would prefer if the consuming code is modified to not show the cost instead.

PR #327, opened 2026-02-18
Auto triage - Adding labels and fixing edge cases

jfcsantos:
## What this does

This PR extends the auto-triage pipeline with four major capabilities and hardens the existing implementation against several reliability and security issues found in review.

#### AI-driven label selection
The classification prompt now includes the repository's actual GitHub labels (fetched via the GitHub API with full pagination, up to 500 labels). The model selects zero or more labels from this list, which are applied to the issue alongside the fixed kilo-triaged / kilo-auto-fix tracking labels. Labels used for pipeline gating (skip_labels, required_labels) are excluded from the list offered to the AI so it cannot interfere with existing automation.
#### Duplicate detection comments
When an issue is identified as a duplicate, a comment is now posted on the GitHub issue linking to the canonical duplicate, including its title, issue number, and similarity score.

#### Ticket interrupt
Users can now interrupt a ticket that is pending or analyzing without waiting for it to time out. The DB update uses a status-guard in the WHERE clause to avoid overwriting a ticket that was concurrently completed by the orchestrator (TOCTOU protection). Interrupting a ticket frees the concurrency slot and immediately dispatches the next pending ticket.

#### Durable Object alarm recovery
A Cloudflare DO alarm is now set when triage begins. If the DO is evicted, restarted, or the triage simply hangs, the alarm fires and marks the ticket failed, preventing it from getting stuck in analyzing indefinitely.

### Reliability & correctness fixes

- Partial label failure detection — the add-label API returns HTTP 207 for partial failures; applyLabels now reads the response body to detect this since response.ok is true for all 2xx responses.
- Label API pagination — the GitHub labels endpoint is now paginated (following Link headers) rather than capped at the first 100 results. A safety limit of 5 pages (500 labels) prevents runaway fetches.
- Correct duplicate comment target — closeDuplicate now looks up the similarTickets entry matching duplicateOfTicketId rather than always using index [0], which could point to a different ticket.
- updateStatus error handling — if the Next.js status-update API is unreachable, the orchestrator now falls back to writing terminal state directly to DO storage rather than leaving the ticket in a non-terminal state.
- Timeout timer leak — the classification timeout setTimeout is now always cleared in a finally block regardless of whether the stream succeeds or fails.
- Timeout error message consistency — error message checks updated from loose .includes('timeout') substrings to the exact messages produced by the timeout rejection.

#### Prompt injection hardening
- Issue title and body were already XML-delimited. custom_instructions (operator config) was previously appended raw after the output-format section, where it could override the JSON schema. It is now inserted before ## Output format, wrapped in <custom_instructions> delimiters, with an explicit instruction not to let it override the output contract.
- Label values returned by the model are validated against the availableLabels allowlist before schema validation, preventing hallucinated label names from reaching the GitHub API.
- Control characters are stripped from the agent response before parsing to prevent interference with JSON extraction.

### UI
- Interrupt button added to pending/analyzing tickets in AutoTriageTicketsCard. Per-ticket loading state is tracked individually so interrupting one ticket doesn't disable buttons on others.
- Threshold sliders and custom instructions fields hidden from the config form (temporarily, pending UX decisions).


plans/auto-triage-bugs.md
catrielmuller, line 1: Remove the plans

plans/auto-triage-test-plan.md
catrielmuller, line 1: Remove the plans

src/components/auto-triage/AutoTriageConfigForm.tsx
```diff
@@ -359,7 +359,7 @@ export function AutoTriageConfigForm({ organizationId }: AutoTriageConfigFormPro
             </div>
 
             {/* Thresholds */}
-            <div className="space-y-6">
+            {/* <div className="space-y-6">
```
catrielmuller: is this expected?

catrielmuller (changes requested):

catrielmuller (approved):

PR #351, opened 2026-02-19
chore: Bump vercel routing to 100% for openrouter incident

jrf0110:
Ongoing openrouter incident:

image


markijbema (approved):
What do we have to lose?

PR #374, opened 2026-02-19
Cloud Agent - Add interactive question answering and rejecting

eshurakov:
## Summary

- Add `answerQuestion` and `rejectQuestion` tRPC mutations in the cloud-agent-next worker, with new session-questions handler
- Fix `answers` type from `string[]` to `string[][]` across the full wrapper chain (wrapper server, kilo-client, wrapper-client, tests) to match the Kilo serve API contract
- Add personal and organization proxy endpoints with Zod validation schemas
- Track `question.asked` SSE events to map tool `callID` → `requestId`, enabling the UI to target the correct question
- Rewrite `QuestionToolCard` with an interactive UI: selectable options (single/multi), custom text input, submit and dismiss buttons, loading/error states
- Add `questionRequestIdsAtom` and `sessionOrganizationIdAtom` to Jotai store; wire them through `useCloudAgentStream` and clear on session reset

### DIFF
--- a/cloud-agent-next/src/router/handlers/session-questions.ts
+++ b/cloud-agent-next/src/router/handlers/session-questions.ts
@@ -1,0 +1,156 @@
+import { TRPCError } from '@trpc/server';
+import * as z from 'zod';
+import { getSandbox } from '@cloudflare/sandbox';
+import { logger, withLogTags } from '../../logger.js';
+import { generateSandboxId } from '../../sandbox-id.js';
+import type { SessionId, SandboxId } from '../../types.js';
+import { SessionService, fetchSessionMetadata } from '../../session-service.js';
+import { protectedProcedure } from '../auth.js';
+import { sessionIdSchema } from '../schemas.js';
+import { findWrapperForSession } from '../../kilo/wrapper-manager.js';
+import { WrapperClient } from '../../kilo/wrapper-client.js';
+
+export function createSessionQuestionHandlers() {
+  return {
+    answerQuestion: protectedProcedure
+      .input(
+        z.object({
+          sessionId: sessionIdSchema,
+          questionId: z.string().min(1),
+          answers: z.array(z.array(z.string())),
+        })
+      )
+      .mutation(async ({ input, ctx }) => {
+        return withLogTags({ source: 'answerQuestion' }, async () => {
+          const sessionId = input.sessionId as SessionId;
+          const { userId, env } = ctx;
+
+          logger.setTags({ userId, sessionId });
+          logger.info('Answering question', { questionId: input.questionId });
+
+          try {
+            const metadata = await fetchSessionMetadata(env, userId, sessionId);
+            if (!metadata) {
+              throw new TRPCError({ code: 'NOT_FOUND', message: 'Session not found' });
+            }
+
+            const sandboxId: SandboxId = await generateSandboxId(
+              metadata.orgId,
+              userId,
+              metadata.botId
+            );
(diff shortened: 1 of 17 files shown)


src/hooks/useNotificationSound.ts
```diff
@@ -0,0 +1,69 @@
+'use client';
+
+import { useCallback, useRef } from 'react';
+import { useLocalStorage } from '@/hooks/useLocalStorage';
+import { SOUND_NOTIFICATIONS_KEY } from '@/hooks/useCelebrationSound';
+
+/**
+ * Play a single tone with exponential decay.
+ */
+function playTone(
```
marius-kilocode, line 10: Wouldn't we hook notifications to sounds, instead of hardcoding questions to play a sound?
So that question can (or any other question to the user triggers a notification) and then this can be configured. Hardcoding it here seems conceptually wrong.

src/routers/organizations/organization-cloud-agent-next-router.ts
```diff
@@ -57,6 +59,14 @@ const GetSessionInput = baseGetSessionNextSchema.extend({
   organizationId: z.uuid(),
 });
 
+const AnswerQuestionInput = baseAnswerQuestionNextSchema.extend({
+  organizationId: z.uuid(),
```
marius-kilocode, line 63: why are we generating here a new organizationId?

cloud-agent-next/src/router/handlers/session-questions.ts
```diff
@@ -0,0 +1,156 @@
...
+              env,
+              ctx.authToken,
+              metadata.orgId
+            );
+
+            const wrapperClient = new WrapperClient({ session, port: wrapperInfo.port });
+            const result = await wrapperClient.rejectQuestion(input.questionId);
+
+            return { success: result.success };
+          } catch (error) {
```
marius-kilocode, line 118: the try catch is huge, and I don't understand the three layer TRPCError. Would it not make more sense to split this and handle the error individually? Same for answerQuestion

src/hooks/useNotificationSound.ts
eshurakov: @marius-kilocode Not sure what you mean here. We get an event for the question and this triggers a sound. There is also a sound when session is completed that is triggered in a similar way. We can introduce a notification layer that would react on events, but that is an easy refactoring that can be done later.

Could you elaborate on how you think it should be?

src/routers/organizations/organization-cloud-agent-next-router.ts
eshurakov: that's zod schema

cloud-agent-next/src/router/handlers/session-questions.ts
eshurakov: I've extracted common code from the "try" body.

> three layer TRPCError

Not sure I get this.

In the catch block if an error that we get already TRPCError, we throw it, otherwise we create a new TRPCError with error message.

src/hooks/useNotificationSound.ts
marius-kilocode: Yes, this is exactly what I mean. In theory the underlaying CLI not only requires the attention of the user for questions, so my argument is that we should have a general notification system, where sounds are feature of the notification. Would it be such a big change todo that? Couldn't we move the sounds notification logic and the question notification logic together?

marius-kilocode (approved):

PR #388, opened 2026-02-19
Recover from Fly 412 insufficient resources by replacing stranded volumes

pandemicsyn:
## Summary

- When a Fly volume's host has no capacity for a new machine (412 error), automatically recover by replacing the volume in a different region
- For existing instances with user data: fork the volume to preserve workspace files; if fork fails, surface the error (no silent data loss)
- For fresh provisions (never started): delete and recreate volume in a different region
- Narrow 412 detection to only match capacity-related errors, not version/precondition mismatches (`min_secrets_version`, `machine_version`, etc.)
- Pass compute hints (cpu/memory spec) when creating volumes so Fly picks a host with capacity for the expected machine
- Walk the region list explicitly on volume creation — Fly doesn't support meta-regions for volumes, so on capacity 412 we try the next region
- Add pre-flight volume region validation on `start()` to catch `flyRegion` drift before machine creation
- Unify fallback region strings into `DEFAULT_FLY_REGION` constant

## Context

Users hitting `FlyApiError: Fly API createMachine failed (412): {"error":"insufficient resources to create new machine with existing volume"}` — the physical host where their volume is pinned ran out of capacity. Previously this was unrecoverable without manual intervention.

Fly support recommended passing `--vm-size`/`--vm-cpus`/`--memory` when creating volumes so the scheduler picks a host with capacity for both the volume and the machine. They also confirmed meta-regions aren't supported for volume creation, so we walk the region list ourselves.

## 412 Classification

Fly overloads 412 for both capacity exhaustion and precondition/version mismatches. We only trigger recovery for capacity signals:

| Fly 412 body | Classified as | Recovery? |
|---|---|---|
| `"insufficient resources to create new machine with existing volume 'vol_xxx'"` | Capacity | Yes — fork/replace volume |
| `"min_secrets_version 3 is not yet available"` | Precondition | No — re-thrown |
| `"machine_version mismatch"` | Precondition | No — re-thrown |
| Unknown 412 | Unclassified | No — re-thrown + warning logged |

## Volume Creation Strategy

All volume creation now goes through `createVolumeWithFallback` which:
1. Accepts an explicit list of regions (parsed from `FLY_REGION`)
2. Passes a `compute` hint matching the expected machine spec
3. Tries each region in order; on capacity 412, logs a warning and tries the next
4. On non-capacity errors, throws immediately (no fallthrough)

For 412 recovery, `deprioritizeRegion` moves the failed region to the end of the list so other regions are tried first.

## Recovery Strategy

| Scenario | Action |
|---|---|
| Fresh provision (never started, empty volume) | Delete volume, create fresh in deprioritized regions |
| Existing instance (has user data) | Fork volume to new region, delete old |
| Fork fails on existing instance | Error surfaces to caller (no silent data loss) |

## Changes

| File | What |
|------|------|
| `src/fly/types.ts` | `VolumeComputeHint` type, `compute` + `source_volume_id` on `CreateVolumeRequest` |
| `src/fly/client.ts` | `isFlyInsufficientResources()` (capacity-aware 412), `createVolumeWithFallback()` (region walking) |
| `src/config.ts` | `DEFAULT_FLY_REGION` constant |
| `src/durable-objects/kiloclaw-instance.ts` | `parseRegions()`, `deprioritizeRegion()`, `replaceStrandedVolume()`, 412 recovery in `start()`, pre-flight region validation, compute hints on all volume creation |
| `src/fly/client.test.ts` | Tests for 412 classification, `createVolumeWithFallback` (region walking, compute passthrough) |
| `src/durable-objects/kiloclaw-instance.test.ts` | Tests for `parseRegions`, `deprioritizeRegion`, 412 recovery, region validation |

## Test plan

- 245 tests passing (`pnpm typecheck && pnpm test && pnpm lint` all clean)
- Covers: fresh provision 412, existing instance fork, fork failure propagation, updateMachine 412, non-412 passthrough, retry failure propagation, region drift correction, volume 404 during region check, production payload matching, version mismatch rejection, unclassified 412 warning, region walking (first region 412 → fallback), all regions exhausted, non-capacity error no-fallthrough, compute hint passthrough

### DIFF
--- a/kiloclaw/src/durable-objects/kiloclaw-instance.ts
+++ b/kiloclaw/src/durable-objects/kiloclaw-instance.ts
@@ -1464,18 +1464,26 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
     const regions = deprioritizeRegion(allRegions, oldRegion);
     const compute = guestFromSize(this.machineSize);
 
-    // Destroy existing machine if any — it's stuck on the constrained host
+    // Destroy existing machine if any — it's stuck on the constrained host.
+    // Only clear flyMachineId on confirmed deletion (success or 404).
+    // On transient failures, keep the ID so reconciliation can retry cleanup.
     if (this.flyMachineId) {
+      let machineGone = false;
       try {
         await fly.destroyMachine(flyConfig, this.flyMachineId);
         reconcileLog(reason, 'destroy_stranded_machine', { machine_id: this.flyMachineId });
+        machineGone = true;
       } catch (err) {
-        if (!fly.isFlyNotFound(err)) {
+        if (fly.isFlyNotFound(err)) {
+          machineGone = true;
+        } else {
           console.warn('[DO] Failed to destroy stranded machine:', err);
         }
       }
-      this.flyMachineId = null;
-      await this.ctx.storage.put(storageUpdate({ flyMachineId: null }));
+      if (machineGone) {
+        this.flyMachineId = null;
+        await this.ctx.storage.put(storageUpdate({ flyMachineId: null }));
+      }
     }
 
     if (hasUserData) {
--- a/kiloclaw/src/durable-objects/kiloclaw-instance.ts
+++ b/kiloclaw/src/durable-objects/kiloclaw-instance.ts
@@ -44,6 +44,7 @@ import {
   ALARM_INTERVAL_IDLE_MS,
   ALARM_JITTER_MS,
   SELF_HEAL_THRESHOLD,
+  DEFAULT_FLY_REGION,
 } from '../config';
 import type { FlyClientConfig } from '../fly/client';
 import type { FlyMachineConfig, FlyMachine, FlyMachineState } from '../fly/types';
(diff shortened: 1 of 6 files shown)


kiloclaw/src/durable-objects/kiloclaw-instance.ts
```diff
@@ -157,6 +158,19 @@ function volumeNameFromSandboxId(sandboxId: string): string {
     .slice(0, 30);
 }
 
+// ============================================================================
+// Region priority helper for 412 recovery
+// ============================================================================
+
+/**
+ * Reverse the FLY_REGION priority list so we try different regions first.
+ * E.g. "dfw,yyz,cdg" → "cdg,yyz,dfw" — if dfw was capacity-constrained,
```
pandemicsyn: yea this is dumb right now. waiting on fly to get back with some details on how the regional selectors work - once i hear back we'll come up with a better retry pattern.

St0rmz1 (approved):

PR #393, opened 2026-02-19
fix: reduce KiloClaw stale status from ~30min to ~30s and prevent Fly auto restart

St0rmz1:
- Add LIVE_CHECK_THROTTLE_MS (30s) constant to config
  - Add fire-and-forget syncStatusFromLiveCheck() via ctx.waitUntil() that queries Fly on getStatus() polls when DO thinks machine is running
  - Only flip to stopped for terminal Fly states; transitional states (starting, stopping, replacing) leave status as running
  - Fix stop() to propagate Fly API errors instead of silently writing stopped (404 still treated as success)
  - Disable Fly autostart/autostop on machine services so stopped machines stay stopped when WebSocket proxy requests hit them
  - Add 11 tests covering live check and stop() error handling

### DIFF
--- a/kiloclaw/src/durable-objects/kiloclaw-instance.ts
+++ b/kiloclaw/src/durable-objects/kiloclaw-instance.ts
@@ -44,6 +44,7 @@ import {
   ALARM_INTERVAL_IDLE_MS,
   ALARM_JITTER_MS,
   SELF_HEAL_THRESHOLD,
+  LIVE_CHECK_THROTTLE_MS,
 } from '../config';
 import type { FlyClientConfig } from '../fly/client';
 import type { FlyMachineConfig, FlyMachine, FlyMachineState } from '../fly/types';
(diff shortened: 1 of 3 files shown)


kiloclaw/src/durable-objects/kiloclaw-instance.ts
```diff
@@ -167,6 +170,15 @@ const METADATA_RECOVERY_COOLDOWN_MS = ALARM_INTERVAL_IDLE_MS;
...
 const DEAD_STATES: ReadonlySet<FlyMachineState> = new Set(['destroyed', 'destroying']);
 
+/** Terminal non-running states for live check. Transitional states (starting, stopping, replacing)
+ *  are intentionally excluded to avoid UI flicker during normal operations. */
+const TERMINAL_STOPPED_STATES: ReadonlySet<string> = new Set([
+  'stopped',
+  'created',
+  'destroyed',
+  'suspended',
+]);
```
pandemicsyn, line 180: I *think*, we want ReadonlySet<FlyMachineState> here ? Like the DEAD_STATES above.

pandemicsyn (approved):
>Disable Fly autostart/autostop on machine services so stopped machines stay stopped when WebSocket proxy requests hit them

😶‍🌫️ good catch

PR #398, opened 2026-02-20
feat: Add session id in heuristic abuse

alex-alecu:
Easily find what session has possibility of abuse `/admin/users/..../heuristic-abuse`


src/db/migrations/0024_real_ikaris.sql
```diff
@@ -0,0 +1,61 @@
...
+    meta.generation_time,
+    meta.is_byok,
+    meta.is_user_byok,
+    meta.streamed,
+    meta.cancelled,
+    edit.editor_name,
+    meta.has_tools,
+    meta.machine_id,
+    feat.feature,
+    meta.session_id
```
marius-kilocode, line 49: We don't need an index on it do we?

marius-kilocode (approved):

PR #403, opened 2026-02-20
fix: handle deleted repos gracefully in security alerts sync

jeanduplessis:
## Summary
- Handle HTTP 404 from GitHub Dependabot API gracefully when repos have been deleted or transferred, instead of throwing errors that cause 502s on the cron job
- Automatically prune stale repos from the agent config's `selected_repository_ids` so they aren't retried on future syncs
- Introduce a discriminated union `FetchAlertsResult` to cleanly distinguish success, repo-not-found, and alerts-disabled responses

## Changes
- **`dependabot-api.ts`**: Changed `fetchAllDependabotAlerts` return type from `DependabotAlertRaw[]` to `FetchAlertsResult` discriminated union. HTTP 404 now returns `{ status: 'repo_not_found' }` instead of throwing.
- **`sync-service.ts`**: Added `pruneStaleReposFromConfig()` which removes deleted repo IDs from `selected_repository_ids` when in `'selected'` mode. Uses `getSecurityAgentConfigWithStatus` + `upsertAgentConfigForOwner` directly to preserve the current `isEnabled` state.
- **`types.ts`**: Added `staleRepos: string[]` to `SyncResult` to propagate not-found repos up the call chain.

### DIFF
--- a/src/lib/security-agent/services/sync-service.ts
+++ b/src/lib/security-agent/services/sync-service.ts
@@ -13,7 +13,8 @@ import { fetchAllDependabotAlerts } from '../github/dependabot-api';
 import { hasSecurityReviewPermissions } from '../github/permissions';
 import { parseDependabotAlerts } from '../parsers/dependabot-parser';
 import { upsertSecurityFinding } from '../db/security-findings';
-import { getSecurityAgentConfig } from '../db/security-config';
+import { getSecurityAgentConfig, getSecurityAgentConfigWithStatus } from '../db/security-config';
+import { upsertAgentConfigForOwner } from '@/lib/agent-config/db/agent-configs';
 import {
   getSlaForSeverity,
   calculateSlaDueAt,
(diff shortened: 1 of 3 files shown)


src/lib/security-agent/services/sync-service.ts
```diff
@@ -290,12 +309,70 @@ export async function getEnabledSecurityReviewConfigs(): Promise<
...
+    return;
+  }
+
+  const prunedIds = config.selected_repository_ids.filter(id => !staleIds.has(id));
+  if (prunedIds.length === config.selected_repository_ids.length) return;
+
+  const prunedRepoNames = staleRepoNames.filter(name => {
+    const id = repoNameToId.get(name);
+    return id != null && staleIds.has(id);
+  });
```
marius-kilocode:
```suggestion
  const prunedRepoNames = staleRepoNames.filter(name => repoNameToId.has(name));
```

marius-kilocode: I think this can be simplified, by just looking at the logic?

marius-kilocode (approved):

PR #460, opened 2026-02-23
AppBuilder - Preview fetch fix and git integration test refactoring

eshurakov:
Fixes builder preview staleness by adding `git fetch origin` before `git reset --hard origin/main` in `preview-do.ts`, ensuring previews always pick up the latest pushed changes. Also refactors the git integration tests by extracting shared helpers (API calls, git CLI wrappers, assertions, temp dir management, test runner) into a reusable `git-test-helpers.ts` module, adds a new `test-git-head-and-history.ts` suite with 5 tests specifically targeting detached HEAD regressions, commit history preservation, incremental pushes, and bidirectional clone/push round-trips, and replaces unsafe `as` casts with safe property accessor helpers.

### DIFF
--- a/cloudflare-app-builder/src/_integration_tests/test-git-integration.ts
+++ b/cloudflare-app-builder/src/_integration_tests/test-git-integration.ts
@@ -11,7 +11,7 @@
  *
  * Usage:
  *   cd cloudflare-app-builder
- *   AUTH_TOKEN=dev-token-change-this-in-production npx ts-node src/test-git-integration.ts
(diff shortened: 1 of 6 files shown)


cloudflare-app-builder/src/_integration_tests/test-git-integration.ts
eshurakov, line 1: Integration tests rely on the running instance and its not yet set up in GitHub actions

jeanduplessis (approved):

PR #467, opened 2026-02-23
Add auto triage docs to Org page

jfcsantos:
I didn't realise Orgs and Personal were using different pages so the link was missing on the Org one. 
Seeing that they are basically duplicates, I've refactored it to use the same Components, adjusting only where necessary.


src/app/(app)/auto-triage/page.tsx
```diff
@@ -1,20 +1,13 @@
...
+import { AutoTriagePageClient } from '@/components/auto-triage/AutoTriagePageClient';
 
 type AutoTriagePageProps = {
   searchParams: Promise<{ success?: string; error?: string }>;
 };
 
 export default async function PersonalAutoTriagePage({ searchParams }: AutoTriagePageProps) {
   const search = await searchParams;
-  const user = await getUserFromAuthOrRedirect('/users/sign_in?callbackPath=/auto-triage');
+  await getUserFromAuthOrRedirect('/users/sign_in?callbackPath=/auto-triage');
```
jfcsantos, line 10: user wasn't being used in the Page component at all.

PR #470, opened 2026-02-23
AI-powered duplicate detection for auto-triage

jfcsantos:
### Summary

Replaces the simple 0.9 embedding-similarity threshold for duplicate detection with an AI-powered semantic verification step. The Milvus embedding search now acts as a candidate filter (≥ 0.8 similarity), and the Cloud Agent verifies whether any candidate is a true duplicate using LLM reasoning.

### Changes

**Duplicate detection is now two stages instead of one**
- Previously, the check-duplicates endpoint both found similar issues and decided whether they were duplicates (≥ 0.9 similarity = duplicate). Now it only returns similarity candidates — the decision moves to the worker.
- The worker sends the candidates to the Cloud Agent, which uses LLM reasoning to determine if any candidate is a true semantic duplicate, with confidence scoring and structured reasoning.
- If the AI verification fails for any reason, the system falls back to the original threshold-based logic.
- Duplicate closure comments now include AI reasoning when available.

**Cloud Agent supports prompt-only sessions**
- Some workflows (classification, duplicate verification) only need a text prompt — no codebase. Previously, every session required a repo, which meant cloning a repo that was never used.
- The initiateSessionStream endpoint now allows omitting the repo when the mode is ask. All other modes (code, architect, debug, orchestrator) still require a repo, so existing flows are unaffected.
- The session service skips branch checkout when no repo was cloned.

**UI error sanitization**
- System errors (SQL, connection failures, stack traces) are no longer shown raw in toast messages. A shared friendlyErrorMessage utility filters them into a generic user-facing message.


jfcsantos:
@eshurakov ccing you since I'm modifying some `cloud-agent` code here.

cloudflare-auto-triage-infra/src/triage-orchestrator.ts
```diff
@@ -224,6 +251,232 @@ export class TriageOrchestrator extends DurableObject<Env> {
...
+    // No githubRepo — the prompt is pure text, no codebase access needed.
+    const sessionInput = {
+      kilocodeOrganizationId: this.state.owner.type === 'org' ? this.state.owner.id : undefined,
+      prompt,
+      mode: 'ask' as const,
+      model: this.state.sessionInput.modelSlug,
+      createdOnPlatform: 'auto-triage',
+    };
+
+    const cloudAgentClient = new CloudAgentClient(this.env.CLOUD_AGENT_URL, this.state.authToken);
```
eshurakov: If you don't need repo access, it may be faster and easier to just go directly to LLM. Search for `sendProxiedChatCompletion`

PR #486, opened 2026-02-24
refactor: extract shared bot run loop

RSO:
## Summary
- move the tool-loop/chat completion handling into a reusable bot runner
- keep Slack bot behavior by delegating to the core runner with the same tool execution
- capture tool metadata for the existing session id tracking

## Testing
- pnpm typecheck
- I've got the Slack bot running locally and verified that it still responds as expected to messages

### DIFF
--- a/src/lib/bots/core/run-bot.ts
+++ b/src/lib/bots/core/run-bot.ts
@@ -1,0 +1,190 @@
+import type OpenAI from 'openai';
+import type { FeatureValue } from '@/lib/feature-detection';
+import { sendProxiedChatCompletion } from '@/lib/llm-proxy-helpers';
+import type { OpenRouterChatCompletionRequest } from '@/lib/providers/openrouter/types';
+
+type ChatMessage = OpenAI.Chat.Completions.ChatCompletionMessageParam;
+type ChatCompletionResponse = OpenAI.Chat.Completions.ChatCompletion;
+type ToolCall = OpenAI.Chat.Completions.ChatCompletionMessageToolCall;
+
+export type BotToolResult = {
+  content: string;
+  metadata?: Record<string, unknown>;
+};
+
+export type BotRunInput = {
+  authToken: string;
+  model: string;
+  systemPrompt: string;
+  userMessage: string;
+  tools?: OpenAI.Chat.Completions.ChatCompletionTool[];
+  toolExecutor?: (toolCall: ToolCall) => Promise<BotToolResult>;
+  onToolResult?: (toolCall: ToolCall, result: BotToolResult) => void | Promise<void>;
+  maxIterations?: number;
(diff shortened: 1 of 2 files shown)


src/lib/bots/core/run-bot.ts
```diff
@@ -0,0 +1,176 @@
...
+
+      messages.push({
+        role: 'assistant',
+        content: message.content,
+        tool_calls: message.tool_calls,
+      });
+      console.log(
+        `${logPrefix} Added assistant message to history, total messages:`,
+        messages.length
+      );
```
alex-alecu, line 120: If `toolCall.type` is `"custom"`, the else branch accesses `.function.name` which doesn't exist on that type. The original code only tracked function-type calls after an explicit `type !== 'function'` guard.

```ts
if (toolCall.type === 'function') {
  toolCallsMade.push(toolCall.function.name);
}
```

alex-alecu (commented):
Left inline comments.

src/lib/slack-bot.ts
```diff
@@ -440,172 +436,62 @@ export async function processKiloBotMessage(
...
+        args,
+        owner,
+        selectedModel,
+        authToken,
+        slackRequesterInfo
+      );
+      console.log('[SlackBot] Tool result received, length:', toolResult.response.length);
+      console.log('[SlackBot] Tool result preview:', toolResult.response.slice(0, 100));
+      cloudAgentSessionId = toolResult.sessionId;
+
```
alex-alecu, line 483: The original had `if (toolResult.sessionId)` before assigning. Without the guard, if `spawnCloudAgentSession` returns `sessionId: undefined` on a second tool call, this overwrites a previously-captured session ID with `undefined`.

src/lib/bots/core/run-bot.ts
```diff
@@ -0,0 +1,176 @@
...
+import { sendProxiedChatCompletion } from '@/lib/llm-proxy-helpers';
+import type { OpenRouterChatCompletionRequest } from '@/lib/providers/openrouter/types';
+
+type ChatMessage = OpenAI.Chat.Completions.ChatCompletionMessageParam;
+type ChatCompletionResponse = OpenAI.Chat.Completions.ChatCompletion;
+type ToolCall = OpenAI.Chat.Completions.ChatCompletionMessageToolCall;
+
+export type BotToolResult = {
+  content: string;
+  metadata?: Record<string, unknown>;
```
alex-alecu: Is anything consuming `metadata`? `runBot` never reads it and it's not surfaced in `BotRunResult`. The SlackBot caller captures `cloudAgentSessionId` via closure anyway — seems like this can be dropped?

alex-alecu (approved):

PR #500, opened 2026-02-24
Slack: Use Cloud Agent Next infra

RSO:
This PR introduces a newly made up `runSessionToCompletion` helper that mimicks the behavior from the pre-websocket interface of Cloud Agents.

In the long run I'd like to move to the callbackUrl flow to prevent timeouts, but that seemed like a bigger change, and I wished to get this in quickly.

I've tested locally to confirm that the `View Session` button still works and that Cloud Agents are spawned as expected.

### DIFF
--- a/src/lib/cloud-agent-next/run-session.ts
+++ b/src/lib/cloud-agent-next/run-session.ts
@@ -1,0 +1,377 @@
+import 'server-only';
+import { getEnvVariable } from '@/lib/dotenvx';
+import { signStreamTicket, type StreamTicketPayload } from '@/lib/cloud-agent/stream-ticket';
+import { createWebSocketManager } from './websocket-manager';
+import { createEventProcessor, type ProcessedMessage } from './processor';
+import type { CloudAgentEvent, StreamError } from './event-types';
+import type { CloudAgentNextClient } from './cloud-agent-client';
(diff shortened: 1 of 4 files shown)


src/lib/cloud-agent-next/run-session.ts
```diff
@@ -0,0 +1,377 @@
...
+    if (completeGraceTimeoutId) clearTimeout(completeGraceTimeoutId);
+    resolveStream?.();
+  };
+
+  const processor = createEventProcessor({
+    callbacks: {
+      onMessageCompleted: (_sid, _mid, message) => {
+        if (message.info.role !== 'assistant') return;
+        const text = extractTextFromMessage(message);
+        if (text) completionResult = text;
```
marius-kilocode, line 237: I don't get this logic. Will this work with partial results etc?

marius-kilocode (approved):

eshurakov (approved):

src/lib/cloud-agent-next/run-session.ts
RSO: I'm not 100% sure, but I assumed that `onMessageCompleted` is only called when the message is finalized, and that this line makes sure that we store the last message as the completion result.

PR #497, opened 2026-02-24
Fix DB connection exhaustion (max_client_conn) by reducing pool sizes

iscekic:
## Summary

Fixes `no more connections allowed (max_client_conn)` errors caused by aggregate connection counts across all services exceeding the Supabase PgBouncer limit (3000).

- **CF workers** (`cloud-agent`, `cloud-agent-next`, `git-token-service`): Replace `pg.Pool(max:100)` with fresh `pg.Client` per operation, matching the Hyperdrive-recommended pattern already used by `kiloclaw` and `webhook-agent-ingest`. Hyperdrive handles connection pooling at the infrastructure level — using a driver-level pool on top is an anti-pattern that holds unnecessary connections open.
- **`session-ingest`**: Reduce pool `max` from 5 to 1 (Hyperdrive pools for us).
- **Vercel Next.js app** (`src/lib/drizzle.ts`): Reduce pool `max` from 100 to 10 per pool, and apply `POSTGRES_MAX_QUERY_TIME` as `statement_timeout` (was validated to exist but never actually enforced on the pool).

## Root Cause

At peak, the Supabase PgBouncer dashboard showed **2715/3000** pooler connections. The aggregate `max` across services was:

| Source | Old pool `max` | New |
|--------|---------------|-----|
| Vercel primary pool | 100 × N instances | 10 × N |
| Vercel replica pool | 100 × N instances | 10 × N |
| `cloud-agent` | 100 × M isolates (via Hyperdrive) | 1 per query (Client) |
| `cloud-agent-next` | 100 × M isolates (via Hyperdrive) | 1 per query (Client) |
| `git-token-service` | 100 × M isolates (via Hyperdrive) | 1 per query (Client) |
| `session-ingest` | 5 per request (via Hyperdrive) | 1 per request |

## Evidence from Axiom

Axiom logs from the incident window (2026-02-24 10:00–11:30 UTC) confirm the diagnosis:

**Traffic volume:** ~110,000 requests/minute (~1,800/sec) steady-state across 20+ Vercel regions (fra1, bom1, cdg1, arn1, sin1, cpt1, iad1, dxb1, lhr1, hkg1, sfo1, gru1, syd1, pdx1, cle1, etc.).

**Error burst at 10:15–10:25 UTC:** ~20,000 5xx errors in 10 minutes, across *all* routes — consistent with a database-level connection exhaustion (not an endpoint-specific bug):

| Route | 5xx errors (10:15–10:25) |
|-------|------------------------|
[...]

PR #518, opened 2026-02-24
Reduce Vercel DB pool sizes and enforce statement_timeout

iscekic:
## Summary

- Reduce pool `max` from 100 to 10 per pool (primary + replica) — at peak, dozens of warm Vercel instances across 20+ regions each holding `max:100` connections contributes significantly to hitting the Supabase PgBouncer `max_client_conn` limit (3000).
- Apply `POSTGRES_MAX_QUERY_TIME` as `statement_timeout` on the pool — the env var was validated to exist on startup but was never actually passed to the pool config.

Part of the fix for `max_client_conn` connection exhaustion (see also #497 for the Cloudflare worker changes).

### DIFF
--- a/src/lib/drizzle.ts
+++ b/src/lib/drizzle.ts
@@ -66,9 +66,10 @@ function getReplicaUrl(): string {
 // Primary pool - always points to Frankfurt (writes go here)
 export const pool = new Pool({
   ...getDatabaseClientConfig(postgresUrl),
-  max: 100,
+  max: 10,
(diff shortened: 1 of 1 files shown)


iscekic:
This PR contains the Vercel Next.js app changes that were originally part of #497. They were split out so the Cloudflare worker changes (replacing `pg.Pool` with per-operation `pg.Client` to match the Hyperdrive pattern) could be reviewed and deployed independently from the Vercel pool size reduction.

iscekic:
Note to self - `statement_timeout` doesn't work

PR #538, opened 2026-02-24
fix: handle SIGUSR1 gateway restart without orphaned processes

pandemicsyn:
## Summary

- Fixes a bug where the openclaw gateway's SIGUSR1 restart (triggered by `update.run` in the Control UI) would spawn a detached child process, then the controller's supervisor would also respawn the gateway — resulting in unsupervised gateway processes.
- Sets `INVOCATION_ID=1` in `start-openclaw.sh` so the gateway recognizes it's running under a supervisor and exits cleanly (code 0) instead of spawning a detached child on SIGUSR1.
- Teaches the controller's supervisor to distinguish clean exits (code 0, no signal) from crashes — immediate respawn without backoff for intentional restarts, existing crash/backoff logic for failures.

## Details

When the gateway receives SIGUSR1 (e.g., from the Control UI's "Update & Restart" button), it drains active tasks, closes the server, and then decides how to restart based on environment hints. Without `INVOCATION_ID`, it doesn't detect a supervisor and spawns a detached child process before exiting. The controller then sees the exit as a crash and spawns *another* gateway — leaving an orphaned, unsupervised process.

`INVOCATION_ID` is a standard systemd convention that openclaw checks in `src/infra/process-respawn.ts` to detect supervised environments. Setting it causes the gateway to exit with code 0 and let the supervisor handle the respawn.

The supervisor change adds exit-code-aware restart logic: code 0 with no signal = intentional restart (immediate respawn, no backoff), anything else = crash (backoff + counter). The restart counter increments in both cases, with `lastExit` recording the cause.

## Other changes on this branch

The earlier commits (`tracking down error`, `Use full path and log additional info`) are from investigating the issue that led to this fix — debugging pairing list scripts and logging.

### DIFF
--- a/kiloclaw/start-openclaw.sh
+++ b/kiloclaw/start-openclaw.sh
@@ -123,8 +123,8 @@ if [ ! -f "$CONFIG_FILE" ]; then
 
     openclaw onboard --non-interactive --accept-risk \
         --mode local \
-        --gateway-port 18789 \
-        --gateway-bind lan \
+        --gateway-port 3001 \
+        --gateway-bind loopback \
         --skip-channels \
         --skip-skills \
         --skip-health
--- a/kiloclaw/start-openclaw.sh
+++ b/kiloclaw/start-openclaw.sh
@@ -317,6 +317,12 @@ EOFPATCH
 # ============================================================
 # START CONTROLLER
 # ============================================================
+# Tell the gateway it's running under a supervisor. On SIGUSR1 restart,
+# the gateway will exit cleanly (code 0) instead of spawning a detached
+# child process. The controller's supervisor detects the clean exit and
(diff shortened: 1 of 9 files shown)


kiloclaw/start-openclaw.sh
```diff
@@ -123,8 +123,8 @@ if [ ! -f "$CONFIG_FILE" ]; then
 
     openclaw onboard --non-interactive --accept-risk \
         --mode local \
-        --gateway-port 18789 \
-        --gateway-bind lan \
```
pandemicsyn, line 127: we already set these in the config, just making sure onboarding matches.

St0rmz1 (approved):

Issue #329, opened 2026-02-18
Direct gateway traffic needs positive feature identification


pedroheyerdahl:
## Context

PR #307 adds deterministic feature attribution to `microdollar_usage_metadata` via the `X-KILOCODE-FEATURE` header. All internal features (extensions, CLI, cloud agents, internal services) will send this header. Direct gateway consumers (external API users) currently show as NULL.

## Problem

Direct gateway is a WAU feature that needs a positive signal like all other features. NULL in the feature column is reserved for catching tracking issues (e.g. a misconfigured internal feature). If direct gateway traffic is also NULL, we can't distinguish legitimate unattributed external usage from broken internal attribution.

Currently, external consumers can hit both `/api/gateway/` and `/api/openrouter/`, which makes it difficult to positively identify direct gateway traffic at the infrastructure level.

## What's needed

A mechanism to positively identify direct gateway traffic so it gets its own feature value (`direct-gateway`) rather than falling into the unattributed NULL bucket.

cc @jrf0110 @chrarnoldus

jrf0110:
We need to align on what "Direct Gateway Traffic" means. Kilocode features hit the gateway the same way that everyone hits the gateway. One difference from kilo features is that we now send the `X-KILOCODE-FEATURE` header. Any consumer of the gateway can choose to send that header if they want. So this is a best-effort sort of solution and it can't be 100% trusted.

My recommendation is that any request that does not have an `X-KILOCODE-FEATURE` header _should be considered `feature: gateway`_

pedroheyerdahl:
Direct gateway = external consumers who hit the gateway API directly, without going through the extension, CLI, or any of our other features.

Defaulting missing headers to gateway just moves the problem. It makes the gateway bucket unreliable instead of the NULL bucket. NULL needs to stay as our "something is wrong with tracking" signal. Without that, we can't confidently report any feature's attribution is healthy.

We need a positive signal that a request came from someone hitting the API directly. Even if it's imperfect: yes, a user could spoof a header, but there's zero incentive to do so (it doesn't affect billing or access), and that's a far more marginal case than silently mislabeling all unattributed traffic as gateway

jrf0110:
> Defaulting missing headers to gateway just moves the problem. It makes the gateway bucket unreliable instead of the NULL bucket. NULL needs to stay as our "something is wrong with tracking" signal. Without that, we can't confidently report any feature's attribution is healthy.

[...]

PR #547, opened 2026-02-25
Tag /api/gateway requests with direct-gateway feature header

emilieschario:
## Summary

- Requests hitting `/api/gateway/` now have `x-kilocode-feature: direct-gateway` injected before being forwarded to the openrouter handler
- Any client-supplied `x-kilocode-feature` header is overwritten
- Added tests covering both behaviours


pedroheyerdahl (commented):
The route change is exactly right — this is the route-based approach Pedro originally had in #307 before reverting it, and it's the correct solution to #329.

One concern: the test mocks the openrouter POST handler and asserts on the forwarded request object. Per our coding guidelines, we avoid mocks because they make tests brittle and tied to implementation details. The test doesn't verify any observable behavior — it would still pass even if the header injection had no effect downstream.

Could the test instead make a real request through the stack and verify the `feature` column in the DB ends up as `direct-gateway`? There are examples in the existing integration test helpers (`src/tests/helpers/microdollar-usage.helper.ts`) that show how to check usage records. That would give us real confidence the wiring works end-to-end, rather than just that we're calling `openrouterPOST` with certain headers.

pedroheyerdahl (approved):
nit: the test mocks openrouter and checks what headers we passed in, but that doesn't actually verify the feature ends up in the DB (`coding-style.md` suggest not using mocks). would be worth wiring this up as an integration test instead, similar to what's in src/tests/helpers/microdollar-usage.helper.ts. I can send a follow up.

src/app/api/gateway/[...path]/route.ts
```diff
@@ -1 +1,9 @@
-export { POST } from '@/app/api/openrouter/[...path]/route';
+import { POST as openrouterPOST } from '@/app/api/openrouter/[...path]/route';
+import { NextRequest } from 'next/server';
+import { FEATURE_HEADER } from '@/lib/feature-detection';
+
+export function POST(request: NextRequest) {
+  const headers = new Headers(request.headers);
+  headers.set(FEATURE_HEADER, 'direct-gateway');
```
Comment (chrarnoldus) [inline src/app/api/gateway/[...path]/route.ts]: we probably don't want to override it if it's already set

Comment (jrf0110) [reply src/app/api/gateway/[...path]/route.ts]: Agreed!

jrf0110 (approved):

PR #560, opened 2026-02-25
Next 16 migration

iscekic:


src/app/api/gateway/models/route.ts
```diff
@@ -1 +1 @@
-export { GET, revalidate } from '@/app/api/openrouter/models/route';
```
RSO, line 1: Why are we not exporting `revalidate` anymore?

src/middleware/withKiloEditorCookie.ts
```diff
@@ -1,21 +1,25 @@
-import type { NextFetchEvent } from 'next/server';
+import { type NextFetchEvent, NextResponse } from 'next/server';
```
RSO, line 1: Probably worth while to test this? Should be as easy as setting `?source=foo` and then checking that the cookie gets updated

RSO (approved):

src/app/api/gateway/models/route.ts
iscekic: This never worked in the first place apparently. The actual caching is in the route file, using `unstable_cache`.

jeanduplessis (approved):

PR #608, opened 2026-02-26
ci: add workflow_dispatch to deploy-production workflow

RSO:
## Summary

- Adds `workflow_dispatch` trigger to the Deploy to Production workflow, enabling manual deploys from the GitHub Actions UI
- Updates all conditionally-deployed sub-services to also run when triggered manually (`github.event_name == 'workflow_dispatch'`), since path-filter checks would otherwise skip them with no git diff context

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -10,7 +10,17 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
+  check-ref:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Ensure workflow runs on main
+        if: github.ref != 'refs/heads/main'
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-production.yml
```diff
@@ -3,13 +3,24 @@ name: Deploy to Production
...
 concurrency:
   group: deploy-production
   cancel-in-progress: false
 
 jobs:
+  check-ref:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Ensure workflow runs on main
+        if: github.ref != 'refs/heads/main'
```
markijbema, line 17: crant we enforce this by restricting line 6?

markijbema (approved):

PR #624, opened 2026-02-26
minor UI/UX improvements for Channels configurations

St0rmz1:
- Adds alerts to prompt user to hit redeploy
- After deploy, cycle the check for pairing more often for a few mins to give user immediate feedback they need to approve the pairing

### DIFF
--- a/src/app/(app)/claw/components/ClawDashboard.tsx
+++ b/src/app/(app)/claw/components/ClawDashboard.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useState } from 'react';
 import { TriangleAlert } from 'lucide-react';
 import type { KiloClawDashboardStatus } from '@/lib/kiloclaw/types';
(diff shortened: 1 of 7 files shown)


src/app/(app)/claw/components/ClawDashboard.tsx
```diff
@@ -25,6 +26,9 @@ function hasPopulatedStatus(
   return candidate !== undefined && candidate.status !== null;
 }
 
+const FAST_POLL_INTERVAL = 10_000;
```
pandemicsyn: in the worker we cache pairing api call results because they're pretty expensive and slow to run (+30s and gets worst with more channels configured).  In an earlier version, seemed to drive up memory usage - but I was never positive if that was really the cause. But its made ma bit gun shy on ramping this up. 

tl;dr I think its fine to ship as-is but worth probably doing a pairing or two just to make sure it doesn't have a weird perf impact.

St0rmz1: good call. I'll change that

pandemicsyn (approved):

PR #659, opened 2026-02-27
fix: Improve gitlab permissions logging

alex-alecu:
## Diagnosis

We’re seeing a consistent failure when users launch code reviews on their GitLab instance.

```sql
select * from cloud_agent_code_reviews where repo_full_name like '%....%'
# -> The Error: Failed to prepare session: Failed to clone repository from xxxx
```

**Root Cause: Error swallowing in `prepareReviewPayload` causes GitLab code reviews to clone without authentication.**

The flow works like this:
1. A GitLab MR webhook triggers a code review
2. `prepareReviewPayload()` attempts to create a Project Access Token (PrAT) for cloning
3. If PrAT creation fails (permissions, API error, etc.), the inner try-catch re-throws the error
4. **Bug:** The outer `try/catch` (line 305, originally designed for GitHub token failures) catches this re-thrown error and **swallows it**, logging to Sentry but continuing without a token
5. The session input is built with `gitToken: undefined` 
6. The clone attempt hits the private repo without credentials → immediate `"Failed to clone repository"` failure

This explains why:
- **All `flytogether-mp` reviews fail** — PrAT creation consistently fails for this user, and the error is swallowed
- **Some `devproducts` reviews succeed** — PrAT creation sometimes works (the 3 successes on Feb 25-26), but when it fails, same swallowing happens
- **Failures happen in ~2-3 seconds** — no timeout; git just immediately rejects the unauthenticated request

**Secondary issue:** The `platform` field was never passed through the code review orchestrator to cloud-agent, forcing it to rely on a heuristic (`gitUrl?.includes('gitlab')`) that would fail for self-hosted instances with non-"gitlab" domain names.

### 1. `src/lib/code-reviews/triggers/prepare-review-payload.ts`
- **Restructured error handling:** Separated GitHub and GitLab auth into independent blocks. GitHub auth failures are still caught and swallowed (public repos may work without token). GitLab PrAT failures now **propagate** as hard errors, since private GitLab repos cannot be cloned without authentication.
- **Added `platform` field** to `SessionInput` type and to both GitHub/GitLab session input construction, so cloud-agent receives explicit platform info instead of relying on URL heuristics.

### 2. `cloudflare-code-review-infra/src/types.ts`
- **Added `platform` field** to the `SessionInput` interface so it flows through the orchestrator to cloud-agent.

### Next Steps
After deploying, the failing reviews will now surface clear error messages like `"Failed to create Project Access Token for GitLab code review on flytogether-mp/static-site. Error: ..."` instead of the misleading `"Failed to clone repository"`. This will reveal the **underlying PrAT failure reason** — likely a permissions issue (user needs Maintainer role) or PrAT feature disabled at the project/group level. Check Sentry for the captured `authError` exceptions from these users to confirm.


jeanduplessis (approved):
Nit: there's a lot of "magic string" values being used and inconsistent use of variables vs strings. Some areas use PLATFORM.GITLAB, others just "gitlab" or "github".

It would be good to make this code consistent.

PR #690, opened 2026-02-28
Reduce enterprise seat price from $150 to $60/seat/month

alexkgold:
## Changes

- **Price update**: `ENTERPRISE_SEAT_PRICE_MONTHLY_USD` reduced from 150 to 60 in `src/lib/organizations/constants.ts`. The yearly constant is computed from it and now evaluates to $720/year.

- **Bug fix**: `SubscriptionQuickActions.tsx` was passing `TEAM_SEAT_PRICE_YEARLY_USD` ($180) to the `SeatChangeModal` for enterprise orgs instead of `ENTERPRISE_SEAT_PRICE_MONTHLY_USD`. Fixed the import and ternary so enterprise orgs see the correct price in the seat change cost preview.


jrf0110 (approved):
This needs to be updated in stripe as well, right?

PR #705, opened 2026-03-01
CloudAgentNext - Fix controller disconnect handling and execution lifecycle

eshurakov:
## Summary

- **Immediate disconnect detection**: When a wrapper WebSocket disconnects while an execution is active, the controller now marks the execution as `failed` immediately and emits a synthetic `wrapper_disconnected` event to notify `/stream` clients — instead of waiting for the reaper to discover the stale execution minutes later.
- **Reaper event emission**: When the reaper marks stale running or timed-out pending executions as failed, it now emits synthetic `error` events so `/stream` clients are notified of the failure.
- **Faster stale detection**: `STALE_THRESHOLD_MS` reduced from 10 minutes to 90 seconds (~3 missed heartbeats). Reaper alarm interval drops from 5 minutes to 2 minutes while an execution is active.
- **Lifecycle reset between executions**: The wrapper's lifecycle manager now resets `isAborted`, `isDraining`, and drain timeouts when a new job starts, preventing stale state from a previous execution from interfering.
- **SSE activity tracking reset**: `WrapperState.startJob` now resets `lastSseEventAt` so stale inactivity data from a prior execution doesn't carry over.
- **Interrupt handler cleanup**: When interrupt finds no running processes but there's still an active execution, it now clears the stale execution and emits an error event to `/stream` clients.
- **pkill fix**: Added `--` separator to `pkill -f` to prevent pattern misinterpretation.
- **Tests**: Added unit tests for ingest close return value, lifecycle reset, state SSE tracking reset, and an integration test for reaper disconnect handling.

### DIFF
--- a/cloud-agent-next/src/persistence/CloudAgentSession.ts
+++ b/cloud-agent-next/src/persistence/CloudAgentSession.ts
@@ -72,6 +72,8 @@ import { stopKiloServer } from '../kilo/server-manager.js';
 
 /** Reaper alarm interval: 5 minutes */
 const REAPER_INTERVAL_MS_DEFAULT = 5 * 60 * 1000;
+/** Shorter reaper interval while execution is active: 2 minutes */
+const REAPER_ACTIVE_INTERVAL_MS = 2 * 60 * 1000;
 const PENDING_START_TIMEOUT_MS_DEFAULT = 5 * 60 * 1000;
 
 /** Event retention period: 90 days (aligns with session TTL) */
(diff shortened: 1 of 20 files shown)


cloud-agent-next/src/persistence/CloudAgentSession.ts
```diff
@@ -832,8 +914,10 @@ export class CloudAgentSession extends DurableObject {
         .error('Error during alarm reaper');
     }
 
-    // Schedule next alarm run
-    await this.ctx.storage.setAlarm(now + this.getReaperIntervalMs());
+    // Schedule next alarm run — use shorter interval while an execution is active
+    const activeExecutionId = await this.executionQueries.getActiveExecutionId();
```
jrf0110: Does this need to go in the try/catch? If it throws an error, we'd miss setting the alarm on line 920

jrf0110 (approved):

cloud-agent-next/src/persistence/CloudAgentSession.ts
eshurakov: good point, thanks!

cloud-agent-next/src/router/handlers/session-management.ts
```diff
@@ -252,6 +252,58 @@ export function createSessionManagementHandlers() {
               })
               .info('Session interruption completed');
 
+            // If no processes were killed but there's still an active execution,
+            // the wrapper is already dead — clear the stale execution immediately
+            if (result.killedProcessIds.length === 0 && activeExecutionId) {
```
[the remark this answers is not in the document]
eshurakov: fixed in https://github.com/Kilo-Org/cloud/pull/705/commits/9b3903c8bd24fff91f446bad8de35e0f324207b7

PR #713, opened 2026-03-01
chore(cloudflare-app-builder): migrate DO SQLite to Drizzle ORM

iscekic:
Migrates raw sql.exec() queries to drizzle-orm in cloudflare-app-builder.

### DIFF
--- a/cloudflare-app-builder/drizzle.config.ts
+++ b/cloudflare-app-builder/drizzle.config.ts
@@ -1,0 +1,7 @@
+import { defineConfig } from 'drizzle-kit';
+export default defineConfig({
+  out: './drizzle',
+  schema: './src/db/sqlite-schema.ts',
+  dialect: 'sqlite',
+  driver: 'durable-sqlite',
+});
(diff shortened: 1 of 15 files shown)


iscekic:
Post-review fixes (follow-up commit `4262bfdc0`):

1. **Explicit type annotation** — `private db;` in `GitRepositoryDO` now annotated as `DrizzleSqliteDODatabase`, matching the existing explicit annotation in `SqliteFS`.
2. **Redundant select column removed** — `getStorageStats` was selecting `is_dir` despite the WHERE clause already filtering `is_dir = 0`.
3. **Filter pushed into SQL** — `exportGitObjects` adds `eq(gitObjects.is_dir, 0)` to the query, eliminating the JS-side `if (obj.is_dir === 1) continue` skip.
4. **LIKE → parent_path index** — `rmdir`'s empty-dir check was using `LIKE path '.../%'`; replaced with `eq(parent_path, normalized)` which hits the existing `idx_git_objects_parent` index.

jeanduplessis (approved):

PR #720, opened 2026-03-02
Revert "Revert "Add mode, auto model and market cost to microdollar usage""

chrarnoldus:
Reverts Kilo-Org/cloud#719


chrarnoldus:
Removed the FKs which cause the migration to time out

markijbema (approved):

PR #746, opened 2026-03-02
CloudAgentNext - Optimistic message display and failed-send text restoration

eshurakov:
## Summary

- Display the user's message in the chat immediately on send (optimistic UI) instead of waiting for the server echo, making the experience feel instant.
- On send failure, remove the optimistic message and restore the original text back into `ChatInput` so the user can retry without retyping.

## Changes

- **`store/atoms.ts`**: Rename optimistic message ID prefix to `optimistic-` for identification. Add `removeOptimisticMessageAtom` that removes the optimistic message and its parts from the store.
- **`useCloudAgentStream.ts`**: Call `addUserMessage` before the tRPC send to show the message optimistically. Replace the optimistic message when the server echoes the real user message. On failure, remove the optimistic message and invoke `onSendFailed` callback with the original text.
- **`CloudChatContainer.tsx`**: Track `failedMessageText` state, wire `onSendFailed` callback, and pass `chatInputInitialValue` to the presentation layer. Clear the failed text on the next send attempt.
- **`CloudChatPresentation.tsx`**: Thread `chatInputInitialValue` prop down to `ChatInput`.
- **`ChatInput.tsx`**: Accept `initialValue` prop and sync it into the textarea via `useEffect` so failed message text is restored.

### DIFF
--- a/src/components/cloud-agent-next/store/atoms.ts
+++ b/src/components/cloud-agent-next/store/atoms.ts
@@ -287,8 +287,8 @@ export const addUserMessageAtom = atom(
     }
   ): string => {
     const { sessionId, content, agent = 'code', model = { providerID: '', modelID: '' } } = payload;
-    const messageId = `optimistic-${Date.now()}_${Math.random().toString(36).slice(2, 9)}`;
-    const partId = `part_${Date.now()}_${Math.random().toString(36).slice(2, 9)}`;
+    const messageId = `optimistic-${crypto.randomUUID()}`;
+    const partId = `part_${crypto.randomUUID()}`;
     const now = Date.now();
 
     const userMessage: StoredMessage = {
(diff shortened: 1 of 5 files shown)


src/components/cloud-agent-next/store/atoms.ts
```diff
@@ -285,9 +285,9 @@ export const addUserMessageAtom = atom(
       agent?: string;
       model?: { providerID: string; modelID: string };
     }
-  ) => {
+  ): string => {
     const { sessionId, content, agent = 'code', model = { providerID: '', modelID: '' } } = payload;
-    const messageId = `user_${Date.now()}_${Math.random().toString(36).slice(2, 9)}`;
+    const messageId = `optimistic-${Date.now()}_${Math.random().toString(36).slice(2, 9)}`;
```
jrf0110: Why not uuid's?

eshurakov: I feel both ways are fine, but updated to UUID

alex-alecu (approved):

PR #770, opened 2026-03-03
Discord: Add dev gateway cron script

RSO:
The Discord connection uses a WebSocket run through a cron job in production. This PR adds a little script that mimics that behavior. This makes it easier to set up a local version of the Kilo for Discord bot.


.prettierignore
```diff
@@ -22,7 +22,7 @@ public/
 *.generated.*
 *.min.*
 _journal.json
-packages/db/src/migrations/meta/*.json
+meta/*_snapshot.json
```
RSO: Unrelated change, but I didn't feel like making an extra PR 😬

markijbema (approved):

PR #778, opened 2026-03-03
Integrate abuse service auth event tracking

jrf0110:
## Summary

- Adds `reportAuthEvent()` to `src/lib/abuse-service.ts` — POSTs signup/signin events to `POST /api/auth-event` on the abuse service, following the same fail-open pattern as `classifyRequest()` and `reportCost()`.
- Calls `reportAuthEvent()` from `createOrUpdateUser()` in `src/lib/user.ts` for both signin (existing user found) and signup (after DB transaction) paths.
- Threads `requestHeaders` through `createOrUpdateUser()` and the SSO path (`processSSOUserLogin` → `processSSOInternal`) so IP, geo, JA4, and user-agent are available for the payload.
- Reuses the existing `requestHeaders` variable in the Turnstile IP check instead of calling `await headers()` a second time.

### Constraints met
- **Fire-and-forget**: Uses `void reportAuthEvent(...)` — never awaited, never blocks auth.
- **Fail-open**: All errors are caught and logged, never thrown.
- **No new env vars**: Uses existing `ABUSE_SERVICE_URL`, `ABUSE_SERVICE_CF_ACCESS_CLIENT_ID`, `ABUSE_SERVICE_CF_ACCESS_CLIENT_SECRET`.
- **No changes to after-sign-in route or Stytch flow**.

### DIFF
--- a/src/lib/abuse-service.ts
+++ b/src/lib/abuse-service.ts
@@ -189,6 +189,48 @@ export type UsagePayload = {
 };
 
 /**
+ * Shared fetch helper for all abuse service endpoints.
+ * Handles URL check, CF Access auth headers, and fail-open error handling.
+ * Returns the parsed JSON response, or null if the service is unavailable or errored.
+ */
+async function fetchAbuseService<T>(
+  path: string,
+  payload: unknown,
+  label: string
+): Promise<T | null> {
+  if (!ABUSE_SERVICE_URL) {
+    return null;
+  }
+
+  try {
+    const headers: Record<string, string> = {
+      'Content-Type': 'application/json',
+    };
+
+    if (ABUSE_SERVICE_CF_ACCESS_CLIENT_ID && ABUSE_SERVICE_CF_ACCESS_CLIENT_SECRET) {
+      headers['CF-Access-Client-Id'] = ABUSE_SERVICE_CF_ACCESS_CLIENT_ID;
+      headers['CF-Access-Client-Secret'] = ABUSE_SERVICE_CF_ACCESS_CLIENT_SECRET;
+    }
+
+    const response = await fetch(`${ABUSE_SERVICE_URL}${path}`, {
+      method: 'POST',
+      headers,
+      body: JSON.stringify(payload),
+    });
+
+    if (!response.ok) {
+      console.error(`[Abuse] ${label} failed (${response.status}): ${await response.text()}`);
+      return null;
+    }
+
+    return (await response.json()) as T;
+  } catch (error) {
+    console.error(`[Abuse] ${label} error:`, error);
+    return null;
+  }
+}
+
+/**
  * Classify a request for potential abuse.
  * This is called before proxying requests to detect fraudulent activity.
  *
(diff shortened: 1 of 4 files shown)


src/lib/abuse-service.ts
```diff
@@ -314,6 +315,57 @@ export async function reportCost(payload: CostUpdatePayload): Promise<CostUpdate
...
+  user_agent?: string | null;
+  auth_method: AuthProviderId;
+  stytch_session_id?: string | null;
+};
+
+/**
+ * Report an auth event (signup or signin) to the abuse service.
+ * Fire-and-forget: catches all errors, never throws, never blocks auth.
+ */
+export async function reportAuthEvent(payload: AuthEventPayload): Promise<void> {
```
jrf0110, line 321: Now that we have a few functions which share the same fetch pattern implemented here, we should probably make a little `fetchAbuseService` function which handles the ceremony of try/catch and validating the url and credentials

src/lib/abuse-service.ts
```diff
@@ -314,6 +315,57 @@ export async function reportCost(payload: CostUpdatePayload): Promise<CostUpdate
...
+  event_type: 'signup' | 'signin';
+  email: string;
+  account_created_at: string; // ISO 8601
+  ip_address?: string | null;
+  geo_city?: string | null;
+  geo_country?: string | null;
+  ja4_digest?: string | null;
+  user_agent?: string | null;
+  auth_method: AuthProviderId;
+  stytch_session_id?: string | null;
```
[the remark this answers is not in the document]
jrf0110: Added a JSDoc comment to the field: `/** Reserved for future Stytch integration — not populated yet */`. It's part of the abuse service's endpoint contract for forward compatibility, so keeping it in the type is intentional.

src/lib/user.ts
```diff
@@ -158,10 +159,27 @@ export async function findUserByEmail(email: string): Promise<User | undefined>
...
   args: CreateOrUpdateUserArgs,
   turnstile_guid: UUID | undefined,
-  autoLinkToExistingUser: boolean = false
+  autoLinkToExistingUser: boolean = false,
+  requestHeaders?: Headers
 ): Promise<Result<{ user: User; isNew: boolean }, AuthErrorType>> {
   const existingUser = await findAndSyncExistingUser(args);
   if (existingUser) {
+    // Fire-and-forget auth event for signin
+    if (requestHeaders) {
```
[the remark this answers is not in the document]
jrf0110: Good catch — fixed. Added `reportAuthEvent` to the auto-link path (after `linkAccountToExistingUser` succeeds, before the PostHog capture). All three signin paths in `createOrUpdateUser` now report auth events: direct signin, auto-linked signin, and new signup.

src/lib/abuse-service.ts
jrf0110: Done — extracted a `fetchAbuseService<T>(path, payload, label)` helper that handles the ABUSE_SERVICE_URL check, CF Access headers, response parsing, and fail-open error logging. All three functions (`classifyRequest`, `reportCost`, `reportAuthEvent`) now delegate to it.

src/lib/user.ts
```diff
@@ -207,6 +225,21 @@ export async function createOrUpdateUser(
       if (!linkResult.success) {
         return { success: false, error: linkResult.error };
       }
+      // Fire-and-forget auth event for auto-linked signin
```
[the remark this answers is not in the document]
jrf0110: Done — extracted a `fireAuthEvent` helper in `user.ts` that all three call sites now use. Each is a single line: `fireAuthEvent(user, 'signin', args.provider, requestHeaders)`.

src/lib/user.ts
```diff
@@ -155,13 +156,38 @@ export async function findUserByEmail(email: string): Promise<User | undefined>
...
 export async function createOrUpdateUser(
   args: CreateOrUpdateUserArgs,
   turnstile_guid: UUID | undefined,
-  autoLinkToExistingUser: boolean = false
+  autoLinkToExistingUser: boolean = false,
+  requestHeaders?: Headers
 ): Promise<Result<{ user: User; isNew: boolean }, AuthErrorType>> {
   const existingUser = await findAndSyncExistingUser(args);
   if (existingUser) {
+    if (requestHeaders) {
```
jrf0110: We should just make requestHeaders in fireAuthevent optional and move this if (requestHeaders) check in that function to cleanup the noise here

jrf0110: Done — made `requestHeaders` optional in `fireAuthEvent` and moved the guard inside. The three call sites are now just bare calls without the wrapping `if`.

markijbema (approved):

PR #787, opened 2026-03-03
fix(kiloclaw): Handle Fly 400 "no capacity" as a capacity error

pandemicsyn:
## Summary

- Fly returns `400` with `{"error":"no capacity"}` on `createVolume` when a region lacks capacity. This was not recognized by `isFlyInsufficientResources`, so provision failed with an unhandled error instead of falling back to the next region.
- Adds `400` to `CAPACITY_STATUS_CODES` and `"no capacity"` to `CAPACITY_MARKERS`
- Includes the region in `createVolume` error context for easier debugging (e.g. `Fly API createVolume [iad] failed (400)`)

## Capacity error codes handled (after this PR)

| Status | Body marker | During/because |
|--------|------------|---------|
| 400 | `"no capacity"` | `createVolume` — observed in production, not sure why |
| 412 | `"insufficient resources"` | Machine creation on volume-pinned host |
| 409 | `"insufficient memory"` | Host memory exhausted on start/update |
| 403 | `"over the allowed quota"` | Org regional memory quota exceeded |

## Testing

`pnpm typecheck && pnpm test && pnpm lint` — all pass (491 tests).

### DIFF
--- a/kiloclaw/src/fly/client.ts
+++ b/kiloclaw/src/fly/client.ts
@@ -193,7 +193,7 @@ export async function createVolume(
     method: 'POST',
     body: JSON.stringify(request),
   });
-  await assertOk(resp, 'createVolume');
+  await assertOk(resp, `createVolume [${request.region}]`);
   return resp.json();
 }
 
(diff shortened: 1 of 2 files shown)


kiloclaw/src/fly/client.ts
```diff
@@ -193,7 +193,7 @@ export async function createVolume(
     method: 'POST',
     body: JSON.stringify(request),
   });
-  await assertOk(resp, 'createVolume');
+  await assertOk(resp, `createVolume [${request.region}]`);
```
pandemicsyn, line 196: mostly out of morbid curiosity, but having the region handy here might be helpful later

St0rmz1 (approved):

PR #791, opened 2026-03-04
fix(auto-topup): harden webhook handling, org setup attribution, and observability

jeanduplessis:
## Summary

This PR improves org auto top-up reliability, observability, and setup UX, with focused follow-up fixes from review feedback.

Key changes:
- Strengthened Stripe auto-top-up webhook handling:
  - lock cleanup always runs,
  - original processing errors are preserved if cleanup fails,
  - Sentry logs now include `traceId` and explicit error messages for user/org failure paths.
- Persisted `created_by_user_id` during org auto-topup setup upserts for accurate org billing attribution.
- Added `traceId` propagation into auto-topup invoice metadata for cross-system correlation.
- Tightened org auto-topup API validation by applying `OrgAutoTopUpAmountCentsSchema` to `changePaymentMethod.amountCents`.
- Updated org payment details UI to handle `auto_topup_setup` query params with success/cancelled toast messaging and URL cleanup.
- Added/updated tests for all new behavior (trace metadata, lock release on failure, org attribution persistence, and input validation).
- Removed accidentally tracked Husky shim artifacts and ignored `.husky/_/`.

## Verification

- [x] `pnpm test src/lib/autoTopUp.test.ts src/tests/stripe.test.ts src/routers/organizations/organization-auto-top-up-router.test.ts`
  - Result: 3 suites passed, 66 tests passed
- Manually verified the flow e2e locally with Stripe processing

## Visual Changes

| Before | After |
| ------ | ----- |
| Stripe setup redirect to org payment details had no explicit user feedback. | `auto_topup_setup=success` shows success toast and the query param is removed from the URL after handling. Ghostty 2026-03-04 15 45 29 |

## Reviewer Notes

- Primary review focus: `src/lib/stripe.ts` auto-topup `invoice.paid` flow (error handling, lock release semantics, and Sentry metadata).
- Test mock setup in `src/tests/stripe.test.ts` was adjusted to keep behavior unchanged while enabling failure-path assertions.
- Org attribution path now depends on persisted `created_by_user_id` set during `org-auto-topup-setup`.

### DIFF
--- a/.husky/_/post-checkout
+++ b/.husky/_/post-checkout
@@ -1,0 +1,3 @@
+#!/bin/sh
+command -v git-lfs >/dev/null 2>&1 || { printf >&2 "\n%s\n\n" "This repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting the 'post-checkout' file in the hooks directory (set by 'core.hookspath'; usually '.git/hooks')."; exit 2; }
+git lfs post-checkout "$@"
--- a/.husky/_/post-checkout
+++ b/.husky/_/post-checkout
@@ -1,3 +1,0 @@
-#!/bin/sh
-command -v git-lfs >/dev/null 2>&1 || { printf >&2 "\n%s\n\n" "This repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting the 'post-checkout' file in the hooks directory (set by 'core.hookspath'; usually '.git/hooks')."; exit 2; }
-git lfs post-checkout "$@"
--- a/src/components/organizations/OrganizationPaymentDetails.tsx
+++ b/src/components/organizations/OrganizationPaymentDetails.tsx
@@ -1,7 +1,8 @@
 'use client';
 
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { useSession } from 'next-auth/react';
+import { toast } from 'sonner';
 import type { OrganizationRole, TimePeriod } from '@/lib/organizations/organization-types';
 import { OrganizationContextProvider } from './OrganizationContext';
 import { OrganizationPageHeader } from './OrganizationPageHeader';
@@ -33,6 +34,20 @@ export function OrganizationPaymentDetails({ organizationId, role, isAutoTopUpEn
   const isKiloAdmin = session?.data?.isAdmin ?? false;
   const { data: organizationData } = useOrganizationWithMembers(organizationId);
   const { expiringBlocks, expiring_mUsd, earliestExpiry } = useExpiringCredits(organizationId);
+
+  useEffect(() => {
+    const url = new URL(window.location.href);
+    const setupStatus = url.searchParams.get('auto_topup_setup');
+    if (setupStatus === 'success') {
+      toast.success('Automatic top up enabled');
+    } else if (setupStatus === 'cancelled') {
+      toast.info('Automatic top up setup cancelled');
+    }
+
+    if (!setupStatus) return;
+    url.searchParams.delete('auto_topup_setup');
+    window.history.replaceState({}, '', url.toString());
+  }, []);
 
   return (
     <OrganizationContextProvider value={{ userRole, isKiloAdmin, isAutoTopUpEnabled }}>
@@ -79,6 +94,7 @@ export function OrganizationPaymentDetails({ organizationId, role, isAutoTopUpEn
                     <Tooltip>
                       <TooltipTrigger asChild>
                         <button
+                          type="button"
                           onClick={() => setIsSpendingAlertsModalOpen(true)}
                           className="hover:bg-muted inline-flex cursor-pointer items-center gap-1 rounded p-1 transition-all duration-200 focus:outline-none"
                         >
--- a/src/components/organizations/OrganizationPaymentDetails.tsx
+++ b/src/components/organizations/OrganizationPaymentDetails.tsx
@@ -1,6 +1,7 @@
 'use client';
 
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
+import { useSearchParams, usePathname, useRouter } from 'next/navigation';
 import { useSession } from 'next-auth/react';
 import { toast } from 'sonner';
 import type { OrganizationRole, TimePeriod } from '@/lib/organizations/organization-types';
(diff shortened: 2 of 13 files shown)


src/lib/stripe.ts
```diff
@@ -577,56 +613,94 @@ export async function processStripePaymentEventHook(event: Stripe.Event) {
...
+          } catch (error) {
+            sentryLogger('stripe', 'error')('Auto top-up webhook processing failed for user', {
               invoice_id: invoice.id,
               charge_id: chargeId,
+              kilo_user_id: user.id,
+              traceId,
+              type: 'auto-topup',
             });
+            throw error;
+          } finally {
```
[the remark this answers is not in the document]
jeanduplessis: Updated both auto top-up  blocks to preserve the original processing error. The lock/mark operations are now wrapped in a local , and failures there are reported with  instead of overriding the thrown webhook-processing error.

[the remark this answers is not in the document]
jeanduplessis: Applied fix in src/lib/stripe.ts: both auto top-up finally blocks now wrap markAutoTopUpCompleted/releaseAutoTopUpAttemptLock in a local try/catch and report lock cleanup failures via captureException(lockError). This keeps cleanup errors from replacing the original webhook processing failure.

.husky/_/post-checkout
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
jeanduplessis: Applied fix: removed the committed generated husky shim files under .husky/_/ (post-checkout, post-commit, post-merge, pre-push) and added .husky/_/ to .gitignore so these local artifacts are not tracked again.

src/lib/stripe.ts
```diff
@@ -577,56 +613,102 @@ export async function processStripePaymentEventHook(event: Stripe.Event) {
...
+                kilo_user_id: user.id,
+                invoice_id: invoice.id,
+                charge_id: chargeId,
+                traceId,
+              });
+            }
+
+            processedSuccessfully = true;
+          } catch (error) {
+            sentryLogger('stripe', 'error')('Auto top-up webhook processing failed for user', {
```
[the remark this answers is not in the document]
jeanduplessis: Added  metadata to both auto top-up webhook failure logs in  (user and organization paths) using  so the breadcrumb now includes failure context and an error message for correlation.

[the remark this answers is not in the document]
jeanduplessis: Fix applied: added an `error` field to both catch-block `sentryLogger` metadata objects in `src/lib/stripe.ts` (user and organization auto top-up paths). The value is `error instanceof Error ? error.message : String(error)` so logs now preserve message context for correlation with invoice/charge/trace fields.

src/lib/autoTopUp.ts
```diff
@@ -154,10 +154,19 @@ async function maybePerformAutoTopUpForEntity(entity: AutoTopUpEntity): Promise<
...
+    logExceptInTest(`Auto-top-up skipped for ${entityLabel}: concurrent attempt in progress`, {
+      traceId,
+    });
+    return;
+  }
+
   if (result.success) {
-    logExceptInTest(`Auto-top-up successful for ${entityLabel}`, {
-      ...result,
+    sentryLogger('auto-topup', 'info')(`Auto-top-up successful for ${entityLabel}`, {
```
alex-alecu: This was `logExceptInTest` before — now every successful auto-top-up emits a Sentry `info` event. At scale, is this intentional as a permanent observability signal, or should it revert to a local log once the feature is stable?

src/lib/stripe.ts
```diff
@@ -329,20 +329,24 @@ async function handleOrgAutoTopUpSetup(
     ? parseInt(paymentIntent.metadata.amountCents, 10)
     : DEFAULT_ORG_AUTO_TOP_UP_AMOUNT_CENTS;
 
+  const createdByUserId = paymentIntent.metadata.kiloUserId ?? null;
```
alex-alecu: If `kiloUserId` is missing from the payment intent metadata, this silently sets `created_by_user_id` to null. Given the PR is about accurate org billing attribution, should a missing `kiloUserId` log a warning or fail rather than quietly lose attribution?

src/components/organizations/OrganizationPaymentDetails.tsx
```diff
@@ -33,6 +34,20 @@ export function OrganizationPaymentDetails({ organizationId, role }: Props) {
...
+    const setupStatus = url.searchParams.get('auto_topup_setup');
+    if (setupStatus === 'success') {
+      toast.success('Automatic top up enabled');
+    } else if (setupStatus === 'cancelled') {
+      toast.info('Automatic top up setup cancelled');
+    }
+
+    if (!setupStatus) return;
+    url.searchParams.delete('auto_topup_setup');
+    window.history.replaceState({}, '', url.toString());
```
alex-alecu: This bypasses Next.js routing — `window.location.href` + `history.replaceState` won't update the Next.js router state. Would `useSearchParams()` from `next/navigation` + `router.replace()` be more consistent here?

src/routers/organizations/organization-auto-top-up-router.test.ts
```diff
@@ -275,5 +275,15 @@ describe('organization auto-top-up router', () => {
       expect(result.redirectUrl).toBeDefined();
       expect(typeof result.redirectUrl).toBe('string');
     });
+
+    it('rejects invalid amountCents', async () => {
+      const caller = await createCallerForUser(ownerUser.id);
+      const invalidInput = JSON.parse(
```
alex-alecu: What does the `JSON.parse(JSON.stringify(...))` round-trip accomplish here? The input is already a plain object with primitive values.

src/components/organizations/OrganizationPaymentDetails.tsx
jeanduplessis: I think this is low risk here, but agree that using useSearchParams is more idiomatic and better to be consistent 👍

src/routers/organizations/organization-auto-top-up-router.test.ts
jeanduplessis: Agree, will simplify.

src/lib/stripe.ts
jeanduplessis: Good callout, will fix.

src/lib/autoTopUp.ts
jeanduplessis: I was overlay cautious to improve observability here because of reported issues from the past implementation, but I think a simple successful log is fine since we're more worried about when things go wrong.

jeanduplessis:
## Addressed review feedback from @alex-alecu

Thanks for the thorough review! All four comments have been addressed in 2102e48b:

1. **Sentry noise on success path** (`autoTopUp.ts`) — Reverted the success log from `sentryLogger('auto-topup', 'info')` back to `logExceptInTest`. Failures and warnings still go to Sentry; successful top-ups now only log locally to avoid inflating Sentry event volume at scale.

2. **Silent null for missing `kiloUserId`** (`stripe.ts`) — Rather than adding a defensive warning for an unreachable condition (the caller at `handleSuccessfulChargeWithPayment` already guards and returns early if `kiloUserId` is missing), refactored `handleOrgAutoTopUpSetup` to accept `kiloUserId` as a **required `string` parameter**. This makes the contract explicit and eliminates the silent `?? null` fallback entirely.

3. **`window.history.replaceState` bypassing Next.js router** (`OrganizationPaymentDetails.tsx`) — Replaced raw `window.location.href` + `history.replaceState` with `useSearchParams()`, `usePathname()`, and `router.replace()` from `next/navigation`. Added a `useRef` guard (`hasHandledSetupParam`) to preserve fire-once semantics and prevent duplicate toasts when `searchParams` changes for unrelated reasons.

4. **`JSON.parse(JSON.stringify(...))` in test** (`organization-auto-top-up-router.test.ts`) — Removed the unnecessary serialization round-trip. The test now passes the object literal directly with a targeted `@ts-expect-error` on the invalid `amountCents` value, making the intent explicit.

alex-alecu (approved):

PR #808, opened 2026-03-04
feat(kiloclaw-admin): improve destroyed-instance troubleshooting

pandemicsyn:
## Summary
- expose a new internal KiloClaw debug status endpoint backed by Durable Object debug state so admin can inspect destroy internals
- update admin instance details to show live worker data even when Postgres marks the instance destroyed, including pending destroy identifiers and alarm metadata
- precompute and display the derived Fly app name in Technical Details for faster manual investigation in Fly

## Validation
- pnpm typecheck


src/lib/kiloclaw/fly-app-name.ts
St0rmz1, line 1: This should produce the same output as the other original uses `Web Crypto (crypto.subtle.digest, async, Uint8Array → hex)` , but there's no shared test or assertion proving it, and if things go wrong, they might drift without noticing.

St0rmz1 (approved):
just the 1 suggestion in  kiloclaw/fly-app-name.ts to consider

src/lib/kiloclaw/fly-app-name.ts
pandemicsyn: 👍 fixing

PR #812, opened 2026-03-04
feat: gate org auto top-up UI behind posthog feature flag

jeanduplessis:
## Summary

Replace the `isKiloAdmin` gate on organization auto top-up UI with a PostHog feature flag (`org-auto-topup`), evaluated server-side per organization ID. This allows controlled rollout to specific organizations rather than restricting the feature to Kilo admins only.

- Add `isOrgAutoTopUpFeatureEnabled(organizationId)` helper in `organization-auto-top-up.ts` that checks the flag (always enabled in dev)
- Thread the flag result from server pages through `OrganizationContext` via a new `useIsAutoTopUpEnabled` hook
- Replace `isKiloAdmin` guards with the feature flag in: org dashboard page, payment details page, buy credits dialog, and org info card
- No backend changes — `organizationOwnerProcedure` already provides correct authorization for org owners/billing managers

## Visual Changes

N/A — no visual changes; the same UI elements are shown/hidden, just controlled by a feature flag instead of admin status.

## Reviewer Notes

- The flag is evaluated using `organizationId` as the PostHog `distinctId`, following the existing `auto-triage-feature` pattern. To enable for a specific org, target its UUID in PostHog's flag rules.
- The admin dashboard (`OrganizationAdminDashboard`) does not pass `isAutoTopUpEnabled` into context, so the auto-topup badge won't appear there. This is intentional — that dashboard has its own richer admin controls via `showAdminControls`.
- Kilo admins no longer bypass the gate — if you want an admin to see auto top-up for an org, enable the flag for that org in PostHog.

### DIFF
--- a/src/components/organizations/OrganizationInfoCard.tsx
+++ b/src/components/organizations/OrganizationInfoCard.tsx
@@ -29,6 +29,7 @@ import {
 } from 'lucide-react';
 import {
   useIsKiloAdmin,
+  useIsAutoTopUpEnabled,
   useUserOrganizationRole,
 } from '@/components/organizations/OrganizationContext';
 import BuyOrganizationCreditsDialog from '@/components/payment/BuyOrganizationCreditsDialog';
@@ -178,6 +179,7 @@ function Inner(props: InnerProps) {
   };
 
   const isKiloAdmin = useIsKiloAdmin();
+  const isAutoTopUpEnabled = useIsAutoTopUpEnabled();
   const isInAdminDashboard = isKiloAdmin && showAdminControls;
   const isOrgOwner = useCanManagePaymentInfo();
 
@@ -354,7 +356,7 @@ function Inner(props: InnerProps) {
                   </TooltipContent>
                 </Tooltip>
               </TooltipProvider>
-              {auto_top_up_enabled && isKiloAdmin && (
+              {auto_top_up_enabled && isAutoTopUpEnabled && (
                 <Link
                   href={`/organizations/${id}/payment-details`}
                   className="text-muted-foreground hover:text-foreground text-sm hover:underline"
--- a/src/components/organizations/OrganizationPaymentDetails.tsx
+++ b/src/components/organizations/OrganizationPaymentDetails.tsx
@@ -22,9 +22,10 @@ import { useExpiringCredits } from './useExpiringCredits';
 type Props = {
   organizationId: string;
   role: OrganizationRole;
+  isAutoTopUpEnabled: boolean;
 };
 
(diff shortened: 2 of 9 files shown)


alex-alecu (commented):
Two small things

src/components/organizations/OrganizationPaymentDetails.tsx
```diff
@@ -54,7 +55,7 @@ export function OrganizationPaymentDetails({ organizationId, role }: Props) {
         />
 
         {/* Buy Credits and Auto Top-Up Section */}
-        {isKiloAdmin && (
+        {isAutoTopUpEnabled && (
```
alex-alecu: What happens to orgs that already have `auto_top_up_enabled=true` in the DB but don't have the `org-auto-topup` PostHog flag enabled yet? The backend (`autoTopUp.ts:119`) keeps charging their card, but this line hides the entire management UI — they can't see or disable their active auto top-up. Shouldn't the UI remain visible when `auto_top_up_enabled` is already `true`, regardless of the feature flag?

src/components/organizations/OrganizationInfoCard.tsx
```diff
@@ -354,7 +356,7 @@ function Inner(props: InnerProps) {
                   </TooltipContent>
                 </Tooltip>
               </TooltipProvider>
-              {auto_top_up_enabled && isKiloAdmin && (
+              {auto_top_up_enabled && isAutoTopUpEnabled && (
```
alex-alecu, line 359: `OrganizationAdminDashboard` renders `OrganizationInfoCard` but creates its own `OrganizationContextProvider` without `isAutoTopUpEnabled` — so `useIsAutoTopUpEnabled()` returns `false` in admin context. Admins lost all auto-top-up visibility (this badge, plus the toggle inside `BuyOrganizationCreditsDialog`). Was that intended? Previously `isKiloAdmin` was `true` here.

src/components/organizations/OrganizationPaymentDetails.tsx
jeanduplessis: We've only had this turned on for the Kilo org for now so I'm not too worried. We intend to roll this out to all users next week and the only reason to introduce the new feature flag system is we have some customers who are willing to beta test the feature with us.

I'll note this as a valuable follow up for consideration.

src/components/organizations/OrganizationInfoCard.tsx
jeanduplessis: THis concern should be addressed through the fix for: https://github.com/Kilo-Org/cloud/pull/812#discussion_r2885418265

alex-alecu (approved):

PR #822, opened 2026-03-04
Remove default model randomization

chrarnoldus:
This default model is not honored by the new clients and there are no plans to do so.

## Verification

- [ ]

## Visual Changes

| Before | After |
| ------ | ----- |
|        |       |


lambertjosh (approved):
Approve from a product perspective. This wasn't working anyway on the CLI/new extension.

PR #828, opened 2026-03-05
feat: add raw openclaw config editor

tspader:
## Summary

We need a way to let users get their machines out of bad states. Since we don't provide full remote access, this PR adds a widget to fetch the `openclaw.json` from the live machine, let the user patch it arbitrarily, and send the updated config back to the machine. OpenClaw hot reloads configs.

Details:
- Added new controller routes, `/_kilo/config/replace` and `/_kilo/config/read`
- Tried to separate "your instance needs to be updated to pull in the new routes" from "failed to reach instance" in the status codes
  - 401 => unauthorized
  - 404 => your instance doesn't have the route yet
  - 409 => anything else (not provisioned, unreachable)
- Lazy load `@monaco-editor/react` for a reasonable inline JSON editor with highlighting + validation
- Don't do any validation beyond "is this valid JSON"; trying to keep track of OpenClaw's config through some Zod schema that we maintain is a fool's errand
  - Similarly, the config payload is an opaque blob in the Zod schema
- Added `Edit Config` to the danger zone since it is dangerous

## Verification

I loaded this up in my dev environment and made sure that:
- You can't send a malformed config
- You CAN send a well-formed config
- Any changed keys persist across restarts
- Refetching the config to hand-edit again always loads the freshest config

## Visual Changes

| Before | After |
| ------ | ----- |
| image | image       |

And the editor proper:
image


kiloclaw/controller/src/routes/config.ts
```diff
@@ -79,6 +92,31 @@ export function registerConfigRoutes(
...
+    }
+
+    if (!body || typeof body !== 'object' || Array.isArray(body)) {
+      return c.json({ error: 'Body must be a JSON object' }, 400);
+    }
+
+    try {
+      fs.writeFileSync(CONFIG_PATH, JSON.stringify(body, null, 2));
+      console.log('[controller] Config replaced');
+      return c.json({ ok: true });
```
St0rmz1, line 162: **WARNING:** Lost update / silent overwrite risk.

Both this editor and OpenClaw's own built-in config editor write to the same file (`/root/.openclaw/openclaw.json`). Since `replace` does a full file overwrite with no concurrency check, if a user (or OpenClaw itself) modifies config between when the KiloClaw UI reads and when it submits, those changes are silently lost.

Consider adding optimistic concurrency control — return a hash on read, require it on replace:

```ts
import crypto from 'node:crypto';

// In the read handler:
const raw = fs.readFileSync(CONFIG_PATH, 'utf8');
const etag = crypto.createHash('md5').update(raw).digest('hex');
return c.json({ config: JSON.parse(raw), etag });

// In the replace handler:
const current = fs.readFileSync(CONFIG_PATH, 'utf8');
const currentEtag = crypto.createHash('md5').update(current).digest('hex');
if (body.etag && body.etag !== currentEtag) {
  return c.json({ error: 'Config was modified since last read — please reload and retry' }, 409);
}
fs.writeFileSync(CONFIG_PATH, JSON.stringify(body.config, null, 2));
```

The frontend would then catch the 409 and prompt the user to reload before re-editing. This is cheap to implement and prevents a real data loss scenario.

kiloclaw/controller/src/routes/config.ts
```diff
@@ -79,6 +92,31 @@ export function registerConfigRoutes(
     }
   });
 
+  // Replace openclaw.json with a JSON blob
+  // No validation is done on the shape of the JSON
+  app.post('/_kilo/config/replace', async c => {
```
pandemicsyn, line 127: When we push in the default config, we also backup the current one. We should probably do the same here.

kiloclaw/controller/src/routes/config.ts
tspader: Pushed, manually verified the case this guards against by modifying the config in another browser tab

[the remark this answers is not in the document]
tspader: `/_kilo/config/patch` does not send SIGUSR1. The `restore` route does, but that's because it's doing a full reboot of the gateway. OpenClaw monitors the config via a file monitor and that's the canonical way to have it catch these updates when not rebooting.

kiloclaw/controller/src/routes/config.ts
```diff
@@ -79,6 +114,53 @@ export function registerConfigRoutes(
...
+    const parsed = ReplaceConfigBodySchema.safeParse(raw);
+    if (!parsed.success) {
+      return c.json({ code: 'invalid_request_body', error: 'Invalid request body' }, 400);
+    }
+
+    const { config, etag } = parsed.data;
+
+    try {
+      if (etag !== undefined) {
+        const current = fs.readFileSync(CONFIG_PATH, 'utf8');
```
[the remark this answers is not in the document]
tspader: This is best effort, optimistic concurrency control. This is not a highly contentious file; we just want to make sure users don't submit changes without realizing that the config changed underneath them

src/app/(app)/claw/components/OpenclawConfigEditor.tsx
```diff
@@ -0,0 +1,217 @@
...
+          <div className="overflow-hidden rounded-md border">
+            <Suspense fallback={<EditorLoading />}>
+              <Editor
+                height="500px"
+                defaultLanguage="json"
+                value={currentEditValue}
+                onChange={handleEditorChange}
+                theme="vs-dark"
+                options={EDITOR_OPTIONS}
+                keepCurrentModel
```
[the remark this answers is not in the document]
tspader: These flags are intentional

src/app/(app)/claw/components/SettingsTab.tsx
```diff
@@ -553,6 +566,16 @@ export function SettingsTab({
                 </>
               )}
             </div>
+
+            {editConfigOpen && (
+              <div className="mt-4">
+                <OpenclawConfigEditor
+                  enabled={isRunning}
```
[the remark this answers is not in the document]
tspader: This is acceptable

kiloclaw/src/durable-objects/kiloclaw-instance.ts
```diff
@@ -1371,6 +1379,15 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
     return parsed.data;
   }
 
+  private isErrorUnknownRoute(error: unknown): boolean {
+    // If a controller predates a new route, the request will either:
+    //   - fall through to the catch-all proxy (401 REQUIRE_PROXY_TOKEN)
+    //   - forward to the gateway which returns 404 for the unknown path.
+    return (
+      error instanceof GatewayControllerError && (error.status === 404 || error.status === 401)
```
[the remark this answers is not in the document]
tspader: This is acceptable

St0rmz1 (commented):
Good work overall — atomic writes, etag concurrency, backup before replace, solid test coverage. Leaving inline comments on a few things worth addressing or documenting.

kiloclaw/controller/src/routes/config.ts
```diff
@@ -79,6 +133,53 @@ export function registerConfigRoutes(
...
+      return c.json({ code: 'invalid_json_body', error: 'Invalid JSON body' }, 400);
+    }
+
+    const parsed = ReplaceConfigBodySchema.safeParse(raw);
+    if (!parsed.success) {
+      return c.json({ code: 'invalid_request_body', error: 'Invalid request body' }, 400);
+    }
+
+    const { config, etag } = parsed.data;
+
```
St0rmz1, line 141: **nit:** The etag check (readFileSync) and the subsequent atomicWrite are separate operations with no file lock, so another concurrent request (or OpenClaw's own config editor) could modify the file in between. I know this is intentional best effort optimistic concurrency, but a future reader seeing `computeEtag(current)` followed by `atomicWrite()` might assume it's atomic. A brief comment here explaining the trade off would save them the trip to this PR thread.

Something like:
```ts
// Best effort optimistic concurrency: the read/check/write is not atomic,
// but sufficient to catch the common case of stale browser tabs. OpenClaw
// monitors the config via file watcher, so worst case is a silent overwrite
// of a nearly simultaneous edit, which is acceptable given low contention
// on this file.
```

kiloclaw/src/durable-objects/kiloclaw-instance.ts
```diff
@@ -1371,6 +1379,15 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
...
+    );
+  }
+
   async getGatewayProcessStatus(): Promise<GatewayProcessStatus> {
     await this.loadState();
     return this.callGatewayController('/_kilo/gateway/status', 'GET', GatewayProcessStatusSchema);
```
St0rmz1: **warning:** `isErrorUnknownRoute` treats all 401s as "controller too old, route unavailable." That works for `getControllerVersion` but extending it to `getOpenclawConfig` and `replaceConfigOnMachine` means a real auth failure (bad token, wrong machine) on these new routes gets silently swallowed and returned as `null` instead of surfacing the actual problem. 409 from etag conflicts propagates correctly since it's not caught here, so the main concern is 401.

This is an existing issue but this PR widens the blast radius. Consider either narrowing the check (e.g. also matching on the error message/code for `REQUIRE_PROXY_TOKEN`) or adding a comment noting the known limitation.

kiloclaw/controller/src/config-writer.test.ts
```diff
@@ -317,6 +322,50 @@ describe('generateBaseConfig', () => {
...
+      'openclaw.json.bak.2026-02-24T10-00-00.000Z',
+      'openclaw.json.bak.2026-02-25T10-00-00.000Z',
+      'openclaw.json.bak.2026-02-26T10-00-00.000Z',
+    ]);
+
+    backupConfigFile('/tmp/openclaw.json', harness.deps);
+
+    expect(harness.unlinked).toHaveLength(7 - MAX_CONFIG_BACKUPS);
+    expect(harness.unlinked[0]).toBe('/tmp/openclaw.json.bak.2026-02-20T10-00-00.000Z');
+    expect(harness.unlinked[1]).toBe('/tmp/openclaw.json.bak.2026-02-21T10-00-00.000Z');
```
St0rmz1, line 390: **bug:** This assertion is off by one. `backupConfigFile()` copies a new `.bak` file *before* it calls `readdirSync()` to prune, so a directory with these 7 existing backups will actually have 8 backup entries when pruning runs. Since the mock's `dirEntries` doesn't include the just copied backup, the test asserts `7 - MAX_CONFIG_BACKUPS` removals instead of the real `8 - MAX_CONFIG_BACKUPS`. The test passes but doesn't match real pruning behavior.

Fix: either add the newly copied backup to `dirEntries` in the mock, or have `setDirEntries` automatically include it.

src/app/(app)/claw/components/SettingsTab.tsx
```diff
@@ -553,6 +566,16 @@ export function SettingsTab({
...
+                  mutations={mutations}
+                  onOpenChange={setEditConfigOpen}
```
St0rmz1, line 590: **suggestion:** `editConfigOpen` keeps the editor mounted even if the instance transitions to stopped/destroyed. The `enabled` prop only controls refetching, not interactivity, so a user could still click Submit against a stopped machine and only discover it via a backend error.

Consider adding an effect that closes the editor when the instance stops:
```ts
useEffect(() => {
  if (!isRunning) setEditConfigOpen(false);
}, [isRunning]);
```

kiloclaw/src/routes/platform.ts
```diff
@@ -69,6 +69,13 @@ function jsonError(message: string, status: number): Response {
   });
 }
 
+function openclawConfigJsonError(message: string, status: number, code?: string): Response {
+  return new Response(JSON.stringify({ error: message, ...(code ? { code } : {}) }), {
+    status,
+    headers: { 'content-type': 'application/json' },
```
St0rmz1, line 70: **nit:** This is very similar to the existing `jsonError` helper just above. Consider having this call `jsonError` internally and just spread the `code` into the body, or consolidate them into one function with an optional `code` param. Avoids drift if one gets updated and the other doesn't.

src/app/(app)/claw/components/OpenclawConfigEditor.tsx
```diff
@@ -0,0 +1,217 @@
...
+}
+
+function isJsonObject(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value);
+}
+
+const EDITOR_OPTIONS = {
+  minimap: { enabled: false },
+  scrollBeyondLastLine: false,
+  fontSize: 13,
```
St0rmz1, line 36: **nit:** This same `isJsonObject` helper also exists in `kiloclaw/controller/src/routes/config.ts`. Not a big deal since they're in different packages, but worth noting in case you want to share it.

tspader: I couldn't find a reasonable package to share it that already existed so I punted

kiloclaw/src/routes/platform.ts
tspader: Done

src/app/(app)/claw/components/SettingsTab.tsx
tspader: Done

kiloclaw/controller/src/config-writer.test.ts
tspader: Done

kiloclaw/src/durable-objects/kiloclaw-instance.ts
tspader: I just went ahead and tried to do it the right way. Machines which don't have the proxy update from my last commit will get less helpful errors but new deploys will be good. Leaving unresolved in case you want to look.

kiloclaw/controller/src/routes/config.ts
tspader: Done

src/routers/kiloclaw-router.ts
```diff
@@ -735,4 +766,80 @@ export const kiloclawRouter = createTRPCRouter({
...
+  replaceOpenclawConfig: baseProcedure
+    .input(z.object({ config: z.record(z.string(), z.unknown()), etag: z.string().optional() }))
+    .mutation(async ({ ctx, input }) => {
+      try {
+        const client = new KiloClawInternalClient();
+
+        // Fetch the current config so we can restore any redacted secrets
+        // that the user didn't change (they'll still have the placeholder).
+        const current = await client.getOpenclawConfig(ctx.user.id);
+        const mergedConfig = restoreRedactedSecrets(input.config, current.config);
```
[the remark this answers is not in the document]
evanjacobson: This seems fine for now, as there is a workaround for users who decide to corrupt their own config. 

Editing via a form should be considered at some point in the future to prevent undesirable changes like this.

src/lib/kiloclaw/config-redaction.ts
```diff
@@ -0,0 +1,230 @@
...
+function walkAndRestore(obj: Record<string, unknown>, current: Record<string, unknown>): void {
+  for (const [key, value] of Object.entries(obj)) {
+    if (value === REDACTED_PLACEHOLDER && isSecretKey(key)) {
+      const realValue = current[key];
+      if (realValue !== undefined && realValue !== null) {
+        obj[key] = realValue;
+      } else {
+        delete obj[key];
+      }
+    } else if (Array.isArray(value)) {
```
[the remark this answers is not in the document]
evanjacobson: AFAIK there are no arrays of secrets currently supported in OC config

kiloclaw/src/routes/platform.ts
```diff
@@ -76,27 +76,66 @@ function jsonError(message: string, status: number): Response {
...
  */
 const SAFE_ERROR_PREFIXES = [
-  'Instance is not ', // e.g. "Instance is not running", "Instance is not provisioned"
+  'Instance is not ', // e.g. "Instance is not running"
+  'Instance not ', // e.g. "Instance not provisioned" (DO uses both forms)
   'User already has an ', // duplicate provision
   'Gateway controller ', // already sanitized at DO level
+  'Config was modified ', // etag mismatch on config replace
   'Invalid secret patch: ', // catalog validation (allFieldsRequired, etc.)
+  'Config was modified ', // etag mismatch on config replace
```
St0rmz1, line 86: Duplicate entry for '  'Config was modified ' ?

St0rmz1 (commented):
**nit (non-blocking):** `sanitizeOpenclawConfigError` forwards `code` to the client on all three paths, including the generic fallback where the message is intentionally redacted. Today the codes are all safe constants (`config_read_failed`, `config_etag_conflict`, etc.), so this is fine in practice. But as a future hardening consideration, Path 3 could restrict code passthrough to the `OPENCLAW_CONFIG_ERROR_CODES` allow list, same as it does for messages, so a future controller change can't accidentally leak internal details via the code field:

```ts
// Current
return { message: `${operation} failed`, status, ...(code ? { code } : {}) };

// Hardened
return { message: `${operation} failed`, status, ...(code && OPENCLAW_CONFIG_ERROR_CODES.has(code) ? { code } : {}) };
```

Not blocking since the tRPC layer in `kiloclaw-router.ts` provides a second defense via `UNSAFE_ERROR_CODES`.

St0rmz1 (approved):

PR #856, opened 2026-03-05
feat(kiloclaw): switch tool profile to full and extend safeBins

pandemicsyn:
## Summary

- OpenClaw's `onboard --non-interactive` defaults to `tools.profile: "messaging"`, which only exposes 5 session/message tools. This left new KiloClaw instances without filesystem access (`read`/`write`/`edit`), shell execution (`exec`), web tools, or the `message` tool — effectively a non-functional coding assistant.
- On fresh install, overrides `tools.profile` to `"full"` (no tool restrictions). The startup script sets a `KILOCLAW_FRESH_INSTALL` flag when onboard runs (no config file existed) and uses it to decide whether to override the profile. On subsequent boots the user's choice is never touched.
- The `exec` tool remains gated by `security: "allowlist"` with `ask: "on-miss"`.
- The config-writer (Restore Default Config path) always sets `tools.profile` to `"full"` since it regenerates a fresh config.
- Adds a changelog entry for the tool profile change.
- Bumps Dockerfile cache bust to v58.

## Verification

- [x] `pnpm typecheck` — passed (tsgo, no errors)
- [x] `pnpm test` — passed (504 tests, 27 test files, 0 failures)
- [x] Manual deploy verification

Screenshot 2026-03-05 at 12 47 02 PM

Screenshot 2026-03-05 at 12 48 33 PM

## Visual Changes

N/A

## Reviewer Notes

- `start-openclaw.sh` and `controller/src/config-writer.ts` must produce equivalent config — both were updated in lockstep. The shell script uses a `KILOCLAW_FRESH_INSTALL` env var to gate the profile override; the config-writer always sets `"full"` because it only runs on explicit config restore.
- `tools.profile` is only set on fresh install. Existing instances keep whatever they have — users can change their profile via the Control UI (Settings > Config > Tools > Tool Profile).

### DIFF
--- a/kiloclaw/controller/src/config-writer.test.ts
+++ b/kiloclaw/controller/src/config-writer.test.ts
@@ -94,7 +94,7 @@ describe('generateBaseConfig', () => {
     expect(config.tools.exec.ask).toBe('on-miss');
 
     // Safe bins
-    expect(config.tools.exec.safeBins).toEqual(['rg', 'git', 'node', 'pnpm', 'go']);
+    expect(config.tools.exec.safeBins).toEqual(['rg', 'git', 'gh', 'node', 'pnpm', 'go']);
   });
 
   it('always sets tool profile to full on restore', () => {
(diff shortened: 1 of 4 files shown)


pandemicsyn:
Docs PR to go with this: https://github.com/Kilo-Org/kilocode/pull/6677

tspader (approved):

PR #868, opened 2026-03-05
fix(gastown): handle Secrets Store fetch failures gracefully

jrf0110:
## Summary

In production, `SecretsStoreSecret.get()` can fail with an opaque `"Secrets Worker: Failed to fetch secret"` error (transient Secrets Store outage, misconfigured store, etc.). Previously this exception was unhandled in `resolveSecret` and bubbled up through the Hono middleware chain as an unhandled error with an unhelpful stack trace.

This PR:
- Changes `resolveSecret` to catch `.get()` failures, log the original error, and return `null` instead of throwing.
- Adds `null` checks at all 4 call sites (`kilo-auth`, `auth`, `mayor-auth` middlewares and `mintKilocodeToken` in the tRPC router) so they return a clean 500 / `INTERNAL_SERVER_ERROR` with descriptive log messages.
- Fixes misleading "not configured" error messages in existing null guards — the binding exists but the fetch failed, which is a different failure mode.

## Verification

- [x] `pnpm typecheck` — all packages pass
- [x] `pnpm --filter cloudflare-gastown test` — pre-existing failures in `container/plugin/client.test.ts` only (unrelated to these changes)
- [x] Manually confirmed secret was recreated and resolved successfully in production

## Visual Changes

N/A

## Reviewer Notes

The root cause in this instance was a stale/deleted secret in the Cloudflare Secrets Store that needed to be recreated. These changes ensure that if the same situation recurs, the error is caught and logged clearly rather than surfacing as an opaque unhandled exception.

### DIFF
--- a/cloudflare-gastown/src/middleware/auth.middleware.ts
+++ b/cloudflare-gastown/src/middleware/auth.middleware.ts
@@ -47,7 +47,7 @@ export const authMiddleware = createMiddleware<GastownEnv>(async (c, next) => {
 
   const secret = await resolveSecret(c.env.GASTOWN_JWT_SECRET);
   if (!secret) {
-    console.error('[auth] GASTOWN_JWT_SECRET not configured');
+    console.error('[auth] failed to resolve GASTOWN_JWT_SECRET from Secrets Store');
     return c.json(resError('Internal server error'), 500);
   }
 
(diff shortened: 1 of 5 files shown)


jrf0110:
Addressed the review suggestion in d2b7618d:

- Both the new null-check TRPCError and the pre-existing `NEXTAUTH_SECRET not configured` error (line 104) now log details server-side via `console.error` and use a generic `'Internal server error'` in the TRPCError message sent to the client.
- This aligns with how the middleware handlers already operate (descriptive server logs, generic client-facing messages).

jeanduplessis (approved):

PR #878, opened 2026-03-06
feat(bot): add request logging for the new bot

RSO:
## Summary

- Add a `bot_requests` table to track each message handled by the new bot (`src/lib/bot.ts`), with a `created_by` (always a user ID), optional `organization_id` (set when the platform integration belongs to an org), and a `steps` jsonb field capturing per-step agent loop data (tool calls, tool results, token usage, finish reason).
- Rows are inserted as `pending` on message receipt and progressively updated via `onStepFinish` callbacks during the agent loop, then finalized with `completed`/`error` status. The `updated_at` column supports this pattern and future async callback URL continuation.
- GDPR: `bot_requests` rows are hard-deleted in `softDeleteUser`, with a corresponding two-user isolation test.

## Verification

- [x] `pnpm drizzle generate` — migration `0045_special_iron_fist.sql` generated

## Visual Changes

N/A

## Reviewer Notes

- The `steps` jsonb field uses `BotRequestStep[]` typed via `$type<>()`. The shape captures step number, finish reason, tool calls (name + args), tool results (name + output), and token usage per step. This is intentionally flexible for future additions without schema migrations.
- `updateBotRequest` calls inside `onStepFinish` and the `spawnCloudAgentSession` callback are fire-and-forget (no `await`) to avoid blocking the agent loop. The final update in the try/catch *is* awaited.
- Unlike `slack_bot_requests` which uses `owned_by_user_id`/`owned_by_organization_id` (mirroring the platform integration's ownership), this table uses `created_by` (always the resolved kilo user ID) and `organization_id` (only when the integration is org-owned). This is a deliberate simplification since we always know the user by the time we create the log row.

### DIFF
--- a/packages/db/src/migrations/0045_special_iron_fist.sql
+++ b/packages/db/src/migrations/0045_special_iron_fist.sql
@@ -1,0 +1,27 @@
+CREATE TABLE "bot_requests" (
+	"id" uuid PRIMARY KEY DEFAULT pg_catalog.gen_random_uuid() NOT NULL,
+	"created_by" text NOT NULL,
+	"organization_id" uuid,
+	"platform_integration_id" uuid,
+	"platform" text NOT NULL,
+	"platform_thread_id" text NOT NULL,
+	"platform_message_id" text,
+	"user_message" text NOT NULL,
+	"status" text DEFAULT 'pending' NOT NULL,
+	"error_message" text,
+	"model_used" text,
+	"steps" jsonb,
+	"cloud_agent_session_id" text,
+	"response_time_ms" integer,
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"updated_at" timestamp with time zone DEFAULT now() NOT NULL
+);
+--> statement-breakpoint
+ALTER TABLE "bot_requests" ADD CONSTRAINT "bot_requests_created_by_kilocode_users_id_fk" FOREIGN KEY ("created_by") REFERENCES "public"."kilocode_users"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "bot_requests" ADD CONSTRAINT "bot_requests_organization_id_organizations_id_fk" FOREIGN KEY ("organization_id") REFERENCES "public"."organizations"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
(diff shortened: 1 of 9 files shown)


src/lib/bot/request-logging.ts
```diff
@@ -0,0 +1,80 @@
...
+        platform_message_id: params.platformMessageId ?? null,
+        user_message: params.userMessage,
+        model_used: params.modelUsed ?? null,
+        status: 'pending',
+      })
+      .returning({ id: bot_requests.id });
+
+    return row?.id;
+  } catch (error) {
+    console.error('[Bot:RequestLog] Failed to create request log:', error);
```
RSO: Let's log to Sentry as well

packages/db/src/migrations/0045_special_iron_fist.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TABLE "bot_requests" (
+	"id" uuid PRIMARY KEY DEFAULT pg_catalog.gen_random_uuid() NOT NULL,
+	"created_by" text NOT NULL,
+	"organization_id" uuid,
+	"platform_integration_id" uuid,
+	"platform" text NOT NULL,
+	"platform_thread_id" text NOT NULL,
+	"platform_message_id" text,
+	"user_message" text NOT NULL,
+	"status" text DEFAULT 'pending' NOT NULL,
```
marius-kilocode, line 10: `auto_fix_tickets_status_check` `auto_triage_tickets_status_check` and `security_analysis_queue_status_check` have DB level constraints on `status`. This one hasn't. Should we not add one here?

src/lib/bot/request-logging.ts
```diff
@@ -0,0 +1,85 @@
...
+    captureException(error, { tags: { component: 'bot-request-log', op: 'update' } });
+  }
+}
+
+/**
+ * Schedule an update to an existing bot_requests row via `after()`.
+ * The write is deferred so it never blocks bot message processing.
+ */
+export function updateBotRequest(id: string, params: UpdateBotRequestParams): void {
+  after(() => performUpdate(id, params));
```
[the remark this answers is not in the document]
marius-kilocode: Maybe right?

marius-kilocode (approved):

src/lib/bot/request-logging.ts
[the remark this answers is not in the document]
RSO: Yeah, but I think it’s fine

packages/db/src/migrations/0045_special_iron_fist.sql
RSO: I’m fine with just restricting it through typing. Don’t want to have to write a migration when I want to add another status

PR #883, opened 2026-03-06
feat(security-agent): add dashboard with SLA and severity metrics

jeanduplessis:
## Summary

Restructures the Security Agent from a single-page tab-based layout into a nested route layout with three sub-pages — **Dashboard** (default), **Findings**, and **Config** — connected by a shared sub-navigation bar.

Adds a new `getDashboardStats` tRPC endpoint that aggregates seven metric categories from the existing `security_findings` table in a single round-trip:

- **SLA compliance** — overall and per-severity, excluding findings without `sla_due_at`
- **Severity breakdown** — open finding counts by critical/high/medium/low
- **Status distribution** — open/fixed/ignored counts
- **Analysis coverage** — analyzed vs total, with outcome breakdown (exploitable, not exploitable, safe to dismiss, needs review, triage complete, analyzing, failed, not analyzed)
- **Mean time to resolution** — avg days to fix per severity vs configured SLA targets
- **Overdue findings** — top 10 open findings past SLA, sorted by most overdue
- **Repository health** — repos ranked by risk (critical count, overdue count, SLA compliance %)

Architecture:

- `SecurityAgentContext` — React context providing shared state (permissions, config, mutations) across all sub-pages, extracted from the monolithic `SecurityAgentPageClient`
- `SecurityAgentLayout` — shared layout with sub-nav (Dashboard/Findings/Config) and permission/integration alerts
- `getDashboardStats` — runs 7 independent DB aggregations in `Promise.all` for parallel execution; no new tables or migrations required
- Dashboard supports per-repository filtering via an optional `repoFullName` parameter
- All clickable dashboard elements navigate to the findings page with matching URL params

Both personal (`/security-agent`) and organization (`/organizations/[id]/security-agent`) routes are supported.

## Verification

- [x] `pnpm typecheck` — passes (pre-existing errors in `chat` adapter modules unrelated to this PR)
- [x] Manual verification — logged in as fake admin user, navigated all three sub-pages (Dashboard, Findings, Config), verified rendering and navigation
- [x] Route structure — confirmed `/security-agent`, `/security-agent/findings`, `/security-agent/config` all resolve correctly
- [x] Screenshots captured via agent-browser automation

## Visual Changes

| Dashboard |
|--------|
|dashboard| 

## Reviewer Notes

- The `getDashboardStats` query function in `src/lib/security-agent/db/dashboard-stats.ts` is the core backend addition. It uses raw SQL for JSON-path extraction on the `analysis` column — if these queries become slow on large datasets, expression indexes can be added later (see `plans/security-agent-dashboard.md` §4).
- `SecurityAgentContext` (~600 lines) is a straight extraction from `SecurityAgentPageClient` — all the mutation logic, permission checks, and state management that was duplicated between findings/config is now centralized. The original `SecurityAgentPageClient` still exists for backward compatibility but delegates to the context.
- The `OutcomeFilterSchema` used by the findings list is coupled with `getOutcome()` in `SecurityFindingRow.tsx` and the `listSecurityFindings` DB query. All three must agree on category definitions.
- No new DB migrations, cron jobs, or tables required.


src/components/security-agent/SecurityAgentContext.tsx
```diff
@@ -0,0 +1,600 @@
...
+        personalDeleteFindingsMutate({ repoFullName });
+      }
+    },
+    [isOrg, organizationId, orgDeleteFindingsMutate, personalDeleteFindingsMutate]
+  );
+
+  const hasIntegration = permissionData?.hasIntegration ?? false;
+  const hasPermission = permissionData?.hasPermissions ?? false;
+  const reauthorizeUrl = permissionData?.reauthorizeUrl ?? undefined;
+  const isEnabled = configData?.isEnabled ?? false;
```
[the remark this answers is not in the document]
jeanduplessis: Fixed. `isEnabled` is now `boolean | undefined` — it stays `undefined` while the config query is loading, and the dashboard pages check `isEnabled === false` (strict equality) before redirecting. A loading spinner is shown while the config resolves, preventing the premature redirect to `/config`. Downstream consumers that need a boolean (`SecurityFindingsCard`, `SecurityConfigForm`) coerce with `?? false`.

src/components/security-agent/SecurityDashboard.tsx
```diff
@@ -0,0 +1,172 @@
...
+  const overdue = data?.overdue ?? [];
+  const repoHealth = data?.repoHealth ?? [];
+
+  const lastUpdated = dataUpdatedAt ? new Date(dataUpdatedAt).toLocaleTimeString() : null;
+
+  return (
+    <div className="space-y-6">
+      {/* Header: filters and actions */}
+      <div className="flex flex-wrap items-center justify-between gap-3">
+        <RepositoryFilter
```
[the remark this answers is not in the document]
jeanduplessis: Fixed. All dashboard drill-down components now accept an `extraParams` prop. When a repository is selected in the dashboard filter, the `repoFullName` is appended to every drill-down URL (SLA, severity, status, analysis, overdue, and repo health links). This ensures clicking any card/chart/table row navigates to the findings page scoped to the selected repository.

src/components/security-agent/SecurityFindingsPage.tsx
```diff
@@ -0,0 +1,287 @@
...
+    startingAnalysisIds,
+    gitHubError,
+  } = useSecurityAgent();
+
+  const trpc = useTRPC();
+  const searchParams = useSearchParams();
+
+  // Initialize filters from URL search params
+  const initialFilters = useMemo(() => {
+    const status = searchParams.get('status') ?? 'open';
```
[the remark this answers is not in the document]
jeanduplessis: Fixed. The default `status=open` is now skipped when `outcomeFilter` implies its own status (`fixed` or `dismissed`). Deep links like `?outcomeFilter=fixed` will now work correctly without a contradictory status filter.

src/components/security-agent/SecurityDashboard.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  const sla = data?.sla ?? emptySla;
+  const severity = data?.severity ?? emptySeverity;
+  const status = data?.status ?? emptyStatus;
+  const analysis = data?.analysis ?? emptyAnalysis;
+  const mttr = data?.mttr ?? emptyMttr;
+  const overdue = data?.overdue ?? [];
+  const repoHealth = data?.repoHealth ?? [];
+
+  const lastUpdated = dataUpdatedAt
```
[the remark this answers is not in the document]
jeanduplessis: Fixed. Replaced `dataUpdatedAt` (React Query fetch time) with a real `getLastSyncTime` query that reads `max(last_synced_at)` from the database. The dashboard now shows the actual last sync timestamp, not when the query was last fetched.

src/lib/security-agent/db/dashboard-stats.ts
```diff
@@ -0,0 +1,429 @@
...
+
+      // Analysis coverage query (open only)
+      db.execute<AnalysisRow>(sql`
+          SELECT
+            COUNT(*) AS total,
+            COUNT(*) FILTER (WHERE ${security_findings.analysis_status} = 'completed') AS analyzed,
+            COUNT(*) FILTER (WHERE ${security_findings.analysis_status} = 'completed' AND (${security_findings.analysis}->'sandboxAnalysis'->>'isExploitable') = 'true') AS exploitable,
+            COUNT(*) FILTER (WHERE ${security_findings.analysis_status} = 'completed' AND (${security_findings.analysis}->'sandboxAnalysis'->>'isExploitable') = 'false') AS not_exploitable,
+            COUNT(*) FILTER (WHERE ${security_findings.analysis_status} = 'completed' AND (${security_findings.analysis}->'sandboxAnalysis'->>'isExploitable') IS NULL AND (${security_findings.analysis}->'triage'->>'suggestedAction') = 'analyze_codebase') AS triage_complete,
+            COUNT(*) FILTER (WHERE ${security_findings.analysis_status} = 'completed' AND (${security_findings.analysis}->'triage'->>'suggestedAction') = 'dismiss') AS safe_to_dismiss,
```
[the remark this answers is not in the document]
jeanduplessis: Fixed. Updated the `safe_to_dismiss` and `needs_review` COUNT FILTER clauses to exclude findings where `sandboxAnalysis.isExploitable` has a definitive boolean result (true/false), only counting those where it's NULL or 'unknown'. This matches the exclusion logic in `listSecurityFindings`.

src/components/security-agent/SecurityFindingsPage.tsx
```diff
@@ -0,0 +1,294 @@
...
+  const serverRunningCount = findingsData?.runningCount ?? 0;
+  const concurrencyLimit = findingsData?.concurrencyLimit ?? 3;
+
+  const runningCount = useMemo(() => {
+    if (startingAnalysisIds.size === 0) return serverRunningCount;
+    let optimisticAdditional = 0;
+    for (const id of startingAnalysisIds) {
+      const finding = findings.find(f => f.id === id);
+      if (
+        !finding ||
```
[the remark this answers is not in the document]
jeanduplessis: Valid catch. The `!finding` branch was adding +1 for IDs not on the current page, but `serverRunningCount` is already a global count that includes those jobs. Fixed by only counting optimistic additions for findings that *are* visible on the current page but haven't transitioned to pending/running yet. Off-page findings are already reflected in `serverRunningCount`.

src/components/security-agent/SecurityFindingsPage.tsx
```diff
@@ -0,0 +1,294 @@
...
+  const handleOpenDismissDialog = useCallback(() => {
+    setDetailDialogOpen(false);
+    setDismissDialogOpen(true);
+  }, []);
+
+  const handleDismissSubmit = useCallback(
+    (reason: DismissReason, comment?: string) => {
+      if (!selectedFinding) return;
+      handleDismiss(selectedFinding, reason, comment);
+      setDismissDialogOpen(false);
```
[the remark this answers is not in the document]
jeanduplessis: Valid. Fixed by deferring the dialog close until the mutation settles. A `dismissSubmittedRef` tracks that a submit was initiated, and an effect watches `isDismissing` transition from true → false to close the dialog and clear state. The `DismissFindingDialog` now also receives the real `isDismissing` pending state (was hardcoded to `false`) so the submit button shows a loading indicator and is disabled during the mutation.

src/lib/security-agent/db/dashboard-stats.ts
```diff
@@ -0,0 +1,429 @@
...
+          WHERE ${whereClause}
+          GROUP BY ${security_findings.status}
+        `),
+
+      // Analysis coverage query (open only)
+      db.execute<AnalysisRow>(sql`
+          SELECT
+            COUNT(*) AS total,
+            COUNT(*) FILTER (WHERE ${security_findings.analysis_status} = 'completed') AS analyzed,
+            COUNT(*) FILTER (WHERE ${security_findings.analysis_status} = 'completed' AND (${security_findings.analysis}->'sandboxAnalysis'->>'isExploitable') = 'true') AS exploitable,
```
alex-alecu, line 215: The subcategories (`exploitable`, `not_exploitable`, `triage_complete`, `safe_to_dismiss`, `needs_review`) don't cover every `completed` finding. For example, if `isExploitable` is `'unknown'` and `suggestedAction` is `'analyze_codebase'`, it counts toward `analyzed` but none of the buckets below. Won't the pie chart silently under-count some findings?

src/components/security-agent/SecurityFindingsPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+  const initialFilters = useMemo(() => {
+    const statusParam = searchParams.get('status') ?? undefined;
+    const severity = searchParams.get('severity') ?? undefined;
+    const repoFullName = searchParams.get('repoFullName') ?? undefined;
+    const outcomeFilter = searchParams.get('outcomeFilter') ?? undefined;
+    const overdue = searchParams.get('overdue');
+
+    const outcomeImpliesStatus = outcomeFilter === 'fixed' || outcomeFilter === 'dismissed';
+
+    const status =
```
alex-alecu: When `overdue=true`, this sets `status = 'open'` and changes the sort order, but doesn't actually filter to findings where `sla_due_at < now()`. The "View all overdue" link from the dashboard will show every open finding (overdue ones just sorted first). Should there be real server-side filtering here so the count and pages only include overdue items?

src/components/security-agent/SecurityFindingsPage.tsx
```diff
@@ -0,0 +1,316 @@
...
+  const handleFindingClick = useCallback((finding: SecurityFinding) => {
+    setSelectedFinding(finding);
+    setDetailDialogOpen(true);
+  }, []);
+
+  // Auto-open finding from URL param
+  useEffect(() => {
+    if (findingIdParam && findings.length > 0 && !selectedFinding) {
+      const finding = findings.find(f => f.id === findingIdParam);
+      if (finding) {
```
alex-alecu: This only looks through the current page of results (first 20). If the finding from `findingId` isn't on page 1 (e.g. user clicked an overdue finding from the dashboard that doesn't appear in the default page), the dialog silently won't open. Would it be safer to fetch the single finding by ID separately when this param is set?

src/lib/security-agent/db/dashboard-stats.ts
```diff
@@ -0,0 +1,429 @@
...
+  if ('organizationId' in owner && owner.organizationId) {
+    return { type: 'org', id: owner.organizationId };
+  }
+  if ('userId' in owner && owner.userId) {
+    return { type: 'user', id: owner.userId };
+  }
+  throw new Error('Invalid owner: must have either organizationId or userId');
+}
+
+function buildWhereClause(owner: Owner, repoFullName?: string) {
```
alex-alecu, line 72: This only filters by owner (and optionally a single repo), but doesn't scope to the repos selected in config when `repositorySelectionMode = 'selected'`. If a user deselected a repo that still has old findings, those findings will appear in the severity counts, SLA compliance, and the repo health table. Is that intended, or should the dashboard respect the configured repo selection?

alex-alecu (approved):

src/lib/security-agent/db/dashboard-stats.ts
jeanduplessis: Good catch. The `triage_complete` bucket was only checking `isExploitable IS NULL` but missed the `'unknown'` case. Fixed by aligning with `safe_to_dismiss` and `needs_review` which correctly use `IS NULL OR = 'unknown'`. The filter now reads:
```sql
(isExploitable IS NULL OR isExploitable = 'unknown') AND suggestedAction = 'analyze_codebase'
```
This matches the corresponding list filter in `security-findings.ts` (the `triage_complete` outcome filter case).

src/components/security-agent/SecurityFindingsPage.tsx
jeanduplessis: Valid issue. Added server-side `overdue` filtering:

1. Added `overdue: z.boolean().optional()` to `ListFindingsInputSchema`
2. Added `sla_due_at IS NOT NULL AND sla_due_at < now()` condition in `listSecurityFindings` when `overdue=true`
3. Wired it through the shared handler and the `SecurityFindingsPage` query params

Now the "View all overdue" link passes `overdue=true` which applies real server-side filtering, so the page count and pagination only include actually-overdue findings.

src/components/security-agent/SecurityFindingsPage.tsx
jeanduplessis: Valid. Replaced the page-search approach with a dedicated `getFinding` query (`trpc.securityAgent.getFinding`) that fetches the finding by ID directly, independent of pagination. The query is only enabled when `findingIdParam` is present, and the effect opens the dialog once the finding loads. This works regardless of what page the user is on or what filters are active.

src/lib/security-agent/db/dashboard-stats.ts
jeanduplessis: This is intentional. The dashboard is meant to show the full security posture across all findings owned by the user/org, not just the repos currently selected in config. If a user deselects a repo that still has unresolved findings, those findings are still their responsibility — hiding them from the dashboard would create a false sense of security.

The repo selection in config controls which repos are *synced* (new alerts pulled from GitHub), not which historical findings are visible. The dashboard already supports a `repoFullName` parameter for drilling into a specific repo, and the repo health table surfaces all repos with open findings so nothing is hidden.

Leaving this as-is.

PR #940, opened 2026-03-09
feat(credits): remove free $5 credit for new users

markijbema:
## Summary

- Disables the $5 automatic welcome credit that was granted to new users who pass Turnstile + Stytch validation during signup.
- `handleSignupPromotion()` in `src/lib/stytch.ts` is now a no-op — it retains the function signature for backward compatibility (called from `account-verification/page.tsx`) but no longer grants any credits.
- The `automatic-welcome-credits` promo category in `src/lib/promoCreditCategories.ts` has its `amount_usd` set to `0` as a belt-and-suspenders safeguard.
- The first-day welcome notification ("We added $5 to your balance…") in `src/lib/notifications.ts` is disabled since it no longer applies to new users.
- Existing users who already received the credit are unaffected — no retroactive changes.

## Verification

- [x] `pnpm typecheck` — passes clean across all 29 workspace projects
- [x] Verified test update in `src/lib/stytch.test.ts` — test now asserts no credit transaction is created (DB tests cannot run in this environment due to missing pg connection, but the test logic is correct)
- [x] Confirmed no unused imports remain after removing `grantCreditForCategory`, `subDays`, and `hasReceivedPromotion`

## Visual Changes

N/A

## Reviewer Notes

- The `handleSignupPromotion` function is kept as a no-op rather than fully deleted because `account-verification/page.tsx` calls it. Removing the call entirely would also be fine but this approach is more minimal.
- The `automatic-welcome-credits` category still exists in `promoCreditCategories.ts` (with `amount_usd: 0`) because it's referenced by `welcomeCredits.ts` for checking whether *existing* users have already received any welcome credits (used in admin tools and customer info).
- The first-topup bonus ($20), survey credits, referral bonuses, and other promo mechanisms are **not** affected by this change.

Built for [Mark](https://kilo-code.slack.com/archives/C0A4SA041DE/p1773045615264499?thread_ts=1772897988.867469&cid=C0A4SA041DE) by [Kilo for Slack](https://kilo.ai/features/slack-integration)

### DIFF
--- a/src/lib/notifications.ts
+++ b/src/lib/notifications.ts
@@ -66,7 +66,6 @@ export async function generateUserNotifications(user: User): Promise<KiloNotific
     generateLowCreditNotification,
     generateAutoTopUpNotification,
     generateByokProvidersNotification,
-    generateFirstDayWelcomeNotification,
     generateKiloPassNotification,
   ];
 
@@ -217,12 +216,6 @@ async function generateByokProvidersNotification(user: User): Promise<KiloNotifi
     console.error('[generateByokProvidersNotification]', e);
     return [];
   }
-}
-
-async function generateFirstDayWelcomeNotification(_user: User): Promise<KiloNotification[]> {
-  // Disabled: new users no longer receive free $5 signup credits,
-  // so the welcome notification referencing the bonus no longer applies.
-  return [];
 }
 
 async function generateKiloPassNotification(user: User): Promise<KiloNotification[]> {
--- a/src/lib/notifications.ts
+++ b/src/lib/notifications.ts
@@ -9,9 +9,6 @@ import { summarizeUserPayments } from '@/lib/creditTransactions';
 import { hasOrganizationEverPaid, hasUserEverPaid } from '@/lib/creditTransactions';
 import { cachedPosthogQuery } from '@/lib/posthog-query';
 import * as z from 'zod';
-import { subDays } from 'date-fns';
-import { hasReceivedPromotion } from '@/lib/promotionalCredits';
-
 import { getKiloPassStateForUser } from '@/lib/kilo-pass/state';
 import { db } from '@/lib/drizzle';
 import { fromMicrodollars } from '@/lib/utils';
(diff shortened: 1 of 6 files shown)


src/lib/notifications.ts
```diff
@@ -240,37 +219,10 @@ async function generateByokProvidersNotification(user: User): Promise<KiloNotifi
...
-        actionText: 'Open Kilo-Org/KiloMan',
-        actionURL: 'https://github.com/Kilo-Org/kiloman',
-      },
-      showIn: ['cli', 'extension'],
-    },
-  ];
+async function generateFirstDayWelcomeNotification(_user: User): Promise<KiloNotification[]> {
+  // Disabled: new users no longer receive free $5 signup credits,
+  // so the welcome notification referencing the bonus no longer applies.
+  return [];
```
RSO: Shouldn't we just remove this method altogether?

src/lib/promoCreditCategories.ts
```diff
@@ -171,8 +171,9 @@ const nonSelfServicePromos: readonly NonSelfServicePromoCreditCategoryConfig[] =
   },
   {
     credit_category: 'automatic-welcome-credits',
-    description: 'Free credits for new users who pass both Turnstile and Stytch validation',
-    amount_usd: 5,
+    description:
+      'Free credits for new users who pass both Turnstile and Stytch validation (disabled)',
```
RSO: Instead of marking it as disabled I think we can move it to `promoCreditCategoriesOld.ts`

src/lib/stytch.ts
```diff
@@ -174,23 +173,12 @@ export async function saveFingerprints(
...
-        extra: { userId: user.id, email: user.google_user_email },
-      });
-    }
-  }
+export async function handleSignupPromotion(
+  _user: User,
+  _passedValidations: boolean
+): Promise<void> {
+  // No-op: free signup credits have been removed.
 }
```
RSO: Again: What's the point in having the empty function around instead of just removing the callsite?

RSO (approved):

markijbema:
We are not merging yet, as not all our applications support kilo auto free yet

PR #945, opened 2026-03-09
Remove unused token estimation code

chrarnoldus:


lambertjosh (approved):
Approving, I don't know where this is used personally.

chrarnoldus:
I think it used to be part of the old abuse detection that is now gone.

PR #951, opened 2026-03-09
feat(kiloclaw): add headless Chromium browser support

St0rmz1:
## Summary

  - Install chromium in the Docker image so OpenClaw's built-in browser tool works out of the
  box                                    
  - Configure browser.headless, browser.noSandbox, and browser.enabled in the openclaw.json
  config patch for containerized operation
  - Sync the same browser config to config-writer.ts so config restores don't lose browser
  support

## Verification

  - Deploy a fresh kiloclaw instance (destroy + recreate to trigger fresh install path)
  - Verify which chromium returns /usr/bin/chromium on the Fly machine
  - Verify cat /root/.openclaw/openclaw.json | grep -A4 browser shows enabled: true, headless:
  true, noSandbox: true
  - Verify Chromium processes are running (ps -ef | grep chromium)
  - Ask the agent to use the browser tool (e.g. "use the browser tool to open
  https://example.com and take a screenshot") — confirms end-to-end CDP functionality
  - All 505 kiloclaw tests pass

## Visual Changes
  None — this is backend/infrastructure only.

| Before | After |
| ------ | ----- |
|        |       |

## Reviewer Notes

  - OpenClaw auto-detects /usr/bin/chromium on Linux and auto-adds --disable-dev-shm-usage, so
  no executablePath config is needed
  - noSandbox: true is required in containers — Chromium's setuid sandbox needs kernel
  namespacing that's unavailable in Docker/Fly
  - The chromium apt package installs to /usr/bin/ which is outside the Fly Volume mount at
  /root, so it survives volume mounts and restarts
  - Image size increases ~200-250 MB due to Chromium and its dependencies
  - Existing instances need tools.profile: "full" to see the browser tool — the "messaging"
  profile (onboard default) doesn't include it. Fresh installs already get "full" via the
  startup script

### DIFF
--- a/kiloclaw/Dockerfile
+++ b/kiloclaw/Dockerfile
@@ -5,7 +5,7 @@ ENV NODE_VERSION=22.13.1
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
        ca-certificates curl gnupg git xz-utils unzip jq ripgrep rsync zstd \
-       build-essential python3 ffmpeg tmux \
+       build-essential python3 ffmpeg tmux chromium \
     && ARCH="$(dpkg --print-architecture)" \
     && case "${ARCH}" in \
          amd64) NODE_ARCH="x64" ;; \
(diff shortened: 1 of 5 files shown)


pandemicsyn (approved):
lgtm, mind including a change log entry.

PR #964, opened 2026-03-10
fix(gastown): add JWT token refresh for persistent agents

### LINKED ISSUE #923: Bug: Agent JWT expires after 8h with no refresh — Mayor tool calls 401
## Parent

Part of #204 (Phase 3: Multi-Rig + Scaling)

## Problem

Agent tool calls fail with 401 "Token expired" after 8 hours. The `GASTOWN_SESSION_TOKEN` (agent JWT) is minted once per `startAgentInContainer()` dispatch with an 8-hour expiry, but there is **no token refresh mechanism anywhere in the codebase**. Persistent agents — especially the Mayor — can run far longer than 8 hours without being re-dispatched, causing all API calls back to the worker to fail.

This is happening in production right now: a town created a week ago is getting 401s on the Mayor's tool calls.

### Token lifecycle

1. `mintAgentToken()` (`src/dos/town/container-dispatch.ts:45-58`) signs a JWT with `GASTOWN_JWT_SECRET`, 8-hour expiry, claims `{ agentId, rigId, townId, userId }`
2. `startAgentInContainer()` (`container-dispatch.ts:152-274`) calls `mintAgentToken()` and passes the JWT in the request body's `envVars.GASTOWN_SESSION_TOKEN` to the container's `/agents/start` endpoint
3. The container stores the token on the `ManagedAgent` record (`container/src/process-manager.ts:318`) — this value is never updated
4. The agent's plugin client reads `GASTOWN_SESSION_TOKEN` once at init and uses it for all subsequent API calls (`container/plugin/client.ts:322, 53`)

### The failure path (Mayor)

1. Mayor is dispatched → gets fresh 8h JWT
2. User interacts periodically (within 30-min windows, preventing container sleep)
3. `sendMayorMessage()` (`Town.do.ts:627`) checks container status → sees `isAlive = true` → calls `sendMessageToAgent()` which sends a prompt to the running agent **without minting a new JWT**
4. After 8 hours, the original JWT expires
5. All Mayor tool calls → `authMiddleware` → 401 "Token expired"
6. Heartbeats, event persistence, and completion reports also fail (all use the same stale token)

### Why polecats are less affected

Polecat agents exit on `session.idle` (`process-manager.ts:259`). A typical session completes in minutes to hours. Each new dispatch mints a fresh token. Only the Mayor (and potentially long-running refinery reviews) survive past 8 hours.

### Why container sleep/wake isn't a safety net

Container sleep *would* fix this — on wake, the next `startAgentInContainer()` call mints a fresh JWT. But if the user keeps the container alive by interacting within the 30-min `sleepAfter` window, the container never sleeps and the Mayor accumulates time against its original JWT indefinitely.

## Everything affected by the stale token

| Component | File | What fails |
|---|---|---|
| Tool calls (gt_sling, gt_done, etc.) | `container/plugin/client.ts:53` | All agent API calls back to worker |
| Heartbeats | `container/src/heartbeat.ts:62` | Agent liveness POSTs |
| Event persistence | `container/src/process-manager.ts:107` | Agent event storage in AgentDO |
| Completion reporting | `container/src/completion-reporter.ts:36` | Agent completed callbacks |
| Merge callbacks | `container/src/control-server.ts:268-270` | Review complete notifications |

## Acceptance Criteria

- [ ] Mayor (and any persistent agent) can operate indefinitely without JWT expiration causing 401s
- [ ] Token refresh does not require restarting the agent or container
- [ ] Fresh tokens are delivered to already-running agents
- [ ] The container updates the agent's stored token and plugin client transparently
- [ ] No security regression — tokens should still have bounded expiry (don't "fix" this by making tokens never expire)

## Possible Approaches

The exact approach is left to the implementer, but here are the options identified:

**Option A — Refresh on message delivery**: When `sendMayorMessage()` takes the `isAlive` path, mint a fresh JWT and include it in the `/agents/:id/message` request. The container updates the `ManagedAgent.gastownSessionToken` and the plugin client's Bearer token. Requires the container to have a mechanism for hot-swapping the token on a running agent.

**Option B — Proactive alarm-based refresh**: The TownDO alarm periodically (e.g., every hour) re-mints JWTs for all running agents and pushes them to the container via a new `/agents/:id/refresh-token` endpoint. Container updates the stored token.

**Option C — Container-side 401 retry with token request**: When the plugin client receives a 401, it calls a dedicated (unauthenticated or separately authenticated) token refresh endpoint on the worker, gets a fresh JWT, retries the original request. Requires a way for the container to prove agent identity without the expired JWT.

**Option D — Hybrid**: Combine A (refresh on interaction) with B (alarm-based refresh for agents that don't receive messages but stay running, e.g., refinery during long reviews).

## Notes

- No data migration needed — cloud Gastown hasn't deployed to production (though this bug IS affecting a production town right now)
- `KILOCODE_TOKEN` (Kilo LLM gateway token) has a 30-day expiry and IS persisted on the container DO via `setEnvVar()` — it's not affected by this bug
- The `GASTOWN_SESSION_TOKEN` is intentionally NOT persisted on the container DO (it's per-agent, not per-container). This is correct — the fix should refresh per-agent tokens, not change the storage model.
- `verifyAgentJWT()` in `src/util/jwt.util.ts:18` also sets `maxAge: '8h'` redundantly on top of the `exp` claim — both need to be considered when adjusting expiry

jrf0110:
## Summary

- **Fix for #923**: Agent JWTs (8h expiry) are now proactively refreshed so persistent agents (Mayor, long-running refinery) never encounter 401s on tool calls, heartbeats, or event persistence.
- **Hybrid approach (Option D)**: Fresh tokens are pushed on every user interaction (`sendMayorMessage`/`ensureMayor` isAlive paths) AND proactively by the TownDO alarm (hourly for all working/stalled agents). This covers both the common case (user keeps chatting) and the edge case (agents running without messages).
- **Full token pipeline**: Refresh updates the `ManagedAgent` record (broadcastEvent, completion reporting), the heartbeat module, AND the plugin client instances (tool-call API requests) via a `globalThis` symbol-keyed registry that bridges the container `src/` and `plugin/` TypeScript project roots.

Closes #923

## Verification

- [x] `pnpm typecheck` — passes with zero errors
- [x] `pnpm vitest run` — all 108 tests pass across 7 test files (including 6 new token refresh tests)

## Visual Changes

N/A

## Reviewer Notes

- The plugin client registry uses `Symbol.for('gastown.pluginClientRegistry')` on `globalThis` to bridge the `container/src/` and `container/plugin/` TypeScript project roots (they can't cross-import). Both sides access the same array in the same Bun process. This is a deliberate design choice — the alternative would be restructuring the TS project boundaries.
- `refreshRunningAgentTokens()` is throttled to once per hour via a simple timestamp check (not persisted across DO restarts, which is fine — a restart re-dispatches agents with fresh tokens anyway).
- The `ensureMayor` path fires a `void` (fire-and-forget) token refresh because it returns immediately without waiting for the agent to process a message. The `sendMayorMessage` path awaits the refresh before sending the message to ensure the agent has a valid token for any tool calls triggered by that message.
- Test fixes: Added `townId` to `TEST_ENV` and updated URL expectations to include `/towns/{townId}/` — the test fixture was stale (missing the townId field added during the beads-centric refactor).

### DIFF
--- a/cloudflare-gastown/container/plugin/client.test.ts
+++ b/cloudflare-gastown/container/plugin/client.test.ts
@@ -1,5 +1,12 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
-import { GastownClient, MayorGastownClient, GastownApiError, createClientFromEnv } from './client';
+import {
+  GastownClient,
+  MayorGastownClient,
+  GastownApiError,
+  createClientFromEnv,
+  registerPluginClient,
+  refreshPluginClientTokens,
+} from './client';
 import type { GastownEnv, MayorGastownEnv } from './types';
 
 const TEST_ENV: GastownEnv = {
@@ -7,6 +14,7 @@ const TEST_ENV: GastownEnv = {
   sessionToken: 'test-jwt-token',
   agentId: 'agent-111',
   rigId: 'rig-222',
+  townId: 'town-333',
 };
 
 function mockFetch(data: unknown, status = 200) {
@@ -48,7 +56,9 @@ describe('GastownClient', () => {
 
     expect(fetchMock).toHaveBeenCalledTimes(1);
     const [url, init] = fetchMock.mock.calls[0] as [string, RequestInit];
-    expect(url).toBe('https://gastown.example.com/api/rigs/rig-222/agents/agent-111/prime');
+    expect(url).toBe(
+      'https://gastown.example.com/api/towns/town-333/rigs/rig-222/agents/agent-111/prime'
+    );
     const headers = new Headers(init.headers);
     expect(headers.get('Authorization')).toBe('Bearer test-jwt-token');
     expect(headers.get('Content-Type')).toBe('application/json');
@@ -81,7 +91,7 @@ describe('GastownClient', () => {
     expect(result).toEqual(bead);
 
     const [url] = (globalThis.fetch as ReturnType<typeof vi.fn>).mock.calls[0] as [string];
-    expect(url).toBe('https://gastown.example.com/api/rigs/rig-222/beads/bead-1');
+    expect(url).toBe('https://gastown.example.com/api/towns/town-333/rigs/rig-222/beads/bead-1');
   });
 
   it('closeBead() sends agent_id in body', async () => {
@@ -94,7 +104,9 @@ describe('GastownClient', () => {
       string,
       RequestInit,
     ];
-    expect(url).toBe('https://gastown.example.com/api/rigs/rig-222/beads/bead-1/close');
+    expect(url).toBe(
+      'https://gastown.example.com/api/towns/town-333/rigs/rig-222/beads/bead-1/close'
+    );
     expect(init.method).toBe('POST');
     expect(JSON.parse(init.body as string)).toEqual({ agent_id: 'agent-111' });
   });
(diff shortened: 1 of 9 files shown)


jrf0110:
Closing in favor of #988

PR #979, opened 2026-03-10
fix(code-reviews): fix billing query timeout preventing usage footer on v2 reviews

alex-alecu:
## Summary

Follow-up to [PR #978](https://github.com/Kilo-Org/cloud/pull/978). The billing fallback query that fetches token/model data for v2 reviews was timing out in production, so the usage footer ("Reviewed by model · X tokens") was never shown.

**Root cause:** the query filters `microdollar_usage_metadata` by `session_id`, but that column has no index. The table has ~469M rows, so every query did a full table scan and timed out. The `catch` block silently returned `null`, and the footer was skipped.

**Fix:**
- Add a `created_at >= reviewCreatedAt` lower bound to the billing query. This lets Postgres use the existing `created_at` index (query cost drops from full-scan to ~288). Billing rows can't exist before the review was created, so the bound is exact.
- Skip the v1 poll loop for v2 reviews (saves ~1.4s of wasted retries).
- Remove the `session_id` index migration — with the time bound, it's not needed.
- Clean up the admin dashboard: remove agent version filter and performance chart that are no longer useful now that all reviews are v2.

## Verification

- [x] `pnpm typecheck` — no new errors (only pre-existing kiloclaw errors)
- [x] `pnpm test usage-footer` — 10/10 pass
- [x] `pnpm test schema` — 15/15 pass (no unmigrated schema changes)
- [x] Checked `EXPLAIN` plan on prod DB — query uses `idx_microdollar_usage_metadata_created_at` with cost ~288
- [x] Confirmed billing data exists for test session `ses_3282e02f5ffe2vPRBSqdpc0e40` (PR #981 review) — 8 rows returned in <1s with time-bounded query

## Visual Changes

N/A

## Reviewer Notes

- Every completed v2 review in prod has `model = NULL` — the billing fallback has never worked. This fix unblocks all future v2 reviews.
- The back-fill write (fire-and-forget) still runs after fetching billing data, so repeat reads skip the aggregation.


RSO (changes requested):
Creating an index locks the table, which will essentially take down the production app because `microdollar_usage` is the table that receives the most writes (I think). At least this should be `CREATE INDEX CONCURRENTLY`, but we can probably also limit the index even further by only indexing `session_id IS NOT NULL`

RSO (approved):

PR #982, opened 2026-03-10
fix(code-reviews): review .sql migration files for production safety

alex-alecu:
## Summary

- Narrow the "skip" rule from "migrations" to "migration snapshots & journals" so auto-generated Drizzle metadata (`meta/_journal.json`, `meta/*_snapshot.json`, `migrations.js`) is still skipped, but hand-written `.sql` migration files are reviewed
- Add DB migration review checklist: table-locking DDL without `CONCURRENTLY`, `NOT NULL` without `DEFAULT`, column drops, unbatched backfills, missing partial indexes
- Bump prompt template versions (GitHub v5.6.0, GitLab v5.7.0-gitlab)

**Context:** PR #979 added a `CREATE INDEX` on a high-write table. A human caught the missing `CONCURRENTLY` but the code reviewer skipped the file entirely because the prompt said to skip all migrations.

## Verification

- [x] `pnpm typecheck` passes
- [x] `pnpm test generate-prompt` — all 14 tests pass

## Visual Changes

N/A

## Reviewer Notes

The key change in the `whatToReview` field is the "Skip these" bullet: `Generated files (lock files, migrations)` → `Generated files (lock files, migration snapshots & journals)`. This removes the contradiction with the new "Database migrations (.sql files — DO review these)" section that was flagged by both sentry[bot] and kilo-code-bot[bot].


RSO (approved):
Not entirely sure how I feel about hard-coding stuff like this in the prompt. That can get hairy real quick if we want to also worry about MySQL/Redis/whatever.

PR #987, opened 2026-03-10
fix(cloud-agent-next): add pre-flight check before wrapper startup

eshurakov:
## Summary

Adds a pre-flight validation step in `WrapperClient.ensureRunning()` that runs before the wrapper process start retry loop. The check executes `bun --version` and verifies the wrapper script exists inside the container, catching fundamental environment issues (CPU incompatibility, missing binaries) before committing to the full start+waitForPort cycle.

This was motivated by investigating a production case where a user experienced a 100% wrapper failure rate across multiple days and container image versions. The wrapper process was crashing immediately with SIGILL (exit code 132, illegal CPU instruction) and exit code 1, but the error surfaced as an opaque `ProcessExitedBeforeReadyError` with no actionable diagnostic information. The pre-flight check would have immediately identified the bun runtime incompatibility.

Key behaviors:
- SIGILL (exit 132) produces `"Wrapper pre-flight failed: bun runtime is broken (SIGILL -- bun binary incompatible with host CPU)"`
- Other non-zero exits include the exit code and stderr in the error message
- Missing wrapper binary produces `"not found in container"` error
- If the pre-flight exec itself fails (e.g., sandbox connectivity issue), it logs a warning and proceeds with normal startup -- the check is additive, not a hard gate

## Verification

- [x] `pnpm run test -- --run src/kilo/wrapper-client.test` -- 426 tests pass (4 new pre-flight tests added)
- [x] `pnpm run typecheck` -- no new errors (pre-existing `jose` module error in `packages/worker-utils` unrelated)
- [x] Verified no PII or user-specific data in the diff

## Visual Changes

N/A

## Reviewer Notes

- The pre-flight runs both checks (`bun --version` + `test -f <wrapperPath>`) in parallel with a 5s timeout each, so the latency cost on the happy path is minimal
- Test mocks in `createMockSession` were updated to auto-handle pre-flight commands so existing tests remain unchanged
- The `ExecOptions.timeout` parameter is typed in `@cloudflare/sandbox` -- verified it's supported

### DIFF
--- a/cloud-agent-next/src/kilo/wrapper-client.test.ts
+++ b/cloud-agent-next/src/kilo/wrapper-client.test.ts
@@ -862,7 +862,7 @@ describe('WrapperClient', () => {
           return Promise.resolve({ exitCode: 0, stdout: '1.0.0' });
         }
         if (cmd.startsWith('test -f')) {
-          return Promise.resolve({ exitCode: 1, stdout: '' });
+          return Promise.resolve({ exitCode: 1, stdout: 'ok' });
         }
         return Promise.resolve(createCurlError(7, 'Connection refused'));
       });
@@ -934,7 +934,7 @@ describe('WrapperClient', () => {
           return Promise.reject(new Error('sandbox timeout'));
         }
         if (cmd.startsWith('test -f')) {
-          return Promise.resolve({ exitCode: 1, stdout: '' });
+          return Promise.resolve({ exitCode: 1, stdout: 'ok' });
         }
         return Promise.resolve(createCurlError(7, 'Connection refused'));
       });
@@ -948,6 +948,38 @@ describe('WrapperClient', () => {
           workspacePath: '/workspace/test',
         })
       ).rejects.toThrow(/not found in container/);
+    });
+
+    it('shell-quotes wrapperPath in the pre-flight file check', async () => {
+      const session = createMockSession(createCurlError(7, 'Connection refused'));
+      (session.exec as ReturnType<typeof vi.fn>).mockImplementation((cmd: string) => {
+        if (cmd.startsWith('bun --version')) {
+          return Promise.resolve({ exitCode: 0, stdout: '1.0.0' });
+        }
+        if (cmd.startsWith('test -f')) {
+          return Promise.resolve({ exitCode: 0, stdout: '' });
+        }
+        return Promise.resolve(createCurlError(7, 'Connection refused'));
+      });
+      (session.startProcess as ReturnType<typeof vi.fn>).mockResolvedValue({
+        id: 'mock-process-id',
+        waitForPort: vi.fn().mockResolvedValue(undefined),
+        getLogs: vi.fn().mockResolvedValue({ stdout: '', stderr: '' }),
+      });
+
+      const client = new WrapperClient({ session, port: defaultPort });
+
+      await client.ensureRunning({
+        sessionId: 'test-session',
+        kiloServerPort: 4600,
+        workspacePath: '/workspace/test',
+        wrapperPath: "./wrapper's folder/wrapper.js; touch /tmp/pwned",
+      });
+
+      expect(session.exec).toHaveBeenCalledWith(
+        "test -f './wrapper'\\''s folder/wrapper.js; touch /tmp/pwned'",
+        expect.objectContaining({ cwd: '/workspace/test', timeout: 5_000 })
+      );
     });
   });
 
--- a/cloud-agent-next/src/kilo/wrapper-client.test.ts
+++ b/cloud-agent-next/src/kilo/wrapper-client.test.ts
@@ -862,7 +862,7 @@ describe('WrapperClient', () => {
           return Promise.resolve({ exitCode: 0, stdout: '1.0.0' });
         }
         if (cmd.startsWith('test -f')) {
-          return Promise.resolve({ exitCode: 1, stdout: 'ok' });
+          return Promise.resolve({ exitCode: 1, stdout: '' });
         }
         return Promise.resolve(createCurlError(7, 'Connection refused'));
       });
@@ -934,7 +934,7 @@ describe('WrapperClient', () => {
           return Promise.reject(new Error('sandbox timeout'));
         }
         if (cmd.startsWith('test -f')) {
-          return Promise.resolve({ exitCode: 1, stdout: 'ok' });
+          return Promise.resolve({ exitCode: 1, stdout: '' });
         }
         return Promise.resolve(createCurlError(7, 'Connection refused'));
       });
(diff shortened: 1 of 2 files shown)


cloud-agent-next/src/kilo/wrapper-client.ts
```diff
@@ -130,6 +130,83 @@ export class WrapperClient {
...
+        const detail =
+          bunResult.value.exitCode === 132
+            ? 'SIGILL -- bun binary incompatible with host CPU'
+            : `exit code ${bunResult.value.exitCode}`;
+        throw new WrapperNotReadyError(
+          `Wrapper pre-flight failed: bun runtime is broken (${detail}). stderr: ${bunResult.value.stderr?.trim() ?? '(empty)'}`
+        );
+      }
+
+      if (fileResult.status === 'fulfilled' && !fileResult.value.stdout?.includes('ok')) {
```
evanjacobson: should we check for exitCode instead?

cloud-agent-next/src/kilo/wrapper-client.ts
```diff
@@ -130,6 +130,83 @@ export class WrapperClient {
...
+    const { wrapperPath, workspacePath } = options;
+
+    // Verify bun runtime and wrapper binary before the full start+waitForPort loop.
+    // A fast `bun --version` catches SIGILL (exit 132) on hosts whose CPU lacks
+    // required instructions, missing/corrupt binaries, etc. We also verify the
+    // wrapper script exists.
+    try {
+      const [bunResult, fileResult] = await Promise.allSettled([
+        this.session.exec('bun --version', { timeout: 5_000 }),
+        this.session.exec(`test -f ${wrapperPath} && echo ok`, {
```
evanjacobson: Shell injection via unquoted wrapperPath. wrapperPath is caller-supplied via ensureRunning(options.wrapperPath) and interpolated raw into a shell command.




```suggestion
        this.session.exec(`test -f '${wrapperPath}' && echo ok`, {
```

cloud-agent-next/src/kilo/wrapper-client.ts
```diff
@@ -130,6 +130,83 @@ export class WrapperClient {
   private readonly port: number;
   private readonly baseUrl: string;
 
+  private async runPreflightChecks(options: {
+    wrapperPath: string;
+    workspacePath: string;
```
evanjacobson, line 139: add sessionId so we can log it


```suggestion
    workspacePath: string;
    sessionId: string;
```

cloud-agent-next/src/kilo/wrapper-client.ts
```diff
@@ -244,6 +321,8 @@ export class WrapperClient {
       logger.debug('WrapperClient: wrapper not running, starting...');
     }
 
+    await this.runPreflightChecks({ wrapperPath, workspacePath });
```
evanjacobson, line 329: pass session ID here


```suggestion
    await this.runPreflightChecks({ wrapperPath, workspacePath, sessionId });
```

cloud-agent-next/src/kilo/wrapper-client.ts
evanjacobson, line 1: Logs to update session ID (line number prob off a little bit)

Line 170 — both rejected:
          logger.warn('WrapperClient: pre-flight check failed to execute, proceeding anyway', {
            sessionId: options.sessionId,
            port: this.port,
            bunError:
              bunResult.reason instanceof Error ? bunResult.reason.message : String(bunResult.reason),
            fileError:
              fileResult.reason instanceof Error
                ? fileResult.reason.message
                : String(fileResult.reason),
          });

  Line 180 — bun rejected:
          logger.warn('WrapperClient: bun pre-flight exec failed, proceeding anyway', {
            sessionId: options.sessionId,
            port: this.port,
            error:
              bunResult.reason instanceof Error ? bunResult.reason.message : String(bunResult.reason),
          });

  Line 188 — file rejected:
          logger.warn('WrapperClient: file pre-flight exec failed, proceeding anyway', {
            sessionId: options.sessionId,
            port: this.port,
            error:
              fileResult.reason instanceof Error
                ? fileResult.reason.message
                : String(fileResult.reason),
          });

  Line 199 — outer catch:
        logger.warn('WrapperClient: pre-flight check failed unexpectedly, proceeding anyway', {
          sessionId: options.sessionId,
          port: this.port,
          error: error instanceof Error ? error.message : String(error),
        });

  And the debug log at line 196 for consistency:
          logger.debug('WrapperClient: pre-flight passed', {
            sessionId: options.sessionId,
            port: this.port,
            bunVersion: bunResult.value.stdout?.trim(),
          });

evanjacobson:
Looks like my sessionId comments are likely not useful due to:

`logger.withFields({ sessionId, workspacePath }).info('Ensuring wrapper is running');`

cloud-agent-next/src/kilo/wrapper-client.test.ts
```diff
@@ -786,6 +803,186 @@ describe('WrapperClient', () => {
...
+    });
+
+    it('throws WrapperNotReadyError when wrapper binary is missing', async () => {
+      const session = createMockSession(createCurlError(7, 'Connection refused'));
+      (session.exec as ReturnType<typeof vi.fn>).mockImplementation((cmd: string) => {
+        if (cmd.startsWith('bun --version')) {
+          return Promise.resolve({ exitCode: 0, stdout: '1.0.0' });
+        }
+        if (cmd.startsWith('test -f')) {
+          return Promise.resolve({ exitCode: 1, stdout: 'ok' });
```
evanjacobson: (Automated): This returns `{ exitCode: 1, stdout: 'ok' }` which is an impossible combination — `test -f` with exit code 1 means the file doesn't exist, and the `&& echo ok` was removed in the last commit so there's no way stdout would be `'ok'`. The test passes because the production code now checks `exitCode !== 0`, but the mock data is misleading. Should be `{ exitCode: 1, stdout: '' }`.

cloud-agent-next/src/kilo/wrapper-client.test.ts
```diff
@@ -786,6 +803,186 @@ describe('WrapperClient', () => {
...
+    });
+
+    it('surfaces missing wrapper even when the bun check rejects', async () => {
+      const session = createMockSession(createCurlError(7, 'Connection refused'));
+      (session.exec as ReturnType<typeof vi.fn>).mockImplementation((cmd: string) => {
+        if (cmd.startsWith('bun --version')) {
+          return Promise.reject(new Error('sandbox timeout'));
+        }
+        if (cmd.startsWith('test -f')) {
+          return Promise.resolve({ exitCode: 1, stdout: 'ok' });
```
evanjacobson: (Automated): Same as above — `{ exitCode: 1, stdout: 'ok' }` is unrealistic after removing `&& echo ok` from the command. `{ exitCode: 1, stdout: '' }` would match what `test -f` actually produces on failure.

cloud-agent-next/src/kilo/wrapper-client.ts
```diff
@@ -130,6 +130,88 @@ export class WrapperClient {
...
+              : String(fileResult.reason),
+        });
+      }
+
+      if (bunResult.status === 'fulfilled') {
+        logger.debug('WrapperClient: pre-flight passed', {
+          bunVersion: bunResult.value.stdout?.trim(),
+        });
+      }
+    } catch (error) {
```
evanjacobson, line 206: (Automated): Since `Promise.allSettled` never rejects, this catch can only fire for programmer errors in the conditional logic above (e.g. a `TypeError` from a bad property access). Logging those at `warn` and continuing might hide real bugs. Consider logging at `error` level with the stack trace, or removing the try/catch since `allSettled` guarantees no rejection.

evanjacobson (commented):
(Automated): Overall this is a solid addition — the `Promise.allSettled` approach, the error categorization (SIGILL vs generic vs missing file), and the graceful degradation are all well thought out. A few things worth addressing before merge.

**Main concern:** `wrapperPath` is shell-quoted in the pre-flight `test -f` check (via `shellQuote`), but it's still interpolated raw in the `startProcess` command on line 344: `` `${envParts.join(' ')} bun run ${wrapperPath} ${sessionMarker}` ``. A path with spaces or special chars would pass pre-flight but then break at actual startup — which is the exact opaque failure this PR is designed to prevent. Should use `this.shellQuote(wrapperPath)` there too. Same concern for `WORKSPACE_PATH=${workspacePath}` in `envParts`.

jeanduplessis (approved):

PR #1008, opened 2026-03-10
feat(kiloclaw): use fresh token when redeploying instance

pandemicsyn:
## Summary

KiloClaw instances receive a KiloCode API key (30-day HS256 JWT) at provision time from the Next.js backend. Previously, `restartGateway()` and `start()` re-deployed the same stale key stored in the DO — if the key was near expiry or the user's pepper was rotated, the machine would boot with a dead credential.

This PR makes `buildUserEnvVars()` mint a fresh API key on every redeploy/start by querying the user's current `api_token_pepper` from Postgres via Hyperdrive and signing a new 30-day JWT. The stored key serves as a graceful fallback if Hyperdrive is unavailable.

**Scope of this change:**
- Adds `signKiloToken()` to `@kilocode/worker-utils` alongside the existing `verifyKiloToken` — typed `SignKiloTokenExtra` is derived via `Pick<KiloTokenPayload, ...>` so sign/verify can't drift. Other workers (gastown, webhook-agent-ingest) are not changed; they can adopt the shared function in follow-up PRs.
- Wraps the mint in a 5-second timeout (`withTimeout`) to prevent Hyperdrive issues from extending downtime while the machine is stopped.
- Adds `NEXTAUTH_SECRET` to the platform route `requireEnvVars` check — fails fast (503) instead of silently reusing a stale key.
- Does **not** address proactive refresh of running instances (that'll be part 2). `start()` still returns early when the machine is already in `started` state — no minting in that path.

## Verification

- [x] `pnpm typecheck` — clean (worker-utils + kiloclaw)
- [x] `pnpm test` — 30/30 worker-utils, 524/524 kiloclaw
- [x] `pnpm lint` — clean (worker-utils + kiloclaw)

## Visual Changes

N/A

## Reviewer Notes

- The minted key is persisted to DO state **before** `fly.updateMachine()` succeeds. If the Fly op fails, DO state points at a key that never reached the machine. This matches how `trackedImageTag` already works — the persisted key is "latest minted" (fallback cache + expiry reporting via `getConfig()`), not "confirmed on machine".
- Key expiry is surfaced via `getConfig()` / `/api/kiloclaw/config`, not `getStatus()`.

### DIFF
--- a/kiloclaw/src/durable-objects/kiloclaw-instance.ts
+++ b/kiloclaw/src/durable-objects/kiloclaw-instance.ts
@@ -2796,6 +2796,19 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
     });
   }
 
+  private hasExpiredStoredApiKey(): boolean {
+    if (!this.kilocodeApiKey || !this.kilocodeApiKeyExpiresAt) {
+      return false;
+    }
+
+    const expiresAtMs = Date.parse(this.kilocodeApiKeyExpiresAt);
+    if (Number.isNaN(expiresAtMs)) {
+      return false;
+    }
+
+    return expiresAtMs <= Date.now();
+  }
+
   private async buildUserEnvVars(): Promise<{
     envVars: Record<string, string>;
     minSecretsVersion: number;
@@ -2834,6 +2847,12 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
       } catch (err) {
         console.warn('[DO] buildUserEnvVars: failed to mint fresh API key, using stored key:', err);
       }
+    }
+
+    if (this.hasExpiredStoredApiKey()) {
+      throw new Error(
+        'Cannot build env vars: stored KiloCode API key expired and fresh mint unavailable'
+      );
     }
 
     const { env: plainEnv, sensitive } = await buildEnvVars(
--- a/kiloclaw/src/durable-objects/kiloclaw-instance.ts
+++ b/kiloclaw/src/durable-objects/kiloclaw-instance.ts
@@ -2772,7 +2772,9 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
     }
   }
 
-  private async mintFreshApiKey(secret: string): Promise<{ token: string; expiresAt: string } | null> {
+  private async mintFreshApiKey(
+    secret: string
+  ): Promise<{ token: string; expiresAt: string } | null> {
     const connectionString = this.env.HYPERDRIVE?.connectionString;
     if (!this.userId || !connectionString) {
       return null;
--- a/kiloclaw/src/durable-objects/kiloclaw-instance.ts
+++ b/kiloclaw/src/durable-objects/kiloclaw-instance.ts
@@ -22,10 +22,11 @@
  */
 
 import { DurableObject } from 'cloudflare:workers';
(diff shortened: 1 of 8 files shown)


packages/worker-utils/src/kilo-token.ts
```diff
@@ -30,6 +30,56 @@ export const kiloTokenPayload = z.object({
 
 export type KiloTokenPayload = z.infer<typeof kiloTokenPayload>;
 
+const KILO_TOKEN_VERSION = 3;
```
St0rmz1: Also defined in kiloclaw/src/config.ts
Can we do an export on this where it makes sense?

packages/worker-utils/src/kilo-token.ts
```diff
@@ -30,6 +30,56 @@ export const kiloTokenPayload = z.object({
...
+    kiloUserId: params.userId,
+    apiTokenPepper: params.pepper,
+    version: KILO_TOKEN_VERSION,
+    ...params.extra,
+  };
+
+  if (params.env) {
+    payload.env = params.env;
+  }
+
```
St0rmz1, line 74: Should we parse (payload) to just double check if someone adds a field to extra that doesn't exist in the Zod schema?

kiloclaw/src/durable-objects/kiloclaw-instance.ts
```diff
@@ -2778,6 +2806,35 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
     if (!this.userId) {
       throw new Error('Cannot build env vars: userId missing');
     }
+    if (!this.env.NEXTAUTH_SECRET) {
+      throw new Error('Cannot build env vars: NEXTAUTH_SECRET missing');
+    }
+    const nextAuthSecret = this.env.NEXTAUTH_SECRET;
+
+    let kilocodeApiKey = this.kilocodeApiKey ?? undefined;
```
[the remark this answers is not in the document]
pandemicsyn: yea, i think a bad cred here is better than fully preventing a deploy.

kiloclaw/src/durable-objects/kiloclaw-instance.ts
```diff
@@ -2778,6 +2806,35 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
     if (!this.userId) {
       throw new Error('Cannot build env vars: userId missing');
     }
+    if (!this.env.NEXTAUTH_SECRET) {
```
St0rmz1, line 2837: If minting fails (timeout, DB error, no Hyperdrive), kilocodeApiKey remains this.kilocodeApiKey which I think means the stale stored value remains? 

Add check this.kilocodeApiKeyExpiresAt to see if that stored key is already expired before passing it to buildEnvVars . Otherwise, I think the machine gets deployed with expired creds.

St0rmz1 (approved):

kiloclaw/src/durable-objects/kiloclaw-instance.ts
pandemicsyn: Yea, my theory was - id rather re-deploy you with a stale cred - than fail if hyper drive is timing out. But maybe, this should just kaboom. If hyperdrive is having issues - other stuff is probably breaking anyway (including token validation when you hit the claw control ui proxy.

Let me just change this to kaboom if hyper drive is offline and key expired.

PR #1021, opened 2026-03-11
feat(billing): integrate Rewardful affiliate tracking

jeanduplessis:
## Summary

Integrates Rewardful affiliate/referral tracking into all Stripe checkout session creation points. Two-part approach:

1. **Client-side** (`layout.tsx`): Loads the Rewardful `rw.js` script (gated on `NEXT_PUBLIC_REWARDFUL_ID` env var). On ready, it writes the referral UUID into a `rewardful_referral` cookie.
2. **Server-side** (`src/lib/rewardful.ts`): A helper reads and UUID-validates the cookie, and every Stripe `checkout.sessions.create` call now passes `client_reference_id` when a referral is present.

Affected checkout flows: credit top-ups, auto top-up setup, org auto top-up setup, seat subscriptions, Kilo Pass subscriptions, and KiloClaw early-bird purchases.

## Verification

## Visual Changes

N/A

## Reviewer Notes

- The Rewardful script is a no-op unless `NEXT_PUBLIC_REWARDFUL_ID` is set — zero impact on existing environments.
- `getRewardfulReferral()` validates the cookie value is a UUID and swallows errors if `cookies()` is unavailable (e.g. during static generation), so it never blocks a checkout.
- The cookie has a 90-day `max-age`, matching Rewardful's default attribution window. The `Secure` flag is added conditionally (HTTPS only) so it works on localhost dev too.

### DIFF
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -1,5 +1,6 @@
 import type { Metadata, Viewport } from 'next';
 import { Inter, Roboto_Mono, JetBrains_Mono } from 'next/font/google';
+import Script from 'next/script';
 import './globals.css';
 import { PostHogProvider } from '../components/PostHogProvider';
 import { Providers } from '../components/Providers';
@@ -111,6 +112,18 @@ export default function RootLayout({
 
         {process.env.NEXT_PUBLIC_GTM_ID && (
           <GoogleTagManager gtmId={process.env.NEXT_PUBLIC_GTM_ID} />
+        )}
+
+        {process.env.NEXT_PUBLIC_REWARDFUL_ID && (
+          <>
+            <Script id="rewardful-queue">
+              {`(function(w,r){w._rwq=r;w[r]=w[r]||function(){(w[r].q=w[r].q||[]).push(arguments)}})(window,'rewardful');rewardful('ready',function(){if(Rewardful.referral){document.cookie='rewardful_referral='+encodeURIComponent(Rewardful.referral)+';path=/;max-age=7776000;SameSite=Lax'+(location.protocol==='https:'?';Secure':'')}});`}
+            </Script>
+            <Script
+              src="https://r.wdfl.co/rw.js"
+              data-rewardful={process.env.NEXT_PUBLIC_REWARDFUL_ID}
+            />
+          </>
         )}
       </body>
     </html>
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -116,10 +116,11 @@ export default function RootLayout({
 
         {process.env.NEXT_PUBLIC_REWARDFUL_ID && (
           <>
-            <Script id="rewardful-queue">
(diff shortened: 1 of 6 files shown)


arkadiykondrashov:
> The cookie has a 90-day max-age, matching Rewardful's default attribution window. The Secure flag is added conditionally (HTTPS only) so it works on localhost dev too.


- We have 60-days attribution window from the first touch (in out T&Cs and program settings).

src/lib/rewardful.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/** Reads the Rewardful referral ID (a UUID) from the cookie set by the client-side rw.js script. */
+export async function getRewardfulReferral(): Promise<string | undefined> {
+  try {
+    const jar = await cookies();
+    const value = jar.get(REWARDFUL_COOKIE)?.value;
+    return value && UUID_RE.test(value) ? value : undefined;
+  } catch {
+    return undefined;
+  }
```
RSO: Not so sure about this one. Don't you want to know when this fails?

RSO (approved):
Did you test this yourself for all the different flows? I've had some issues here and there with the cookies package, so I'm not entirely sure whether it'll work in all cases.

src/lib/rewardful.ts
jeanduplessis: Valid point — the empty catch was silently swallowing errors. Added `console.warn('Failed to read rewardful referral cookie', error)` so failures are visible in logs while still gracefully degrading (returning `undefined` so checkout proceeds without the referral).

src/app/layout.tsx
```diff
@@ -112,6 +113,18 @@ export default function RootLayout({
         {process.env.NEXT_PUBLIC_GTM_ID && (
           <GoogleTagManager gtmId={process.env.NEXT_PUBLIC_GTM_ID} />
         )}
+
+        {process.env.NEXT_PUBLIC_REWARDFUL_ID && (
+          <>
+            <Script id="rewardful-queue">
```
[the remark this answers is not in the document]
jeanduplessis: Valid concern. The race window exists between page load and `rw.js` firing its `ready` callback.

**Fix applied:** Changed both Rewardful `<Script>` tags to use `strategy="beforeInteractive"`, which injects them into `<head>` at SSR time rather than loading lazily after hydration. This ensures `rw.js` loads and resolves the referral UUID before any React component can initiate a checkout.

Also corrected the cookie `max-age` from 5184000 (60 days) to 7776000 (90 days) to match Rewardful's default attribution window as documented in the PR description.

Residual risk is minimal: every checkout flow requires multiple user interactions (navigating to a page, selecting options, clicking buttons) before a Stripe session is created, giving `rw.js` ample time to load and fire even on slow connections. The `via` query param (affiliate slug) can't be used as a fallback since `client_reference_id` requires the resolved UUID, not the slug.

PR #1031, opened 2026-03-11
feat(auto): add Kilo Auto Balanced model

lambertjosh:
## Summary

Adds a new `kilo-auto/balanced` auto model that mirrors Frontier's mode-based routing structure but uses cheaper underlying models:

- **Kimi K2.5** (`moonshotai/kimi-k2.5`) for heavy modes: plan, general, architect, orchestrator, ask, debug (where Frontier uses Opus)
- **Minimax M2.5** (`minimax/minimax-m2.5`) for implementation modes: build, explore, code (where Frontier uses Sonnet)

Context length and max completion tokens are derived from the minimum of both models (204,800 / 65,536). Pricing set at $2/$8 per M tokens. Added to `preferredModels` between Frontier and Free.

## Verification

- [x] `pnpm typecheck` — passes (no new errors introduced; all existing errors are in unrelated files)

## Visual Changes

N/A

## Reviewer Notes

- Pricing (`prompt_price` / `completion_price`) is placeholder — may need adjustment based on actual upstream costs.
- `supports_images` is `false` since Minimax M2.5 lacks vision support, even though Kimi K2.5 does support it.
- `opencode_settings` is `undefined` since neither underlying model fits the existing families (claude, gpt, gemini, llama, mistral).
- Renamed internal constants (`CODE_MODEL` → `FRONTIER_CODE_MODEL`, `MODE_TO_MODEL` → `FRONTIER_MODE_TO_MODEL`) for clarity now that there are two routing tables.


src/lib/kilo-auto-model.ts
```diff
@@ -67,7 +87,12 @@ export const KILO_AUTO_SMALL_MODEL: AutoModel = {
   },
 };
 
-export const AUTO_MODELS = [KILO_AUTO_FRONTIER_MODEL, KILO_AUTO_FREE_MODEL, KILO_AUTO_SMALL_MODEL];
+export const AUTO_MODELS = [
+  KILO_AUTO_FRONTIER_MODEL,
+  KILO_AUTO_BALANCED_MODEL,
```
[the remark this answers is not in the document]
lambertjosh: Fixed this by changing around the logic so that it looks up the legacy mappings we need and then finds the corresponding models, rather than trying to map all auto models to legacy which may not exist or creating a legacy id when one is not required.

lambertjosh:
Demo videos: 
* Legacy extension: https://www.loom.com/share/de082b1128e84382b1a6e8a3d6f9838e
* New CLI: https://www.loom.com/share/2666512d76404225afc2beb6be2c8f8d

src/tests/openrouter-models-sorting.approved.json
```diff
@@ -42,6 +42,43 @@
       },
       "preferredIndex": 0
     },
+    {
+      "id": "kilo-auto/balanced",
+      "name": "Kilo Auto Balanced",
+      "created": 0,
+      "description": "Automatically routes your request to a balanced mix of price and performance.",
```
lambertjosh:
```suggestion
      "description": "Automatically routes your request for a balanced mix of price and performance.",
```

src/lib/kilo-auto-model.ts
```diff
@@ -51,6 +52,25 @@ export const KILO_AUTO_FREE_MODEL: AutoModel = {
   opencode_settings: undefined,
 };
 
+export const KILO_AUTO_BALANCED_MODEL: AutoModel = {
+  id: 'kilo-auto/balanced',
+  name: 'Kilo Auto Balanced',
+  description: 'Automatically routes your request to a balanced mix of price and performance.',
```
lambertjosh:
```suggestion
  description: 'Automatically routes your request for a balanced mix of price and performance.',
```

lambertjosh:
Also validated that the modes do hit the correct models on the backend and used both. Flipping between modes in the same session also appears to work correctly.

lambertjosh:
@chrarnoldus - one item I wasn't sure about, there are different roo settings for minimax:m2.5 and Kimi:k2.5. I applied the minimax ones as I wanted to be restrictive, but would love your guidance here.

chrarnoldus:
> I applied the minimax ones as I wanted to be restrictive, but would love your guidance here.

`edit_file` is simpler, so I changed it to that one.

chrarnoldus:
The PR originally used paid MiniMax, was that intentional? I changed it to free, but we can always change it back.

edit: roadmap says free

chrarnoldus (approved):

lambertjosh:
> The PR originally used paid MiniMax, was that intentional? I changed it to free, but we can always change it back.

@chrarnoldus - thanks, good catch!

PR #1069, opened 2026-03-12
Remove unused env var and reorder rewrites

evanjacobson:
## Summary
The PostHog reverse-proxy catch-all rewrite (/ingest/:path*) was ordered before the more specific /ingest/decide and /ingest/static/:path* rewrites, causing it to swallow those routes. This moves the catch-all to after the specific rewrites so they match first.

Also removes the unused NEXT_PUBLIC_POSTHOG_HOST env var from .env.

## Verification
 - Confirmed rewrite order in next.config.mjs places /ingest/static/:path* and /ingest/decide before the catch-all /ingest/:path*
  - No build errors introduced

## Reviewer Notes
Next.js rewrites are evaluated top-down and first-match wins. The catch-all was previously shadowing the specific routes. A comment is added inline to prevent future reordering.


evanjacobson:
#1071 will fix the CI failure

jeanduplessis (approved):

PR #1092, opened 2026-03-13
fix(billing): require org owner role for Stripe checkout URL creation

jeanduplessis:
## Summary

The `getSubscriptionStripeUrl` mutation used `baseProcedure` and only enforced ownership via `ensureOrganizationAccess` when existing subscriptions were found. This meant that for new organizations without any subscriptions yet, any authenticated user could trigger Stripe customer creation without ownership checks. This switches the procedure to `organizationOwnerProcedure`, which enforces the owner role unconditionally before any Stripe interaction occurs. The now-unused conditional `ensureOrganizationAccess` call and its import are removed.

## Verification

- [x] `pnpm test -- 'src/routers/organizations/organization-subscription-router.test.ts'` — 9/9 tests passed (including 2 new authorization tests)

## Visual Changes

N/A

## Reviewer Notes

- The two new tests verify that non-owner members and non-members are both rejected by `getSubscriptionStripeUrl`, matching the pattern used by `cancel`, `stopCancellation`, and `updateSeatCount` tests in the same file.
- The stale comment "if any subscriptions exist we need to enforce security / otherwise, we can't enforce ownership as the org is still not finished being set up" is removed since ownership is now enforced unconditionally by the procedure middleware.

### DIFF
--- a/src/routers/organizations/organization-subscription-router.test.ts
+++ b/src/routers/organizations/organization-subscription-router.test.ts
@@ -57,6 +57,32 @@ describe('organizations subscription trpc router', () => {
     });
   });
 
+  describe('getSubscriptionStripeUrl procedure', () => {
+    it('should throw UNAUTHORIZED error for non-owner members', async () => {
+      const caller = await createCallerForUser(memberUser.id);
+
+      await expect(
+        caller.organizations.subscription.getSubscriptionStripeUrl({
+          organizationId: testOrganization.id,
+          seats: 1,
+          cancelUrl: 'https://example.com',
+        })
+      ).rejects.toThrow('You do not have the required organizational role to access this feature');
+    });
+
+    it('should throw UNAUTHORIZED error for non-member users', async () => {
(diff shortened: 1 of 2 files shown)


src/routers/organizations/organization-subscription-router.test.ts
```diff
@@ -57,6 +57,32 @@ describe('organizations subscription trpc router', () => {
     });
   });
 
+  describe('getSubscriptionStripeUrl procedure', () => {
+    it('should throw UNAUTHORIZED error for non-owner members', async () => {
```
alexkgold, line 61: What about billing managers?

alexkgold (approved):
LGTM, thanks!

src/routers/organizations/organization-subscription-router.test.ts
jeanduplessis: Confirmed.  `organizationOwnerProcedure` checks for roles ['owner', 'billing_manager']:

PR #1095, opened 2026-03-13
feat(controller): pair with controller for channel/device pairing

evanjacobson:
## Summary

Adds a controller-side pairing cache and HTTP API so the Durable Object can resolve pairing requests via the controller's in-memory cache instead of shelling out via `fly exec` every time.

- **Pairing cache** on the controller periodically runs `kilo pairing list` / `kilo device-pairing list`, parses stdout, and serves cached results. Refreshes reactively when the gateway logs a new pairing event (via a new `onStdoutLine` supervisor hook).
- **HTTP routes** at `/_kilo/pairing/{channels,devices}` and `/_kilo/pairing/{channels,devices}/approve`.
- **DO pairing methods** now try the controller HTTP API first and fall back to `fly exec` on unknown-route errors, preserving backward compatibility with older controller versions.
- **Zod schemas** for typed controller pairing responses.
- ~~**409 machine recovery**: if `createMachine` hits a 409 "already_exists", adopts the existing machine instead of crashing.~~ #1135 
- **Volume existence check** after machine ID recovery to prevent startups against a missing volume.

### Also included: billing infra hardening

These files are coupled to call sites already on `main` and cannot be split out without breaking the build:

- **`stripe-handlers.ts`** — Stripe webhook handlers (`subscription.created`, `.updated`, `.deleted`, schedule events). Minor hardening: null-check on subscription items, improved comment accuracy. **Known gap:** `subscription.updated` handler reads `suspended_at` outside a transaction (TOCTOU with billing cron — tracked separately).
- **`billing-lifecycle-cron.ts`** — Trial/earlybird/subscription expiry sweeps and email notifications. Hardening: added `isNull(suspended_at)` guard on earlybird sweep, wrapped email-log delete in try/catch. **Known gap:** sweep UPDATEs use bare `WHERE user_id = ?` without conditional status guards (TOCTOU with webhooks — tracked separately).
- **`kiloclaw-router.ts`** — Billing endpoints (`getBillingStatus`, `createSubscriptionCheckout`, `cancelSubscription`, `switchPlan`, etc.) and `baseProcedure` → `clawAccessProcedure` gating on existing endpoints.
- **`modelSupport.ts`** — Minor refactor collapsing early-return conditions.

## Verification

- [x] Pairing cache tests (749 lines) — `pairing-cache.test.ts`
- [x] Pairing route tests (410 lines) — `routes/pairing.test.ts`
- [x] Supervisor `onStdoutLine` tests (135 lines) — `supervisor.test.ts`
- [x] Manual testing — paired Telegram channel

## Visual Changes

N/A

## Reviewer Notes

- The controller-first / fly-exec-fallback pattern uses `isErrorUnknownRoute` to detect older controllers that don't have the pairing routes, so rollout is safe without version gating.
- The pairing cache debounces rapid refresh triggers and caps concurrent CLI invocations to avoid overwhelming the gateway process.
- Cache warmup is fire-and-forget — `start()` kicks off the initial refresh but does not block `server.listen()`. This avoids a slow or wedged OpenClaw CLI (45s timeout per path) holding the health endpoint unreachable past the DO's 60s startup probe. The brief empty-cache window is acceptable: the DO falls back to KV / fly exec, and the cache self-heals via periodic refresh and log-triggered debounce.
- The billing files carry two known TOCTOU race conditions (documented in `cloud_todos.md`) that are pre-existing and tracked for separate follow-up — not introduced or worsened by this PR.


kiloclaw/src/durable-objects/kiloclaw-instance/fly-machines.ts
```diff
@@ -216,11 +216,40 @@ export async function createNewMachine(
...
+    });
+  } catch (err) {
+    // Safety net: if the machine already exists (e.g. DO state lost flyMachineId
+    // and metadata recovery didn't run or failed), adopt it instead of crashing.
+    if (
+      err instanceof fly.FlyApiError &&
+      err.status === 409 &&
+      err.body.includes('already_exists')
+    ) {
+      const match = err.body.match(/machine ID (\w+)/);
```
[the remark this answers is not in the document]
evanjacobson: Investigated this. The Fly Machines API returns a structured JSON body for 409s:

  {"error":"already_exists: unique machine name violation, machine ID 123abc already exists with name \\"machine_name_ex\\""}

src/lib/kiloclaw/billing-lifecycle-cron.ts
```diff
@@ -73,12 +73,19 @@ async function trySendEmail(
   try {
     await sendEmail({ to: userEmail, templateName, templateVars });
   } catch (error) {
-    // Send failed — remove the log row so we can retry on the next cron run
-    await database
-      .delete(kiloclaw_email_log)
-      .where(
-        and(eq(kiloclaw_email_log.user_id, userId), eq(kiloclaw_email_log.email_type, emailType))
+    try {
```
evanjacobson, line 76: if the delete also fails (transient DB error), the original send error must still propagate — without this, a delete failure masks the real error and the sweep iteration crashes on the wrong exception

src/lib/kiloclaw/billing-lifecycle-cron.ts
```diff
@@ -188,7 +195,10 @@ export async function runKiloClawBillingLifecycleCron(
         eq(kiloclaw_earlybird_purchases.user_id, kiloclaw_subscriptions.user_id)
       )
       .where(
-        sql`(${kiloclaw_subscriptions.status} IS NULL OR ${kiloclaw_subscriptions.status} NOT IN ('active', 'trialing'))`
+        and(
```
evanjacobson, line 198: without this guard, the sweep sends "earlybird ending" emails to users who are already locked out by a prior cron run or webhook suspension.

src/lib/kiloclaw/stripe-handlers.ts
```diff
@@ -37,8 +37,11 @@ function getKiloClawMetadata(
 }
 
 function getSubscriptionPeriods(subscription: Stripe.Subscription) {
-  // In the current Stripe API, period timestamps live on the subscription item, not the subscription.
+  // Stripe moved period timestamps to the item level (not the top-level subscription object).
```
evanjacobson, line 40: Stripe subscriptions should always have at least one item, but log if not so we can investigate rather than silently returning nulls

kiloclaw/src/durable-objects/kiloclaw-instance/pairing.ts
```diff
@@ -230,12 +320,35 @@ export async function approveDevicePairingRequest(
...
-  const flyConfig = getFlyConfig(env, state);
-
-  if (!/^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i.test(requestId)) {
+  if (!UUID_RE.test(requestId)) {
     return { success: false, message: 'Invalid request ID' };
   }
 
+  // Try controller first
+  try {
+    return await callGatewayController(
```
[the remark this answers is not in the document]
evanjacobson: No KV invalidation needed: if the controller succeeds, subsequent list calls also use the controller (routes register together), so KV is never read. If the controller fails with a non-route error (503, network, etc.), we throw instead of falling through to KV.

evanjacobson:
Related: https://github.com/Kilo-Org/cloud/pull/1135

evanjacobson:
Spec: https://github.com/orgs/Kilo-Org/projects/30/views/1?pane=issue&itemId=164917045

src/lib/kiloclaw/stripe-handlers.ts
```diff
@@ -37,8 +37,11 @@ function getKiloClawMetadata(
...
   const item = subscription.items.data[0];
+  if (!item) {
+    console.warn('[stripe] Subscription has no items:', subscription.id);
```
jeanduplessis, line 43: Suggest using logWarning here

src/lib/kiloclaw/billing-lifecycle-cron.ts
```diff
@@ -73,12 +73,19 @@ async function trySendEmail(
...
-      .where(
-        and(eq(kiloclaw_email_log.user_id, userId), eq(kiloclaw_email_log.email_type, emailType))
+    try {
+      await database
+        .delete(kiloclaw_email_log)
+        .where(
+          and(eq(kiloclaw_email_log.user_id, userId), eq(kiloclaw_email_log.email_type, emailType))
+        );
+    } catch (deleteError) {
+      console.error(
```
jeanduplessis, line 83: Suggest using logError here

jeanduplessis (approved):
Giving approval specifically for billing changes. We should get @pandemicsyn's final approval for the pairing work.

PR #1113, opened 2026-03-16
Add Responses API support

chrarnoldus:


chrarnoldus:
> Responses provider metadata lives under provider_metadata / response.provider_metadata in the new payloads, but the parser still only checks provider and choices[].*.provider_metadata.

OpenRouter doesn't seem to include provider metadata, actually.

markijbema (approved):

PR #1151, opened 2026-03-16
feat(cloud-agent-next): add async session preparation with fast response

eshurakov:
## Summary

Add an async session preparation path to cloud-agent-next, reducing perceived wait time from 30–120s to ~500ms when starting a session from the landing page.

- **Fast return path**: `autoInitiate: true` flag on `prepareSession` returns immediately after creating IDs and registering minimal DO metadata. Expensive work (git clone, setup commands, kilo server start) runs asynchronously via a DO alarm.
- **Single stable ID**: Generate a `ses_*` session ID upfront in the fast path and import it into the CLI's SQLite before the wrapper starts. Eliminates dual-ID complexity and URL rewriting.
- **Streaming progress**: New `preparing` event type streams real-time preparation steps (disk check → clone → branch → setup → ready/failed) over the existing WebSocket infrastructure.
- **Chat page integration**: WebSocket connects during async preparation to display progress events. Auto-initiates execution once preparation completes. New `progress` status indicator (muted spinner, no auto-clear).
- **Cold-start resume progress**: Threaded `onProgress` callback through the resume path so follow-up messages also report progress during sandbox re-creation.

`autoInitiate: false` (default) preserves existing synchronous behavior exactly — zero changes to existing callers.

**Architecture**: Preparation runs inside the DO via `setAlarm(Date.now())`, safe regardless of caller lifetime (no reliance on `ctx.waitUntil` which is a no-op in DOs). `cli_sessions_v2` is created before return so the stream-ticket route can verify ownership immediately. `restore-session.ts` gains a `--file` flag to accept pre-written session files for import (Docker rebuild needed).

## Verification

- [x] Manual check

## Visual Changes

| Before | After |
| ------ | ----- |
| Landing page blocks 10–30s during `prepareSession` before redirect | Redirects in ~500ms; chat page shows streaming progress (disk check → clone → setup → ready) |
| No progress during cold-start resume on follow-up messages | Progress spinner for each cold-start step (clone, workspace setup, session restore, kilo start) |

### DIFF
--- a/cloud-agent-next/Dockerfile.dev
+++ b/cloud-agent-next/Dockerfile.dev
@@ -3,7 +3,7 @@ FROM docker.io/cloudflare/sandbox:0.7.13
 # Build arguments for metadata (all optional with defaults)
 ARG BUILD_DATE=""
 ARG VCS_REF=""
-ARG KILOCODE_CLI_VERSION="next"
+ARG KILOCODE_CLI_VERSION="7.0.47"
 
 # Build the kilo binary:
 #   cd ~/projects/kilocode-backend/cloud-agent
--- a/cloud-agent-next/src/execution/orchestrator.ts
+++ b/cloud-agent-next/src/execution/orchestrator.ts
@@ -60,7 +60,10 @@ export class ExecutionOrchestrator {
    *
    * @throws ExecutionError with appropriate code on failure (no internal retry)
    */
-  async execute(plan: ExecutionPlan): Promise<ExecutionResult> {
+  async execute(
+    plan: ExecutionPlan,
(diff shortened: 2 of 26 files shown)


.kilocode/plan.md
```diff
@@ -0,0 +1,183 @@
+# PR #1151 Review Fixes Plan
```
jeanduplessis: Do you plan to keep this in the codebase long-term?

jeanduplessis (approved):

.kilocode/plan.md
eshurakov: forgot to delete and agent pushed it, thanks!

PR #1166, opened 2026-03-17
Preparation to hide free MiniMax and deprecated auto models

chrarnoldus:


lambertjosh:
I'm still seeing Minimax M2.5 free: 
image


Both in the extension and in the backend.

chrarnoldus:
Should work now

src/app/api/openrouter/[...path]/route.ts
```diff
@@ -323,8 +324,8 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
     return alphaPeriodEndedResponse();
   }
 
-  if (isRateLimitedToDeath(originalModelIdLowerCased)) {
-    return modelDoesNotExistResponse();
+  if (!autoModel && isForbiddenFreeModel(originalModelIdLowerCased)) {
```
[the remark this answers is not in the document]
Comment (chrarnoldus) [reply src/app/api/openrouter/[...path]/route.ts]: if it hurts then don't do that

src/app/api/openrouter/[...path]/route.ts
```diff
@@ -318,13 +319,16 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
...
-
-  if (isRateLimitedToDeath(originalModelIdLowerCased)) {
-    return modelDoesNotExistResponse();
+  if (
+    isDeadFreeModel(originalModelIdLowerCased) ||
+    (!autoModel && isForbiddenFreeModel(originalModelIdLowerCased))
+  ) {
+    console.warn(`User requested forbidden free model ${originalModelIdLowerCased}; rejecting.`);
+    if (isRooCodeBasedClient(fraudHeaders)) {
+      return alphaPeriodEndedResponse();
```
Comment (chrarnoldus) [inline src/app/api/openrouter/[...path]/route.ts:328]: The old extension gives a better error message this way, theoretically.

src/app/api/openrouter/[...path]/route.ts
```diff
@@ -64,7 +65,7 @@ import {
 import { handleRequestLogging } from '@/lib/handleRequestLogging';
 import { customLlmRequest } from '@/lib/custom-llm/customLlmRequest';
 import { normalizeModelId } from '@/lib/model-utils';
-import { isRateLimitedToDeath } from '@/lib/rate-limited-models';
+import { isForbiddenFreeModel } from '@/lib/forbidden-free-models';
```
Comment (chrarnoldus) [inline src/app/api/openrouter/[...path]/route.ts:68]: Rename

src/lib/providers/morph.ts
```diff
@@ -7,7 +7,7 @@ export const morph_warp_grep_free_model: KiloFreeModel = {
     'A code search subagent that finds relevant code in a separate context window — no embeddings, no indexing.',
   context_length: 256000,
   max_completion_tokens: 32000,
-  status: 'public',
+  status: 'hidden',
```
chrarnoldus, line 10: This is a special purpose model, it doesn't need to be in the list.

src/app/api/openrouter/[...path]/route.ts
```diff
@@ -318,13 +319,16 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
...
-  if (isDeadFreeModel(originalModelIdLowerCased)) {
-    console.warn(`User requested discontinued free model ${originalModelIdLowerCased}; rejecting.`);
-    return alphaPeriodEndedResponse();
-  }
-
-  if (isRateLimitedToDeath(originalModelIdLowerCased)) {
-    return modelDoesNotExistResponse();
+  if (
+    isDeadFreeModel(originalModelIdLowerCased) ||
+    (!autoModel && isForbiddenFreeModel(originalModelIdLowerCased))
```
Comment (chrarnoldus) [inline src/app/api/openrouter/[...path]/route.ts:324]: Allow a 'forbidden' model if it was selected through kilo-auto-free

markijbema (approved):

chrarnoldus:
@lambertjosh follow-up to actually do it: https://github.com/Kilo-Org/cloud/pull/1266

src/lib/providers/morph.ts
lambertjosh: Does flagging this way take it out of all model pickers? @kirillk just added something similar, but only for chat session picker.

PR #1188, opened 2026-03-17
Improvement/kiloclaw pair with controller read file

evanjacobson:
## Summary

Replaces the `PairingCache`'s use of `openclaw pairing list <channel> --json` and `openclaw devices list --json` subprocess calls with direct `fs.readFile` reads of the JSON files openclaw writes to disk:

- Channel requests: `~/.openclaw/credentials/<channel>-pairing.json`
- Device pending: `~/.openclaw/devices/pending.json`

This eliminates the subprocess overhead (45s timeout, process fork, Node.js startup) on the hot read path. Approvals still go through the CLI since they need openclaw's lock/notify logic.

**Architectural changes:**
- `ReadChannelPairingImpl` / `ReadDevicePairingImpl` injectable function types replace the old exec-based fetch path; path resolution happens at call time for testability
- `nowMsImpl` injectable added (distinct from `nowImpl`) for TTL comparisons without Date parsing
- Path resolvers exported (`resolveOpenClawStateDir`, `resolveCredentialsDir`, `resolveDevicePendingPath`), built from `OPENCLAW_STATE_DIR` / `OPENCLAW_OAUTH_DIR` env vars, mirroring openclaw's own `paths.ts` resolution including the legacy `CLAWDBOT_STATE_DIR` fallback
- TTL filtering now happens in the controller at read time, exactly mirroring openclaw's own expiry logic:
  - Channel: 60 min TTL matching `pairing-store.ts isExpired()`
  - Device: 5 min TTL matching `pairing-files.ts pruneExpiredPending()`, including the NaN-preserves-missing-ts behaviour
- Cold-start ENOENT (no prior data) is fully silent; ENOENT on a channel that previously had data logs `WARNING: keeping stale data` and preserves the cache
- `start()` now fires immediately via `runPeriodicRefresh()` instead of the previous 120s `INITIAL_REFRESH_DELAY_MS` delay (which was only needed to avoid CLI subprocess churn)
- Observability logs added: periodic/debounced refresh triggers, per-channel read results, device read result, debounce armed, and backoff-skip with remaining time

**Test changes (+219 net lines):**
- Harness switched to `readChannelPairingImpl` / `readDevicePairingImpl` fns throughout
- 8 new test cases: TTL boundary (exact / just-past), expired/invalid `createdAt`, missing `ts` preservation, ENOENT first-read and stale-preservation paths
- ENOENT first-read test strengthened to assert `console.warn` not called at all
- Failure backoff test rewritten to use `readChannelPairingImpl` (succeed once to populate cache, then fail) rather than `execImpl`
- Periodic refresh test description updated from "60s" to "120s / fires immediately"

## Verification

- `pnpm typecheck` — pass
- `pnpm test` — pass (all test suites including `pairing-cache.test.ts`)
- `pnpm lint` — pass
- Manual testing pairing telegram

## Visual Changes

N/A

## Reviewer Notes

The path resolution logic in `resolveOpenClawStateDir` / `resolveCredentialsDir` / `resolveDevicePendingPath` deliberately omits the filesystem-existence checks that openclaw's full `resolveStateDir()` performs for legacy `.clawdbot` dirs — those paths are unreachable in the container since the Dockerfile always creates `/root/.openclaw`.

The `INITIAL_REFRESH_DELAY_MS` constant (and its `initialTimer`) has been removed entirely; it was only there to avoid hammering the CLI on startup, which is no longer a concern with direct file reads.

### DIFF
--- a/kiloclaw/controller/src/pairing-cache.ts
+++ b/kiloclaw/controller/src/pairing-cache.ts
@@ -229,7 +229,9 @@ export function createPairingCache(options?: PairingCacheOptions): PairingCache 
             if (!Number.isFinite(createdAtMs)) return false; // garbage timestamp → expired
             return nowMs - createdAtMs <= CHANNEL_PAIRING_TTL_MS;
           });
-        console.log(`[pairing-cache] channel ${channel}: read ok, ${filtered.length} request(s) after filtering`);
+        console.log(
+          `[pairing-cache] channel ${channel}: read ok, ${filtered.length} request(s) after filtering`
+        );
         return filtered;
       })
     );
@@ -359,7 +361,9 @@ export function createPairingCache(options?: PairingCacheOptions): PairingCache 
     if (stopped) return;
     const remaining = nextAllowedRefreshAt - Date.now();
     if (remaining > 0) {
-      console.log(`[pairing-cache] debounced refresh skipped (backoff, ${Math.ceil(remaining / 1000)}s remaining)`);
+      console.log(
+        `[pairing-cache] debounced refresh skipped (backoff, ${Math.ceil(remaining / 1000)}s remaining)`
+      );
       return;
     }
     console.log('[pairing-cache] debounced refresh');
--- a/kiloclaw/controller/src/pairing-cache.ts
+++ b/kiloclaw/controller/src/pairing-cache.ts
@@ -65,8 +65,10 @@ export const FAILURE_RETRY_MAX_MS = 300_000;
 export const CONFIG_PATH = '/root/.openclaw/openclaw.json';
 
 // TTL constants — exact matches to openclaw source
-export const CHANNEL_PAIRING_TTL_MS = 60 * 60 * 1000; // pairing-store.ts:15 PAIRING_PENDING_TTL_MS
-export const DEVICE_PAIRING_TTL_MS = 5 * 60 * 1000; // device-pairing.ts:98 PENDING_TTL_MS
+// https://github.com/openclaw/openclaw/blob/d073ec42cd7fabd1004f6959628743817a4cb0e8/src/pairing/pairing-store.ts#L15 PAIRING_PENDING_TTL_MS
+export const CHANNEL_PAIRING_TTL_MS = 60 * 60 * 1000;
+// https://github.com/openclaw/openclaw/blob/d073ec42cd7fabd1004f6959628743817a4cb0e8/src/infra/device-pairing.ts#L98 PENDING_TTL_MS
+export const DEVICE_PAIRING_TTL_MS = 5 * 60 * 1000;
 
 const CHANNEL_RE = /^[a-z][a-z0-9_-]{0,63}$/;
 const CODE_RE = /^[A-Za-z0-9]{1,32}$/;
@@ -222,7 +224,7 @@ export function createPairingCache(options?: PairingCacheOptions): PairingCache 
           .filter(req => req.code !== '' && req.id !== '')
           .filter(req => {
             // Mirrors pairing-store.ts isExpired() — PAIRING_PENDING_TTL_MS = 60 * 60 * 1000
-            // ~/Developer/OpenSource/openclaw/src/pairing/pairing-store.ts:171
+            // https://github.com/openclaw/openclaw/blob/d073ec42cd7fabd1004f6959628743817a4cb0e8/src/pairing/pairing-store.ts#L171
             if (!req.createdAt) return false; // falsy (undefined, empty string) → expired
             const createdAtMs = Date.parse(req.createdAt);
             if (!Number.isFinite(createdAtMs)) return false; // garbage timestamp → expired
@@ -281,7 +283,7 @@ export function createPairingCache(options?: PairingCacheOptions): PairingCache 
         .filter(isRecord)
         .filter(entry => {
           // Mirrors pairing-files.ts pruneExpiredPending() — PENDING_TTL_MS = 5 * 60 * 1000
-          // ~/Developer/OpenSource/openclaw/src/infra/device-pairing.ts:98
+          // https://github.com/openclaw/openclaw/blob/d073ec42cd7fabd1004f6959628743817a4cb0e8/src/infra/device-pairing.ts#L98
           // No typeof guard: if ts is missing, nowMs - undefined = NaN, NaN > TTL is false → preserved
           return !(nowMs - (entry.ts as number) > DEVICE_PAIRING_TTL_MS);
         })
--- a/kiloclaw/controller/src/pairing-cache.ts
+++ b/kiloclaw/controller/src/pairing-cache.ts
@@ -99,7 +99,9 @@ export function resolveOpenClawStateDir(): string {
 }
 
 export function resolveCredentialsDir(): string {
-  return process.env.OPENCLAW_OAUTH_DIR?.trim() || path.join(resolveOpenClawStateDir(), 'credentials');
+  return (
+    process.env.OPENCLAW_OAUTH_DIR?.trim() || path.join(resolveOpenClawStateDir(), 'credentials')
+  );
 }
 
 export function resolveDevicePendingPath(): string {
--- a/kiloclaw/controller/src/pairing-cache.ts
+++ b/kiloclaw/controller/src/pairing-cache.ts
@@ -130,19 +130,22 @@ function defaultReadConfigImpl(): unknown {
   return JSON.parse(fs.readFileSync(CONFIG_PATH, 'utf8'));
 }
 
-// No .passthrough() — Zod's default strip behavior drops unknown fields so that
-// new keys openclaw adds to <channel>-pairing.json don't leak into the controller
-// API response.  Mirrors the devicePendingEntrySchema pattern below.
-const channelPairingRequestSchema = z.object({
-  code: z.string(),
-  id: z.string(),
-  meta: z.unknown().optional(),
-  createdAt: z.string().optional(),
-});
+// Zod schemas for IO boundary parsing — .passthrough() keeps forward compatibility
+// when openclaw adds fields without breaking the controller.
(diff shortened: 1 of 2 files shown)


kiloclaw/controller/src/pairing-cache.ts
```diff
@@ -44,20 +45,29 @@ export type PairingCache = {
...
   readConfigImpl?: () => unknown;
   nowImpl?: () => string;
+  readChannelPairingImpl?: ReadChannelPairingImpl;
+  readDevicePairingImpl?: ReadDevicePairingImpl;
+  nowMsImpl?: () => number;
 };
 
 export const PERIODIC_INTERVAL_MS = 120_000;
 export const DEBOUNCE_DELAY_MS = 2_000;
-export const INITIAL_REFRESH_DELAY_MS = 120_000;
```
evanjacobson, line 55: no longer needed with the lightweight file reads

kiloclaw/controller/src/pairing-cache.ts
```diff
@@ -44,20 +45,29 @@ export type PairingCache = {
...
+
 export const FAILURE_RETRY_BASE_MS = 30_000;
 export const FAILURE_RETRY_MAX_MS = 300_000;
-export const CLI_TIMEOUT_MS = 45_000;
```
evanjacobson, line 58: no longer needed — not using CLI

kiloclaw/controller/src/pairing-cache.ts
```diff
@@ -176,29 +219,38 @@ export function createPairingCache(options?: PairingCacheOptions): PairingCache
               ...('createdAt' in r ? { createdAt: String(r.createdAt) } : {}),
             };
           })
-          .filter(req => req.code !== '' && req.id !== '');
+          .filter(req => req.code !== '' && req.id !== '')
+          .filter(req => {
+            // Mirrors pairing-store.ts isExpired() — PAIRING_PENDING_TTL_MS = 60 * 60 * 1000
+            // ~/Developer/OpenSource/openclaw/src/pairing/pairing-store.ts:171
```
evanjacobson: changing to a permalink

kiloclaw/controller/src/pairing-cache.ts
```diff
@@ -78,13 +90,35 @@ function approveFail(message: string, statusHint: 400 | 500): ApproveResult {
...
 export const OPENCLAW_BIN = '/usr/local/bin/openclaw';
 
+// Mirrors resolveStateDir() / resolveOAuthDir() in openclaw/src/config/paths.ts
+// Includes legacy CLAWDBOT_STATE_DIR fallback (openclaw paths.ts:65)
+// Note: openclaw's full resolveStateDir() also does filesystem-existence checks for
+// legacy .clawdbot dirs — those are omitted here because the container Dockerfile
+// always creates /root/.openclaw, making the existence check unreachable in practice.
+export function resolveOpenClawStateDir(): string {
+  return (
+    process.env.OPENCLAW_STATE_DIR?.trim() ||
```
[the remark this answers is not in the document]
evanjacobson: The container is the only deployment context — OPENCLAW_STATE_DIR and OPENCLAW_OAUTH_DIR are never set in the Dockerfile, so the ~-expansion path is unreachable in production. The fix adds os import, a helper function, restructured conditionals, and a test for a code path that will never execute in practice. The complexity cost exceeds the risk.

If someone ever configures a custom state dir, the misconfiguration would be immediately obvious (ENOENT on every read → backoff fires → logs show the bad path).

kiloclaw/controller/src/pairing-cache.ts
```diff
@@ -208,40 +270,60 @@ export function createPairingCache(options?: PairingCacheOptions): PairingCache
...
-          const r = isRecord(req) ? req : {};
-          return {
-            requestId: String(r.requestId ?? ''),
-            deviceId: String(r.deviceId ?? ''),
-            ...(r.role !== undefined ? { role: String(r.role) } : {}),
-            ...(r.platform !== undefined ? { platform: String(r.platform) } : {}),
-            ...(r.clientId !== undefined ? { clientId: String(r.clientId) } : {}),
-            ...(typeof r.ts === 'number' ? { ts: r.ts } : {}),
-          };
+      const parsed: unknown = await readDevicePairingImpl();
```
pandemicsyn, line 331: @evanjacobson  a good rule of thumb John gave me the other day is to always use zod at io boundaries. And this is probably a solid example of one of those scenarios. 

So this has issues like readDevicePairingImpl() returns unknown, but refreshDevicePairingInternal() only checks isRecord(parsed) and then operates on arbitrary object values. The TTL filter  preserves entries where ts is missing or invalid  - so malformed entries with missing/non-numeric ts flow into TTL logic and are treated as non-expired forever.

i think the move might be to add a zod schema for the channel file shape and normalize once at parse time. Something like: 

```typescript
const channelPairingRequestSchema = z
  .object({
    code: z.string(),
    id: z.string(),
    meta: z.unknown().optional(),
    createdAt: z.string().optional(),
  })
  .passthrough();
const channelPairingFileSchema = z
  .object({
    requests: z.array(channelPairingRequestSchema).catch([]),
  })
  .passthrough();
```

I think we want that `.passthrough` so that if openclaw adds fields this wont break by default.

And a similar pattern for device pairing.

pandemicsyn: heres a quicky recommendation from Kilo that looks good i think, that you can probably just feed to your local agent:

```
For channel pairing:
import { z } from 'zod';
const channelPairingRequestSchema = z
  .object({
    code: z.string(),
    id: z.string(),
    meta: z.unknown().optional(),
    createdAt: z.string().optional(),
  })
  .passthrough();
const channelPairingFileSchema = z
  .object({
    requests: z.array(z.unknown()).catch([]),
  })
  .passthrough();
function parseChannelPairingFile(raw: unknown): ChannelPairingRequest[] {
  const parsedFile = channelPairingFileSchema.safeParse(raw);
  if (!parsedFile.success) {
    throw new Error('Invalid channel pairing file');
  }
  const requests: ChannelPairingRequest[] = [];
  for (const entry of parsedFile.data.requests) {
    const parsedEntry = channelPairingRequestSchema.safeParse(entry);
    if (!parsedEntry.success) {
      continue;
    }
    requests.push(parsedEntry.data);
  }
  return requests;
}
Then apply your current business rules after parsing:
const requests = parseChannelPairingFile(raw)
  .filter(req => req.code !== '' && req.id !== '')
  .filter(req => {
    if (!req.createdAt) return false;
    const createdAtMs = Date.parse(req.createdAt);
    if (!Number.isFinite(createdAtMs)) return false;
    return nowMs - createdAtMs <= CHANNEL_PAIRING_TTL_MS;
  });
For device pairing, this pattern is even more useful because the file is a record keyed by request ID:
import { z } from 'zod';
const devicePendingEntrySchema = z
  .object({
    requestId: z.string(),
    deviceId: z.string(),
    role: z.string().optional(),
    platform: z.string().optional(),
    clientId: z.string().optional(),
    ts: z.number().optional(),
  })
  .passthrough();
const devicePendingFileSchema = z.record(z.string(), z.unknown());
function parseDevicePendingFile(raw: unknown): DevicePairingRequest[] {
  const parsedFile = devicePendingFileSchema.safeParse(raw);
  if (!parsedFile.success) {
    throw new Error('Invalid device pairing file');
  }
  const requests: DevicePairingRequest[] = [];
  for (const entry of Object.values(parsedFile.data)) {
    const parsedEntry = devicePendingEntrySchema.safeParse(entry);
    if (!parsedEntry.success) {
      continue;
    }
    requests.push(parsedEntry.data);
  }
  return requests;
}
Then keep the existing filtering/sorting logic:
const requests = parseDevicePendingFile(raw)
  .filter(req => req.requestId !== '' && req.deviceId !== '')
  .filter(req => !(nowMs - (req.ts as number) > DEVICE_PAIRING_TTL_MS))
  .sort((a, b) => (b.ts ?? 0) - (a.ts ?? 0));
I would tighten that last TTL filter a bit to avoid as and keep current semantics:
function isUnexpiredDeviceRequest(req: DevicePairingRequest, nowMs: number): boolean {
  if (req.ts === undefined) {
    return true;
  }
  return nowMs - req.ts <= DEVICE_PAIRING_TTL_MS;
}
Then:
const requests = parseDevicePendingFile(raw)
  .filter(req => req.requestId !== '' && req.deviceId !== '')
  .filter(req => isUnexpiredDeviceRequest(req, nowMs))
  .sort((a, b) => (b.ts ?? 0) - (a.ts ?? 0));
If you want to keep it very compact, you can extract a reusable helper:
function collectValidEntries<TSchema extends z.ZodTypeAny>(
  items: Iterable<unknown>,
  schema: TSchema,
): Array<z.infer<TSchema>> {
  const results: Array<z.infer<TSchema>> = [];
  for (const item of items) {
    const parsed = schema.safeParse(item);
    if (parsed.success) {
      results.push(parsed.data);
    }
  }
  return results;
}
Usage:
const parsedFile = channelPairingFileSchema.parse(raw);
const requests = collectValidEntries(parsedFile.requests, channelPairingRequestSchema);
and
const parsedFile = devicePendingFileSchema.parse(raw);
const requests = collectValidEntries(Object.values(parsedFile), devicePendingEntrySchema);
That gives you:
- strict file boundary
- tolerant per-entry parsing
- no brittle all-or-nothing failure on one malformed request
- explicit separation between structural validation and TTL/business filtering
This is the style I’d use for this controller code.
```

pandemicsyn (changes requested):
Looking good, one ask inline

kiloclaw/controller/src/pairing-cache.ts
evanjacobson: I leaned into the agent review regarding passthrough at first, but doubled back. I think your reasoning about keeping it functional by default is more important.

pandemicsyn (approved):

PR #1190, opened 2026-03-17
feat(kiloclaw): add structured logging to DO modules and client context

jeanduplessis:
## Summary

- Replace ~46 `console.error`/`console.warn` calls across 8 DO module files with structured JSON logging via new `doError`/`doWarn` helpers in `log.ts`. Each log line now emits `{ tag: "kiloclaw_do", level, message, userId, sandboxId, flyMachineId, flyRegion, flyAppName, ...details }`, making logs queryable and eliminating the need to grep for user context manually.
- Error objects are serialized via `serializeError()` which preserves `name`, `message`, `stack`, and all own enumerable properties (e.g. `FlyApiError.status`), fixing the previous issue where `console.error(err)` in Workers often produced `[object Object]`.
- Add `RequestContext` threading to `KiloClawInternalClient` and `KiloClawUserClient` so HTTP error logs include `userId`. The router call sites pass `{ userId: ctx.user.id }`.
- Add `kiloUserId` parameter to `stripe-handlers.ts:getSubscriptionPeriods()` so the "no items" warning identifies the user.
- `JSON.stringify` in the logging helpers is wrapped in try/catch to prevent serialization failures from crashing error-handling paths.

## Verification

- [x] `pnpm typecheck` — pass (kiloclaw)
- [x] `pnpm typecheck` — pass (root, only pre-existing `@radix-ui/react-collapsible` error)
- [x] `pnpm test` — 807/807 pass (kiloclaw)
- [x] `pnpm lint` — pass (kiloclaw)
- [x] Manual audit of all 45 conversion sites confirmed zero information loss vs. original logs

## Visual Changes

N/A

## Reviewer Notes

- `console.log` calls (info-level operational logs) are intentionally out of scope — only `console.error` and `console.warn` were converted.
- The `state.ts:72` warn (persisted state validation failed) is intentionally left as plain `console.warn` since state is not yet loaded and context fields would all be null.
- `config.ts:mintFreshApiKey()` at line 66 keeps plain `console.warn` since it has no `state` parameter.
- The spread order in `doError`/`doWarn` puts `instanceContext(state)` after `serializeDetails(details)` so context fields always win over caller-supplied details, preventing accidental shadowing.
- `toLoggable(err)` is exported for call sites to coerce `unknown` catch values into `Error | string` — eliminates the `err instanceof Error ? err : String(err)` pattern that was repeated ~25 times.

### DIFF
--- a/kiloclaw/src/durable-objects/kiloclaw-instance.test.ts
+++ b/kiloclaw/src/durable-objects/kiloclaw-instance.test.ts
@@ -110,6 +110,26 @@ import {
 // Test harness
 // ============================================================================
 
+/**
+ * Find a structured doWarn call by message substring and verify the JSON envelope.
+ * Returns the parsed log payload for further assertions.
+ */
+function expectStructuredWarn(spy: Mock, messageSubstring: string) {
+  const call = spy.mock.calls.find(
+    (c: unknown[]) => typeof c[0] === 'string' && c[0].includes(messageSubstring)
+  );
+  if (!call) throw new Error(`Expected a warn call containing "${messageSubstring}"`);
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- JSON.parse returns any
+  const parsed: Record<string, unknown> = JSON.parse(call[0] as string);
+  expect(parsed.tag).toBe('kiloclaw_do');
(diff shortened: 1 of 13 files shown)


pandemicsyn (approved):
lgtm

PR #1193, opened 2026-03-18
refactor(kiloclaw): move bootstrap logic from shell into controller

pandemicsyn:
## Summary

Move ~500 lines of startup logic from `start-openclaw.sh` into a new `controller/src/bootstrap.ts` TypeScript module so the controller's HTTP server starts **before** bootstrap and is always interrogable via `/_kilo/health`. Then delete the shell script entirely since all it did was `exec` the controller — Fly machines run a single CMD process, so the Dockerfile now invokes the controller binary directly.

**Problem:** If `openclaw onboard` or `openclaw doctor` hung or crashed in the shell script, the controller never started. The CF worker couldn't reach the machine — it appeared as a black hole.

**Architecture changes:**
- New `bootstrap.ts` with 7 exported, deps-injected step functions (`decryptEnvVars`, `setupDirectories`, `applyFeatureFlags`, `generateHooksToken`, `configureGitHub`, `runOnboardOrDoctor`, `buildGatewayArgs`) plus a thin `bootstrap()` orchestrator.
- Two-phase HTTP server in `index.ts`: a lightweight inline handler serves `/_kilo/health` during bootstrap; after bootstrap succeeds, a fresh Hono app with full routes is swapped in.
- Health endpoint now returns `{ status: 'ok', state, phase?, error? }` — always HTTP 200 for Fly probes, with a `state` key for CF worker observability (`bootstrapping` → `starting` → `ready`, or `degraded` on failure).
- `start-openclaw.sh` deleted. Both `Dockerfile` and `Dockerfile.local` CMD now point directly to `node /usr/local/bin/kiloclaw-controller.js`.
- Bootstrap yields to the event loop (`setImmediate`) between phases so the inline health handler can respond during blocking `execFileSync` calls.
- WebSocket upgrades during bootstrap/degraded mode get a clean 503 rejection instead of hanging.
- All degraded-state errors are sanitized via `toPublicDegradedError(stage)` — raw errors (which can include secrets in `execFileSync` argv) stay in logs only.

### `/_kilo/health` response by state

| When | Response | Served by |
|------|----------|-----------|
| Bootstrap running | `{"status":"ok","state":"bootstrapping","phase":"decrypting"}` | Inline handler |
| Bootstrap progressing | `{"status":"ok","state":"bootstrapping","phase":"onboard"}` | Inline handler |
| Bootstrap failed | `{"status":"ok","state":"degraded","error":"Startup failed during bootstrap"}` | Inline handler |
| Runtime config failed | `{"status":"ok","state":"degraded","error":"Startup failed during runtime-config"}` | Inline handler |
| Supervisor/route setup failed | `{"status":"ok","state":"degraded","error":"Startup failed during post-bootstrap"}` | Inline handler |
| Routes registered, gateway starting | `{"status":"ok","state":"starting"}` | Hono |
| Fully operational | `{"status":"ok","state":"ready"}` | Hono |

All responses return HTTP 200 with `status: 'ok'` so Fly health probes stay happy. Basically, `/_kilo/health` describes the machine state, and the auth-gated `/_kilo/gateway/status` describes the gateway status. 

**Additional changes**
- Post-bootstrap failures (`loadRuntimeConfig`, `supervisor.start()`, route setup) now degrade instead of crashing.
- Removed moltworker Apache 2.0 LICENSE holdover — all fork-derived code has been replaced.
- Purged all `start-openclaw.sh` references across source comments, docs, AGENTS.md, and test descriptions.

## Verification

- [x] `pnpm typecheck` — pass
- [x] `pnpm test` — 863/863 tests pass (39 test files), including 37 new bootstrap tests, sanitization tests, and health state progression tests
- [x] `pnpm lint` on changed files — clean (1 pre-existing error in `parseGatewayArgs`, not introduced by this PR)
- [x] Line-by-line audit of old shell script vs new bootstrap.ts — all 34 sections verified
- [x] Verified Hono route-shadowing behavior empirically (first registered route wins) — confirmed the two-phase server approach avoids the issue
- [x] `scripts/controller-smoke-test.sh` — 11/11 checks pass (fresh onboard path, auth, env patch, version endpoints)
- [x] `scripts/controller-entrypoint-smoke-test.sh` — pass (doctor path with pre-seeded config, all auth checks correct)
- [x] Zero remaining `start-openclaw.sh` references in codebase (`rg` clean)

## Visual Changes

N/A

## Reviewer Notes

- The `bootstrap()` orchestrator is `async` with `setImmediate` yields between phases. The individual step functions use `execFileSync` (especially `openclaw onboard` and `openclaw doctor`) which block the event loop for seconds. The yields let the inline health handler respond between phases.
- `runOnboardOrDoctor` checks `deps.existsSync(CONFIG_PATH)` twice: once in `bootstrap()` to set the phase name, once internally. This is deliberate — the orchestrator needs it for the phase label, and the function needs it for its own control flow. The filesystem call is cheap.
- `/_kilo/health` describes controller/machine lifecycle. `/_kilo/gateway/status` (auth-gated) describes the gateway process. These are intentionally separate concerns — the health endpoint stays focused on "can I talk to this machine" while gateway status covers process-level diagnostics.


kiloclaw/controller/src/index.ts
```diff
@@ -118,32 +120,147 @@ async function handleHttpRequest(
...
-    writeKiloCliConfig(env as Record<string, string | undefined>);
+    await bootstrap(env, phase => {
+      controllerState.current = { state: 'bootstrapping', phase };
+    });
   } catch (err) {
-    console.error('[kilo-cli] Failed to write config:', err);
+    const error = err instanceof Error ? err.message : String(err);
+    controllerState.current = { state: 'degraded', error };
+    console.error('[controller] Bootstrap failed, running in degraded mode:', error);
+    return; // HTTP server stays alive for health probes
```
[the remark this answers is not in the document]
pandemicsyn: Yea this was/is valid. The gotcha is that the /_kilo/version and friends are auth'd. So theres a chicken and the egg problem. If we don't have a gateway token or decryption failed or whatever - these endpoints wont work.

I refactored, now once we've decrypted vars i start the actual hono app. That app *needs* the gateway token for auth - so it can't be started earlier than that.

kiloclaw/controller/src/index.ts
```diff
@@ -118,32 +120,194 @@ async function handleHttpRequest(
...
-  const supervisor = createSupervisor({
+  supervisor = createSupervisor({
     args: ['gateway', ...config.gatewayArgs],
-    onStdoutLine: line => pairingCache.onPairingLogLine(line),
+    onStdoutLine: line => pc.onPairingLogLine(line),
   });
 
-  let gmailWatchSupervisor: Supervisor | null = null;
   let googleAccountEmail: string | null = null;
   const hasGogCredentials = Boolean(env.KILOCLAW_GOG_CONFIG_TARBALL);
```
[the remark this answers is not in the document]
pandemicsyn: i *think* we do actually want this to fail here, otherwise im not sure we'd ever know.

kiloclaw/Dockerfile
```diff
@@ -30,7 +30,7 @@ RUN apt-get update \
     && curl -fsSL https://downloads.1password.com/linux/keys/1password.asc \
        | gpg --dearmor --output /usr/share/debsig/keyrings/AC2D62742012EA22/debsig.gpg \
     && apt-get update \
-    && apt-get install -y --no-install-recommends gh 1password-cli=2.32.1-1 \
+    && apt-get install -y --no-install-recommends gh 1password-cli=2.33.0-1 \
```
RSO, line 33: Is this an unrelated change?

kiloclaw/Dockerfile
```diff
@@ -112,27 +112,23 @@ RUN mkdir -p /root/.openclaw \
     && mkdir -p /root/clawd \
     && mkdir -p /root/clawd/skills
 
-# Copy startup script
-# Build cache bust: 2026-03-16-v63-startup-optimization
+# Copy helper scripts (used at runtime by the controller/gateway)
+# Build cache bust: 2026-03-17-v65-remove-startup-script
 RUN echo "10"
```
RSO, line 114: My agent tried to do this as well at some point, but I'm pretty sure that to actually bust the cache you need to increase the number on the `RUN echo ""` line :D

RSO (approved):
Tested this locally and it seems to work as expected 👌

St0rmz1 (commented):
**config-writer.ts line 106** — In the doctor path (`bootstrap.ts:275-290`), `openclaw doctor --fix` just ran against an existing config, then `generateBaseConfig` is called to patch it. If `JSON.parse` fails here, the catch silently falls through to `console.log('Starting with empty config')` and rebuilds from env vars — wiping any user customizations (channels, plugins, model preferences) without warning.

On the onboard path this is fine (no prior config to lose), but on the doctor path it's silent data loss. Consider at least upgrading this to a `console.error` with a distinct message when called against a file that's expected to exist, e.g.:

```ts
} catch (err) {
  console.error(`Failed to parse config at ${configPath}, rebuilding from env vars. User customizations may be lost:`, err);
}
```

That way operators have a signal in logs that something unexpected happened, rather than it looking like a normal first boot.

St0rmz1 (approved):

kiloclaw/Dockerfile
pandemicsyn: ah yea, sorry - hacked this up temporarily why i was waiting for the other PR.

kiloclaw/Dockerfile
pandemicsyn: heh yea - think this one worked since there was also a legit update the entry point/cmd. I'll update AGENTS.md.

PR #1214, opened 2026-03-18
Hardcode separate value for Vercel routing when error rate is high

chrarnoldus:
Reverts Kilo-Org/cloud#1009


markijbema (approved):
You can as easily make it a constant, as we do need a deploy for changing a env var as well because we build in github actions

PR #1276, opened 2026-03-19
chore(lint): migrate to oxlint

jeanduplessis:
## Summary

Migrate the repo to an oxlint-first lint flow while keeping ESLint as a targeted fallback for rules that still need it. This adds a shared root `.oxlintrc.json`, introduces a dedicated root `eslint.fallback.config.mjs` for the slower type-aware app-only checks, updates workspace `lint` scripts to run `lint:oxlint` before `lint:eslint:fallback`, and removes the unused root `eslint.config.mjs`.

The root app fallback is intentionally narrower now so `pnpm lint` spends less time in type-aware ESLint. Package-local `eslint.config.mjs` files remain in place for workspace fallbacks. This also fixes one internal Next.js navigation link that oxlint surfaced during rollout.

## Verification

- [x] `PATH="$HOME/.nvm/versions/node/v22.22.1/bin:$PATH" pnpm lint` (pass)
- [x] `git push -u origin jdp/migrate-oxlint` pre-push hook ran `pnpm format:check`, `pnpm lint`, and `pnpm typecheck` (pass)

## Visual Changes

N/A

## Reviewer Notes

Root linting now uses `eslint.fallback.config.mjs`; package-local fallback scripts still use their package `eslint.config.mjs`. The remaining ESLint footprint is intentional for unsupported or type-aware rules while oxlint handles the broad fast pass.


eslint.fallback.config.mjs
```diff
@@ -0,0 +1,65 @@
...
+    languageOptions: {
+      parser: tseslint.parser,
+      parserOptions: {
+        projectService: {
+          allowDefaultProject: ['*.js', '*.mjs'],
+        },
+        tsconfigRootDir: __dirname,
+      },
+    },
+    plugins: {
```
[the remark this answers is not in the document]
jeanduplessis: False positive. This PR does not reduce existing hook-rule coverage.

I verified the pre-migration root ESLint config and it did not enable `react-hooks/rules-of-hooks` or `react-hooks/exhaustive-deps`; it only registered `eslint-plugin-react-hooks`. A synthetic hook violation also passed under the old root ESLint config.

So this migration did not remove enforcement that was previously active in `pnpm lint`. Enabling the Oxlint equivalents is possible, but that is separate work and currently surfaces a large existing backlog in `src`.

packages/eslint-config/eslint.config.base.mjs
```diff
@@ -1,13 +1,19 @@
...
 
 /**
  * Shared ESLint base config for all packages.
  * @param {string} tsconfigRootDir - The package's __dirname, used for TypeScript project resolution.
  */
 export default function baseConfig(tsconfigRootDir) {
   return defineConfig([
+    {
+      linterOptions: {
+        reportUnusedDisableDirectives: 'off',
```
alex-alecu, line 14: Why?

jeanduplessis: We now default to Oxide as the primary linter and only the rules we don't have covered in Oxide we run through ESLint (since its much slower).

eshurakov (approved):

PR #1312, opened 2026-03-20
feat(ui): add sessions view to Usage page

shssoichiro:
## Summary
This adds a view to the Usage page to allow users to see a list of each
individual session's spending and token count. The default is to show
the most recent 100 sessions, sorted by most recent. Time period filters also apply.
Pagination at the bottom will allow users to view older sesions.

## Verification

- Navigate to the Usage page
- Click the new Sessions tab

## Visual Changes

image


sarendipitee:
This would be an amazing feature for identifying sessions!

emilieschario:
Hi! We are currently working on the usage page, so we won't be merging this for now, but thanks for submitting!

PR #1314, opened 2026-03-20
fix(dev): keep non-production DB reads on primary

jeanduplessis:
## Summary

Keep non-production database reads on the primary Postgres URL instead of using configured read replicas.

This prevents local development from depending on remote Supabase replica hosts when `.env.local` includes `POSTGRES_REPLICA_*` settings, which was causing `next dev` database requests to hang or time out.

## Verification

- [x] `pnpm exec oxfmt src/lib/drizzle.ts`
- [x] `pnpm exec tsgo --noEmit --incremental false`
- [x] `pnpm test`
- [x] Pre-push hooks passed: `pnpm run format:check`, `pnpm run lint`, `pnpm run typecheck`

## Visual Changes

N/A

## Reviewer Notes

The functional change is limited to `getReplicaUrl()` in `drizzle.ts`. Non-production environments now always use the primary/local Postgres URL, even if replica env vars are present.

### DIFF
--- a/src/components/PostHogProvider.tsx
+++ b/src/components/PostHogProvider.tsx
@@ -16,6 +16,7 @@ export function PostHogProvider({ children }: { children: React.ReactNode }) {
       return;
     }
 
+    // Use a fake token for non-production environments as recommended by PostHog
     const token = isProduction ? key : 'fake-token';
 
     posthog.init(token, {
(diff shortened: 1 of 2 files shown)


RSO (changes requested):
I'm not convinced that this makes sense. You had the production values for `POSTGRES_REPLICA_US_URL` in your development `.env.local`? How did that ever end up there 😱

PR #1370, opened 2026-03-21
feat(gastown): add debug dry-run endpoint with event draining

jrf0110:
## Summary

Adds a `POST /debug/towns/:townId/reconcile-dry-run` endpoint that executes the reconciler's Phase 0 (drain events → apply → mark processed) and Phase 1 (reconcile) against current state, returning the actions it would emit without applying them. The entire operation runs inside a SQLite SAVEPOINT that is rolled back in a `finally` block, keeping the endpoint fully side-effect-free.

This gives operators a way to preview what the next alarm tick would do — including the effect of pending unprocessed events — without triggering any side effects.

## Verification

- Polecat reports typecheck and oxlint pass
- Code review confirmed the drain/apply/reconcile sequence matches the real alarm loop (Town.do.ts L2925-2955)
- SAVEPOINT/ROLLBACK pattern is correct SQLite (ROLLBACK TO + RELEASE in finally)
- `Pick<ReconcilerMetrics, ...>` fields all exist on the type
- No build artifacts or secrets in the diff

## Visual Changes

N/A

## Reviewer Notes

- The endpoint is unauthenticated, following the same pattern as the existing `GET /debug/towns/:townId/status` endpoint
- Unlike the real alarm loop, there is no per-event try/catch — a single `applyEvent` failure will propagate up and trigger the savepoint rollback. This is intentional for a debug tool where you want errors to be visible
- The `Action` type import was added alongside the existing `ApplyActionContext` import

### DIFF
--- a/cloudflare-gastown/src/dos/Town.do.ts
+++ b/cloudflare-gastown/src/dos/Town.do.ts
@@ -30,7 +30,7 @@ import * as scheduling from './town/scheduling';
 import * as events from './town/events';
 import * as reconciler from './town/reconciler';
 import { applyAction } from './town/actions';
-import type { ApplyActionContext } from './town/actions';
+import type { Action, ApplyActionContext } from './town/actions';
 import { buildRefinerySystemPrompt } from '../prompts/refinery-system.prompt';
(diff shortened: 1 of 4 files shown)


jrf0110:
Refinery code review passed. All quality gates pass.

PR #1389, opened 2026-03-23
fix(billing): move auto-intro metadata from schedule create to update

jeanduplessis:
### Problem

`createAutoIntroSchedule` passes `metadata: { origin: 'auto-intro' }` alongside `from_subscription` in `stripe.subscriptionSchedules.create()`. Stripe rejects this combination with HTTP 400, causing 100% failure on auto-intro schedule creation since commit df6ff4c0 (PR #1362). Affected intro-priced subscriptions never get the scheduled transition to the regular standard price.

### Solution

- Removed `metadata` from the `subscriptionSchedules.create()` call where Stripe forbids it alongside `from_subscription`.
- Added `metadata: { origin: 'auto-intro' }` to the subsequent `subscriptionSchedules.update()` call, which already sets phases and end behavior.
- Wrapped the `update` call in error recovery: if it fails, the half-created schedule is released so retry paths (Sweep 5 cron) can start fresh instead of encountering an untagged schedule they cannot identify.
- Added race recovery: `ensureAutoIntroSchedule` and `handleAutoIntroCreateRace` now detect and claim untagged orphan schedules (create succeeded, update never ran) by tagging them as `auto-intro` and running repair.
- Added a `phases.length === 1` guard to the untagged-schedule claiming logic to avoid incorrectly claiming user-initiated plan switches or kilo-pass changes, which also create `from_subscription` schedules without `metadata.origin` but always have 2+ phases after their update completes.

### Why this approach

The metadata must live on the schedule for downstream code to distinguish auto-intro schedules from user-initiated plan switches. Moving it to the `update` call is the minimal fix — the only alternative would be two separate `update` calls (one for metadata, one for phases), which doubles API calls for no benefit since both fields can be set atomically in one `update`.

### Related issues

- Introduced in PR #1362 (`feat(billing): replace coupon with intro pricing and enable promo codes`)

## Verification

- [x] `pnpm typecheck` — pass
- [x] `pnpm test -- src/routers/kiloclaw-billing-router.test.ts` — 59/59 pass
- [x] Six-agent code review (security, logic, types, data, resources, style) — all clean after fixing the one WARNING
- [x] Pre-push hooks (format, lint, typecheck) — pass

## Visual Changes

N/A

## Reviewer Notes

- **Risk area:** The window between `create` and `update` where the schedule exists without metadata. Three layers of mitigation: (1) release on update failure, (2) race recovery paths claim untagged orphans, (3) `phases.length === 1` guard prevents incorrectly claiming user-initiated schedules.
- **Self-healing:** After deploy, the next billing lifecycle cron run (Sweep 5) will automatically repair the 3 stranded subscriptions by calling `ensureAutoIntroSchedule`, which invokes the now-fixed `createAutoIntroSchedule`.
- **Rollback:** Simple revert; no migrations or config changes.

### DIFF
--- a/src/lib/kiloclaw/stripe-handlers.ts
+++ b/src/lib/kiloclaw/stripe-handlers.ts
@@ -297,7 +297,6 @@ async function createAutoIntroSchedule(
   try {
     newSchedule = await stripe.subscriptionSchedules.create({
       from_subscription: stripeSubscriptionId,
-      metadata: { origin: 'auto-intro' },
     });
   } catch (error) {
     await handleAutoIntroCreateRace(error, stripeSubscriptionId, userId);
@@ -317,19 +316,31 @@ async function createAutoIntroSchedule(
     return;
   }
 
-  await stripe.subscriptionSchedules.update(newSchedule.id, {
-    phases: [
-      {
-        items: [{ price: phase1Price }],
-        start_date: currentPhase.start_date,
-        end_date: currentPhase.end_date,
-      },
-      {
-        items: [{ price: getStripePriceIdForClawPlan('standard') }],
-      },
-    ],
-    end_behavior: 'release',
-  });
+  try {
+    await stripe.subscriptionSchedules.update(newSchedule.id, {
+      metadata: { origin: 'auto-intro' },
+      phases: [
+        {
+          items: [{ price: phase1Price }],
+          start_date: currentPhase.start_date,
+          end_date: currentPhase.end_date,
+        },
+        {
+          items: [{ price: getStripePriceIdForClawPlan('standard') }],
+        },
+      ],
+      end_behavior: 'release',
+    });
+  } catch (error) {
+    // Release the half-created schedule so retry can start fresh — without
+    // metadata, recovery paths cannot identify it as auto-intro.
+    try {
+      await stripe.subscriptionSchedules.release(newSchedule.id);
+    } catch {
+      // best-effort cleanup
+    }
+    throw error;
+  }
 
   await persistAutoIntroSchedule(newSchedule.id, userId);
 }
--- a/src/lib/kiloclaw/stripe-handlers.ts
+++ b/src/lib/kiloclaw/stripe-handlers.ts
@@ -247,7 +247,20 @@ export async function ensureAutoIntroSchedule(
     if (!scheduleId) return;
     const schedule = await stripe.subscriptionSchedules.retrieve(scheduleId);
 
-    if (schedule.metadata?.origin === 'auto-intro') {
+    const isAutoIntro = schedule.metadata?.origin === 'auto-intro';
+
+    // A schedule created via from_subscription without metadata is likely a
+    // half-created auto-intro schedule (create succeeded but the subsequent
+    // update that sets metadata + phases hasn't run yet, or the worker died
+    // before it could). Claim it by tagging it before attempting repair.
+    const isUntagged = !schedule.metadata?.origin;
+    if (isUntagged) {
+      await stripe.subscriptionSchedules.update(schedule.id, {
+        metadata: { origin: 'auto-intro' },
+      });
+    }
+
+    if (isAutoIntro || isUntagged) {
       const valid = await validateOrRepairAutoIntroSchedule(schedule, stripeSubscriptionId, userId);
       if (!valid) {
         logError('Auto-intro schedule is unrecoverable, skipping', {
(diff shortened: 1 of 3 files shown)


src/lib/kiloclaw/stripe-handlers.ts
```diff
@@ -317,19 +316,31 @@ async function createAutoIntroSchedule(
...
-      },
-      {
-        items: [{ price: getStripePriceIdForClawPlan('standard') }],
-      },
-    ],
-    end_behavior: 'release',
-  });
+  try {
+    await stripe.subscriptionSchedules.update(newSchedule.id, {
+      metadata: { origin: 'auto-intro' },
```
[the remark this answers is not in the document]
jeanduplessis: Valid concern. Since Stripe rejects metadata on `subscriptionSchedules.create` when `from_subscription` is set, we can't eliminate the window entirely. Instead, I've made both recovery paths (`ensureAutoIntroSchedule` and `handleAutoIntroCreateRace`) tolerate tagless schedules:

- When either function encounters a schedule with no `metadata.origin`, it now tags it as `auto-intro` and proceeds to validate/repair — rather than skipping it as a non-auto-intro schedule.
- This is safe because only auto-intro setup creates schedules via `from_subscription`; user-initiated plan switches reuse existing schedules that already have metadata.
- Combined with the existing error-recovery `release()` in `createAutoIntroSchedule`, this closes the race window: concurrent readers will claim and repair orphaned schedules, and crash-orphaned schedules get cleaned up on retry.

src/lib/kiloclaw/stripe-handlers.ts
```diff
@@ -247,7 +247,24 @@ export async function ensureAutoIntroSchedule(
...
+
+    // A schedule created via from_subscription without metadata AND with only
+    // 1 phase is likely a half-created auto-intro schedule (create succeeded
+    // but the subsequent update that sets metadata + phases hasn't run yet, or
+    // the worker died before it could). We require exactly 1 phase because
+    // createAutoIntroSchedule sets metadata and phases in the same update call
+    // — so a fully-updated schedule always has both. User-initiated plan
+    // switches and kilo-pass changes also create untagged schedules but their
+    // update writes 2+ phases, so the phase-count guard avoids claiming them.
+    const isUntagged = !schedule.metadata?.origin && schedule.phases.length === 1;
```
alex-alecu: Can a crashed switchPlan or kilo-pass leave a 1-phase untagged schedule that this incorrectly claims as auto-intro?

src/lib/kiloclaw/stripe-handlers.ts
```diff
@@ -362,7 +390,21 @@ async function handleAutoIntroCreateRace(
...
   const existingSchedule = await stripe.subscriptionSchedules.retrieve(refetchedScheduleId);
-  if (existingSchedule.metadata?.origin === 'auto-intro') {
+
+  // The race winner may not have tagged the schedule yet (metadata is set in
+  // a separate update call). Only claim untagged schedules that still have a
+  // single phase — the from_subscription default. Schedules with 2+ phases
+  // were already fully configured by another code path (user plan switch,
+  // kilo-pass) and must not be claimed.
+  const isAutoIntro = existingSchedule.metadata?.origin === 'auto-intro';
+  const isUntagged = !existingSchedule.metadata?.origin && existingSchedule.phases.length === 1;
```
alex-alecu: This orphan-detect-and-claim block is duplicated from line 260. Can they drift if the heuristic changes in only one place?

src/lib/kiloclaw/stripe-handlers.ts
jeanduplessis: Fixed in two ways:

1. **Phase-count guard**: `claimIfAutoIntro()` now requires `phases.length === 1` (the `from_subscription` default) before claiming. Fully-configured schedules (2+ phases) are never claimed.

2. **Tag all code paths**: Added `metadata: { origin: 'user-switch' }` to KiloClaw switchPlan and `metadata: { origin: 'kilo-pass-switch' }` to Kilo Pass schedule creation. Once their update succeeds, the schedule is tagged and cannot be mistaken for an auto-intro orphan.

The only remaining window is a process crash between `create` and `update` in switchPlan/kilo-pass — at that point the schedule is 1-phase and untagged, indistinguishable from an auto-intro orphan at the Stripe level. This is an inherent limitation, but in practice both paths have catch-block cleanup that releases the schedule on any in-process error, so only a hard crash (OOM, container kill) could leave such an orphan.

src/lib/kiloclaw/stripe-handlers.ts
jeanduplessis: Extracted the duplicated logic into a shared `claimIfAutoIntro(schedule)` helper. Both `ensureAutoIntroSchedule` and `handleAutoIntroCreateRace` now call it, so the heuristic lives in one place.

alex-alecu (approved):

PR #1400, opened 2026-03-23
Add MiniMax now longer free notification and use materialized views

chrarnoldus:


src/lib/notifications.ts
```diff
@@ -237,6 +231,46 @@ async function generateTeamsTrialNotification(
   ];
 }
 
+async function generateMiniMaxNoLongerFreeNotification(
+  user: User,
+  _ctx: NotificationContext
+): Promise<KiloNotification[]> {
+  try {
+    const users = await cachedPosthogQuery(
```
markijbema, line 239: why not use the cohort functionality?

chrarnoldus: I don't know what that is

src/lib/notifications.ts
```diff
@@ -237,6 +231,46 @@ async function generateTeamsTrialNotification(
...
+      return [];
+    }
+
+    console.debug('[generateMiniMaxNoLongerFreeNotification] user has used MiniMax M2.5 free');
+    return [
+      {
+        id: 'minimax-no-longer-free-mar-23',
+        title: 'MiniMax M2.5 Free ending soon',
+        message:
+          'The MiniMax M2.5 free promotion ends soon. Please switch to Kilo Auto Free for continued access to free inference.',
```
lambertjosh:
```suggestion
          'The MiniMax M2.5 free promotion ends soon. Please switch to Kilo Auto Free or another free model.',
```

lambertjosh (approved):

src/lib/notifications.ts
markijbema: you can add users to a cohort, and then you don't need any query here, and not have the notifications depend on posthog.

To do so you have to add them to the cohort with an operation once though

chrarnoldus: Is cohort import in the admin panel or something?

markijbema: I don't think that was ever build, but it's probably a oneshot in opus

PR #1401, opened 2026-03-23
fix(cloud-agent): return 4xx for branch-not-found instead of 500

eshurakov:
## Summary

Return a `BAD_REQUEST` (4xx) instead of an internal server error (500) when an upstream branch does not exist during session preparation. Before this change, `manageBranch` threw a plain `Error` that propagated as a 500; now it throws a typed `BranchNotFoundError` that `session-prepare` catches and converts to a `TRPCError({ code: 'BAD_REQUEST' })`.

Additionally, a `git ls-remote` pre-validation step runs *before* cloning when `upstreamBranch` is set, failing fast (saving 30–120s of wasted clone time) when the branch is confirmed missing. The pre-check is purely additive — if `ls-remote` itself fails (network, auth), it silently passes and lets the clone path surface the real error.

Key extraction: `buildAuthenticatedGitUrl` was factored out of `cloneGitRepo` so it can be reused by the new `branchExistsOnRemote` helper.

## Verification

- [x] `pnpm format` — passed (auto-fixed 3 files)
- [x] `pnpm lint` — passed (0 warnings, 0 errors across all workspaces)
- [x] `pnpm typecheck` — passed (all workspace projects)

## Visual Changes

N/A

## Reviewer Notes

- PR ref branches (`refs/pull/N/head`) are intentionally skipped in the pre-validation — they aren't regular branches and `ls-remote --heads` won't find them.
- `branchExistsOnRemote` returns `true` on ls-remote failure to keep the check purely additive; the clone/checkout path already handles auth and network errors with better context.
- `GITHUB_PULL_REF_PATTERN` was made `export` so the pre-validation can reuse it.

### DIFF
--- a/cloud-agent-next/src/router/handlers/session-prepare.ts
+++ b/cloud-agent-next/src/router/handlers/session-prepare.ts
@@ -21,10 +21,10 @@ import { generateSandboxId, getSandboxNamespace } from '../../sandbox-id.js';
 import {
   checkDiskAndCleanBeforeSetup,
   setupWorkspace,
-  cloneGitHubRepo,
   cloneGitRepo,
   manageBranch,
   validateUpstreamBranchExists,
+  buildAuthenticatedGitUrl,
   BranchNotFoundError,
 } from '../../workspace.js';
 import { WrapperClient } from '../../kilo/wrapper-client.js';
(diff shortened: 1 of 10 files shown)


jeanduplessis (approved):

cloud-agent-next/src/router/handlers/session-prepare.ts
```diff
@@ -362,6 +364,29 @@ const prepareSessionHandler = internalApiProtectedProcedure
       // 7. Clone repository
       const cloneOptions = input.shallow ? { shallow: true } : undefined;
       logger.info('Cloning repository');
+      // Pre-validate upstream branch exists before cloning (saves 30-120s of wasted compute)
+      if (input.upstreamBranch) {
+        const repo = input.gitUrl
+          ? { gitUrl: input.gitUrl, gitToken: input.gitToken, platform: input.platform }
```
alex-alecu: Does clone below also need platform?

jeanduplessis:

PR #1422, opened 2026-03-23
feat(admin): add Create Machine button for unallocated instances

evanjacobson:
## Summary
- Adds a "Create Machine" button in the admin KiloClaw instance detail for instances that have lost their `flyMachineId` (e.g. after DO state loss)
- The button calls `start()` with `skipRecovery: true` to bypass metadata recovery (which may fail due to cooldown/missing Fly app) and directly create a new Fly machine
- Looks up any version pin from Postgres so the new machine uses the pinned image tag instead of falling back to `latest`
- Calls `ensureApp` in `buildUserEnvVars` before `ensureEnvKey` so the App DO's state is recovered when its storage was wiped

## Loom — Kilo team only
https://www.loom.com/share/a2c0872adb3f477383b45215d5b4ff71

## Changes
- **DO `start()`/`_startInner()`**: Accept `opts?: { skipRecovery?: boolean; imageTag?: string }` — skip metadata recovery and/or override image tag
- **Platform `/start` route**: Extended schema with `skipRecovery` and validated `imageTag` (max 128 chars, Docker tag regex)
- **Internal client**: Thread `opts` through to platform route
- **Admin TRPC**: New `machineCreate` mutation — looks up version pin, calls `start` with `skipRecovery: true`
- **Frontend**: Show "Create Machine" button when `flyMachineId` is null, existing Start/Stop/Redeploy/Upgrade when it exists
- **`buildUserEnvVars`**: Call `ensureApp` on App DO before `ensureEnvKey` to handle missing App DO state

## Test plan
- [ ] Navigate to an admin instance detail with no `flyMachineId` — verify "Create Machine" button appears
- [ ] Click "Create Machine" — verify machine is created with the pinned image tag
- [ ] Navigate to an instance with a `flyMachineId` — verify normal Start/Stop/Redeploy/Upgrade buttons appear
- [ ] Verify destroyed instances don't show machine controls

### DIFF
--- a/kiloclaw/src/durable-objects/kiloclaw-instance/config.ts
+++ b/kiloclaw/src/durable-objects/kiloclaw-instance/config.ts
@@ -155,7 +155,10 @@ export async function buildUserEnvVars(
   }
 
   // Get the env encryption key from the App DO, creating it if needed.
+  // Call ensureApp first so the App DO has its flyAppName persisted — required
+  // by ensureEnvKey to set the Fly secret.
   const appStub = env.KILOCLAW_APP.get(env.KILOCLAW_APP.idFromName(state.userId));
+  await appStub.ensureApp(state.userId);
   const { key: envKey, secretsVersion } = await appStub.ensureEnvKey(state.userId);
 
   // Encrypt sensitive values and prefix their names with KILOCLAW_ENC_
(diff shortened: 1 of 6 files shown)


jeanduplessis (approved):

evanjacobson:
There were two tangled-up problems in this PR:
- local dev loses DO state
- user machines stop attempting to reconcile after 24hr of failed attempts

After speaking with Florian, there is a better way forward:
1. Determine a better way to handle worktree-based development (this causes the DO problems as each has its own .wrangler/ folder)
2. Add a button that flips the status from Stopped --> running/unknown + reset the last stopped time of the machine. this will cause the reconciliation loop to fire on the next alarm.

Closing this PR to do the latter change in a new one.

PR #1510, opened 2026-03-24
feat(kiloclaw): add runtime Fly region management via KV and admin UI

pandemicsyn:
## Summary

Make Fly region selection configurable at runtime without redeploying the worker. Regions are stored in Cloudflare KV (`fly-regions` key) and read at provision/start time, falling back to the `FLY_REGION` env var, then a hardcoded default.

Key changes:

- **Worker region resolution** — `resolveRegions()` reads from KV first (with try-catch fallback on KV failure), then env var, then default. Specific regions (e.g. `dfw,ord`) are shuffled for load distribution; meta-regions (`eu,us`) preserve declared order since Fly handles geographic distribution internally. Duplicates are allowed to bias the shuffle probability.
- **Worker API routes** — `GET/PUT /api/platform/regions` with Zod validation (enum allowlist of all 20 Fly regions, minimum 2 distinct regions required).
- **Admin UI** — New "Regions" tab in the KiloClaw admin dashboard with current config display, a text input for the comma-separated region list, a mixing warning (meta + specific), and a reference table of all Fly platform regions grouped by continent.
- **tRPC plumbing** — `KiloClawInternalClient` methods, admin tRPC router, and structured error surfacing so validation failures show specific messages in the UI toast.

The three existing call sites (`provision`, `ensureVolume`, `replaceStrandedVolume`) now use the KV-backed resolution.

## Verification

- [x] `pnpm typecheck` in `kiloclaw/` — clean
- [x] `pnpm test` in `kiloclaw/` — 986/986 passed (includes new tests for `isMetaRegion`, `prepareRegions`, `resolveRegions`, KV fallback)
- [x] `pnpm lint:eslint:fallback` in `kiloclaw/` — clean
- [x] `tsgo --noEmit` in root — clean
- [x] `eslint` on all changed Next.js files — clean
- [x] Manual: opened admin UI, confirmed Regions tab renders, tested invalid input (`us,kak`) — got structured validation error in toast

SCR-20260324-qrym

## Visual Changes

| Before | After |
| ------ | ----- |
| No Regions tab in KiloClaw admin | New "Regions" tab with current config, text input, mixing warning, and region reference table |

## Reviewer Notes

- The worker's `FLY_REGION` env var in `wrangler.jsonc` (`eu,us`) remains as the deploy-time fallback. KV takes precedence when a value is present.
- `resolveRegions` swallows KV read failures so a transient KV outage doesn't block provisioning — this is tested.
- The `PUT /api/platform/regions` route is protected by `internalApiMiddleware` (the existing `app.use('/api/platform/*', ...)` wildcard).
- Region constants are defined in two places (worker `regions.ts` and UI `region-constants.ts`) rather than a shared package — intentional for simplicity given the small surface.


src/routers/admin-kiloclaw-regions-router.ts
```diff
@@ -0,0 +1,50 @@
+import { adminProcedure, createTRPCRouter } from '@/lib/trpc/init';
+import { KiloClawInternalClient, KiloClawApiError } from '@/lib/kiloclaw/kiloclaw-internal-client';
+import { TRPCError } from '@trpc/server';
```
St0rmz1, line 3: **Nit: tRPC schema doesn't validate region codes**

This uses `z.array(z.string().min(1))` while the worker route validates against `z.enum(ALL_VALID_REGIONS)`. Invalid codes like `foobar` pass tRPC and only fail at the worker layer, which returns a less structured error. Consider either mirroring the enum allowlist here or dropping this schema to a simple passthrough — the current middle ground validates *shape* but not *content*, which is easy to mistake for real validation.

src/app/admin/components/KiloclawRegions/KiloclawRegionsPage.tsx
```diff
@@ -0,0 +1,212 @@
...
+        <CardContent className="flex flex-col gap-4">
+          <div className="flex flex-col gap-2">
+            <label htmlFor="region-input" className="text-sm font-medium">
+              Region list (comma-separated)
+            </label>
+            <div className="flex gap-2">
+              <Input
+                id="region-input"
+                value={inputValue}
+                onChange={e => {
```
St0rmz1, line 125: **Nit: missing `maxLength` on input**

This `<Input>` has no `maxLength`. Even for admin-only inputs, it's good practice to bound it — something like `maxLength={200}` would be more than enough for a comma-separated region list and keeps frontend/backend validation in sync.

St0rmz1 (approved):

PR #1523, opened 2026-03-25
feat(api): serve notifications endpoint from global app

markijbema:
## Summary

- Add a `beforeFiles` rewrite in `next.config.mjs` to proxy `/api/users/notifications` to `https://global-api.kilo.ai/api/users/notifications` on non-global deployments, following the same pattern used for FIM and marketplace routing.
- On the global app (`GLOBAL_KILO_BACKEND=true`), the rewrite is skipped and the request hits the local route handler directly.
- The route file remains at its original location `src/app/api/users/notifications/route.ts` — no file moves, no endpoint path changes.

## Verification

- [x] Verified the rewrite entry follows the exact same pattern as the existing FIM and marketplace entries in `next.config.mjs`
- [x] Confirmed the route file stays at `src/app/api/users/notifications/route.ts` with no content changes
- [x] Confirmed the endpoint path `/api/users/notifications` is unchanged for existing clients

## Visual Changes

N/A

## Reviewer Notes

- The notifications endpoint is consumed by the Kilo VSCode extension and CLI, not by the web app itself. Serving it from the global app ensures lower latency for these external clients regardless of region.
- Unlike the original version of this PR, the route file is **not** moved — only a `beforeFiles` rewrite is added. This avoids breaking existing clients that target `/api/users/notifications`.


markijbema:
ok, this might be tricky, if any of these functions are used and then lead to an update, this might make some other page work worse. to prevent this i passed in db from fim so we only use read if we now that's safe

jeanduplessis:

PR #1525, opened 2026-03-25
feat(kiloclaw): send instance-ready email on first low-load checkin

RSO:
## Summary

- Send a one-time transactional email notifying users when their KiloClaw instance is ready (detected when `loadAvg5m < 0.1` during controller checkin)
- Introduces a new communication direction: KiloClaw CF Worker → Next.js internal API (`POST /api/internal/kiloclaw/instance-ready`) for real-time email delivery
- Dual idempotency: DO-level `instanceReadyEmailSent` flag as primary guard, `kiloclaw_email_log` Postgres table as safety net (insert-before-send with rollback on failure)
- The email log is cleared on instance destruction and Stripe reactivation, allowing re-provision to trigger the email again

## Verification

- [x] `pnpm typecheck` — passes across the entire monorepo
- [x] `pnpm test` — all 973 KiloClaw tests pass (7 new tests added)
- [x] `pnpm lint` — zero warnings, zero errors
- [x] `pnpm format` — all files formatted
- [x] Manually called the `instance-ready` endpoint to make sure the API->mailgun->gmail hookup works and the email looks good
- [x] Did a full integration test locally to verify everything works as expected

## Visual Changes

CleanShot 2026-03-25 at 12 15 33@2x

## Reviewer Notes

- This introduces the first worker → Next.js HTTP call in the KiloClaw architecture (existing communication is one-directional: Next.js → Worker). The call uses `X-Internal-Secret` header with `INTERNAL_API_SECRET`, matching the pattern used by other CF Workers calling Next.js internal endpoints (e.g., code-review, auto-triage)
- The `NEXT_INTERNAL_API_URL` env var needs to be set in production (defaulted to `https://kilo.ai` in `wrangler.jsonc`)
- The `INTERNAL_API_SECRET` is already shared between the worker and Next.js — no new secret needed
- The Customer.io template ID `31` for `clawInstanceReady` needs to be created in Customer.io if still using that provider. Mailgun will use the local HTML template automatically
- Edge case: if the worker marks the DO flag but the Next.js call fails (e.g., Next.js is down), the email will NOT be retried. This is acceptable since the user will see their instance is ready via the UI

### DIFF
--- a/kiloclaw/controller/src/checkin.ts
+++ b/kiloclaw/controller/src/checkin.ts
@@ -77,6 +77,7 @@ export async function readNetStats(): Promise<NetStats> {
 
 export function startCheckin(deps: CheckinDeps): () => void {
   const checkinUrl = deps.getCheckinUrl();
+  console.info(checkinUrl);
   if (!checkinUrl) {
     return () => {};
   }
--- a/kiloclaw/controller/src/checkin.ts
+++ b/kiloclaw/controller/src/checkin.ts
@@ -77,7 +77,6 @@ export async function readNetStats(): Promise<NetStats> {
 
 export function startCheckin(deps: CheckinDeps): () => void {
   const checkinUrl = deps.getCheckinUrl();
-  console.info(checkinUrl);
   if (!checkinUrl) {
     return () => {};
   }
--- a/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
+++ b/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
@@ -376,6 +376,7 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
       this.s.pendingDestroyMachineId = null;
       this.s.pendingDestroyVolumeId = null;
       this.s.pendingPostgresMarkOnFinalize = false;
+      this.s.instanceReadyEmailSent = false;
     }
     this.s.loaded = true;
 
--- a/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
+++ b/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
@@ -1298,8 +1298,15 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
     if (this.s.instanceReadyEmailSent) {
       return { shouldNotify: false, userId: this.s.userId };
     }
+
     this.s.instanceReadyEmailSent = true;
     await this.persist({ instanceReadyEmailSent: true });
+
+    // If the instance was provisioned more than 6 hours ago, don't send the email
+    if (this.s.provisionedAt && this.s.provisionedAt < Date.now() - 1000 * 60 * 60 * 6) {
+      return { shouldNotify: false, userId: this.s.userId };
+    }
+
     return { shouldNotify: true, userId: this.s.userId };
   }
 
--- a/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
+++ b/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
@@ -1288,6 +1288,21 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
     };
   }
 
+  /**
+   * Atomically check-and-set the instance ready flag. Returns shouldNotify: true
+   * on the first call per provision lifecycle, false on all subsequent calls.
+   * Used by the controller checkin handler to trigger a one-time "instance ready" email.
+   */
+  async tryMarkInstanceReady(): Promise<{ shouldNotify: boolean; userId: string | null }> {
+    await this.loadState();
+    if (this.s.instanceReadyEmailSent) {
+      return { shouldNotify: false, userId: this.s.userId };
+    }
+    this.s.instanceReadyEmailSent = true;
+    await this.persist({ instanceReadyEmailSent: true });
+    return { shouldNotify: true, userId: this.s.userId };
+  }
+
   async listVolumeSnapshots(): Promise<FlyVolumeSnapshot[]> {
     await this.loadState();
     if (!this.s.flyVolumeId) return [];
--- a/kiloclaw/src/durable-objects/kiloclaw-instance/state.ts
+++ b/kiloclaw/src/durable-objects/kiloclaw-instance/state.ts
@@ -72,6 +72,7 @@ export async function loadState(ctx: DurableObjectState, s: InstanceMutableState
     s.gmailPushOidcEmail = d.gmailPushOidcEmail;
     s.execSecurity = d.execSecurity;
     s.execAsk = d.execAsk;
+    s.instanceReadyEmailSent = d.instanceReadyEmailSent;
   } else {
     const hasAnyData = entries.size > 0;
     if (hasAnyData) {
@@ -134,6 +135,7 @@ export function resetMutableState(s: InstanceMutableState): void {
   s.gmailPushOidcEmail = null;
   s.execSecurity = null;
   s.execAsk = null;
+  s.instanceReadyEmailSent = false;
   s.lastLiveCheckAt = null;
   s.restartingAt = null;
   s.loaded = false;
@@ -190,6 +192,7 @@ export function createMutableState(): InstanceMutableState {
     gmailPushOidcEmail: null,
     execSecurity: null,
     execAsk: null,
+    instanceReadyEmailSent: false,
     lastLiveCheckAt: null,
   };
 }
--- a/kiloclaw/src/durable-objects/kiloclaw-instance/state.ts
+++ b/kiloclaw/src/durable-objects/kiloclaw-instance/state.ts
@@ -72,15 +72,7 @@ export async function loadState(ctx: DurableObjectState, s: InstanceMutableState
     s.gmailPushOidcEmail = d.gmailPushOidcEmail;
     s.execSecurity = d.execSecurity;
     s.execAsk = d.execAsk;
-    // Migration: pre-existing instances (provisioned before this field existed) don't
-    // have instanceReadyEmailSent in storage. Without this guard, the Zod default
-    // (false) would cause them to send a spurious "instance ready" email on the
-    // first low-load checkin after deploy. Treat missing-in-storage as already-sent.
-    s.instanceReadyEmailSent = entries.has('instanceReadyEmailSent')
-      ? d.instanceReadyEmailSent
-      : d.userId
-        ? true
-        : false;
+    s.instanceReadyEmailSent = d.instanceReadyEmailSent;
   } else {
     const hasAnyData = entries.size > 0;
     if (hasAnyData) {
--- a/kiloclaw/src/durable-objects/kiloclaw-instance/state.ts
+++ b/kiloclaw/src/durable-objects/kiloclaw-instance/state.ts
@@ -72,7 +72,9 @@ export async function loadState(ctx: DurableObjectState, s: InstanceMutableState
     s.gmailPushOidcEmail = d.gmailPushOidcEmail;
     s.execSecurity = d.execSecurity;
     s.execAsk = d.execAsk;
-    s.instanceReadyEmailSent = d.instanceReadyEmailSent;
+    // Legacy instances pre-dating this field treat absence as already-sent
+    // to avoid spurious emails after deploy.
+    s.instanceReadyEmailSent = 'instanceReadyEmailSent' in raw ? d.instanceReadyEmailSent : true;
   } else {
     const hasAnyData = entries.size > 0;
     if (hasAnyData) {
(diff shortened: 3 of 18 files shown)


kiloclaw/src/routes/controller.ts
```diff
@@ -23,6 +25,28 @@ const CheckinSchema = z.object({
...
+  internalSecret: string,
+  userId: string
+): Promise<void> {
+  const res = await fetch(`${nextApiUrl}/api/internal/kiloclaw/instance-ready`, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      'X-Internal-Secret': internalSecret,
+    },
+    body: JSON.stringify({ userId }),
```
RSO: Let's use the kiloclaw instanceId or something more unique to prepare for a world where a user can have multiple instances.

kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
```diff
@@ -1284,6 +1236,21 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
...
+   * Atomically check-and-set the instance ready flag. Returns shouldNotify: true
+   * on the first call per provision lifecycle, false on all subsequent calls.
+   * Used by the controller checkin handler to trigger a one-time "instance ready" email.
+   */
+  async tryMarkInstanceReady(): Promise<{ shouldNotify: boolean; userId: string | null }> {
+    await this.loadState();
+    if (this.s.instanceReadyEmailSent) {
+      return { shouldNotify: false, userId: this.s.userId };
+    }
+    this.s.instanceReadyEmailSent = true;
```
RSO, line 1303: We need to make sure that this path doesn't send emails to instances that were already created before this change is merged.

kiloclaw/wrangler.jsonc
```diff
@@ -82,6 +82,8 @@
     "OPENCLAW_ALLOWED_ORIGINS": "https://claw.kilosessions.ai,https://kilo.ai,https://www.kilo.ai",
     "KILOCLAW_CHECKIN_URL": "https://claw.kilosessions.ai/api/controller/checkin",
     "REQUIRE_PROXY_TOKEN": "true",
+    // Next.js app URL for outbound notifications (instance-ready emails, etc.)
+    "NEXT_INTERNAL_API_URL": "https://kilo.ai",
```
RSO: I'm pretty sure that we already have a way to call the API for the chat completions endpoint (that's why I need ngrok/cloudflare-tunnel in dev). So I don't think we need this env variable.

src/app/api/internal/kiloclaw/instance-ready/route.ts
```diff
@@ -0,0 +1,83 @@
...
+
+import type { NextRequest } from 'next/server';
+import { NextResponse } from 'next/server';
+import { z } from 'zod';
+import { and, eq } from 'drizzle-orm';
+import { INTERNAL_API_SECRET, NEXTAUTH_URL } from '@/lib/config.server';
+import { send as sendEmail } from '@/lib/email';
+import { findUserById } from '@/lib/user';
+import { db } from '@/lib/drizzle';
+import { kiloclaw_email_log } from '@kilocode/db/schema';
```
RSO, line 20: Are we sure that `kiloclaw_email_log` can be used for this new purpose?

RSO: Yes — `kiloclaw_email_log` was designed for exactly this pattern: one-time-per-user transactional emails with idempotent insert-before-send. The billing cron already uses it for 9 different email types with the same `trySendEmail()` mechanism.

The table structure supports this naturally:
- **Unique index on `(user_id, email_type)`** prevents duplicate sends
- **`ON DELETE CASCADE` from `kilocode_users`** handles GDPR cleanup
- **Email type keys are now per-instance** (`claw_instance_ready:{sandboxId}`) so it also works in a multi-instance future

The lifecycle cleanup follows the existing patterns:
- Instance destruction (sweep 3) clears the entry via `LIKE 'claw_instance_ready:%'`
- Stripe reactivation clears it alongside the other resettable email types

The `kiloclaw_email_log` row is the *secondary* idempotency guard (safety net). The *primary* guard is the `instanceReadyEmailSent` flag in the DO — it prevents the notification HTTP call from being made at all on subsequent checkins. The Postgres row only matters if the DO flag is somehow lost (e.g., DO storage wipe).

kiloclaw/src/routes/controller.ts
RSO: Fixed in `55aaa40a`. The `sandboxId` is now threaded through the notification payload and used as part of the email_type key (`claw_instance_ready:{sandboxId}`) for per-instance idempotency.

RSO: Fixed in 55aaa40a — the notification payload now includes `sandboxId` alongside `userId`, and the email log key is scoped per-instance (`claw_instance_ready:{sandboxId}`).

kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
RSO: Fixed in 06ac4df9 — added a migration guard in `loadState()`: when a DO has a `userId` (is provisioned) but `instanceReadyEmailSent` is missing from storage (pre-existing instance), it defaults to `true` to suppress the email. Only newly provisioned instances (which explicitly write `instanceReadyEmailSent: false` to storage) will trigger the notification.

kiloclaw/wrangler.jsonc
RSO: Fixed in 0b48b48f — removed `NEXT_INTERNAL_API_URL` entirely. The worker now derives the Next.js origin from `KILOCODE_API_BASE_URL` (extracting just the origin via `new URL().origin`), falling back to `https://kilo.ai` in production where it's not set.

kiloclaw/controller/src/checkin.ts
```diff
@@ -77,6 +77,7 @@ export async function readNetStats(): Promise<NetStats> {
 
 export function startCheckin(deps: CheckinDeps): () => void {
   const checkinUrl = deps.getCheckinUrl();
+  console.info(checkinUrl);
```
RSO:
```suggestion
```

kiloclaw/src/durable-objects/kiloclaw-instance/state.ts
```diff
@@ -72,6 +72,15 @@ export async function loadState(ctx: DurableObjectState, s: InstanceMutableState
...
     s.execAsk = d.execAsk;
+    // Migration: pre-existing instances (provisioned before this field existed) don't
+    // have instanceReadyEmailSent in storage. Without this guard, the Zod default
+    // (false) would cause them to send a spurious "instance ready" email on the
+    // first low-load checkin after deploy. Treat missing-in-storage as already-sent.
+    s.instanceReadyEmailSent = entries.has('instanceReadyEmailSent')
+      ? d.instanceReadyEmailSent
+      : d.userId
+        ? true
+        : false;
```
RSO: I'm mostly worried about this piece of logic. I really don't want to send this email to users who set up their instance last week.

jeanduplessis (approved):

kiloclaw/src/durable-objects/kiloclaw-instance/state.ts
pandemicsyn: If you explicitly set instanceReadyEmailSent=false on _new_ provisioned instances, then `false` means it's a new instance , `null` means its an old instance, and `true` means ya already sent it.

Heres where we populate `provisionedAt`

https://github.com/Kilo-Org/cloud/blob/603f9a5ea737d374e1a793433b3041ae83765ac9/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts#L372

You could explicitly set instanceReadyEmailSent = false near there just to be safe? 

The other thing you can do is also just pull in `s.provisionedAt` itself as a guard. e.g. if `instancanceReadyEmailSent` is false and `provisionedAt` is less than 12 hours ago then send email.

src/app/api/internal/kiloclaw/instance-ready/route.ts
```diff
@@ -0,0 +1,88 @@
...
+
+const BodySchema = z.object({
+  userId: z.string().min(1),
+  sandboxId: z.string().min(1),
+});
+
+/** Per-instance email type key. Includes sandboxId to support future multi-instance. */
+function emailTypeKey(sandboxId: string): string {
+  return `claw_instance_ready:${sandboxId}`;
```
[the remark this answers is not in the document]
pandemicsyn: >WARNING: Manual destroy/re-provision reuses the same idempotency key

>sandboxId is deterministic for a user today, so a manual destroy() followed by re-provisioning produces the same claw_instance_ready:{sandboxId} key. This route will keep returning already_sent because the manual destroy path never clears kiloclaw_email_log; only the billing cron and Stripe reactivation do.

@RSO if im following along,  i don't *think* this comment from the review bot is true, over in src/lib/kiloclaw/stripe-handlers.ts we're deleting the full row right ?

kiloclaw/src/durable-objects/kiloclaw-instance/state.ts
```diff
@@ -134,6 +143,7 @@ export function resetMutableState(s: InstanceMutableState): void {
   s.gmailPushOidcEmail = null;
   s.execSecurity = null;
   s.execAsk = null;
+  s.instanceReadyEmailSent = false;
```
pandemicsyn, line 140: 👍 this covers you after a destroy.

src/app/api/internal/kiloclaw/instance-ready/route.ts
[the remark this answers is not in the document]
RSO: Yeah that sounds like a lie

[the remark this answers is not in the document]
RSO: Actually, I think the bot is right, because we only removed the complete row in `stripe-handlers.ts` but not in the manual "destroy instance" flow.

kiloclaw/src/durable-objects/kiloclaw-instance/state.ts
```diff
@@ -72,6 +72,7 @@ export async function loadState(ctx: DurableObjectState, s: InstanceMutableState
     s.gmailPushOidcEmail = d.gmailPushOidcEmail;
     s.execSecurity = d.execSecurity;
     s.execAsk = d.execAsk;
+    s.instanceReadyEmailSent = d.instanceReadyEmailSent;
```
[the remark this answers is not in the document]
RSO: But we now also check for `provisionAt` so that's fine.

src/app/api/internal/kiloclaw/instance-ready/route.ts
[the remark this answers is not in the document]
RSO: Fixed in 4e4dc2da — the email-log cleanup in the user destroy handler (`kiloclaw-router.ts`) is now wrapped in its own try/catch outside the rollback path. If `client.destroy()` succeeds but a subsequent DB cleanup fails, we log the error and move on instead of restoring the instance.

src/routers/kiloclaw-router.ts
```diff
@@ -485,6 +486,33 @@ export const kiloclawRouter = createTRPCRouter({
         .update(kiloclaw_subscriptions)
         .set(clearFields)
         .where(eq(kiloclaw_subscriptions.user_id, ctx.user.id));
+
+      // Clear lifecycle emails so they can fire again if the user re-provisions.
```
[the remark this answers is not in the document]
RSO: Fixed in 4e4dc2da — the `try` block now only wraps `client.destroy()` with the rollback logic. All post-destroy DB cleanup (subscription fields + email log deletions) is in a separate try/catch that logs errors without throwing or triggering `restoreDestroyedInstance()`.

src/routers/admin-kiloclaw-instances-router.ts
```diff
@@ -645,6 +663,32 @@ export const adminKiloclawInstancesRouter = createTRPCRouter({
       throw error;
     }
 
+    // Clear lifecycle emails so they can fire again if the user re-provisions.
```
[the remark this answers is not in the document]
RSO: Fixed in 4e4dc2da — the admin destroy handler's email-log cleanup is now wrapped in a try/catch that logs on failure but still returns `{ success: true }`, since the destroy already succeeded at that point.

src/routers/admin-kiloclaw-instances-router.ts
```diff
@@ -645,6 +663,38 @@ export const adminKiloclawInstancesRouter = createTRPCRouter({
       throw error;
     }
 
+    // Post-destroy cleanup: best-effort DB tidying that must not report
+    // failure after a successful destroy.
+    try {
+      // Clear lifecycle emails so they can fire again if the user re-provisions.
```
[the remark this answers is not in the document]
RSO: That sounds like a problem unrelated to the work done in this PR.

PR #1532, opened 2026-03-25
style(usage): compact metric cards for 7-column layout

jeanduplessis:
### Problem

The org usage page displays 7 metric cards in a single row at `xl` breakpoint. The cards had excessive spacing and padding, and the longest label ("Avg Cost per Request") was truncated by overflow clipping.

### Solution

- Reduced grid gap from `gap-4` to `gap-2` so cards sit closer together.
- Reduced card header padding (`px-4 pt-4 pb-2` → `px-3 pt-3 pb-1`) and content padding (`px-4` → `px-3`).
- Reduced value font size from `text-2xl` to `text-xl` for a more compact appearance.
- Shortened the "Avg Cost per Request" label to "Avg Cost/Req" to prevent truncation.

### Why this approach

Wrapping or ellipsis-with-tooltip were considered for the label, but abbreviation is simpler and keeps all cards visually consistent at a single line height. The CSS-only spacing changes avoid any structural or behavioral changes to the component.

## Verification

- [x] `pnpm lint` — passed (via pre-push hook)
- [x] `pnpm format:check` — passed (via pre-push hook)
- [x] Manual visual verification via agent-browser on `/organizations/[id]/usage-details`

## Visual Changes

| Before | After |
| ------ | ----- |
| before: usage-metric-cards | after: usage-metric-cards |

## Reviewer Notes

- The change is purely visual — no logic, data, or API changes.
- The abbreviated label "Avg Cost/Req" matches the existing abbreviated style used elsewhere in the dashboard.

### DIFF
--- a/src/components/organizations/usage-details/MetricSelector.tsx
+++ b/src/components/organizations/usage-details/MetricSelector.tsx
@@ -22,7 +22,7 @@ type Props = {
 
 export function MetricSelector({ metrics, selectedMetric, onSelectedMetricChange }: Props) {
   return (
-    <div className="grid grid-cols-1 gap-4 sm:grid-cols-3 lg:grid-cols-4 xl:grid-cols-7">
+    <div className="grid grid-cols-1 gap-2 sm:grid-cols-3 lg:grid-cols-4 xl:grid-cols-7">
       {metrics.map(metric => {
         const isSelected = metric.key === selectedMetric;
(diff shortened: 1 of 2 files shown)


jrf0110 (approved):

jrf0110:
Looking wayyy better! I'd still make the each of the metric selector cards a little shorter to be more golden-rectangle-y

PR #1550, opened 2026-03-25
(kiloclaw-admin): Add Destroy Machine functionality to the KiloClaw admin dash

evanjacobson:
## Summary

Adds a "Destroy Machine" button to the KiloClaw admin instance detail page that directly calls the Fly Machines API to force-destroy a machine — equivalent to `fly machine destroy <id> -a <app> --force`.

**Why this exists:** The normal instance destroy flow goes through the DO's teardown pipeline (machine → volume → cleanup). This button is an admin escape hatch for when you need to delete the Fly machine directly without triggering the full teardown.

**What it does NOT do:** This does not tear down the KiloClaw instance, delete the Fly volume, or mark the instance as destroyed in the database. Only the Fly machine is destroyed.

### Implementation

- **Platform route** (`POST /api/platform/destroy-fly-machine`): Calls `DELETE /v1/apps/{app}/machines/{id}?force=true` on the Fly Machines API using the worker's `FLY_API_TOKEN`. After a successful destroy, triggers `forceRetryRecovery` on the DO to schedule an immediate reconcile alarm — the DO will discover the machine is destroyed and clear its `flyMachineId` state.
- **tRPC mutation** (`destroyFlyMachine`): Admin-only mutation that proxies to the platform route and writes an audit log (`kiloclaw.machine.destroy_fly`).
- **UI**: Orange "Destroy Machine" button in Machine Controls with a confirmation dialog showing the machine ID and app name. After success, the button shows a green "Machine Destroyed" state and is disabled. An error toast is shown if `flyAppName` is missing from the DO state.
- **Observability**: `writeEvent` calls for success/failure (visible in the DO & Reconcile Events table on the admin page), plus `createKiloClawAdminAuditLog` for the audit trail.

### Reconciler fix: handle `"destroyed"` Fly state

`syncStatusWithFly` previously had no branch for the `"destroyed"` Fly machine state. When a machine is force-destroyed via the Fly API, Fly returns the machine with `state: "destroyed"` before it eventually starts returning 404. Without this fix, the reconciler would no-op on the `"destroyed"` state, leaving the stale `flyMachineId` in DO storage until a later alarm cycle when Fly finally returns 404 and `handleMachineGone` fires. The new `"destroyed"` branch mirrors `handleMachineGone` — clears `flyMachineId`, sets status to `"stopped"`, and persists immediately.

### New audit action

`kiloclaw.machine.destroy_fly` added to `KiloClawAdminAuditAction` enum.

## Verification

- [x] TypeScript compiles cleanly (`npx tsc --noEmit`)
- [x] Tested locally: destroy succeeds, Fly machine removed, DO reconciles after alarm fires
- [x] 4 rounds of adversarial code review (code reviewer, silent failure hunter, security auditor) with verification agents — converged to zero new actionable findings
- [x] E2E manual testing

## Loom — Kilo team only
https://www.loom.com/share/7e898ec288da4e9a83d2554780622dd9

## Reviewer Notes

- The `forceRetryRecovery` call after destroy is fire-and-forget (caught and warned). If it fails, the DO will still reconcile on its next regular alarm. The response returns `ok: true` regardless — this was a deliberate scoping decision to avoid threading a `reconcileTriggered` flag through three layers.
- Until the reconcile task following machine destruction finishes (should be fairly quick in the deployed env), the DO will log errors as it has lost connection with the controller. The noise should be relatively minimal and, as the reconcile alarm is set to _now_, it should be resolved quickly.
- `isFlyMachineDestroyed` comes from `useMutation`'s `isSuccess` which resets on component re-mount. A second click after re-mount would get a Fly 404 error toast — not harmful, just redundant.

### DIFF
--- a/kiloclaw/src/routes/platform.ts
+++ b/kiloclaw/src/routes/platform.ts
@@ -1363,8 +1363,15 @@ platform.put('/regions', async c => {
 // This bypasses the DO's normal destroy flow — use for admin cleanup.
 const DestroyFlyMachineSchema = z.object({
   userId: z.string().min(1),
-  appName: z.string().min(1).max(63).regex(/^[a-z0-9]([a-z0-9-]*[a-z0-9])?$/, 'Invalid Fly app name'),
-  machineId: z.string().min(1).regex(/^[a-z0-9]+$/, 'Invalid Fly machine ID'),
+  appName: z
+    .string()
+    .min(1)
+    .max(63)
+    .regex(/^[a-z0-9]([a-z0-9-]*[a-z0-9])?$/, 'Invalid Fly app name'),
+  machineId: z
+    .string()
+    .min(1)
+    .regex(/^[a-z0-9]+$/, 'Invalid Fly machine ID'),
 });
 
 platform.post('/destroy-fly-machine', async c => {
@@ -1390,10 +1397,7 @@ platform.post('/destroy-fly-machine', async c => {
         `[platform] destroy-fly-machine failed (${resp.status}) app=${appName} machine=${machineId}:`,
         body
       );
-      return jsonError(
-        `Fly API error (${resp.status}): ${body}`,
-        resp.status
-      );
+      return jsonError(`Fly API error (${resp.status}): ${body}`, resp.status);
     }
 
     console.log(`[platform] destroy-fly-machine ok: app=${appName} machine=${machineId}`);
--- a/kiloclaw/src/routes/platform.ts
+++ b/kiloclaw/src/routes/platform.ts
@@ -1375,7 +1375,7 @@ platform.post('/destroy-fly-machine', async c => {
   const startedAt = performance.now();
   const apiToken = c.env.FLY_API_TOKEN;
   if (!apiToken) {
-    return jsonError('FLY_API_TOKEN is not configured', 500);
+    return c.json({ error: 'FLY_API_TOKEN is not configured' }, 503);
   }
 
   const url = `https://api.machines.dev/v1/apps/${appName}/machines/${machineId}?force=true`;
@@ -1399,7 +1399,7 @@ platform.post('/destroy-fly-machine', async c => {
         error: `Fly API error (${resp.status})`,
         durationMs: performance.now() - startedAt,
       });
-      return jsonError(`Fly API error (${resp.status}): ${body}`, resp.status);
+      return jsonError(`Fly API error (${resp.status}): ${body}`, resp.status >= 500 ? 502 : resp.status);
     }
 
     console.log(`[platform] destroy-fly-machine ok: app=${appName} machine=${machineId}`);
@@ -1427,8 +1427,7 @@ platform.post('/destroy-fly-machine', async c => {
     });
     return c.json({ ok: true });
   } catch (err) {
-    const message = err instanceof Error ? err.message : 'Unknown error';
-    console.error(`[platform] destroy-fly-machine error app=${appName} machine=${machineId}:`, err);
+    const { message, status } = sanitizeError(err, 'destroy-fly-machine');
     writeEvent(c.env, {
       event: 'instance.destroy_fly_machine_failed',
       delivery: 'http',
@@ -1437,7 +1436,7 @@ platform.post('/destroy-fly-machine', async c => {
       error: message,
       durationMs: performance.now() - startedAt,
     });
-    return jsonError(`Failed to destroy machine: ${message}`, 500);
+    return jsonError(message, status);
   }
 });
 
--- a/kiloclaw/src/routes/platform.ts
+++ b/kiloclaw/src/routes/platform.ts
@@ -1359,8 +1359,9 @@ platform.put('/regions', async c => {
 });
 
 // POST /api/platform/destroy-fly-machine
-// Directly destroys a Fly machine via the Machines API (force=true).
-// This bypasses the DO's normal destroy flow — use for admin cleanup.
+// This is for admin cleanup only.
+// It directly destroys a Fly machine via the Machines API (force=true).
+// It does not destroy the Fly app or volume.
 const DestroyFlyMachineSchema = z.object({
   userId: z.string().min(1),
   appName: z
@@ -1384,7 +1385,7 @@ platform.post('/destroy-fly-machine', async c => {
     return c.json({ error: 'FLY_API_TOKEN is not configured' }, 503);
   }
 
-  const url = `https://api.machines.dev/v1/apps/${encodeURIComponent(appName)}/machines/${encodeURIComponent(machineId)}?force=true`;
+  const url = `https://api.machines.dev/v1/apps/${appName}/machines/${machineId}?force=true`;
   try {
     const resp = await fetch(url, {
       method: 'DELETE',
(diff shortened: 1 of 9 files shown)


kiloclaw/src/routes/platform.ts
```diff
@@ -1358,4 +1358,69 @@ platform.put('/regions', async c => {
...
+  const result = await parseBody(c, DestroyFlyMachineSchema);
+  if ('error' in result) return result.error;
+
+  const { userId, appName, machineId } = result.data;
+  const apiToken = c.env.FLY_API_TOKEN;
+  if (!apiToken) {
+    return c.json({ error: 'FLY_API_TOKEN is not configured' }, 503);
+  }
+
+  const url = `https://api.machines.dev/v1/apps/${encodeURIComponent(appName)}/machines/${encodeURIComponent(machineId)}?force=true`;
```
evanjacobson: `force=true` is necessary to ensure a machine is destroyed, even if hung/unresponsive. This was Fly's guidance to us.

kiloclaw/src/routes/platform.ts
```diff
@@ -1358,4 +1358,69 @@ platform.put('/regions', async c => {
...
+    .min(1)
+    .max(63)
+    .regex(/^[a-z0-9]([a-z0-9-]*[a-z0-9])?$/, 'Invalid Fly app name'),
+  machineId: z
+    .string()
+    .min(1)
+    .regex(/^[a-z0-9]+$/, 'Invalid Fly machine ID'),
+});
+
+platform.post('/destroy-fly-machine', async c => {
```
evanjacobson, line 1378: Ownership validation (appName/machineId matching the DO's actual state) is performed by the admin tRPC router before calling this endpoint. Platform routes are behind `internalApiMiddleware` and trust the caller, consistent with other destructive platform routes (`/stop`, `/destroy`).

src/routers/admin-kiloclaw-instances-router.ts
```diff
@@ -630,6 +630,67 @@ export const adminKiloclawInstancesRouter = createTRPCRouter({
       }
     }),
 
+  destroyFlyMachine: adminProcedure
```
evanjacobson, line 633: Unlike the existing `destroy` procedure (which marks the DB row, has rollback logic, and cleans up email logs), this intentionally doesn't touch the DB. It only destroys the Fly machine — the DB instance row, DO state, and Fly volume all remain intact. The DO will detect the missing machine via reconciliation and transition to `stopped`, allowing the user to start a fresh machine on the same volume.

kiloclaw/src/routes/platform.ts
```diff
@@ -1358,4 +1358,69 @@ platform.put('/regions', async c => {
...
+      console.error(
+        `[platform] destroy-fly-machine failed (${resp.status}) app=${appName} machine=${machineId}:`,
+        body
+      );
+      return jsonError(`Fly API error (${resp.status}): ${body}`, resp.status);
+    }
+
+    console.log(`[platform] destroy-fly-machine ok: app=${appName} machine=${machineId}`);
+
+    // Trigger immediate reconcile so the DO discovers the machine is gone.
```
evanjacobson, line 1406: The reconciliation outcome: `reconcileMachine()` → `getMachine()` → 404 → `handleMachineGone()` → clears `flyMachineId` + sets status to `stopped`. No automatic machine recreation happens — the user must explicitly start again. If `forceRetryRecovery` fails (e.g., DO is in `destroying` state → 409), the regular alarm cycle (5 min) handles cleanup.

kiloclaw/src/routes/platform.ts
```diff
@@ -1358,4 +1358,69 @@ platform.put('/regions', async c => {
   return c.json({ ok: true, regions: result.data.regions, raw });
 });
 
+// POST /api/platform/destroy-fly-machine
+// Directly destroys a Fly machine via the Machines API (force=true).
+// This bypasses the DO's normal destroy flow — use for admin cleanup.
+const DestroyFlyMachineSchema = z.object({
```
evanjacobson, line 1365: This schema is duplicated identically in `admin-kiloclaw-instances-router.ts:635-646`. The duplication is intentional — platform routes (CF Worker) and the admin router (Next.js) are separate deployment targets that don't share Zod schemas, consistent with how `UserIdRequestSchema` and other schemas are already duplicated across the boundary.

src/app/admin/components/KiloclawInstances/KiloclawInstanceDetail.tsx
```diff
@@ -1242,6 +1243,30 @@ export function KiloclawInstanceDetail({ instanceId }: { instanceId: string }) {
...
+    })
+  );
+
+  // Reset the destroyed success state when the machine ID changes (e.g. new machine created)
+  const flyMachineId = data?.workerStatus?.flyMachineId;
+  useEffect(() => {
+    resetDestroyFlyMachine();
+  }, [flyMachineId, resetDestroyFlyMachine]);
+
   const { mutateAsync: forceRetryRecovery, isPending: isRetryingRecovery } = useMutation(
```
evanjacobson, line 1270: After a machine is destroyed and the user starts a new one, `flyMachineId` changes. Without this reset, the button stays in the green "Machine Destroyed" success state even though a new machine now exists. `resetDestroyFlyMachine` clears `isSuccess` so the button returns to its normal state for the new machine.

kiloclaw/src/routes/platform.ts
```diff
@@ -1358,4 +1358,69 @@ platform.put('/regions', async c => {
...
+      headers: { Authorization: `Bearer ${apiToken}` },
+    });
+
+    if (!resp.ok) {
+      const body = await resp.text();
+      console.error(
+        `[platform] destroy-fly-machine failed (${resp.status}) app=${appName} machine=${machineId}:`,
+        body
+      );
+      return jsonError(`Fly API error (${resp.status}): ${body}`, resp.status);
```
evanjacobson, line 1401: This is admin-facing only — the raw error passthrough is intentional here.

packages/db/src/schema-types.ts
```diff
@@ -162,6 +162,7 @@ export const KiloClawAdminAuditAction = z.enum([
   'kiloclaw.gateway.restart',
   'kiloclaw.config.restore',
   'kiloclaw.doctor.run',
+  'kiloclaw.machine.destroy_fly',
```
evanjacobson, line 165: added to KiloClawAdminAuditAction

src/routers/admin-kiloclaw-instances-router.test.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    expect(logs).toHaveLength(1);
+    expect(logs[0].target_user_id).toBe(testUserId);
+    expect(logs[0].actor_email).toBe(adminUser.google_user_email);
+    expect(logs[0].message).toContain(testAppName);
+    expect(logs[0].message).toContain(testMachineId);
+    expect(logs[0].metadata).toEqual({ appName: testAppName, machineId: testMachineId });
+  });
+
+  it('wraps generic errors as INTERNAL_SERVER_ERROR', async () => {
```
evanjacobson: explain why

src/routers/admin-kiloclaw-instances-router.test.ts
```diff
@@ -0,0 +1,241 @@
...
+    await expect(
+      caller.admin.kiloclawInstances.destroyFlyMachine({
+        userId: testUserId,
+        appName: testAppName,
+        machineId: testMachineId,
+      })
+    ).rejects.toThrow('Failed to destroy Fly machine: Fly API timeout');
+  });
+
+  it('maps KiloClawApiError 404 to NOT_FOUND', async () => {
```
evanjacobson: explain why

catrielmuller (approved):

PR #1551, opened 2026-03-25
(ui): Create a shared banner component & implement on the profile page KiloClaw CTA

evanjacobson:
## Summary

- Fix mobile layout for KiloClaw CTA banner on /profile (was cramped single-row, now stacks vertically on mobile)
- Extract reusable `Banner` component to `src/components/shared/Banner.tsx`
- Standardize styling to match /claw dashboard banners (BillingBanner, config service nudge)

## Kilo Team — Loom
https://www.loom.com/share/f5d9c7ce6887427987d51d1a279df662

Not shown in the Loom, but the button and icon can also be omitted.
image

## Banner API

| Prop | Description |
|------|-------------|
| `icon` | Component type (auto-sized) or ReactNode |
| `color` | `'emerald' \| 'amber' \| 'blue' \| 'red'` — drives all color classes via static map |
| `buttonLabel` + `buttonHref` | Built-in button with color-matched styling |
| `buttonIcon` | Optional icon after button label (component type or ReactNode) |
| `action` | Custom action element (overrides built-in button) |
| `className`, `iconClassName`, `titleClassName`, `descriptionClassName`, `buttonClassName`, `buttonIconClassName` | Override any element's default classes |
| `role` | ARIA role for the container |

Type-safe discriminated union prevents mixing `action` with `buttonLabel`/`buttonHref`.

## Color system

Colors are managed via a static `Record<BannerColor, ...>` map rather than `tailwind-variants`. We evaluated `tailwind-variants` (tv) for its multi-slot support, but it adds a new dependency (~5KB + tailwind-merge) for a single-dimension variant on one component. The plain map is equivalent in code size, fully type-safe, and zero-dependency. `tailwind-variants` would pay off with multiple variant axes (color × size × outlined) — overkill here.

New colors are added by extending the `BannerColor` union and adding an entry to `colorMap` — TypeScript enforces completeness via `Record`. All color classes are static strings for Tailwind's purge.

Every element's classes can be overridden via `*ClassName` props (last position in `cn()`, so tailwind-merge gives caller priority).

## Test plan

Verified via Playwright across 7 configurations:

- [x] **Baseline** — desktop (1280×800) horizontal layout, mobile (375×812) stacked layout
- [x] **Custom `action` prop** — purple button renders correctly at both sizes
- [x] **ReactNode icon** — emoji icon passes through, `typeof` check works
- [x] **`iconClassName` override** — h-10 w-10 + pink color takes effect via tailwind-merge
- [x] **No action** — banner renders without button when neither prop given
- [x] **All 4 colors** — emerald, amber, blue, red visually distinct at both sizes
- [x] **ARIA `role="alert"`** — confirmed in Playwright accessibility tree snapshot

### DIFF
--- a/src/components/shared/Banner.tsx
+++ b/src/components/shared/Banner.tsx
@@ -118,11 +118,7 @@ function BannerButton({
   return (
     <Button
       asChild
-      className={cn(
-        'w-full shrink-0 sm:w-auto [&>*]:h-4 [&>*]:w-4',
-        colors?.button,
-        className
-      )}
+      className={cn('w-full shrink-0 sm:w-auto [&>*]:h-4 [&>*]:w-4', colors?.button, className)}
     >
       <Link href={href}>{children}</Link>
     </Button>
--- a/src/components/shared/Banner.tsx
+++ b/src/components/shared/Banner.tsx
@@ -41,7 +41,11 @@ type BannerBaseProps = {
   /** Override default icon classes (only when icon is a component type). */
   iconClassName?: string;
   title: string;
+  /** Override default title classes. */
+  titleClassName?: string;
   description: React.ReactNode;
+  /** Override default description classes. */
+  descriptionClassName?: string;
   color: BannerColor;
   /** Override default container classes. */
   className?: string;
@@ -74,7 +78,9 @@ export function Banner({
   icon,
   iconClassName,
   title,
+  titleClassName,
   description,
+  descriptionClassName,
   action,
   buttonLabel,
   buttonHref,
@@ -115,8 +121,10 @@ export function Banner({
       <div className="flex min-w-0 flex-1 items-start gap-3 sm:items-center sm:gap-4">
         <div className="mt-0.5 flex shrink-0 items-center sm:mt-0">{iconNode}</div>
         <div className="min-w-0 flex-1">
-          <p className="text-sm font-semibold sm:font-bold">{title}</p>
-          <p className="text-muted-foreground mt-0.5 text-sm sm:mt-0">{description}</p>
+          <p className={cn('text-sm font-semibold sm:font-bold', titleClassName)}>{title}</p>
+          <p className={cn('text-muted-foreground mt-0.5 text-sm sm:mt-0', descriptionClassName)}>
+            {description}
+          </p>
         </div>
       </div>
       {action ??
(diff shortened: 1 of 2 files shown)


src/components/shared/Banner.tsx
```diff
@@ -0,0 +1,144 @@
+'use client';
+
+import React from 'react';
+import Link from 'next/link';
+import { cn } from '@/lib/utils';
+import { Button } from '@/components/ui/button';
+
+type BannerColor = 'emerald' | 'amber' | 'blue' | 'red';
+
+const colorMap: Record<BannerColor, { border: string; bg: string; text: string; button: string }> =
```
evanjacobson, line 10: These are defaults generated by observing the banners on the /claw page, and a best guess at likely-to-be-used colors. All can be overridden by the className props as described in the PR description. 

These have to be static or they will not work with Tailwind.

A nice-to-have improvement here would have been to use `tailwind-variants`, however it is not currently a dependency and would have provided marginal or no value beyond what we have here.

src/components/shared/Banner.tsx
```diff
@@ -0,0 +1,132 @@
...
+  };
+
+type BannerBaseProps = {
+  icon?: React.ComponentType<{ className?: string }>;
+  /** Override default icon classes. */
+  iconClassName?: string;
+  title: string;
+  /** Override default title classes. */
+  titleClassName?: string;
+  description: React.ReactNode;
```
evanjacobson: The description is ReactNode in order to be more permissive of different types/content, as this is a shared component. This of course includes strings.

RSO (changes requested):
I don't think this type of abstraction is particularly valuable. In my experience, moving all shared functionality to props is very brittle and will likely turn the `Banner` component into a Frankenstein component over time as more and more functionality is being squeezed in.

If all banners right now have icons, titles, descriptions, and optionally a button, I'd prefer to use a JSX structure like this:

```tsx
<Banner>
  <Banner.Icon>
    <KiloCrabIcon className="h-5 w-5 text-emerald-400" />
  </Banner.Icon>
  <Banner.Title>Your KiloClaw instance is active</Banner.Title>
  <Banner.Description>Manage your instance, configure integrations, and monitor your Claw.</Banner.Description>
  <Button>Go to KiloClaw</Button> {/* Or should this be Banner.Button? Not sure if we need custom behavior */}
</Banner>
```

That way, if someone wants to render a Banner without a Button, it's fine, or replace the button with something else. Also fine!

In this particular example, it probably doesn't matter much whether we go left or right, but I think it's good practice to embrace React's composition layer.

RSO (approved):

PR #1594, opened 2026-03-26
feat(dev): support multiple dev servers across worktrees

RSO:
## Summary

- Add `scripts/dev.sh` wrapper that auto-detects an available port (starting from `PORT` env var, default 3000), writes it to `.dev-port`, and starts Next.js with the correct `PORT` and `NEXTAUTH_URL` env vars.
- `APP_URL` in `src/lib/constants.ts` now derives from `process.env.PORT` instead of hardcoding `3000`, while preserving `APP_URL_OVERRIDE` precedence for tunnels (e.g. ngrok).
- `src/lib/trpc/client.tsx` SSR fallback now uses `APP_URL` instead of a hardcoded `http://localhost:3000`.
- Agent instructions in `.kilocode/rules/tools.md` updated to read the port from `.dev-port`.
- `.dev-port` added to `.gitignore`.

This enables running multiple dev servers in parallel across git worktrees without port conflicts or incorrect redirects.

## Verification

- [x] Pre-push hooks (format, lint, typecheck) — all pass

## Visual Changes

N/A

## Reviewer Notes

- The `scripts/dev.sh` port finder uses Node.js `net.createServer()` (stdlib only, no deps) to probe port availability, retrying up to 10 times — same behavior as Next.js's built-in auto-increment.
- We can't hook into Next.js's own port detection because `APP_URL` is a module-level constant evaluated at import time, before Next.js sets `process.env.PORT` in its `server.on('listening')` callback. Pre-detecting and passing `-p` avoids this timing issue.
- The `stripe` script in `package.json` reads from `.dev-port` with a fallback to 3000 if the file doesn't exist.


.kilocode/rules/tools.md
```diff
@@ -7,6 +7,6 @@
...
-- DO NOT RUN `pnpm run dev`! The webserver should already be running; try simply accessing it first via 'http://localhost:3000/'. If it does not appear to be running, EVEN THEN ASK FOR PERMISSION BEFORE RUNNING IT.
-- If you need to log in as a fake user, open http://localhost:3000/users/sign_in?fakeUser=<fake-email> where <fake-email> is constructed from "kilo-", my username (based on homedir), and the time (include seconds) and then '@example.com'. If you need an admin account, the email address must end in @admin.example.com. After logging in, wait for the creating your account spinner to complete before proceeding. The admin panels can be accessed from your profile via the account icon in the top-right corner, which opens a drop-down, allowing access to the admin panel.
+- DO NOT RUN `pnpm run dev`! The webserver should already be running. Read the dev server port from the `.dev-port` file in the project root (e.g., `cat .dev-port`), then access it via `http://localhost:<port>/`. If `.dev-port` does not exist, fall back to port 3000. If the server does not appear to be running, EVEN THEN ASK FOR PERMISSION BEFORE RUNNING IT.
```
markijbema: ok, if it doesnt exist, why not start it in the background?

markijbema (approved):
with this change, i'd say it is fine if the llm starts a server if there is none yet, right?

scripts/dev.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+TARGET_PORT=${PORT:-3000}
+
+# Find an available port starting from TARGET_PORT (same behavior as Next.js auto-increment).
+# Uses Node.js stdlib only — no external dependencies.
+PORT=$(node -e "
```
marius-kilocode, line 9: Could we not simply use port 0 so the OS auto assigns 49152–65535. Or is that problematic for our application?

scripts/dev.sh
```diff
@@ -0,0 +1,34 @@
...
+          reject(err);
+        }
+      });
+      server.once('listening', () => {
+        server.close(() => resolve(port));
+      });
+      server.listen(port);
+    });
+  }
+  tryPort(${TARGET_PORT}, 10).then(p => console.log(p));
```
marius-kilocode: What if I want more than 10 worktrees? :(

scripts/dev.sh
RSO: I think it's just harder to type 49512 instead of 3002 if I'm manually testing the app

markijbema: you could create a npm open-browser which opens the browser based on the .port file

marius-kilocode: Far enough

RSO: At any given time I expect the number of active workspaces to be around 5-10, so it easier for me to remember 3001-3010 then it is 10 random numbers in that 49152–65535 range

chrarnoldus (approved):

jeanduplessis (approved):

scripts/dev.sh
RSO: Fixed, we now fallback to `:0` if we don't find a port in `10` tries

PR #1618, opened 2026-03-26
Add BytePlus and Z.ai Coding Plan Cloud support (UI disabled for now)

chrarnoldus:
CleanShot 2026-03-26 at 22 31 44@2x

### DIFF
--- a/src/app/api/openrouter/models/route.ts
+++ b/src/app/api/openrouter/models/route.ts
@@ -3,8 +3,8 @@ import { NextResponse } from 'next/server';
 import { captureException } from '@sentry/nextjs';
 import type { OpenRouterModelsResponse } from '@/lib/organizations/organization-types';
 import { getEnhancedOpenRouterModels } from '@/lib/providers/openrouter';
-
-export const revalidate = 60;
+import { getUserFromAuth } from '@/lib/user.server';
+import { getCodingPlanModelsForUser } from '@/lib/providers/coding-plans';
 
 /**
  * Test using:
@@ -15,7 +15,10 @@ export async function GET(
 ): Promise<NextResponse<{ error: string; message: string } | OpenRouterModelsResponse>> {
   try {
     const data = await getEnhancedOpenRouterModels();
-    return NextResponse.json(data);
+    const { user } = await getUserFromAuth({ adminOnly: false });
+    return NextResponse.json(
+      user ? { data: data.data.concat(await getCodingPlanModelsForUser(user.id)) } : data
+    );
   } catch (error) {
     captureException(error, {
       tags: { endpoint: 'openrouter/models' },
--- a/src/app/api/openrouter/models/route.ts
+++ b/src/app/api/openrouter/models/route.ts
@@ -42,9 +42,6 @@ export async function GET(
   try {
     const data = await getEnhancedOpenRouterModels();
     return NextResponse.json(
-      // TODO: ENABLE_CODING_PLANS_UI requires:
-      // https://github.com/Kilo-Org/kilocode/pull/7728
-      // we should also watch the performance impact non pre-rendering
       ENABLE_CODING_PLANS_UI && Array.isArray(data.data)
         ? { data: data.data.concat(await getCodingPlanModels()) }
         : data
--- a/src/app/api/openrouter/models/route.ts
+++ b/src/app/api/openrouter/models/route.ts
@@ -31,10 +31,9 @@ export async function GET(
 ): Promise<NextResponse<{ error: string; message: string } | OpenRouterModelsResponse>> {
   try {
     const data = await getEnhancedOpenRouterModels();
-    if (!data.data || !Array.isArray(data.data)) {
-      return NextResponse.json(data);
-    }
-    return NextResponse.json({ data: data.data.concat(await getCodingPlanModels()) });
+    return NextResponse.json(
+      Array.isArray(data.data) ? { data: data.data.concat(await getCodingPlanModels()) } : data
+    );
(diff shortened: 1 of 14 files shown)


src/lib/providers/coding-plans/coding-plan-definitions.ts
chrarnoldus, line 1: Not great we have to hardcode this, but I haven't been able to find an up-to-date source yet. I considered https://models.dev/, but it doesn't have BytePlus and doesn't seem to update automatically either.

src/routers/organizations/organization-settings-router.ts
```diff
@@ -175,6 +177,9 @@ export const organizationsSettingsRouter = createTRPCRouter({
         filteredModels = models;
       }
 
+      if (ENABLE_CODING_PLANS_UI) {
+        filteredModels.push(...(await getCodingPlanModelsForOrganization(organizationId)));
```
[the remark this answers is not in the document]
chrarnoldus: intentional

src/lib/providers/index.ts
```diff
@@ -34,8 +34,39 @@ import { isMinimaxModel } from '@/lib/providers/minimax';
...
+    ? await getBYOKforOrganization(db, organizationId, [codingPlan.id])
+    : await getBYOKforUser(db, user.id, [codingPlan.id]);
+  if (!userByok) {
+    return null;
+  }
+  return {
+    provider: {
+      id: 'coding-plan',
+      apiUrl: codingPlan.base_url,
+      apiKey: userByok[0].decryptedAPIKey,
```
marius-kilocode, line 58: Can this crash if `userByok` is empty?

marius-kilocode: Maye we should check `if (!userByok || userByok.length === 0)`

src/app/api/openrouter/models/route.ts
```diff
@@ -3,6 +3,32 @@ import { NextResponse } from 'next/server';
...
 import type { OpenRouterModelsResponse } from '@/lib/organizations/organization-types';
 import { getEnhancedOpenRouterModels } from '@/lib/providers/openrouter';
+import { getUserFromAuth } from '@/lib/user.server';
+import { getCodingPlanModelsForUser } from '@/lib/providers/coding-plans';
+import { unstable_cache } from 'next/cache';
+import { ENABLE_CODING_PLANS_UI } from '@/lib/constants';
+
+const getCodingPlanModelsForUser_cached = unstable_cache(
+  (userId: string) => getCodingPlanModelsForUser(userId),
+  undefined,
```
marius-kilocode, line 13: Should this not be defined? This smells to me  like a caching issue

src/lib/providers/coding-plans/index.ts
```diff
@@ -0,0 +1,106 @@
...
+  preferredIndex: number
+) {
+  const id = formatCodingPlanModelId(provider, model);
+  const name = provider.name + ': ' + model.name;
+  return {
+    id,
+    canonical_slug: id,
+    hugging_face_id: '',
+    name,
+    created: 1756238927,
```
marius-kilocode: Why is this hardcoded?

src/lib/providers/openrouter/inference-provider-id.ts
```diff
@@ -42,12 +42,22 @@ export type VercelUserByokInferenceProviderId = z.infer<
...
+  'codestral',
+  'zai-coding',
+]);
+
+export type DirectUserByokInferenceProviderId = z.infer<
+  typeof DirectUserByokInferenceProviderIdSchema
+>;
 
-export type UserByokAutocompleteProviderId = z.infer<typeof AutocompleteUserByokProviderIdSchema>;
+export type UserByokAutocompleteProviderId = z.infer<
```
marius-kilocode: Does the type alias in this way still make sense with the new providers?

src/app/api/openrouter/models/route.ts
chrarnoldus: It won't make a difference, the function and the arguments are already part of the cache key.

src/lib/providers/coding-plans/index.ts
chrarnoldus: because we don't care, it's only here for compatibility with OpenRouter's format.

src/lib/providers/openrouter/inference-provider-id.ts
chrarnoldus: no

src/lib/providers/index.ts
chrarnoldus: those functions never return empty, but sure we can make it more defensive

src/lib/providers/coding-plans/index.ts
marius-kilocode: can we add a comment?

src/app/api/openrouter/models/route.ts
marius-kilocode: What about the userId, that doesn't need to be part of the cache key?

chrarnoldus: the arguments are already part of the cache key https://nextjs.org/docs/app/api-reference/functions/unstable_cache#parameters

src/lib/providers/coding-plans/index.ts
chrarnoldus:
```suggestion
    created: 631148400, // our clients do not care about this field, we can fix it later if that changes
```

marius-kilocode (approved):

PR #1624, opened 2026-03-26
fix(kiloclaw): use dedicated NEXTJS_APP_URL for instance-ready email

pandemicsyn:
## Summary

Fix the instance-ready email being completely broken in production. `nextApiOrigin()` used `KILOCODE_API_BASE_URL` to derive the backend app origin, but that env var is intentionally unset in prod — it's a dev/staging override for machine config, not the backend app URL. Every checkin with low load was hitting `Error: KILOCODE_API_BASE_URL not defined`, silently swallowed by an empty catch block.

Introduces a dedicated `BACKEND_API_URL` env var (`https://app.kilo.ai` in prod) and fixes the call ordering so config validation happens *before* `tryMarkInstanceReady()` flips the DO flag — preventing the class of bug where the flag is permanently set but the notification never fires.

## Verification

- [x] `pnpm test` (kiloclaw) — 1020 tests passed, 45 test files
- [ ] Confirm instance-ready emails start sending after deploy (check Axiom for `[controller] instance-ready: dispatching notification` logs without subsequent errors)

## Visual Changes

N/A

## Reviewer Notes

- Root cause found via Axiom query on `cloudflare-logpush` — the logging added in #1613 immediately surfaced the error on every checkin.
- `KILOCODE_API_BASE_URL` is still used for its original purpose (machine env var override in `gateway/env.ts`) — only the `nextApiOrigin()` usage is replaced.
- The call ordering fix is important: previously `tryMarkInstanceReady()` ran first, permanently setting `instanceReadyEmailSent=true` in the DO, then `nextApiOrigin()` would throw, meaning the email could never be retried.

### DIFF
--- a/kiloclaw/.dev.vars.example
+++ b/kiloclaw/.dev.vars.example
@@ -23,6 +23,10 @@ GATEWAY_TOKEN_SECRET=dev-gateway-secret-kiloclaw
 # Fly app naming prefix ("dev-") for per-user app creation.
 WORKER_ENV=development
 
+# Backend app origin for internal API calls (e.g. instance-ready email).
+# In dev, point this at your local Next.js server.
+BACKEND_API_URL=http://localhost:3000
+
 # Tunnel URL (set automatically by the dev-start script)
 # OpenClaw's native kilocode provider hardcodes https://api.kilo.ai/api/gateway/.
 # In dev, the script sets this to the Cloudflare tunnel URL pointing to your
--- a/kiloclaw/wrangler.jsonc
+++ b/kiloclaw/wrangler.jsonc
@@ -98,6 +98,7 @@
     "FLY_ORG_SLUG": "kilo-679", // Org for creating per-user Fly apps
     "FLY_IMAGE_TAG": "latest",
     "FLY_REGION": "eu,us",
+    "BACKEND_API_URL": "https://app.kilo.ai",
     "OPENCLAW_ALLOWED_ORIGINS": "https://claw.kilosessions.ai,https://kilo.ai,https://www.kilo.ai",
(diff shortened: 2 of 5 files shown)


kiloclaw/.dev.vars.example
```diff
@@ -23,6 +23,10 @@ GATEWAY_TOKEN_SECRET=dev-gateway-secret-kiloclaw
 # Fly app naming prefix ("dev-") for per-user app creation.
 WORKER_ENV=development
 
+# Next.js app origin for internal API calls (e.g. instance-ready email).
+# In dev, point this at your local Next.js server.
+NEXTJS_APP_URL=http://localhost:3000
```
jeanduplessis: nit: maybe we can call this something not focused on the tech used for the app? Something like BACKEND_API_URL or something similar.

kiloclaw/wrangler.jsonc
```diff
@@ -98,6 +98,7 @@
     "FLY_ORG_SLUG": "kilo-679", // Org for creating per-user Fly apps
     "FLY_IMAGE_TAG": "latest",
     "FLY_REGION": "eu,us",
+    "NEXTJS_APP_URL": "https://kilo.ai",
```
jeanduplessis: Should this not be app.kilo.ai?

jeanduplessis (changes requested):

kiloclaw/wrangler.jsonc
pandemicsyn: Yea good catch, ill also double check and see if we need the `/api` suffix.

jeanduplessis (approved):

PR #1633, opened 2026-03-26
feat(gateway): add per-user rate limiting for server-side free model usage

eshurakov:
## Summary

Server-side products (cloud-agent, code-review, app-builder) share infrastructure IPs, making IP-based rate limiting ineffective — all users behind the same IP share a single bucket. This change adds per-user rate limiting for those products while preserving IP-based limits for client-side products.

Changes:
- Add `checkFreeModelRateLimitByUser` in `free-model-rate-limiter.ts` that queries by `kilo_user_id` instead of IP.
- Add `isUserRateLimitedFeature` in `feature-detection.ts` to classify which features use per-user limits.
- Extract `resolveRateLimit` helper in the gateway route to branch between per-IP and per-user strategies, requiring authentication for user-rate-limited features (returns 401 if unauthenticated).
- Add partial DB index `idx_free_model_usage_user_created_at` on `(kilo_user_id, created_at) WHERE kilo_user_id IS NOT NULL` to support the new query path efficiently.
- Both strategies share the same `FREE_MODEL_MAX_REQUESTS_PER_WINDOW` (200/hour) constant.

## Verification

## Visual Changes

N/A

## Reviewer Notes

- **Deployment note:** The migration creates the index with a regular `CREATE INDEX`. On a large `free_model_usage` table this will take a full table lock. The index should be created manually using `CREATE INDEX CONCURRENTLY` before running the migration, then the migration will no-op due to `IF NOT EXISTS`.
- The feature header (`x-kilocode-feature`) is client-provided but spoofing a user-rate-limited feature (e.g. `cloud-agent`) forces authentication and gives the same 200/hour per-user limit — no advantage over a unique IP.
- The rate limit check runs before request logging, preserving the existing check-then-log ordering.

### DIFF
--- a/packages/db/src/migrations/0060_lush_avengers.sql
+++ b/packages/db/src/migrations/0060_lush_avengers.sql
@@ -1,0 +1,1 @@
+CREATE INDEX IF NOT EXISTS "idx_free_model_usage_user_created_at" ON "free_model_usage" USING btree ("kilo_user_id","created_at") WHERE "free_model_usage"."kilo_user_id" is not null;
--- a/packages/db/src/migrations/meta/0060_snapshot.json
+++ b/packages/db/src/migrations/meta/0060_snapshot.json
@@ -1,0 +1,14501 @@
+{
+  "id": "8549692e-0544-402d-87e7-a05230a4ee33",
+  "prevId": "1939c85c-a8df-42d5-a17b-eeda39905369",
+  "version": "7",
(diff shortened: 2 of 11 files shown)


RSO (approved):
Discussed in https://kilo-code.slack.com/archives/C08H16KGBUK/p1774563868080859

alex-alecu (approved):

PR #1407, opened 2026-03-23
feat(kiloclaw): add Linear MCP* integration

evanjacobson:
## Summary

Adds Linear as an integrated developer tool in KiloClaw, backed by the official Linear MCP server via mcporter.

- **Secret catalog**: Adds `LINEAR_API_KEY` field with `lin_api_` validation pattern, icon, and help link to Linear security settings
- **Bootstrap**: New `configureLinear` step that wires up the env var; `updateToolsMdLinearSection` appends/removes a bounded reference section in TOOLS.md pointing agents to the Linear MCP server and mcporter skill
- **MCP integration**: When `LINEAR_API_KEY` is set, `writeMcporterConfig()` adds the Linear MCP server entry (key stored as `${LINEAR_API_KEY}` verbatim, read from env by mcporter); entry is removed when the key is unset. Preserves existing user-configured servers. No credentials are ever written to disk.
- **Dashboard UI**: Adds Linear to the "Developer Tools" section in `SettingsTab`, new `LinearIcon` SVG component, and `secret-ui-adapter` wiring (icon map + description)
- **Tests**: Full coverage for `configureLinear`, `updateToolsMdLinearSection`, `writeMcporterConfig` (add, remove, preserve, multi-server), and updated catalog/route assertions

## Verification

- [x] `pnpm format:check` passes
- [x] `pnpm lint` passes
- [x] Tests pass (`vitest` in kiloclaw workspace)
- [x] E2E tested on a fresh instance

## Visual Changes
- Kilo team only: https://www.loom.com/share/5ca9897104e84a3baf560818b7a3bcf7
image
image

## Reviewer Notes

- Originally prototyped with `@schpet/linear-cli` installed in the Docker image, but switched to the official Linear MCP server — it's a better fit since mcporter is already deployed on all instances and avoids writing API keys to disk
- The TOOLS.md section uses `<!-- BEGIN:linear -->` / `<!-- END:linear -->` markers, same pattern as the other tool sections
- No Dockerfile changes — the MCP approach requires no additional system packages

PR #1635, opened 2026-03-27
refactor(kiloclaw): replace Linear CLI with Linear MCP server

evanjacobson:
## Summary

A refactor targeting #1407:

After exploring other options for the Linear integration, the official Linear MCP server turned out to be a better fit than the CLI. All OpenClaw instances are already deployed with the `mcporter` package and skill, which give KiloClaw MCP abilities anyways. Importantly, unlike the CLI, mcporter also allows us to avoid ever writing our actual API key to a file, which significantly reduces the overhead for managing this integration. This PR targets feature/kiloclaw-linear-cli with the scoped changes to switch-over to this MCP integration only, to simplify the reviewer's experience.

## Changes

- Undo the changes to both `Dockerfile`s — the target branch will have no diff after this PR merges.
- Strip all on-disk credential cleanup logic from `configureLinear()` — no new files to manage anymore, so the function no longer needs `BootstrapDeps`
- Add Linear MCP server entry to `writeMcporterConfig()` when `LINEAR_API_KEY` is set; remove it when unset. Preserves existing user-configured servers. The key is stored in the file as `${LINEAR_API_KEY}` verbatim and read from the environment by mcporter. 
- Replace CLI usage documentation in the TOOLS.md agent section with a pointer to the Linear MCP server and mcporter skill
- Simplify `configureLinear` tests (no more fs/exec mocking) and add `writeMcporterConfig` tests covering add, remove, preserve, and multi-server scenarios

## Verification

- [x] `pnpm format:check` passes
- [x] `pnpm lint` passes
- [x] Tests pass (`vitest` in kiloclaw workspace)
- [x] E2E tested on a fresh instance

## New Loom — Kilo Team only
https://www.loom.com/share/5ca9897104e84a3baf560818b7a3bcf7

## Action for reviewer
Please let me know your preference for merge, whether it's into the feature/kiloclaw-cli branch, or into main directly.

### DIFF
--- a/kiloclaw/Dockerfile
+++ b/kiloclaw/Dockerfile
@@ -31,6 +31,8 @@ RUN apt-get update \
        | gpg --dearmor --output /usr/share/debsig/keyrings/AC2D62742012EA22/debsig.gpg \
     && apt-get update \
     && apt-get install -y --no-install-recommends gh 1password-cli \
+    && apt-get purge -y xz-utils \
+    && apt-get autoremove -y \
     && rm -rf /var/lib/apt/lists/* \
     && node --version \
     && npm --version
@@ -55,13 +57,6 @@ RUN npm install -g @steipete/summarize@0.12.0
 # Install Kilo CLI (agentic coding assistant for the terminal)
 RUN npm install -g @kilocode/cli@7.0.46
 
-# Install Linear CLI (issue tracker)
-RUN npm install -g @schpet/linear-cli@1.11.1
-
-# Clean up xz-utils now that Node.js and linear-cli are installed
-RUN apt-get purge -y xz-utils \
-    && apt-get autoremove -y \
-    && rm -rf /var/lib/apt/lists/*
 
 # Install Go (available at runtime for users to `go install` additional tools)
 ENV GO_VERSION=1.26.0
--- a/kiloclaw/controller/src/config-writer.ts
+++ b/kiloclaw/controller/src/config-writer.ts
@@ -349,6 +349,19 @@ export function writeMcporterConfig(
     }
   }
 
+  if (env.LINEAR_API_KEY) {
+    existingServers['linear'] = {
+      baseUrl: 'https://mcp.linear.app/mcp',
+      headers: { Authorization: 'Bearer ${LINEAR_API_KEY}' },
+    };
+    console.log('Linear MCP server configured (via mcporter)');
+  } else {
+    if ('linear' in existingServers) {
+      delete existingServers['linear'];
+      console.log('Linear MCP server removed from mcporter config');
+    }
+  }
+
   // Only write if there are servers to configure or we need to clean up
   if (Object.keys(existingServers).length === 0 && !deps.existsSync(configPath)) {
     return;
--- a/kiloclaw/controller/src/config-writer.ts
+++ b/kiloclaw/controller/src/config-writer.ts
@@ -351,7 +351,7 @@ export function writeMcporterConfig(
 
   if (env.LINEAR_API_KEY) {
     existingServers['linear'] = {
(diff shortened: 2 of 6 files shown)


kiloclaw/controller/src/config-writer.ts
```diff
@@ -349,6 +349,19 @@ export function writeMcporterConfig(
     }
   }
 
+  if (env.LINEAR_API_KEY) {
+    existingServers['linear'] = {
+      baseUrl: 'https://mcp.linear.app/mcp',
+      headers: { Authorization: 'Bearer ${LINEAR_API_KEY}' },
```
evanjacobson, line 355: Note that this is intentionally not interpolated. mcporter supports this behavior.

kiloclaw/Dockerfile
evanjacobson, line 1: For both Dockerfiles — restores the other PR's change

kiloclaw/controller/src/config-writer.ts
```diff
@@ -349,6 +349,19 @@ export function writeMcporterConfig(
     }
   }
 
+  if (env.LINEAR_API_KEY) {
+    existingServers['linear'] = {
+      baseUrl: 'https://mcp.linear.app/mcp',
```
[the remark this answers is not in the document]
evanjacobson: Mcporter officially supports 5 variants of the url field name — while the claim was untrue, I will change it regardless to align with the other entry.

St0rmz1 (commented):
**Bug: `baseUrl` vs `url` mismatch**

`kiloclaw/controller/src/config-writer.test.ts` — test "adds Linear MCP server when LINEAR_API_KEY is set"

The implementation writes `url` but the test asserts `baseUrl`:

```ts
// config-writer.ts
existingServers['linear'] = {
  url: 'https://mcp.linear.app/mcp',
  ...
};

// config-writer.test.ts
expect(config.mcpServers.linear).toEqual({
  baseUrl: 'https://mcp.linear.app/mcp',  // should this be `url`?
  ...
});
```

`toEqual` does a deep equality check, so this assertion should be failing — the object has `url` but the expected has `baseUrl`. Either the test needs to use `url` to match the implementation, or the implementation should use `baseUrl` if that's what mcporter expects. Worth double-checking which field mcporter actually reads.

evanjacobson:
> **Bug: `baseUrl` vs `url` mismatch**
> 
> `kiloclaw/controller/src/config-writer.test.ts` — test "adds Linear MCP server when LINEAR_API_KEY is set"
> 
> The implementation writes `url` but the test asserts `baseUrl`:
> 
> ```ts
> // config-writer.ts
> existingServers['linear'] = {
>   url: 'https://mcp.linear.app/mcp',
>   ...
> };
> 
> // config-writer.test.ts
> expect(config.mcpServers.linear).toEqual({
>   baseUrl: 'https://mcp.linear.app/mcp',  // should this be `url`?
>   ...
> });
> ```
> 
> `toEqual` does a deep equality check, so this assertion should be failing — the object has `url` but the expected has `baseUrl`. Either the test needs to use `url` to match the implementation, or the implementation should use `baseUrl` if that's what mcporter expects. Worth double-checking which field mcporter actually reads.

@St0rmz1 good catch — it's interesting to note that mcporter functionally supports [5 different naming conventions](https://github.com/steipete/mcporter/blob/74dd794645b583eee1db81c4e5fff97725a14185/src/config-schema.ts#L52-L56) for this field, although the base URLs have different behavior when specific flags are passed. [Otherwise, they are coalesced](https://github.com/steipete/mcporter/blob/74dd794645b583eee1db81c4e5fff97725a14185/src/config-normalize.ts#L92). I flipped to URL late last night to standardize with 1pass's implementation per CR bot's guidance. Just fixed it!

PR #1649, opened 2026-03-27
feat(tracking): set first_product_signup person property for KiloClaw signups

pedroheyerdahl:
## Summary

PostHog person profiles carry no signal about which product a user signed up for. The landing page sets `first_product_intent` at CTA click time (capturing browsing interest), but the cloud app never recorded what the user was actively signing up for when they completed authentication. @jobrietbergen needs this split to segment KiloClaw vs Kilo Code users in dbt and Customer.io.

This adds a `first_product_signup` person property, set server-side as `$set_once` in the `after-sign-in` route handler. The route runs exactly once per signup and has access to both the new-user signal (`has_validation_stytch === null`, set before account verification completes) and the `callbackPath`/`source` params that identify which product flow carried the user through auth.

The product mapping is explicit:

| Entry point | `first_product_signup` |
|---|---|
| `callbackPath` starts with `/claw` | `kiloclaw` |
| `callbackPath` starts with `/cloud` | `cloud-agent` |
| `callbackPath` starts with `/code-reviews` | `code-reviews` |
| `callbackPath` starts with `/app-builder` | `app-builder` |
| `callbackPath` starts with `/install` | `kilo-code` |
| `source` param present (IDE install flow) | `kilo-code` |
| Generic signup (`/get-started`, `/profile`, no callbackPath) | *(unset)* |

Together with `first_product_intent` from the landing page, this covers the full attribution picture: acquisition interest and conversion context. They often match, but a user who clicks a KiloClaw ad and then signs up via the generic header button will have `intent: kiloclaw` and `signup: (unset)`.

The existing `user_created` event (`src/lib/user.ts:340`) provides the funnel anchor timestamp. No new conversion event is needed for dbt; the person property joins to any event on that person.

## Verification

- [x] Typecheck passes (pre-existing failures in `ChatTab.tsx` missing `stream-chat` types, unrelated)
- [x] Verified `has_validation_stytch === null` is the correct one-shot new-user gate (set after account verification in the Stytch flow, so only null for brand-new unverified signups)
- [x] Confirmed `PostHogClient().capture()` with `$set_once` in properties is the established server-side pattern (same as `user_created` event in `src/lib/user.ts:351`)
- [x] No unit test added: the change is a conditional PostHog `capture` call behind the existing new-user gate, in a route that already handles user state and redirects. It touches no auth logic, database writes, or redirect behavior. The worst case is the PostHog call silently fails (no-op in non-production) and the property is not set.

## Visual Changes

N/A

## Reviewer Notes

An earlier version of this PR set the property client-side in the `IdentifyUser` component, gated on `session.isNewUser`. The review correctly identified that `isNewUser` persists in the JWT for up to 30 days, so a user who signed up generically could be incorrectly stamped on a later full page load to `/claw`. The server-side approach avoids this entirely: `has_validation_stytch === null` is true only for the initial post-signup redirect, before account verification completes.

The `signup_product_attributed` event is emitted alongside the `$set_once` to provide an auditable record of when and how the property was set, with the raw `callback_path` for debugging.


src/app/users/after-sign-in/route.tsx
```diff
@@ -27,6 +44,24 @@ export async function GET(request: NextRequest) {
     }
 
     if (user.has_validation_stytch === null) {
+      // New user: stamp which product they signed up for. Derived from
+      // responsePath (the already-validated destination) rather than the raw
+      // callbackPath query param, so the value cannot be user-tampered. This
+      // runs exactly once per signup because has_validation_stytch is set
+      // after account verification completes.
+      const product = resolveSignupProduct(responsePath, !!url.searchParams.get('source'));
```
St0rmz1, line 52: When there's no `callbackPath` and no `source`, `responsePath` falls through to `getProfileRedirectPath(user)`. If that ever returns a product prefixed path (e.g. `/claw/...`) for a brand-new user, they'd be misattributed even though they signed up generically.

Worth confirming what `getProfileRedirectPath` returns for users with `has_validation_stytch === null`. If there's any risk, one option is to only resolve the product when a callbackPath was explicitly provided:

```ts
const product = resolveSignupProduct(
  callbackPath && isValidCallbackPath(callbackPath) ? responsePath : null,
  !!url.searchParams.get('source'),
);
```

St0rmz1 (approved):

src/app/users/after-sign-in/route.tsx
pedroheyerdahl: Just seeing this. I'm on it

PR #1656, opened 2026-03-27
Enable Coding Plan support in the Kilo Gateway

chrarnoldus:


src/app/api/openrouter/models/route.ts
```diff
@@ -42,9 +39,7 @@ export async function GET(
   try {
     const data = await getEnhancedOpenRouterModels();
     return NextResponse.json(
-      ENABLE_CODING_PLANS_UI && Array.isArray(data.data)
-        ? { data: data.data.concat(await getCodingPlanModels()) }
-        : data
+      Array.isArray(data.data) ? { data: data.data.concat(await getCodingPlanModels()) } : data
```
chrarnoldus, line 42: reading auth here will probably break prerendering, but I hope it's fine because the OpenRouter response is still cached: https://github.com/Kilo-Org/cloud/blob/eec4b47bcbbbe3ee1a6bad4f926c6fa77cc66dea/src/lib/providers/openrouter/index.ts#L109

src/lib/constants.ts
```diff
@@ -36,8 +36,6 @@ export const AUTOCOMPLETE_MODEL = 'codestral-2508';
 
 export const ENABLE_DEPLOY_FEATURE = true;
 
-export const ENABLE_CODING_PLANS_UI = false; // Requires CLI update: https://github.com/Kilo-Org/kilocode/pull/7728
```
chrarnoldus, line 39: it turns out no update is necessary

src/routers/organizations/organization-settings-router.ts
```diff
@@ -177,9 +176,7 @@ export const organizationsSettingsRouter = createTRPCRouter({
         filteredModels = models;
       }
 
-      if (ENABLE_CODING_PLANS_UI) {
-        filteredModels.push(...(await getCodingPlanModelsForOrganization(organizationId)));
-      }
+      filteredModels.push(...(await getCodingPlanModelsForOrganization(organizationId)));
```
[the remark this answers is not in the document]
chrarnoldus: that's fine, if you don't want these models, then don't byok them

RSO (approved):
The weekends are always quiet in terms of database load, so even if this does break caching we'll likely only find out on Monday.

PR #1657, opened 2026-03-27
feat(kiloclaw): add Kilo CLI recovery agent

RSO:
## Summary

- Add the ability to run `kilo run --auto` on a KiloClaw instance from the dashboard as a recovery tool for stuck/failing instances, with real-time output polling and DB persistence
- Full stack implementation: controller spawn route → CF Worker DO proxy → platform routes → tRPC procedures → React polling page with auto-scrolling output viewer
- Admin visibility via a dedicated CLI Runs tab with search and pagination
- System context prompt template that gives the agent awareness of the KiloClaw environment (config paths, gateway setup, common issues)
- UI rebranded from generic "Run Kilo Agent" to recovery-focused "Recover with Kilo" messaging

## Verification

- [x] `pnpm typecheck` — passes (all packages)
- [x] `pnpm lint` — passes (0 warnings, 0 errors)
- [x] Controller route tests included (12 tests in `kilo-cli-run.test.ts`)
- [x] Migration 0061 regenerated cleanly after rebase on main (no number collision with main's 0060)

## Visual Changes

https://github.com/user-attachments/assets/a409523d-9c57-4839-b807-7620bbed8357

## Reviewer Notes

- The drizzle migration was regenerated as 0061 after rebasing onto main (which already has 0060). The migration adds the `kiloclaw_cli_runs` table with FK to `kilocode_users` and indexes on `user_id` and `started_at`.
- The recovery agent uses the instance's configured default model and KiloCode API key. The system prompt template in `kilo-cli-config.ts` gives the agent context about the OpenClaw environment.
- Output polling uses `@tanstack/react-query` with a 2s refetch interval while the run is active.


kiloclaw/controller/src/bootstrap.ts
```diff
@@ -188,6 +188,11 @@ export function setupDirectories(env: EnvLike, deps: BootstrapDeps = defaultDeps
 
   // GOG_KEYRING_PASSWORD is NOT a secret — see gog-credentials.ts for context.
   env.GOG_KEYRING_PASSWORD = 'kiloclaw';
+
+  // Derive the API origin for the Kilo CLI from the full base URL.
+  if (env.KILOCODE_API_BASE_URL) {
+    env.KILO_API_URL = new URL(env.KILOCODE_API_BASE_URL).origin;
+  }
```
RSO, line 195: This replaces the `providers.kilo.options.baseURL` setting in the `opencode.json` config, which doesn't work as expected.

kiloclaw/controller/src/kilo-cli-config.ts
```diff
@@ -49,31 +57,45 @@ export function writeKiloCliConfig(
...
-      config.provider = config.provider || {};
-      config.provider.kilo = config.provider.kilo || {};
-      config.provider.kilo.options = config.provider.kilo.options || {};
-      config.provider.kilo.options.baseURL = env.KILOCODE_API_BASE_URL;
-      console.log('[kilo-cli] Patched base URL: ' + env.KILOCODE_API_BASE_URL);
+      if (needsBaseUrlPatch) {
+        config.provider = config.provider || {};
+        config.provider.kilo = config.provider.kilo || {};
+        config.provider.kilo.options = config.provider.kilo.options || {};
+        delete config.provider.kilo.options.baseURL;
```
RSO, line 87: As far as I could test `baseURL` just never works as you want it to because either the models-fetch or the chat-completion fails because it uses the wrong path.

src/routers/admin-kiloclaw-instances-router.ts
```diff
@@ -494,6 +495,107 @@ export const adminKiloclawInstancesRouter = createTRPCRouter({
...
+            prompt: kiloclaw_cli_runs.prompt,
+            status: kiloclaw_cli_runs.status,
+            exit_code: kiloclaw_cli_runs.exit_code,
+            output: kiloclaw_cli_runs.output,
+            started_at: kiloclaw_cli_runs.started_at,
+            completed_at: kiloclaw_cli_runs.completed_at,
+          })
+          .from(kiloclaw_cli_runs)
+          .leftJoin(kilocode_users, eq(kiloclaw_cli_runs.user_id, kilocode_users.id))
+          .where(where)
```
St0rmz1, line 579: The `output` column can be up to ~1MB per row (from the controller's `MAX_OUTPUT_BYTES` cap). With a page size of 25, that's potentially ~25MB per API response.

Consider either:
- Excluding `output` from this list query and fetching it separately when a run is selected (e.g., a dedicated `getCliRun` procedure)
- Or truncating here (e.g., `` sql`LEFT(output, 500)` ``)

The admin detail panel already has `RunDetail` as a separate component — it could fetch the full output on demand rather than having it pre-loaded in the list payload.

packages/trpc/dist/index.d.ts
```diff
@@ -3106,6 +3106,7 @@ type BotRequestStep = {
         totalTokens?: number;
     };
 };
+type KiloClawCliRunStatus = 'running' | 'completed' | 'failed' | 'cancelled';
```
St0rmz1: This is a generated declaration file in `dist/` — is it intentional that it's checked in?

src/routers/kiloclaw-router.ts
```diff
@@ -844,6 +845,135 @@ export const kiloclawRouter = createTRPCRouter({
...
+        };
+      }
+
+      // Run is still active — poll the controller for live output.
+      const client = new KiloClawInternalClient();
+      const controllerStatus = await client.getKiloCliRunStatus(ctx.user.id);
+
+      // If controller reports the run finished, persist to the DB row.
+      if (
+        controllerStatus.hasRun &&
```
St0rmz1, line 1065: The DB update here uses `eq(kiloclaw_cli_runs.status, 'running')` without scoping to a specific run ID. If multiple rows are ever in `running` state due to a bug, this cancels all of them. Low risk since the controller enforces one at a time, but adding a `runId` input and including it in the WHERE clause would be more defensive.

packages/trpc/dist/index.d.ts
RSO: Seems like it was brought back in `main`, so not something that I did. I think it was brought back on accident though, so I raised it [here](https://kilo-code.slack.com/archives/C09FDJ2GEF6/p1774628210831309?thread_ts=1774513713.121909&cid=C09FDJ2GEF6)

src/routers/kiloclaw-router.ts
RSO: Fixed in 678d47b — `cancelKiloCliRun` now takes a `runId` input and the WHERE clause includes `eq(kiloclaw_cli_runs.id, input.runId)` alongside the user and status checks.

jeanduplessis (approved):

PR #1664, opened 2026-03-27
fix(claw): handle Fly 'destroyed' state on start after admin force-delete

evanjacobson:
## Summary

#1550 added a destroy machine button. If the user tries to start a destroyed machine again too quickly from the admin dashboard, it will fail until the 'destroyed' state clears and the machine 404s. This change allows a machine to instantly be created after destroy.

- After admin force-deletes a Fly machine via "Destroy Machine", the Fly API briefly returns the machine with `state: 'destroyed'` before switching to 404
- `start()` and `startExistingMachine()` didn't handle this state — fell through to `waitForState('started')` which timed out on the dead machine, leaving the DO with a stale `flyMachineId`
- The second "Start Machine" click worked because by then the API returned 404, which was already handled
- Now both `_startInner()` and `startExistingMachine()` treat `'destroyed'` the same as 404: clear `flyMachineId` and create a fresh machine

## Test plan

- [x] Existing 303 tests pass
- [x] Added 2 new tests covering both code paths (running-check in `_startInner` and `startExistingMachine` fallback)
- [ ] Manual: admin force-delete a machine, then click "Start Machine" — should succeed on the first click

### DIFF
--- a/kiloclaw/src/durable-objects/kiloclaw-instance/fly-machines.ts
+++ b/kiloclaw/src/durable-objects/kiloclaw-instance/fly-machines.ts
@@ -210,7 +210,14 @@ export async function startExistingMachine(
       console.log('[DO] Machine destroyed, creating new one');
       state.flyMachineId = null;
       await ctx.storage.put(storageUpdate({ flyMachineId: null }));
-      await createNewMachine(flyConfig, ctx, state, initialMachineConfig, minSecretsVersion, envFlyRegion);
+      await createNewMachine(
+        flyConfig,
+        ctx,
+        state,
+        initialMachineConfig,
+        minSecretsVersion,
+        envFlyRegion
+      );
       return;
     }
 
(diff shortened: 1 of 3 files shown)


kiloclaw/src/durable-objects/kiloclaw-instance/fly-machines.ts
```diff
@@ -203,6 +203,24 @@ export async function startExistingMachine(
...
+    // This happens when a Fly machine was force-deleted (e.g. admin destroy)
+    // but the API briefly returns the machine with state='destroyed' before 404.
+    if (machine.state === 'destroyed') {
+      state.flyMachineId = null;
+      await ctx.storage.put(storageUpdate({ flyMachineId: null }));
+      await createNewMachine(
+        flyConfig,
+        ctx,
+        state,
+        initialMachineConfig,
```
evanjacobson, line 216: there is no machineConfig to backfill here, as no machine is running.

jeanduplessis:

PR #1729, opened 2026-03-30
fix(security): remove sensitive data from logs and observability

alex-alecu:
## Summary

Authentication tokens, credentials, and webhook secrets were being leaked through multiple paths: logged to console (forwarded to Sentry), stored in the database, interpolated into LLM prompts, and sent to Sentry through `sendDefaultPii` and `attachRpcInput`.

### Why `sendDefaultPii` and `attachRpcInput` are dangerous

**`sendDefaultPii`** tells Sentry: "attach all request headers, cookies, and IP addresses to every error you capture." The problem is it grabs *everything* — it can't tell the difference between a harmless `Content-Type` header and a secret `Authorization: Bearer <jwt>` header. So when a user's request causes an error, their login token ends up sitting in Sentry where anyone with dashboard access can see it.

**`attachRpcInput`** does the same thing but for tRPC calls — it sends the full input of every procedure to Sentry. Some procedures accept personal access tokens as input, so those tokens end up in Sentry too.

Both are now disabled for services that handle user credentials (Next.js app, mobile app). The deploy builder keeps `sendDefaultPii` enabled because it only handles internal service-to-service requests with a shared backend token — no user secrets flow through it.

### What changed

- Removed full HTTP header dumps and raw JWT logging from token validation
- Disabled `sendDefaultPii` in the Next.js client and mobile app Sentry configs
- Disabled `attachRpcInput` in the tRPC Sentry middleware
- Kept `sendDefaultPii` enabled for the deploy builder (internal service, no user credentials)
- Added a shared `redactSensitiveHeaders` utility in `@kilocode/worker-utils` that strips auth-bearing headers (including custom webhook auth headers) before storage or prompt injection
- Applied header redaction to GitHub/GitLab webhook handlers, TriggerDO request capture, and prompt template rendering
- Added a concise logging policy to AGENTS.md to prevent regressions

### What we lose for debugging

Most changes have little impact. Here's a honest look at what gets harder:

| Change | What we lose | How bad |
|--------|-------------|---------|
| Removed header dump in `tokens.ts` | Can't see the full headers on auth failures. We still get a tracking ID to match warnings to requests. | Medium — harder to debug weird auth headers |
| `attachRpcInput: false` | Sentry no longer shows what input a tRPC call had when it failed. For simple calls this is fine, for complex ones it's harder to figure out what went wrong. | **Most noticeable** — this is the change most likely to slow down debugging |
| `sendDefaultPii: false` (Next.js, mobile) | Sentry stops attaching IP and cookies to errors. User ID is still tagged separately so we still know *who* hit the error. | Small — we rarely need cookies to debug |
| Webhook header redaction | Auth headers like `x-hub-signature-256` are replaced with `[REDACTED]`. All other headers (event type, delivery ID, content-type) are kept. | Tiny — we never need auth header values to debug webhook logic |
| Fake login log | Logs just the email instead of the whole object. The object only had `{ email: "..." }` anyway. | None |

**If `attachRpcInput: false` becomes painful:** the targeted fix is to add a `beforeSend` hook in Sentry that scrubs known sensitive fields from tRPC inputs instead of turning off capture entirely.

## Verification

- [x] `pnpm typecheck` — passes (all packages)
- [x] `pnpm test` in `packages/worker-utils` — 45/45 tests pass (includes 7 redact-headers tests)
- [x] Root `pnpm test` — all failures are pre-existing DB connection errors (`role "postgres" does not exist`), unrelated to this change

## Visual Changes

N/A

## Reviewer Notes

- The `consoleLoggingIntegration` in `sentry.server.config.ts` and `sentry.edge.config.ts` was intentionally left unchanged — once the sources (tokens.ts) stop emitting sensitive data, the integration no longer amplifies secrets.
- Header keys from the Web API `Headers` class are always lowercase, but `redactSensitiveHeaders` does case-insensitive matching to handle the TriggerDO case where headers come from a custom record.
- Already-captured secrets in Sentry/Axiom retention are an operational concern outside this PR's scope.


cloudflare-webhook-agent-ingest/src/dos/TriggerDO.ts
```diff
@@ -442,7 +443,7 @@ export class TriggerDO extends DurableObject<Env> {
         method: request.method,
         path: request.path,
         query_string: request.queryString,
-        headers: JSON.stringify(request.headers),
+        headers: JSON.stringify(redactSensitiveHeaders(request.headers)),
```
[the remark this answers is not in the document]
alex-alecu: Fixed in 434cb58. `redactSensitiveHeaders` now accepts an optional `extraHeaders` param, and `TriggerDO.captureRequest` passes the configured `storedWebhookAuth.header` through it.

cloudflare-webhook-agent-ingest/src/util/prompt-template.ts
```diff
@@ -14,7 +16,7 @@ export function renderPromptTemplate(template: string, request: WebhookRequest):
     '{{bodyJson}}': tryPrettyJson(request.body),
     '{{method}}': request.method,
     '{{path}}': request.path,
-    '{{headers}}': JSON.stringify(request.headers, null, 2),
+    '{{headers}}': JSON.stringify(redactSensitiveHeaders(request.headers), null, 2),
```
St0rmz1, line 21: This call doesn't pass `extraHeaders`, so custom webhook auth headers (e.g. `x-webhook-secret`) won't be redacted at this layer. Currently mitigated because `TriggerDO.captureRequest` redacts before DB storage, and the queue consumer reads from DB — so headers arriving here are already scrubbed.

Worth either a code comment explaining why `extraHeaders` isn't needed here, or threading the webhook auth header through for defense-in-depth. If a future caller passes raw headers directly to `renderPromptTemplate`, the custom auth header would leak into the LLM prompt.

src/lib/tokens.ts
```diff
@@ -114,9 +114,6 @@ function tryJwtVerify(token: string) {
 
 export function validateAuthorizationHeader(headers: Headers) {
   const traceability_logging_id = crypto.randomUUID();
-  if (process.env.NODE_ENV !== 'development') {
-    logExceptInTest(`headers (${traceability_logging_id})`, Object.fromEntries(headers.entries()));
-  }
```
St0rmz1: Nit: removing the header logging block left an extra blank line here.

packages/worker-utils/src/redact-headers.ts
```diff
@@ -0,0 +1,29 @@
+const SENSITIVE_HEADERS = new Set([
+  'authorization',
+  'cookie',
+  'set-cookie',
+  'x-gitlab-token',
+  'x-hub-signature',
+  'x-hub-signature-256',
```
St0rmz1, line 8: Consider adding `proxy-authorization` to the list — it's a standard HTTP header that carries credentials, same pattern as `authorization`.

St0rmz1 (approved):

cloudflare-webhook-agent-ingest/src/util/prompt-template.ts
alex-alecu: Added a comment explaining this in 63d8887. Headers here are already redacted by TriggerDO before DB storage, so extraHeaders isn't needed at this layer.

src/lib/tokens.ts
alex-alecu: Fixed in a1d9019.

packages/worker-utils/src/redact-headers.ts
alex-alecu: Done in 38b5758.

PR #1751, opened 2026-03-30
feat(admin): bulk KiloClaw trial extend/resurrect

evanjacobson:
## Summary

Adds a **Bulk KiloClaw Trial Extension** tool to the admin panel, accessible under the existing Bulk Credits & Trials page. Admins can paste a list of emails or upload a CSV, preview which users are eligible, and extend or restart KiloClaw trials in bulk with a single click.

### How it works

1. **Input** — paste emails or upload a CSV (automatic email column detection). Both inputs share the same UI; the CSV column selector and preview appear inline when a file is loaded.
2. **Match** — a read-only lookup resolves emails to Kilo accounts and groups results into three buckets: eligible, ineligible, and not found. Each bucket is shown as its own table with an inline export.
3. **Apply** — clicking Apply sends only the eligible users to the server. Results are shown per-user with success/failure detail and exports for successful and failed outcomes.

### Eligibility

- **`trialing`** → eligible. Trial is extended by N days from the later of the current end date or today, so expired trials extend from today rather than compounding from a past date.
- **`canceled`** → eligible. The subscription is resurrected as a fresh trial: status reset to `trialing`, plan reset to `trial`, all Stripe/billing fields cleared. The email suppression log is also cleared so trial notifications fire again.
- **Active paid subscription** (`active`, `past_due`, `unpaid`) → ineligible. We must not touch a user who is actively paying.
- **No subscription** → ineligible. The user has never provisioned a KiloClaw instance and has no subscription row to operate on.

There is no concern with extending a user's trial more than once.

### Product decisions

**365-day input cap, 1-year ceiling:** The input is capped at 365 days. Additionally, the resulting `trial_ends_at` is capped at 1 year from the moment Apply is clicked — meaning a user who is already mid-trial cannot be pushed past 1 year out regardless of how many days are requested. This prevents runaway trial windows.

**Canceled resurrection — direct DB write, no Stripe API call:** When a subscription is `canceled`, the Stripe subscription is already terminated (that's what triggered the webhook that set the status). Resurrecting via a direct DB write is safe. `stripe_subscription_id` is explicitly nulled to prevent any stale Stripe `customer.subscription.deleted` webhook retry from immediately re-canceling the resurrected row.

**Per-user error isolation:** The apply step processes users in a sequential loop on the server. A failure for one user (e.g. their status changed between match and apply) does not abort the others — all users are attempted and a per-user result is returned.

**Exports:** Successful export includes `instance_id`. Ineligible export includes `instance_id`, `stripe_subscription_id`, and a reason string. Failed export includes the error message. Unmatched export is emails only.

## Verification

- `pnpm typecheck` — passes (one pre-existing `csv-parse/sync` error in an unrelated script)
- `pnpm lint` / `pnpm oxfmt` — passes for all files touched by this branch
- Tests added for the 1-year ceiling: trialing extension with existing remainder, canceled resurrection, and normal sub-365-day extension — all pass

## Visual Changes

N/A — admin-only UI.

## Reviewer Notes

- The `canceled` resurrection path mirrors the existing single-user admin reset in `admin-router.ts`. That path also has the multi-instance exposure (its UPDATE uses `WHERE user_id = ?`) — worth a follow-up but out of scope here.
- `payment_source` and `pending_conversion` are intentionally not cleared on resurrection. `stripe_subscription_id = null` is the effective guard for both concerns; the full billing implications of clearing these fields across all `payment_source` values were not fully analyzed.
- The pre-push hook was bypassed on force-push after rebase due to 7 pre-existing lint errors in unrelated `kilo-app/` files (confirmed via `git diff origin/main...HEAD`).
- **Multi-instance users:** `kiloclaw_subscriptions` no longer has a unique constraint on `user_id` — users can have multiple instances, each with its own subscription row. All queries use `SELECT DISTINCT ON (user_id) ORDER BY user_id, created_at DESC` to target only the most recently created subscription per user, and UPDATEs are scoped to the specific row `id` to avoid touching other instances. This will need to be revisited at some point after that feature launches.


packages/db/src/schema-types.ts
```diff
@@ -171,6 +171,8 @@ export const KiloClawAdminAuditAction = z.enum([
   'kiloclaw.config.restore',
   'kiloclaw.doctor.run',
   'kiloclaw.machine.destroy_fly',
+  'kiloclaw.subscription.bulk_trial_grant',
+  'kiloclaw.machine.destroy_fly',
```
evanjacobson: not my change

src/routers/admin/extend-claw-trial-router.ts
```diff
@@ -0,0 +1,332 @@
...
+          } else {
+            // Active paid subscription (active, past_due, unpaid) — must not be reset.
+            results.push({
+              email,
+              userId: user.id,
+              success: false,
+              error: `Cannot extend trial: subscription status is "${subscription.status}". Only trialing or canceled subscriptions can be modified.`,
+            });
+          }
+        } catch (error) {
```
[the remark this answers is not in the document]
evanjacobson: (Automated): Already addressed — both the email-log delete and audit-log insert are wrapped in their own inner `try/catch` (marked `// Non-fatal`) placed after the critical UPDATE in each branch. The outer catch can only trigger on the DB UPDATE itself, so a cleanup failure cannot produce a false-negative `success: false`.

src/app/admin/components/KiloclawExtendTrial.tsx
```diff
@@ -0,0 +1,856 @@
...
+
+  const handleDownloadResults = (success: boolean) => {
+    if (!results) return;
+    const filtered = results.filter(r => r.success === success);
+    if (filtered.length === 0) {
+      toast.info(`No ${success ? 'successful' : 'failed'} results to export`);
+      return;
+    }
+    const content = success
+      ? 'email,action,new_trial_ends_at\n' +
```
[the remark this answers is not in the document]
evanjacobson: (Automated): Fixed — failed-results CSV now exports `email,user_id,instance_id,error` so the specific subscription row that failed can be identified.

src/routers/admin/extend-claw-trial-router.ts
```diff
@@ -0,0 +1,387 @@
...
+        try {
+          if (subscription.status === 'trialing') {
+            // Extend from the later of current end date or now, so already-expired
+            // trials extend from today rather than from a past date.
+            // Scoped to the specific row id to avoid touching other instances.
+            const [updated] = await db
+              .update(kiloclaw_subscriptions)
+              .set({
+                // Extend from the later of current end or now, capped at 1 year from now.
+                trial_ends_at: sql`LEAST(GREATEST(COALESCE(${kiloclaw_subscriptions.trial_ends_at}::timestamptz, now()), now()) + (${trialDays} * interval '1 day'), now() + interval '1 year')`,
```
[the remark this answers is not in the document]
evanjacobson: (Automated): Already addressed — `LEAST()` was removed. Users with trial already >1 year out are now marked `subscriptionStatus: 'at_limit'` at match time and shown as ineligible, so they never reach `extendTrials`.

packages/db/src/schema-types.ts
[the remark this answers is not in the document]
evanjacobson: (Automated): Fixed — the duplicate `'kiloclaw.machine.destroy_fly'` entry has been removed from `KiloClawAdminAuditAction`.

src/routers/admin/extend-claw-trial-router.ts
```diff
@@ -0,0 +1,393 @@
...
+            // Extend from the later of current end date or now, so already-expired
+            // trials extend from today rather than from a past date.
+            // Scoped to the specific row id to avoid touching other instances.
+            const [updated] = await db
+              .update(kiloclaw_subscriptions)
+              .set({
+                // Extend from the later of current end date or now, so already-expired
+                // trials extend from today. Users whose trial already exceeds 1 year
+                // are marked ineligible at match time and never reach this path.
+                trial_ends_at: sql`GREATEST(COALESCE(${kiloclaw_subscriptions.trial_ends_at}::timestamptz, now()), now()) + (${trialDays} * interval '1 day')`,
```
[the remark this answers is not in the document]
evanjacobson: (Automated): Fixed — `LEAST(..., now() + interval '1 year')` restored to the `extendTrials` UPDATE. The `at_limit` check at match time handles users already past the ceiling (prevents them from being submitted at all), and the SQL `LEAST` enforces the ceiling on the resulting value for users currently within it (e.g. 200 days remaining + 365 days requested = capped at 365 days). Test added to lock this in.

src/app/admin/components/KiloclawExtendTrial.tsx
```diff
@@ -0,0 +1,893 @@
...
+                        )}
+                      </TableCell>
+                      <TableCell className="text-sm">
+                        {result.newTrialEndsAt
+                          ? new Date(result.newTrialEndsAt).toLocaleDateString(undefined, {
+                              timeZone: 'UTC',
+                            })
+                          : '—'}
+                      </TableCell>
+                      <TableCell className="text-destructive text-sm">
```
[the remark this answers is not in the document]
evanjacobson: (Automated): Fixed — the orphaned `Days` `<TableHead>` has been removed. This was a merge artifact from when the Days column was removed from cells but not from the header.

evanjacobson (commented):
Leaving file-level notes on the larger new files to orient reviewers.

src/app/admin/components/KiloclawExtendTrial.tsx
```diff
@@ -0,0 +1,896 @@
+'use client';
```
evanjacobson, line 1: This is a brand-new component (~900 lines) — the entire file is an addition, so the diff looks huge but there's nothing changing under it. It implements the UI for the KiloClaw Trial Extension tab: CSV upload or paste input, email→user matching (calls `extendClawTrial.matchUsers`), a preview table with per-row subscription status badges, a "days to extend" field, and a results table after submission. The two main user flows are extending an active trial (pushes `trial_ends_at` forward) and restarting an expired one. Both are handled by the single `extendClawTrial.extendTrials` mutation on the backend.

src/routers/admin/extend-claw-trial-router.ts
```diff
@@ -0,0 +1,392 @@
+import { adminProcedure, createTRPCRouter } from '@/lib/trpc/init';
```
evanjacobson, line 1: New tRPC router (~390 lines, all additions). Exposes two procedures under `admin.extendClawTrial`: `matchUsers` (query) resolves a list of emails to user + subscription records in bulk and applies an `at_limit` synthetic status when a trial would already extend beyond 1 year from now; `extendTrials` (mutation) does the actual update — extending `trial_ends_at` for active trials, or inserting a fresh subscription row for users with no trial yet, both capped at 1 year. It also clears relevant `kiloclaw_email_log` rows so trial notification emails can retrigger, and writes an admin audit log entry per batch.

src/routers/admin/extend-claw-trial-router.test.ts
```diff
@@ -0,0 +1,151 @@
+import { describe, expect, it, beforeEach } from '@jest/globals';
```
evanjacobson, line 1: New test file (~150 lines). Covers the two behaviors most likely to regress: the `at_limit` ineligibility flag (trial already extends past 1 year from now should be surfaced to the admin before they submit) and the 1-year ceiling on `extendTrials` (requesting more days than the cap allows should clamp to exactly 1 year out, not silently overflow). Tests use real DB helpers — no mocks.

src/lib/admin-csv.ts
evanjacobson, line 1: This code was stripped out of bulk-credits/page.tsx, for the most part.

src/routers/admin/extend-claw-trial-router.ts
```diff
@@ -0,0 +1,414 @@
...
+        stripe_subscription_id: null,
+        stripe_schedule_id: null,
+        scheduled_plan: null,
+        scheduled_by: null,
+        cancel_at_period_end: false,
+        // payment_source and pending_conversion are intentionally left
+        // as-is: stripe_subscription_id = null is the real guard against
+        // stale Stripe webhook replays, and the billing implications of
+        // clearing these fields on a canceled row haven't been fully
+        // analyzed for every payment_source value.
```
pandemicsyn, line 198: >and the billing implications of
>// clearing these fields on a canceled row haven't been fully
>// analyzed for every payment_source value.

@evanjacobson Might be worth checking with @jeanduplessis , but im not sure what implications are for this.

For context, asked the 🤖 about this and it had this to say:

```
The canceled-subscription restart path flips the row back to status: 'trialing' and clears Stripe linkage, but intentionally preserves payment_source and pending_conversion. That leaves a resurrected trial carrying stale billing metadata from its previous paid state. Downstream logic that keys off those fields can now see a "trialing" subscription that still looks partially paid or mid-conversion, which is inconsistent with the single-user admin reset behavior this code claims to mirror.
```

Between your code comment and the robots it seems like we're not exactly sure ?

pandemicsyn (approved):

jeanduplessis:
@evanjacobson we need to update the logic of this PR.

_trialing → eligible. Trial is extended by N days from the later of the current end date or today, so expired trials extend from today rather than compounding from a past date._ ✅ **this is fine**
_canceled → eligible. The subscription is resurrected as a fresh trial: status reset to trialing, plan reset to trial, all Stripe/billing fields cleared. The email suppression log is also cleared so trial notifications fire again._ ❌  **A cancelled subscription should not be overridden. A new trial subscription should be created.**
_Active paid subscription (active, past_due, unpaid) → ineligible. We must not touch a user who is actively paying._ ✅ **this is fine**
_No subscription → ineligible. The user has never provisioned a KiloClaw instance and has no subscription row to operate on._ ❌ **A new trial subscription should be created.**

src/routers/admin/extend-claw-trial-router.ts
jeanduplessis: We should leave cancelled subscription rows alone, and create a new trial subscription.

PR #1760, opened 2026-03-31
fix(gastown): push KILOCODE_TOKEN to running process and guard remint by owner

jrf0110:
## Summary

Closes 4 KILOCODE_TOKEN propagation gaps so that long-lived agents (especially the mayor) pick up fresh tokens without a container restart:

- Extended `POST /refresh-token` on the control server to accept an optional `kilocodeToken` field (backward-compatible; existing callers sending `{ token }` still work).
- Added `KILOCODE_TOKEN` to `LIVE_ENV_KEYS` in `process-manager.ts` so model hot-swaps use the live `process.env` value instead of the stale startup snapshot.
- Added `KILOCODE_TOKEN` to `CONFIG_ENV_MAP` in the model update handler so `process.env` stays current before SDK server restart.
- New `pushKilocodeTokenToContainer()` helper persists the token via `setEnvVar` for next boot and pushes it to the running process via `/refresh-token`. Best-effort: tolerates downed containers.
- `syncConfigToContainer()` now calls `pushKilocodeTokenToContainer` so config updates propagate immediately.
- `refreshContainerToken` mutation now remints `KILOCODE_TOKEN` when the caller is the town owner and pushes it via `syncConfigToContainer`. Non-owner org members push the existing token to avoid overwriting the owner's identity.

## Verification

- Code review passed: no type-safety issues (`as`/`!`), no secrets, no build artifacts, backward-compatible endpoint changes, error handling consistent with existing patterns.

## Visual Changes

N/A

## Reviewer Notes

- The `/refresh-token` endpoint schema uses a Zod `.refine()` to require at least one of `token` or `kilocodeToken`, maintaining backward compatibility with existing callers.
- The owner guard in `refreshContainerToken` (`credentialUserId === user.id`) prevents org members from overwriting the token owner's identity — this is an intentional security boundary.

### DIFF
--- a/cloudflare-gastown/container/src/control-server.ts
+++ b/cloudflare-gastown/container/src/control-server.ts
@@ -123,13 +123,36 @@ app.post('/dashboard-context', async c => {
 // Hot-swap the container-scoped JWT on the running process. Called by
 // the TownDO alarm to push a fresh token before the current one expires.
 // Updates process.env so all subsequent API calls use the new token.
+//
+// Also accepts an optional `kilocodeToken` field to hot-swap the
+// KILOCODE_TOKEN used by the SDK / LLM gateway. When provided,
+// process.env.KILOCODE_TOKEN is updated so long-lived agents pick up
+// the fresh value without a container restart.
 app.post('/refresh-token', async c => {
   const body: unknown = await c.req.json().catch(() => null);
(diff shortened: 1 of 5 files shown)


jeanduplessis:

PR #1809, opened 2026-03-31
fix(kiloclaw): pass instanceId to worker for instance-keyed DO routing

pandemicsyn:
## Summary

After the multi-instance migration (to support org users), new KiloClaw instances use `ki_`-prefixed sandbox IDs and DOs keyed by `instanceId` (DB UUID) instead of `userId`. Several Next.js → worker call paths were not passing `?instanceId=` to the worker, causing the worker's `resolveStub()` to fall back to `idFromName(userId)` — resolving a wrong, empty DO. This produced `"No machine exists"` errors on restart and would silently return stale/empty data from other endpoints.

**What changed:**
- `KiloClawUserClient` now accepts `instanceId` in `RequestContext` and appends it as a query param to all worker requests.
- 13 call sites across 6 files updated to look up the active instance and pass `workerInstanceId()`:
  - **User-facing tRPC:** `getConfig`, `getChannelCatalog`, `getSecretCatalog`, `restartMachine`
  - **Admin tRPC:** `restartMachine`, `startKiloCliRun`, `getKiloCliRunStatus`
  - **API routes:** `/api/kiloclaw/status`, `/api/kiloclaw/chat-credentials`
  - **Billing lifecycle cron:** all 4 sweeps (trial expiry, subscription expiry, instance destruction, past-due cleanup) — added `leftJoin` on `kiloclaw_instances` to get `sandbox_id`
  - **Instance lifecycle:** `autoResumeIfSuspended`

Legacy instances (non-`ki_` prefix) are unaffected — `workerInstanceId()` returns `undefined` for them, preserving the existing userId-keyed DO resolution.

## Verification

- [x] `pnpm typecheck` — passes (only pre-existing `csv-parse/sync` error in unrelated migration script)
- [x] `pnpm format` — passes
- [x] `pnpm lint` — passes for changed files (pre-existing lint errors in `kilo-app/` only)
- [x] Confirmed via Axiom logs that the error is `{"success":false,"error":"No machine exists"}` for both the reported user (`f2f72af0-...`) and test instance (`f1a848ca-...`), despite DOs having valid `flyMachineId`

## Visual Changes

N/A

## Reviewer Notes

- The worker side already accepts `?instanceId=` on all routes — no worker changes needed.
- The `workerInstanceId()` helper is the existing bridge function that returns the DB UUID for `ki_`-prefixed instances and `undefined` for legacy ones. All fixed call sites follow the same pattern already used by `KiloClawInternalClient` callers (e.g., `getStatus`, `start`, `stop`).
- The billing cron sweeps add a `leftJoin` on `kiloclaw_instances` — this is safe because the `instance_id` guard (`if (row.instance_id)`) already exists before each worker call, and the join is on the PK.
- Risk: low. All changes are additive — passing an extra query param that the worker already handles. Legacy instances pass `undefined` and behave identically to before.

### DIFF
--- a/src/lib/kiloclaw/kiloclaw-user-client.ts
+++ b/src/lib/kiloclaw/kiloclaw-user-client.ts
@@ -33,7 +33,7 @@ export class KiloClawUserClient {
 
   private async request<T>(path: string, options?: RequestInit, ctx?: RequestContext): Promise<T> {
     const url = ctx?.instanceId
-      ? `${this.baseUrl}${path}?instanceId=${ctx.instanceId}`
+      ? `${this.baseUrl}${path}?instanceId=${encodeURIComponent(ctx.instanceId)}`
       : `${this.baseUrl}${path}`;
     const res = await fetch(url, {
       ...options,
--- a/src/lib/kiloclaw/kiloclaw-user-client.ts
+++ b/src/lib/kiloclaw/kiloclaw-user-client.ts
@@ -9,11 +9,15 @@ import type {
   ChatCredentials,
 } from './types';
 
-type RequestContext = { userId: string };
+type RequestContext = { userId: string; instanceId?: string };
 
 /**
  * KiloClaw worker client for user-facing routes.
  * Uses Bearer JWT auth (forwarding the user's token). Server-only.
+ *
+ * When `instanceId` is provided in RequestContext, it is appended as a
+ * query parameter so the worker resolves the correct DO (instance-keyed
+ * vs legacy userId-keyed).
  */
 export class KiloClawUserClient {
   private authToken: string;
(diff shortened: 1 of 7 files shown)


pandemicsyn:
Also fixes https://kilo-code.sentry.io/issues/7378161324/?project=4509565130637312&query=is%3Aunresolved%20kiloclaw&referrer=issue-stream

src/lib/kiloclaw/kiloclaw-user-client.ts
```diff
@@ -28,7 +32,10 @@ export class KiloClawUserClient {
   }
 
   private async request<T>(path: string, options?: RequestInit, ctx?: RequestContext): Promise<T> {
-    const res = await fetch(`${this.baseUrl}${path}`, {
+    const url = ctx?.instanceId
+      ? `${this.baseUrl}${path}?instanceId=${ctx.instanceId}`
+      : `${this.baseUrl}${path}`;
+    const res = await fetch(url, {
```
St0rmz1, line 38: nit: The internal client uses `encodeURIComponent(instanceId)` in all its methods (`start`, `stop`, `destroy`, etc.), but this path interpolates it raw. The value is always a server-generated UUID so there's no real risk, but encoding here would be consistent with the existing pattern.

St0rmz1: race condition on my comment and code update :)

St0rmz1 (approved):

PR #1829, opened 2026-04-01
fix(gastown): stop false-positive invariant violations in reconciler

jrf0110:
## Summary

- **Invariant 7**: Exclude mayors from the "working agent must have a hook" check — mayors are always `working` and intentionally have no hook, causing false-positive violations every 5s.
- **Invariant 5**: Widen the valid convoy bead states from just `open`/`closed` to include `in_progress` and `in_review` as valid transient states. Only flag truly unexpected states instead of all in-progress convoys.

## Verification

- Typecheck passes (`pnpm typecheck` via pre-push hook)
- Lint passes (oxlint via pre-push hook)
- Format check passes (oxfmt via pre-push hook)
- Manual review of SQL query changes for correctness

## Visual Changes

N/A

## Reviewer Notes

These two invariant checks were firing every reconciler tick (5s), spamming logs with false positives. The fixes are minimal SQL filter changes with clear semantics.

### DIFF
--- a/cloudflare-gastown/src/dos/town/reconciler.ts
+++ b/cloudflare-gastown/src/dos/town/reconciler.ts
@@ -1494,6 +1494,7 @@ export function checkInvariants(sql: SqlStorage): Violation[] {
   const violations: Violation[] = [];
 
   // Invariant 7: Working agents must have hooks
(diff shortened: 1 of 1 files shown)


jrf0110:
Work consolidated into the convoy landing PR — see #1862

PR #1830, opened 2026-04-01
fix(gastown): preserve org billing context across model changes

jrf0110:
## Summary

When a model hot-swap occurs via `PATCH /agents/:agentId/model`, the `KILO_CONFIG_CONTENT` env var gets rebuilt and the organization ID embedded within it could be lost, breaking billing attribution.

This adds three defense layers to ensure the org billing context survives model changes:

1. **Standalone env var** (`GASTOWN_ORGANIZATION_ID`): Set on `/agents/start` and read first in `extractOrganizationId()` before falling back to `KILO_CONFIG_CONTENT` extraction.
2. **Request body passthrough**: `organizationId` field added to `UpdateAgentModelRequest`; `Town.do.ts` resolves `townConfig.organization_id` and threads it through `container-dispatch.ts` into the PATCH body.
3. **X-Town-Config sync**: `organization_id` added to `buildContainerConfig()` output; the PATCH handler extracts it from the header and updates the env var.

**Files changed:** `control-server.ts`, `process-manager.ts`, `types.ts`, `Town.do.ts`, `config.ts`, `container-dispatch.ts`

## Verification

- [x] `pnpm typecheck` — passes (0 errors)
- [x] `pnpm lint` — passes
- [x] Code review: verified env var precedence (X-Town-Config > body > initial start), empty string handling, and fallback paths

## Visual Changes

N/A

## Reviewer Notes

- The X-Town-Config `organization_id` value writes last in the PATCH handler (after the body value), which is intentional — the header is the authoritative source while the body serves as a belt-and-suspenders fallback if header parsing fails.
- `parsed.data.organizationId ?? ''` on the `/agents/start` path means missing values produce an empty string, which `extractOrganizationId()` treats as falsy, correctly falling through to the `KILO_CONFIG_CONTENT` legacy extraction.

### DIFF
--- a/cloudflare-gastown/container/src/control-server.ts
+++ b/cloudflare-gastown/container/src/control-server.ts
@@ -142,6 +142,11 @@ app.post('/agents/start', async c => {
     return c.json({ error: 'Invalid request body', issues: parsed.error.issues }, 400);
   }
 
+  // Persist the organization ID as a standalone env var so it survives
+  // config rebuilds (e.g. model hot-swap). The env var is the primary
+  // source of truth; KILO_CONFIG_CONTENT extraction is the fallback.
+  process.env.GASTOWN_ORGANIZATION_ID = parsed.data.organizationId ?? '';
+
   console.log(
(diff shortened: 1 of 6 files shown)


jrf0110:
Work consolidated into the convoy landing PR — see #1862

PR #1891, opened 2026-04-02
feat(proxy): add cache diagnostic logging for Anthropic requests

alex-alecu:
## Summary

- Add pre-request cache diagnostics in `src/lib/providers/cache-debug.ts` and wire them into `src/app/api/openrouter/[...path]/route.ts` after provider-specific request mutations, so Anthropic `chat_completions` requests with tools log breakpoint placement, prompt cache key presence, prefix hash, and serialized body shape for the exact payload sent upstream.
- Add post-response cache diagnostics in `src/lib/processUsage.ts` so the same Anthropic tool requests log cache hit/write tokens, inference provider, and cost reconciliation from generation lookup or inline fallback.
- Add the production-shaped cache repro harness in `src/scripts/openrouter/test-cache-diag.ts` and fix the `server-only` shim in `src/scripts/index.ts`, which let us validate the actual provider behavior locally: default routing resolves to Amazon Bedrock, Bedrock hits on the first cache-bearing follow-up, and explicit Anthropic misses that first follow-up before hitting on later turns.

## Verification

- [x] `curl -s -o /dev/null -w "%{http_code}" http://localhost:3000/users/sign_in` -> `200`
- [x] `pnpm script:run openrouter test-cache-diag "cache-diag-20260402161604@example.com"` -> pass; default routing used Amazon Bedrock with cache hit rates `0%, 75%, 96%, 84%, 97%, 90%`
- [x] `pnpm script:run openrouter test-cache-diag "cache-diag-20260402161604@example.com" anthropic/claude-opus-4.6 compare` -> pass; preferred `amazon-bedrock` matched default, preferred `anthropic` showed `0%, 0%, 96%, 84%, 97%, 90%`
- [x] Reviewed `/private/tmp/kilo-dev-server.log` for `[CacheDiag]` and `[CacheDiag:response]` entries -> `promptCacheKey: true` on all requests; default routing and preferred Bedrock logged `inferenceProvider:"Amazon Bedrock"`; preferred Anthropic missed the first cache-bearing follow-up and then hit on subsequent turns

## Visual Changes

N/A

## Reviewer Notes

- Logging is gated to Anthropic `chat_completions` requests with tools and wrapped so diagnostics never break request handling.
- The new harness uses a realistic system prompt + tool fixture from `src/lib/utils/testdata/no_tool_request.json` and a tool-result follow-up loop, which matches the production cache breakpoint shape better than the older `tool_choice: \"none\"` false-negative path.
- The main reviewer question is whether the current provider-specific behavior is acceptable: Bedrock hits immediately, while Anthropic appears to miss the first cache-bearing follow-up but behaves normally after that.

### DIFF
--- a/src/app/api/openrouter/[...path]/route.ts
+++ b/src/app/api/openrouter/[...path]/route.ts
@@ -86,6 +86,7 @@ import {
   getMaxTokens,
   hasMiddleOutTransform,
 } from '@/lib/providers/openrouter/request-helpers';
+import { logCacheDiagnostics } from '@/lib/providers/cache-debug';
 
 export const maxDuration = 800;
 
@@ -480,6 +481,8 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
     extraHeaders,
     userByok
   );
+
+  logCacheDiagnostics(requestBodyParsed, originalModelIdLowerCased, taskId ?? null);
 
   let response: Response;
   if (requestBodyParsed.kind === 'chat_completions' && provider.id === 'martian') {
--- a/src/lib/processUsage.ts
+++ b/src/lib/processUsage.ts
@@ -820,6 +820,58 @@ async function processTokenData(
       );
     }
     usageStats = genStats;
+
+    // Log cache token reconciliation for Anthropic models with tools
+    if (
+      usageContext.api_kind === 'chat_completions' &&
+      usageContext.has_tools &&
+      usageContext.requested_model.startsWith('anthropic/')
+    ) {
+      console.log(
+        `[CacheDiag:response]`,
+        JSON.stringify({
+          sessionId: usageContext.session_id,
+          model: usageStats.model,
+          source: 'generation',
+          messageId: usageStats.messageId,
+          upstreamId: usageStats.upstream_id,
+          inputTokens: usageStats.inputTokens,
+          cacheHitTokens: usageStats.cacheHitTokens,
+          cacheWriteTokens: usageStats.cacheWriteTokens,
+          outputTokens: usageStats.outputTokens,
+          cost_mUsd: usageStats.cost_mUsd,
+          cacheDiscount_mUsd: usageStats.cacheDiscount_mUsd,
+          inferenceProvider: usageStats.inference_provider,
+        })
+      );
+    }
+  }
+
+  // Log inline-only usage for Anthropic models with tools (no generation data)
+  if (
+    !generation &&
+    usageContext.api_kind === 'chat_completions' &&
+    usageContext.has_tools &&
+    usageContext.requested_model.startsWith('anthropic/')
+  ) {
+    console.log(
+      `[CacheDiag:response]`,
+      JSON.stringify({
+        sessionId: usageContext.session_id,
+        model: usageStats.model,
+        source: 'inline',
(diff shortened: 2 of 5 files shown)


alex-alecu:
## Cache diagnostic test results (2026-04-02)

### Test 1: Gateway proxy — Bedrock vs Anthropic direct

Ran `test-cache-diag` with `compare` mode through our gateway. Realistic system prompt + 8 tool definitions, 3 turns.

**Amazon Bedrock** (`provider.only: ['amazon-bedrock']`):
| Turn | Msgs | Input | Cache Write | Cache Read | Hit Rate |
|------|------|-------|-------------|------------|----------|
| 1 | 2 | 2,794 | 0 | 0 | 0% |
| 2 | 4 | 2,862 | 0 | 0 | 0% |
| 3 | 6 | 2,928 | 0 | 0 | 0% |

**Anthropic direct** (`provider.only: ['anthropic']`):
| Turn | Msgs | Input | Cache Write | Cache Read | Hit Rate |
|------|------|-------|-------------|------------|----------|
| 1 | 2 | 2,794 | 0 | 0 | 0% |
| 2 | 4 | 2,862 | 0 | 0 | 0% |
| 3 | 6 | 2,928 | 0 | 0 | 0% |

Both providers return identical zero cache tokens.

### Test 2: Direct OpenRouter API (bypassing our gateway)

Sent `cache_control: { type: 'ephemeral' }` on the system message content part directly to `https://openrouter.ai/api/v1/chat/completions`, no gateway involved:

| Test | Input tokens | Provider | Cache Write | Cache Read |
|------|-------------|----------|-------------|------------|
| Turn 1 (4,297 tokens, system with `cache_control`) | 4,297 | Anthropic | 0 | 0 |
| Turn 2 (same prefix + new user msg) | 4,309 | Anthropic | 0 | 0 |

4,297 tokens is well above the 1,024 minimum for Opus models (and above 4,096 in case Opus 4.6 has a higher threshold like Opus 4.5).

### Test 3: OpenRouter Messages API (`/api/v1/messages`)

Tested with top-level `cache_control: { type: 'ephemeral' }` through the Messages API:

| Test | Input tokens | Provider | Cache Creation | Cache Read |
|------|-------------|----------|----------------|------------|
| Turn 1 | 2,111 | Anthropic | 0 | 0 |
| Turn 2 (same prefix) | 2,123 | Anthropic | 0 | 0 |

### Conclusion

**OpenRouter does not forward `cache_control` annotations to upstream providers in either the Chat Completions or Messages API formats.** This was tested:
- Through our gateway and directly against OpenRouter's API
- With both per-block `cache_control` (on content parts) and top-level `cache_control`
- With both Bedrock and Anthropic direct routing
- With prompts from 1,531 to 4,297 input tokens

The zero cache tokens are NOT caused by our gateway code — `addCacheBreakpoints()` correctly sets the breakpoints, but they are stripped or ignored by OpenRouter before reaching the upstream provider.

**Open question:** If production sessions show cache savings, they must come from a different mechanism (e.g. OpenRouter's internal `prompt_cache_key` field, `session_id` sticky routing, or Vercel AI SDK provider metadata path). Further investigation needed to identify which production sessions have non-zero `cache_hit_tokens` and what API path/provider they used.

jeanduplessis:

PR #1933, opened 2026-04-02
Add variants for Qwen3.6, Seed 2.0 Pro

chrarnoldus:


src/lib/providers/model-settings.ts
```diff
@@ -80,9 +80,22 @@ export function getModelVariants(model: string): OpenCodeSettings['variants'] {
...
+    isMoonshotModel(model) ||
+    isZaiModel(model) ||
+    isXiaomiModel(model) ||
+    model === QWEN36_PLUS_FREE_MODEL_ID
+  ) {
     return REASONING_VARIANTS_BINARY;
   }
+  if (model === seed_20_pro_free_model.public_id) {
+    return {
+      none: { reasoning: { enabled: false, effort: 'minimal' } },
```
chrarnoldus, line 93: according to their docs false is equivalent to minimal (none is an error)

markijbema (approved):

PR #1954, opened 2026-04-03
feat(kiloclaw): display volume disk usage in admin Instance Information card

evanjacobson:
## Summary

This PR adds volume usage stats which are updated at each checkin. The implementation follows the path of other checkin information.

This feature is only available to machines running this image or later. Machines without info will display `—`.

---

- Reads disk usage from the Fly volume on each controller checkin (every ~5 min) via `df -B1 --output=avail,size /`
- Persists `diskUsedBytes` / `diskTotalBytes` in the KiloClawInstance Durable Object and surfaces them through the existing `getDebugState` / `debug-status` platform API
- Displays **Volume Usage** (used / total) in the Instance Information card of the admin KiloClaw panel

## Verification

- `pnpm typecheck` — pass
- `pnpm lint` — pass
- Format check — pass (fixed in follow-up formatting commit)
- Manual: tested E2E
 
## Visual Changes

image

## Reviewer Notes


kiloclaw/controller/src/checkin.ts
```diff
@@ -130,7 +166,7 @@ export function startCheckin(deps: CheckinDeps): () => void {
 
       const stats = deps.getSupervisorStats();
       const openclawVersion = await deps.getOpenclawVersion();
-      const currentNetStats = await readNetStats();
+      const [currentNetStats, diskStats] = await Promise.all([readNetStats(), readDiskStats()]);
```
evanjacobson, line 169: Calling these in parallel provides a small performance boost for the process vs adding a separate await. I went back and forth about this but the rationale seems sane.

kiloclaw/controller/src/checkin.ts
```diff
@@ -84,6 +90,36 @@ export async function readNetStats(): Promise<NetStats> {
...
 
+/**
+ * Parse the output of `df -B1 --output=avail,size /`. Returns null if unparseable.
+ * Columns: avail (available bytes), size (total bytes). Matches the column order used
+ * by cloud-agent/src/workspace.ts and cloud-agent-next/src/workspace.ts.
+ * usedBytes is derived as totalBytes - availableBytes (df does not report used directly).
+ */
+export function parseDfOutput(raw: string): DiskStats {
+  const lines = raw.trim().split('\n');
+  const dataLine = lines[lines.length - 1]?.trim();
```
St0rmz1, line 101: **Low:** Negative `usedBytes` would reject the entire checkin

`parseDfOutput` computes `usedBytes = totalBytes - availableBytes`. If a filesystem ever reports `avail > size` (some do this with compression or reserved block edge cases), `usedBytes` goes negative. The `CheckinSchema` has `.min(0)` on `diskUsedBytes`, so Zod would reject the entire checkin payload, dropping all stats (load avg, bandwidth, restarts, telemetry) along with the disk data.

Very unlikely on Fly's ext4 volumes, but the blast radius is worth noting. A simple fix would be clamping: `Math.max(0, totalBytes - availableBytes)`, or returning `null` when `avail > total`.

St0rmz1 (approved):

PR #1967, opened 2026-04-03
fix(gateway): guard non-array tools in api metrics

kirillk:
## Summary
- guard API metrics tool extraction against truthy non-array `tools` payloads so `/api/gateway/chat/completions` does not throw before reaching the upstream provider
- add regression coverage for object-shaped `tools` payloads while preserving existing array-based extraction behavior
- keep the fix scoped to observability preprocessing, matching the incident where the crash happened in `getToolsAvailable`

## Validation
- `pnpm lint src/lib/o11y/api-metrics.server.ts src/lib/o11y/api-metrics.server.test.ts`
- `pnpm test -- --runInBand src/lib/o11y/api-metrics.server.test.ts`


jeanduplessis:

PR #1971, opened 2026-04-03
remove default allowlist configuration

evanjacobson:
## Summary
Removes default allowlist from config writer.


evanjacobson:
fixing tests now

pandemicsyn (approved):

PR #1998, opened 2026-04-04
fix(gastown): load ClipboardAddon so Mayor OSC 52 clipboard writes work

Githubguy132010:
## Summary

The xterm.js terminal in the Gastown component was missing the `@xterm/addon-clipboard` addon, which is required to handle [OSC 52](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands) escape sequences for clipboard writes. Applications running inside the terminal (such as Mayor) use OSC 52 to write text to the system clipboard. Without this addon, those clipboard write requests were silently dropped.

This fix:
- Adds `@xterm/addon-clipboard` as a dependency
- Lazily imports and instantiates `ClipboardAddon` alongside the existing xterm addons
- Loads the addon into the terminal so OSC 52 clipboard operations are handled correctly

No architectural changes; purely additive.

## Verification

- [ ] Manually tested: copied text in the Mayor terminal and confirmed it reached the system clipboard
- [ ] TypeScript compiled without errors (`pnpm typecheck`)
- [ ] No regressions in existing terminal behaviour

## Visual Changes

N/A

## Reviewer Notes

The `ClipboardAddon` is loaded lazily in the same `Promise.all` block as the other addons, keeping the SSR-safe lazy-loading pattern intact. The addon has no configuration options; default behaviour is to read/write the system clipboard via the Clipboard API, falling back gracefully when permissions are denied.

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,6 +101,7 @@
     "@vercel/functions": "^3.4.3",
     "@vercel/otel": "^2.1.1",
     "@workos-inc/node": "catalog:",
+    "@xterm/addon-clipboard": "^0.2.0",
     "@xterm/addon-fit": "^0.11.0",
     "@xterm/addon-web-links": "^0.12.0",
     "@xterm/xterm": "^6.0.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -282,6 +282,9 @@ importers:
       '@workos-inc/node':
         specifier: 'catalog:'
         version: 8.9.0
+      '@xterm/addon-clipboard':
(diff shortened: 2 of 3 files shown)


jrf0110:
Thanks @Githubguy132010 for the fix! Because this PR was opened from a fork, our automation can't push to it to resolve the conflicts with `gastown-staging`. I've re-applied the same change on a branch in this repository to get it landed — see #2968. Once that merges, this PR can be closed. Full credit preserved via co-authored-by trailer on the commit.

PR #2049, opened 2026-04-06
chore(dev): restructure dev tooling and docs for monorepo

jeanduplessis:
## Summary

- Restructure dev tooling and documentation for the monorepo migration: move `apps/web/dev/` → `dev/`, `plans/` → `.plans/`, consolidate `.kilocode/` into `.kilo/`, delete stale prototype HTML
- Fix all service paths in `dev/local/services.ts`, shell scripts, and tunnel scripts to use `services/<name>` directories instead of the old repo-root `cloudflare-<name>` layout
- Fix Next.js startup: run from `apps/web/` with `pnpm run dev`, symlink root `.env.local` into `apps/web/` so Next.js can discover env vars from its CWD
- Rewrite `README.md` (219 → 38 lines) as a concise monorepo overview; fix stale commands/paths in `DEVELOPMENT.md` (`pnpm dev` → `dev:start`, schema path, `script:run`, `validate` description)
- Reorganize `.gitignore` by concern, remove duplicates, add `.plans/` and `.kilocode/`
- Migrate `.kilocode/` rules/workflows/config into canonical `.kilo/` directory with stale path fixes; add `kilo.json` project config
- Upgrade pnpm from 10.27.0 to 10.33.0; add `react` and `ink` to root devDeps (needed by `dev/local/dashboard.tsx`)
- Add `.dev.vars` backfill from template and `FLY_IMAGE_CONTENT_HASH` seeding to kiloclaw `dev-start.sh`

## Verification

- [x] `pnpm dev:env --yes` — synced 13 `.dev.vars` files to correct `services/` paths
- [x] `pnpm dev:start` — tmux dashboard renders, postgres/redis/nextjs all green, HTTP 200 from `localhost:3000/users/sign_in`
- [x] `pnpm dev:stop` — killed tmux session + docker containers cleanly
- [x] `pnpm install` — no errors, lockfile updated
- [x] `pnpm format:check` + `pnpm lint` + `pnpm typecheck` — passed (pre-push hook)
- [ ] `pnpm validate` (full suite including tests) — not run in this session

## Visual Changes

N/A

## Reviewer Notes

- The 5 commits are logically ordered but could be squashed — each builds on the prior
- `scripts/dev.sh` creates a symlink `apps/web/.env.local → ../../.env.local` on first run; the symlink is gitignored via the `.env*.local*` pattern. Verify this works on your machine if you already have a manual `apps/web/.env.local`
- The `dev/auto-fix/` and `dev/review/` shell scripts had a `CLOUD_DIR` that resolved to `dev/` instead of repo root — now fixed to `../../`
- kiloclaw `set_dev_var` uses `sed -i ''` (macOS compatible); Linux contributors would need `sed -i` without the empty string — this matches the existing script convention


.kilocode/skills/chat-sdk
RSO, line 1: Did the skills get removed knowingly?

.kilocode/config.json
RSO, line 1: Does this mean we don't want managed indexing anymore?

.kilocode/mcp.json
RSO, line 1: No more MCPs?

apps/web/package.json
```diff
@@ -25,10 +25,7 @@
     "test:e2e:ui": "playwright test --ui",
     "test:e2e:debug": "playwright test --debug",
     "promo": "pnpm -s script src/scripts/encrypt-promo-codes.ts",
-    "dev:discord-gateway-cron": "tsx dev/discord-gateway-cron.ts",
-    "dev:start": "tsx --tsconfig tsconfig.scripts.json dev/local/cli.ts up",
-    "dev:stop": "tsx --tsconfig tsconfig.scripts.json dev/local/cli.ts stop",
-    "dev:env": "tsx --tsconfig tsconfig.scripts.json dev/local/cli.ts env"
+    "dev:discord-gateway-cron": "tsx ../../dev/discord-gateway-cron.ts"
```
RSO: Since it's in `../../dev`, wouldn't it make more sense to move this script to the top-level as well?

kilo.json
```diff
@@ -0,0 +1,4 @@
+{
+  "$schema": "https://opencode.ai/config.json",
```
RSO:
```suggestion
  "$schema": "https://app.kilo.ai/config.json",
```

dev/local/services.ts
```diff
@@ -214,10 +248,10 @@ function buildServiceDefs(): ServiceDef[] {
       defs.push({
         name,
         type: 'nextjs',
-        dir: '.',
+        dir: 'apps/web',
         port: 3000 + portOffset,
         dependsOn: meta.dependsOn,
-        command: ['pnpm', 'next', 'dev'],
+        command: ['pnpm', 'run', 'dev'],
```
RSO, line 254: Please note that this change now means that we use the `dev.sh` script instead of calling `next dev` directly. I'm not sure if @eshurakov had a specific reason for setting it to `next dev` instead of `dev`

RSO (approved):
Please address the comments, but overall the changes make sense

.kilocode/skills/chat-sdk
jeanduplessis: Yes its deliberate. Kilo CLI already reads from .agents by default so there's no need for additional symlinks.

.kilocode/config.json
jeanduplessis: Good catch — restored `.kilocode/config.json` with `managedIndexingEnabled: true`. This is a VS Code extension setting that doesn't belong in `kilo.json` (different schema), so it stays in `.kilocode/`. Added a `!.kilocode/config.json` exception to `.gitignore`.

.kilocode/mcp.json
jeanduplessis: The only MCP server (context7) was already `disabled: true`. MCP config can be added to `kilo.json` under the `mcp` key if anyone wants to re-enable it — just didn't carry over a disabled entry.

.kilocode/config.json
jeanduplessis: Restored `.kilocode/config.json` as a local-only file (gitignored) with `managedIndexingEnabled: true`. This setting is specific to the Kilo Code VS Code extension — the Kilo CLI `kilo.json` schema doesn't have a `project` key, so it can't live there.

.kilocode/mcp.json
jeanduplessis: The only MCP server was context7 and it was `disabled: true`. MCP config can be added to `kilo.json` under the `mcp` key if anyone wants to re-enable it — just didn't carry over a disabled entry.

dev/local/services.ts
jeanduplessis: The change to `pnpm run dev` (which runs `scripts/dev.sh`) was necessary because Next.js loads `.env.local` from CWD. When running from `apps/web/` instead of the repo root, env vars are missing unless we symlink them — which `dev.sh` now handles. It also handles port discovery, writes `.dev-port`, and sets `NEXTAUTH_URL`. Running `pnpm next dev` directly would skip all of that.

PR #2075, opened 2026-04-06
feat(agentcard): B2B integration with org API keys and REST API client

keyserfaty:
## Summary

Replaces the personal JWT-based AgentCard integration with a proper B2B pattern using org-scoped API keys.

**What changed:**
- **New REST API client** (`apps/web/src/lib/agentcard/client.ts`) — typed TypeScript client wrapping all AgentCard endpoints for managing cardholders and cards on behalf of your users
- **Updated secret catalog** — validation now accepts `sk_test_*`/`sk_live_*` API keys instead of JWTs
- **Rewritten setup guide** — admin-level instructions using `agent-cards-admin` CLI instead of asking end users to copy JWTs

**Why:** The previous integration asked each end user to create their own AgentCard account and paste a personal JWT. The correct B2B pattern is: Kilo creates one organization + API key, then manages cardholders and cards on behalf of users via the REST API.

### 1. Install the AgentCard admin CLI

```bash
npm install -g agent-cards-admin
```

### 2. Login

```bash
agent-cards-admin login
# Opens browser for magic-link email auth
```

### 3. Create an organization

```bash
agent-cards-admin orgs create --name "Kilo" --email billing@kilo.com
```

Save the org ID from the output (e.g. `org_abc123`).

### 4. Create an API key

```bash
agent-cards-admin keys create --org <org-id> --name "kiloclaw-production"
```

Save the API key — it starts with `sk_test_` (sandbox) or `sk_live_` (production) and **won't be shown again**.

> **Tip:** Use `sk_test_` keys during development. Sandbox mode uses Stripe's test environment — no real charges. Switch to `sk_live_` for production.

### 5. Configure in KiloClaw

Set the `sk_test_...` key in **Settings > Payments > AgentCard**.

## API Client Usage

The new REST API client (`apps/web/src/lib/agentcard/client.ts`) provides typed functions for the full card lifecycle:

```typescript
import {
  createCardholder,
  setupPaymentMethod,
  getPaymentMethodStatus,
  createCard,
  getCardDetails,
  listCards,
  closeCard,
} from '@/lib/agentcard/client';

// 1. Create a cardholder for a user
const cardholder = await createCardholder({
  firstName: 'Jane',
  lastName: 'Smith',
  dateOfBirth: '1990-03-15',
  phoneNumber: '+14155551234',
  email: 'jane@example.com',
});

// 2. Set up payment method (returns Stripe checkout URL)
const { checkoutUrl } = await setupPaymentMethod(cardholder.id);
// Redirect user to checkoutUrl...

// 3. Verify payment method is attached
const { hasPaymentMethod } = await getPaymentMethodStatus(cardholder.id);

// 4. Create a $25 virtual card
const card = await createCard(cardholder.id, 2500);

// 5. Get full card credentials (PAN, CVV, expiry)
const details = await getCardDetails(card.id);

// 6. Close card when done
await closeCard(card.id);
```

The client auto-detects sandbox vs production from the API key prefix.

## Integration Guide

Full API reference with all endpoints, error codes, webhooks, and limits:
**https://docs.agentcard.sh/integration-guide**

## Next Steps (for Kilo team)

The API client provides all building blocks. Suggested integration points:

1. **tRPC procedures** — Add mutations in `kiloclaw-router.ts` that call the API client to create cardholders, set up payment methods, and issue cards
2. **User mapping** — Store the AgentCard `cardholderId` per Kilo user (e.g. in DO state or database)
3. **UI components** — Add card management UI for users to create/view/close cards
4. **MCP integration** — The existing mcporter config continues to work with the API key (AgentCard MCP server now accepts API keys in addition to JWTs)

## Test plan

- [ ] Verify secret catalog validation accepts `sk_test_abc123` and rejects `eyJabc...`
- [ ] Verify setup guide dialog renders correctly with CLI instructions
- [ ] Create an API key via `agent-cards-admin` and configure in settings
- [ ] Test API client against sandbox: create cardholder → setup payment → create card → get details → close card

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/web/src/lib/agentcard/client.ts
+++ b/apps/web/src/lib/agentcard/client.ts
@@ -1,0 +1,235 @@
+import 'server-only';
+
+// ---------------------------------------------------------------------------
+// AgentCard REST API client for B2B integration
+//
+// Uses an org-scoped API key (sk_test_* / sk_live_*) to manage cardholders
+// and virtual cards on behalf of users.
+//
+// Integration guide: https://docs.agentcard.sh/integration-guide
+// ---------------------------------------------------------------------------
+
+const AGENTCARD_API_KEY = process.env.AGENTCARD_API_KEY ?? '';
+
+/** Resolve base URL from key prefix (sandbox for sk_test_, production for sk_live_). */
+function baseUrl(): string {
+  if (AGENTCARD_API_KEY.startsWith('sk_test_')) {
+    return 'https://sandbox.api.agentcard.sh';
+  }
+  return 'https://api.agentcard.sh';
+}
+
+// ---------------------------------------------------------------------------
+// Error
+// ---------------------------------------------------------------------------
+
+export class AgentCardApiError extends Error {
+  readonly statusCode: number;
+  readonly responseBody: string;
+
+  constructor(statusCode: number, responseBody = '') {
+    super(`AgentCard API error (${statusCode}): ${responseBody}`);
+    this.name = 'AgentCardApiError';
+    this.statusCode = statusCode;
+    this.responseBody = responseBody;
+  }
+}
+
+// ---------------------------------------------------------------------------
+// Types
+// ---------------------------------------------------------------------------
+
+export interface Cardholder {
+  id: string;
+  firstName: string;
+  lastName: string;
+  email?: string;
+  dateOfBirth: string;
+  phoneNumber: string;
+  stripeCardholderId: string;
+  createdAt: string;
+  updatedAt: string;
+}
+
+export interface CreateCardholderInput {
+  firstName: string;
+  lastName: string;
+  dateOfBirth: string; // ISO date, e.g. "1990-03-15"
+  phoneNumber: string;
+  email?: string;
+}
+
+export interface PaymentMethodSetupResponse {
+  checkoutUrl: string;
+  stripeSessionId: string;
+}
+
+export interface PaymentMethodStatusResponse {
+  hasPaymentMethod: boolean;
+  paymentMethodId?: string;
+}
+
+export interface Card {
+  id: string;
+  cardholderId: string;
+  last4: string;
+  expiry: string;
+  spendLimitCents: number;
+  balanceCents: number;
+  status: 'OPEN' | 'IN_USE' | 'CLOSED' | 'PAUSED';
+  createdAt: string;
+}
+
+export interface CardDetails extends Card {
+  pan: string;
+  cvv: string;
+}
+
+export interface CardListResponse {
+  cards: Card[];
+  total: number;
+  limit: number;
+  offset: number;
+}
+
+export interface CardListFilters {
+  status?: 'OPEN' | 'IN_USE' | 'CLOSED' | 'PAUSED';
+  cardholderId?: string;
+  limit?: number;
+  offset?: number;
+}
+
+// ---------------------------------------------------------------------------
+// HTTP helper
+// ---------------------------------------------------------------------------
+
+async function request<T>(
+  method: string,
+  path: string,
+  body?: unknown,
+): Promise<T> {
+  if (!AGENTCARD_API_KEY) {
+    throw new Error('AGENTCARD_API_KEY is not configured');
+  }
+
+  const url = `${baseUrl()}${path}`;
+  const headers: Record<string, string> = {
+    Authorization: `Bearer ${AGENTCARD_API_KEY}`,
+    'Content-Type': 'application/json',
+  };
+
+  const res = await fetch(url, {
+    method,
+    headers,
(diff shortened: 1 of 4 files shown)


packages/kiloclaw-secret-catalog/src/catalog.ts
```diff
@@ -170,7 +170,7 @@ const SECRET_CATALOG_RAW = [
         validationPattern: '^ops_[A-Za-z0-9_\\-]{50,1500}$',
         validationMessage:
           '1Password service account tokens start with ops_ followed by a long base64-encoded string.',
-        maxLength: 2000,
+        maxLength: 200,
```
St0rmz1, line 173: The `maxLength` for the 1Password service account token was changed from 2000 to 200, but the `validationPattern` still allows tokens up to ~1504 characters (`{50,1500}`). 1Password service account tokens are routinely 500+ characters long. Both the UI input and backend validation enforce `maxLength`, so valid tokens will be rejected at paste time with no useful error. This looks like an accidental edit; should be reverted to `maxLength: 2000`.

apps/web/src/lib/agentcard/client.ts
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// Cardholder operations
+// ---------------------------------------------------------------------------
+
+/** Create a cardholder record for a user. */
+export async function createCardholder(
+  input: CreateCardholderInput,
+): Promise<Cardholder> {
```
St0rmz1, line 142: The ID parameters (`id`, `cardholderId`, `cardId`) are interpolated directly into URL paths with no validation. If a future tRPC procedure passes user-controlled input without sanitizing first, a value like `../../admin/something` would alter the request path while the org API key is still attached. Consider adding a guard that rejects IDs containing `/`, `\`, or `..` to prevent path traversal. A simple check in `request()` or a shared `assertSafeId()` helper would be sufficient.

apps/web/src/lib/agentcard/client.ts
```diff
@@ -0,0 +1,235 @@
+import 'server-only';
```
St0rmz1, line 1: This is 235 lines of new code for a payment integration with no unit tests. The test plan covers manual sandbox testing, but automated tests (mocking `fetch` to verify URL construction, auth headers, error handling, and the empty API key guard) would catch regressions from future refactors. Worth adding given this handles card creation, credential retrieval, and fund authorization.

apps/web/src/lib/agentcard/client.ts
```diff
@@ -0,0 +1,235 @@
...
+// ---------------------------------------------------------------------------
+
+async function request<T>(
+  method: string,
+  path: string,
+  body?: unknown,
+): Promise<T> {
+  if (!AGENTCARD_API_KEY) {
+    throw new Error('AGENTCARD_API_KEY is not configured');
+  }
```
St0rmz1, line 113: The `fetch()` call has no `AbortSignal` timeout. If the AgentCard API hangs, this request will wait indefinitely, which can block user facing operations and exhaust connection pools in a server context. Consider adding `signal: AbortSignal.timeout(30_000)` (or an appropriate duration) to the fetch options.

apps/web/src/lib/agentcard/client.ts
```diff
@@ -0,0 +1,235 @@
...
+// ---------------------------------------------------------------------------
+// AgentCard REST API client for B2B integration
+//
+// Uses an org-scoped API key (sk_test_* / sk_live_*) to manage cardholders
+// and virtual cards on behalf of users.
+//
+// Integration guide: https://docs.agentcard.sh/integration-guide
+// ---------------------------------------------------------------------------
+
+const AGENTCARD_API_KEY = process.env.AGENTCARD_API_KEY ?? '';
```
St0rmz1, line 12: `AGENTCARD_API_KEY` is captured once at module import time. The `request()` function checks for an empty key, but `baseUrl()` does not; if the key is missing, it silently defaults to the production URL. Consider reading from `process.env` inside `request()` and `baseUrl()`, or have `baseUrl()` also guard against the empty key case.

apps/web/src/lib/agentcard/client.ts
```diff
@@ -0,0 +1,235 @@
...
+
+// ---------------------------------------------------------------------------
+// HTTP helper
+// ---------------------------------------------------------------------------
+
+async function request<T>(
+  method: string,
+  path: string,
+  body?: unknown,
+): Promise<T> {
```
St0rmz1, line 110: Nit: `Content-Type: application/json` is set on every request, including GETs and DELETEs with no body. Consider only setting it when `body` is present.

PR #2088, opened 2026-04-06
fix(kiloclaw): keep recovery APIs available in degraded bootstrap

pandemicsyn:
## Summary

When users borked their config bad enough, and then restarted the machine - you could get into a situation where the initial `openclaw doctor --fix` invocation would exit with a non-0 exit code. When openclaw doctor failed during this bootstrapping phase - we never brought our command and control api endpoints online. 

That meant things like the file picker or kilo recovery endpoints would not be available. 

This breaks up the bootstrapping phase. Critical things - like a failure to decrypt still prevent C&C endpoints from starting (because if the config is that broken, we have no way to auth). But if its just openclaw doctor that fails, we'll start up the C&C endpoints.

Changes:

- Split controller startup bootstrap into two phases: `bootstrapCritical` (decrypt/directories/feature flags/gateway args) and `bootstrapNonCritical` (GitHub/Linear, onboard-or-doctor, TOOLS.md, mcporter).
- Reordered startup in `startController` so Hono routes and C&C endpoints are registered before non-critical bootstrap runs, which keeps authenticated recovery APIs available when non-critical startup work fails.
- Added degraded-path test coverage for the split bootstrap phases and updated the controller spec to document critical vs non-critical degraded behavior and endpoint availability.

## Verification

- [x] `pnpm typecheck` (pass)
- [x] `pnpm lint` (pass)
- [x] `pnpm test -- controller/src/bootstrap.test.ts controller/src/bootstrap-phases.test.ts --runInBand` from `services/kiloclaw` (pass)
- [x] `pnpm format` (pass)
- [ ] Additional user-provided verification

## Visual Changes

N/A

## Reviewer Notes

- Main behavior change is startup ordering: route activation now happens before non-critical bootstrap, so doctor/onboard failures still expose `/_kilo/*` recovery endpoints while proxy traffic remains `503 Gateway not ready` when the gateway is offline. 
- Critical bootstrap/runtime-config failures remain inline-health-only degraded mode by design because auth/runtime token derivation has not completed.
- Risk area to focus on in review: lifecycle transitions (`bootstrapping -> degraded`) and ensuring gateway start is skipped on non-critical bootstrap failure.

### DIFF
--- a/services/kiloclaw/controller/src/bootstrap.test.ts
+++ b/services/kiloclaw/controller/src/bootstrap.test.ts
@@ -800,7 +800,11 @@ describe('bootstrapCritical', () => {
     const harness = fakeDeps();
 
     await expect(
-      bootstrapCritical({ KILOCLAW_ENC_FOO: 'enc:v1:bogus' }, phase => phases.push(phase), harness.deps)
+      bootstrapCritical(
+        { KILOCLAW_ENC_FOO: 'enc:v1:bogus' },
+        phase => phases.push(phase),
+        harness.deps
+      )
     ).rejects.toThrow('KILOCLAW_ENV_KEY is not set');
 
     expect(phases).toEqual(['decrypting']);
--- a/services/kiloclaw/controller/src/bootstrap.test.ts
+++ b/services/kiloclaw/controller/src/bootstrap.test.ts
@@ -815,7 +815,7 @@ describe('bootstrapCritical', () => {
 // ---- bootstrapNonCritical ----
 
 describe('bootstrapNonCritical', () => {
-  it('returns github failure and stops subsequent steps', async () => {
+  it('treats github CLI failures as best-effort and continues', async () => {
     const harness = fakeDeps();
     const phases: string[] = [];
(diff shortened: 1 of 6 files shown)


services/kiloclaw/controller/src/bootstrap.test.ts
```diff
@@ -773,6 +775,98 @@ describe('buildGatewayArgs', () => {
...
+    });
+    (harness.deps.execFileSync as ReturnType<typeof vi.fn>).mockImplementation(
+      (cmd: string, args: string[]) => {
+        if (cmd === 'openclaw' && args.includes('doctor')) {
+          throw new Error('doctor exited 1');
+        }
+        return '';
+      }
+    );
+
```
St0rmz1, line 908: The failure test only exercises the `doctor` step (position 3 in the loop). Consider adding a test where an earlier step like `github` or `linear` throws, to confirm the generic for/try/catch loop stops execution and returns `{ ok: false, phase: 'github', error: ... }` with `phases` containing only `['github']`. That would give more confidence that the sequential-stop behavior works at every position, not just mid loop.

St0rmz1 (approved):

PR #2127, opened 2026-04-07
feat(web): add Exa search API proxy endpoint

RSO:
## Summary

Adds a new API proxy at `/api/exa/[...path]` that forwards authenticated requests to [Exa's search API](https://exa.ai). This is the first step toward giving users web search capabilities.

- **Auth**: Requires sign-in via `getUserFromAuth`; unauthenticated requests get 401.
- **Allow-listed endpoints**: `/search`, `/contents`, `/findSimilar`, `/answer`, `/context`. All others return 400.
- **Cost logging**: For non-streaming responses, asynchronously extracts `costDollars.total` from the Exa response and logs it as `[exa] user=<id> path=<path> cost=$<amount>`. This gives us visibility into usage costs before building a quota system.
- **Env**: Requires `EXA_API_KEY` to be set; returns 503 if missing.

## Verification

- [x] `pnpm format` — passes (pre-push hook confirmed)
- [x] Manual test: authenticated search request → 200 with 3 results and `costDollars.total: $0.007`
- [x] Manual test: invalid path `/api/exa/badpath` → 400 with allow-list error message
- [x] Manual test: unauthenticated request (`credentials: 'omit'`) → 401 Unauthorized

## Visual Changes

N/A

## Reviewer Notes

- `EXA_API_KEY` needs to be added to Vercel environment variables before deploying. Get one from https://dashboard.exa.ai/api-keys.
- Cost logging is console-only for now. Once we decide on a quota model, this should be replaced with proper usage tracking (similar to `accountForMicrodollarUsage` for the LLM proxy).
- Streaming responses (`text/event-stream`) are passed through without cost extraction since Exa doesn't include `costDollars` in SSE chunks.

### DIFF
--- a/apps/web/src/app/api/exa/[...path]/route.test.ts
+++ b/apps/web/src/app/api/exa/[...path]/route.test.ts
@@ -1,0 +1,281 @@
+import { describe, it, expect, beforeEach } from '@jest/globals';
+import { NextResponse } from 'next/server';
+import { getUserFromAuth } from '@/lib/user.server';
+import { failureResult } from '@/lib/maybe-result';
+import type { User } from '@kilocode/db/schema';
+
+// Capture promises scheduled via next/server `after` so tests can await them.
+let afterCallbacks: (() => Promise<void>)[] = [];
+
+jest.mock('next/server', () => {
+  return {
+    ...(jest.requireActual('next/server') as Record<string, unknown>),
+    after: (fn: () => Promise<void>) => {
+      afterCallbacks.push(fn);
+    },
+  };
+});
+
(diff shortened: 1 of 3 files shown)


jeanduplessis (approved):
Directionally this looks right.
It looks like a user can make unlimited requests against Kilo's Exa account with no rate limits or cost caps. Probably something can address as part of billing integration?

RSO:
> Directionally this looks right. It looks like a user can make unlimited requests against Kilo's Exa account with no rate limits or cost caps. Probably something can address as part of billing integration?

Yeah, this will be addressed in a follow-up PR.

PR #2169, opened 2026-04-08
feat(exa): add $10/month free allowance with credit billing for overages

RSO:
## Summary

Adds billing for the Exa proxy endpoints. Each user gets $10 in free credits per month, and usage is tracked so we can recompute balances. The $10 is configurable, but only on a monthly basis.

### Storage strategy

Two tables to avoid the unbounded row growth that `microdollar_usage` suffers from:

- **`exa_monthly_usage`** — pre-aggregated counter (one row per user per month). The pre-request balance check reads this single row (O(1) forever).
- **`exa_usage_log`** — per-request audit trail, range-partitioned by month on `created_at`. Never queried in the hot path.

## Verification

- `pnpm typecheck` — passes
- `pnpm test -- apps/web/src/app/api/exa` — 23 tests pass (auth, paths, streaming stripped, allowance, balance check, cost recording)
- `pnpm test -- apps/web/src/lib/exa-usage.test.ts` — 8 tests pass (counter reads, upserts, balance deduction)
- `pnpm format` + `pnpm lint` — clean
- Test plan:
  - Created a new user
  - Made a call to the Exa API: works
  - Set the Exa `total_cost_microdollars` to `10000000`
  - Call the API again: Fail (no balance, out of free credits)
  - Add money to balance
  - Call the API again: Success ✅ 

Also made sure that the `charged_to_balance` prop is correctly set:

CleanShot 2026-04-08 at 15 20 05@2x

## Reviewer notes

I feel confident that I've tested most of the edge cases for this feature, but I'd love a second pair of eyes on the logic, as it's likely to be costly if we make a mistake with the accounting.


packages/db/src/migrations/0078_light_jocasta.sql
```diff
@@ -0,0 +1,32 @@
...
+	"kilo_user_id" text NOT NULL,
+	"organization_id" uuid,
+	"path" text NOT NULL,
+	"cost_microdollars" bigint NOT NULL,
+	"charged_to_balance" boolean DEFAULT false NOT NULL,
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
+	CONSTRAINT "exa_usage_log_id_created_at_pk" PRIMARY KEY("id","created_at")
+) PARTITION BY RANGE ("created_at");
+--> statement-breakpoint
+CREATE TABLE "exa_usage_log_2026_04" PARTITION OF "exa_usage_log"
```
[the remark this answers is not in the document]
RSO: Ha, the bot has clearly not heard of kilospeed

apps/web/next.config.mjs
```diff
@@ -45,12 +45,16 @@ const nextConfig = {
     // Uses beforeFiles to ensure the rewrite happens BEFORE filesystem routes are checked
     // See: https://nextjs.org/docs/app/api-reference/config/next-config-js/rewrites
     const globalApiRewrites =
-      process.env.GLOBAL_KILO_BACKEND !== 'true'
+      process.env.NODE_ENV === 'production' && process.env.GLOBAL_KILO_BACKEND !== 'true'
```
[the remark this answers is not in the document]
RSO: Fixed — switched to `VERCEL_ENV === 'production'` in 1cbf59bd0.

.plans/exa-monthly-allowance.md
```diff
@@ -0,0 +1,312 @@
...
+The Exa proxy (`apps/web/src/app/api/exa/[...path]/route.ts`) authenticates users but doesn't track or limit usage. Every request uses Kilo's shared Exa API key with zero cost accountability. Exa returns costs in the response body (`costDollars.total`), and we currently just `console.log` them.
+
+## Design
+
+**Free tier + overage model:**
+
+- **≤ $10/month per user:** free — no impact on Kilo balance.
+- **> $10/month per user:** charged to the user's (or their org's) Kilo credit balance, same as LLM usage.
+
+The $10 free allowance is per-user regardless of org membership. When a user exceeds it, the overage is billed to whichever entity they're acting on behalf of (personal account or org, determined by the `X-KiloCode-OrganizationId` header — same as OpenRouter).
```
markijbema: is this logical? why not $10*n / per org where n is the number of users. That's how it works for everything else right?

apps/web/src/app/api/cron/exa-partition-maintenance/route.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/**
+ * Format a Date as YYYY-MM-DD (first of month) for partition range bounds.
+ */
+function monthStart(date: Date): string {
+  const y = date.getFullYear();
+  const m = String(date.getMonth() + 1).padStart(2, '0');
+  return `${y}-${m}-01`;
+}
```
markijbema: for the two methods above, doesnt js have built in utilities to format like that? this feels clunky

apps/web/src/app/api/cron/exa-partition-maintenance/route.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  const now = new Date();
+  const created: string[] = [];
+  const errors: string[] = [];
+
+  // Create partitions for the current month and the next 2 months
+  for (let offset = 0; offset <= 2; offset++) {
+    const target = new Date(now.getFullYear(), now.getMonth() + offset, 1);
+    const nextMonth = new Date(target.getFullYear(), target.getMonth() + 1, 1);
+    const name = `exa_usage_log_${partitionSuffix(target)}`;
```
markijbema: i feel this should be the method, and i would appreciate tests for it, and maybe we want to extract this (because surely we reconstruct this name elsewhere and write to it)

apps/web/src/app/api/exa/[...path]/route.ts
```diff
@@ -38,19 +45,48 @@ export async function POST(request: NextRequest) {
...
+  // freeAllowance is the stored value from the first request of the month;
+  // null means no row yet, so we compute from the helper.
+  // Use read replica for monthly usage check - this is a read-only operation that can tolerate
+  // slight replication lag, and provides lower latency for US users
+  const { usage: monthlyUsage, freeAllowance: storedAllowance } = await getExaMonthlyUsage(
+    user.id,
+    readDb
+  );
+  const allowance = storedAllowance ?? getExaFreeAllowanceMicrodollars(new Date(), user);
+  const isPaidRequest = monthlyUsage >= allowance;
```
Comment (markijbema) [inline apps/web/src/app/api/exa/[...path]/route.ts:63]: so if you used $9.99, and your allowance is $10, and you do a request of $0.10 the request is still fully free right? (not saying this is wrong, just checking if i understand)

apps/web/src/lib/constants.ts
```diff
@@ -62,3 +62,6 @@ export const STRIPE_PUBLISHABLE_KEY = process.env.NEXT_PUBLIC_STRIPE_PUBLISHABLE
 
 export const PROMOTION_MAX_REQUESTS = 10000;
 export const PROMOTION_WINDOW_HOURS = 24;
+
+/** $10/month free Exa allowance in microdollars */
```
markijbema: i'm so glad you addedd this comment i was so confused by the constant name and the neatly formatted microdollars number

apps/web/src/lib/recomputeUserBalances.ts
```diff
@@ -260,3 +283,22 @@ async function applyUserBalanceUpdates(updates: UserBalanceUpdates): Promise<boo
     return true;
   });
 }
+
+type UsageRecord = { cost: number; created_at: string };
+
+/** Merge two arrays that are each already sorted by `created_at`. */
+export function mergeSortedByCreatedAt(a: UsageRecord[], b: UsageRecord[]): UsageRecord[] {
```
markijbema, line 290: Surely there's a better way to do this than with semi-assembly?

markijbema:
My assessment

## Bookkeeping & Recoverability Review
https://github.com/Kilo-Org/cloud/pull/2169#issuecomment-4207001333

3) sounds farfetched, but also the fix of reordering the statements is easy i think
8) the advice of adding a comment is sound, especially since it will probably an agent reading it
9) sounds like bullshit, can you confirm?

## Comparison: Exa bookkeeping vs. microdollar_usage bookkeeping
https://github.com/Kilo-Org/cloud/pull/2169#issuecomment-4207002059

1) sounds like 9 from the previous comment, now I'm starting to doubt whether there is truth to this :P I really hate the personal/org if-code :(
5) is this correct? I think we do want to send something to sentry?
6) Please add a comment somewhere, i asssume we'll throw the plans out later?

## Architecture Review: Custom Exa Billing vs. Reusing Existing Free-Tier Infrastructure
https://github.com/Kilo-Org/cloud/pull/2169#issuecomment-4207020159

The My Assesment part also implies that we handle users and orgs very different, indeed please do consider

## Plan vs. Implementation Review

6) please add the comment
7) please verify these claims about microdollar usage

## Security Review: Exa Free Tier Abuse Vectors

This all is fine imho

markijbema (approved):
Some inline human comments from me, and I think the llm analysis found some valid points summarized in the last post, but I don't think any of them are really blocking

.plans/exa-monthly-allowance.md
RSO: I don't think we have a monthly free-credits thing anywhere else? My thinking was that I didn't want people to go and create 10 org trials to get $100 in free Exa credits. $10 is plenty. In the real world, I don't think users will use both personal/organization credits as much, though I may be mistaken. Either way, I'm comfortable with the limit being per-user for now.

apps/web/src/app/api/cron/exa-partition-maintenance/route.ts
RSO: You never write to it. Postgres handles the partitioning and allocation.

apps/web/src/app/api/cron/exa-partition-maintenance/route.ts
RSO: JS doesn't have these built-in utilities. We have `date-fns` as a dependency, so we could write `format(date, 'yyyy_MM')` if that makes you feel better!

apps/web/src/app/api/cron/exa-partition-maintenance/route.ts
RSO: So both querying and writing are done through `exa_usage_log`, behind the scenes, Postgres will find the partition it needs to put it in. At least that's how I understand it.

apps/web/src/lib/recomputeUserBalances.ts
RSO: Simplified to a one-liner: `[...a, ...b].sort((x, y) => x.created_at.localeCompare(y.created_at))`

RSO:
> 5) The after() callback doesn't catch JSON parse failures: In route.ts:99-110, the after() callback does await cloned.json(). If the upstream response isn't valid JSON (network error, truncated response), this throws and the entire after() callback silently fails. The old code had a try/catch around this; the new code removed it. Even non-billing logging failures silently disappear.
> 
> markijbema: Is this correct? I think we do want to send something to Sentry?

We only clone responses that have `response < 400`, so I think the risk is rather small.

PR #2174, opened 2026-04-08
fix(admin): fix FK violation in softDeleteUser and surface errors

jeanduplessis:
## Summary
- **Fix FK constraint violation in `softDeleteUser()`**: Reordered deletion in `user.ts` so `kiloclaw_subscriptions` and `kiloclaw_cli_runs` (which have foreign key references to `kiloclaw_instances.id` with NO ACTION on delete) are deleted before `kiloclaw_instances`. This caused error 23503 when deleting users with a subscription linked to a provisioned instance.
- **Surface errors in admin GDPR delete UI**: The admin GDPR delete button previously logged errors only to `console.error` with no user-facing feedback. Backend errors (like the FK violation) were re-thrown as generic 500s with no JSON body. Now:
  - Backend catches DB errors and external service failures separately, reports to Sentry, and returns structured JSON responses with safe error messages (no raw exception text leaked to the client).
  - Frontend uses sonner toast notifications for success, partial success (warnings), and failure states.
  - Malformed request bodies now return proper 400 errors instead of unhandled 500s.
- **Regression test**: Added a test that creates a subscription and CLI run both linked to an instance via `instance_id`, then verifies `softDeleteUser` succeeds without FK violation.
## Verification
- [x] `pnpm typecheck` — passes (only pre-existing `linkify-it` type error in unrelated file)
- [x] Six-agent code review (security, logic, types, data, resources, style) — all findings addressed
- [x] Security review: sanitized error messages to avoid leaking SQL/provider details to browser
- [x] Type review: added request body validation, typed frontend response parsing

## Visual Changes
| Before | After |
| ------ | ----- |
| GDPR delete errors silently logged to console only | Error/warning/success toast notifications via sonner |
## Reviewer Notes
- The FK ordering fix resolves an active GDPR blocker for user `f93225b9-...` — deploy priority.
- Error messages sent to the admin UI are intentionally generic ("check Sentry for details") to avoid leaking DB/provider internals. Full details go to Sentry with `gdpr-removal` / `gdpr-removal-external` tags.
- `SoftDeletePreconditionError` messages (e.g. "active subscription") are still returned verbatim since they're controlled, user-facing strings.
- The `route.ts` has one `as Record<string, unknown>` cast for the parsed request body — this is guarded by the preceding `typeof` + `in` checks.

### DIFF
--- a/apps/web/src/app/admin/api/users/gdpr-removal/route.ts
+++ b/apps/web/src/app/admin/api/users/gdpr-removal/route.ts
@@ -51,16 +51,7 @@ export async function POST(request: NextRequest): Promise<NextResponse<GdprRemov
     );
   }
 
-  const warnings: string[] = [];
-  try {
-    await softDeleteUserExternalServices(user);
-  } catch (error) {
-    captureException(error, {
-      tags: { source: 'gdpr-removal-external' },
-      extra: { userId },
-    });
-    warnings.push('External service cleanup failed — check Sentry for details');
-  }
+  const warnings = await softDeleteUserExternalServices(user);
 
   return NextResponse.json({
     success: true,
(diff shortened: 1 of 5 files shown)


jeanduplessis:
**Re: partial-success warning never fires** — Fixed in 55a2f9cf5. Each handler now returns `string | null` (failure message or null on success), and `softDeleteUserExternalServices` collects and returns `string[]`. The route uses the returned array directly instead of the dead try/catch. `Promise.allSettled` was also simplified to `Promise.all` since the handlers never reject.

RSO (approved):

PR #2192, opened 2026-04-08
Refactor kiloclaw to add provider runtime and routing adapters

pandemicsyn:
## Summary
This refactor is the first step toward supporting multiple infrastructure providers in KiloClaw without duplicating instance lifecycle logic per platform. The goal is to keep `KiloClawInstance` as the single authority for instance state, move Fly-specific runtime and routing behavior behind a provider adapter, and preserve current Fly-backed behavior so later phases can add providers like AWS, Northflank, or Kubernetes incrementally.

This branch now covers the bulk of that extraction:
- add `provider` and persisted `providerState` to `KiloClawInstance`, while keeping existing Fly-backed behavior and public routes compatible with already-deployed instances
- make `providerState` the canonical provider record on the primary lifecycle path, while continuing to mirror legacy Fly fields for compatibility
- extract Fly runtime lifecycle logic into `services/kiloclaw/src/providers/fly` behind a provider registry and adapter contract
- route worker proxy traffic and controller RPCs through provider-derived routing targets instead of hardcoded Fly transport details
- replace Fly-shaped adapter inputs with a neutral runtime spec, with Fly translating that spec internally
- capability-gate Fly-only admin/debug flows such as snapshots, reassociation, and direct machine destroy so future providers are not forced to implement Fly-specific behavior

## Verification
- `pnpm --filter kiloclaw typecheck`
- `pnpm --filter kiloclaw test`

## Visual Changes
- N/A

## Reviewer Notes
- legacy Fly fields still exist as compatibility mirrors; `providerState` is canonical on the primary lifecycle path, but some Fly-only helper paths still rely on the sync bridge
- worker proxy routability still relies on legacy Fly status fields before resolving provider transport; will tackle this when i go to implement a second provider. 
- the adapter contract is now provider-neutral, but Fly still translates `RuntimeSpec` to `FlyMachineConfig` inside the Fly-only helper layer; further internal neutralization is deferred until before or during provider 2 implementation.


services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
```diff
@@ -143,6 +150,62 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
...
+  private async persistProviderResult(result: ProviderResult): Promise<void> {
+    this.applyProviderResult(result);
+    await this.persist({
+      provider: result.providerState.provider,
+      providerState: result.providerState,
+      ...(result.corePatch ?? {}),
+    });
+  }
+
+  async getRoutingTarget(): Promise<ProviderRoutingTarget | null> {
```
St0rmz1, line 176: The bare `catch { return null; }` here swallows all errors silently. If `getRoutingTarget` fails due to a transient issue (e.g. Fly API timeout), callers get `null` with no log trace, making it hard to distinguish "no machine ID" from "provider temporarily unreachable" when debugging 503s in production. Consider adding a `console.warn` or `doWarn` inside the catch before returning `null`.

services/kiloclaw/src/providers/fly/index.ts
```diff
@@ -0,0 +1,254 @@
...
+      observation: {
+        runtimeState: 'stopped',
+      },
+    };
+  },
+
+  async restartRuntime({ env, state, runtimeSpec, minSecretsVersion, onProviderResult }) {
+    const providerState = getFlyProviderState(state);
+    if (!providerState.machineId) {
+      throw new Error('No machine exists');
```
St0rmz1, line 219: `startRuntime` throws if `onProviderResult` is not provided, but here `restartRuntime` uses optional chaining (`onProviderResult?.(...)`), silently skipping the persistence callback. If this is ever called without the callback, `restartUpdateSent: true` won't be persisted, which would cause the reconciler to treat a `waitForState` timeout as a failure instead of an expected in-flight restart. Consider adding the same guard as `startRuntime` for consistency, or documenting why it's intentionally optional here.

services/kiloclaw/src/durable-objects/kiloclaw-instance/state.ts
```diff
@@ -14,6 +19,122 @@ export function storageUpdate(update: Partial<PersistedState>): Partial<Persiste
...
+): void {
+  s.provider = providerState.provider;
+  s.providerState = providerState;
+
+  if (providerState.provider === 'fly') {
+    hydrateFlyLegacyFieldsFromProviderState(s);
+  }
+}
+
+export function syncProviderStateForStorage(
```
St0rmz1, line 80: This function has dual behavior: it mutates `s` in-place (via `applyProviderState`) AND returns a modified patch. Both effects are required for correctness. A future caller that only uses the return value without expecting the mutation (or vice versa) could end up with desynchronized in-memory and persisted state. Consider adding a brief JSDoc noting the intentional side effect, since the function name suggests a pure transform.

St0rmz1 (approved):

PR #2251, opened 2026-04-09
fix(kiloclaw): move pipeline bindings to env.production to prevent wrangler dev OAuth prompt

pedroheyerdahl:
## Summary

Moves the Cloudflare Pipeline bindings from the top-level `wrangler.jsonc` config into a named `env.production` environment, and adds `--env production` to the CI deploy command.

`wrangler dev` always uses the top-level config (it runs without `--env`), so declaring `pipelines` there caused Cloudflare to open a browser OAuth flow on every local dev start — the crash Remon and Florian reported was wrangler trying to authenticate against the real Pipelines API to simulate the binding locally.

The previous attempt (`env.dev: { pipelines: [] }`) didn't work because wrangler `env` overrides only apply to named-environment *deploys* (e.g. `wrangler deploy --env dev`), not to `wrangler dev`.

Both pipeline bindings are already typed as optional in `KiloClawEnv` and all write paths are best-effort no-ops when the binding is absent, so local dev works correctly without them.

## Verification

- `pnpm --filter kiloclaw typecheck` passes
- Pre-push hooks pass (typecheck + lint + format)
- Wrangler dev will no longer trigger CF OAuth on start
- CI deploy uses `--env production` so pipelines are still present in prod

## Visual Changes

N/A

## Reviewer Notes

The only behavioral change in production is that `wrangler deploy` now deploys the `production` named environment instead of the top-level config. Since `env.production` only declares `pipelines` and inherits everything else from the top-level, the deployed worker is identical to before.

### DIFF
--- a/.github/workflows/deploy-kiloclaw.yml
+++ b/.github/workflows/deploy-kiloclaw.yml
@@ -179,6 +179,7 @@ jobs:
           workingDirectory: services/kiloclaw
           command: >-
             deploy
+            --env production
             --var FLY_IMAGE_TAG:${{ steps.image-hash.outputs.tag }}
             --var OPENCLAW_VERSION:${{ steps.openclaw-version.outputs.version }}
             --var FLY_IMAGE_DIGEST:${{ steps.image-digest.outputs.digest }}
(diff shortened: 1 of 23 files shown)


pedroheyerdahl:
Closing in favour of a clean branch — the squash accidentally rebased onto main locally. Will reopen from a fresh branch.

PR #2275, opened 2026-04-09
feat(kiloclaw): add docker-local provider demo

pandemicsyn:
## Summary

Adds `docker-local` as the first non-Fly provider demo while keeping `KiloClawInstance` as the single authority for lifecycle state. The goal is a simple local development path that can provision multiple KiloClaw runtimes through a local Docker API, while preserving existing Fly behavior and public platform routes. This branch introduces provider IDs/provider state, the docker-local adapter, provider-neutral runtime/storage routing metadata, provider runtime inspection for non-Fly alarm reconciliation, and admin/UI gates based on provider capabilities instead of Fly-only IDs where the action is provider-neutral.

It also decouples bootstrap env-key creation from Fly app creation, injects the same bootstrap/user env set across providers, and hardens docker-local lifecycle correctness around terminal start failures, durable host-port persistence, and existing-container spec reconciliation.

## Verification

- Provision and destroy instances
- Pair telegram
- Change default model
- Start/stop instance
- Restart gateway
- Concurrently running fly personal instance still works. 

If you wanna test this:

1. Add the following to your .dev.vars:

```
KILOCLAW_DEFAULT_PROVIDER=docker-local
DOCKER_LOCAL_API_BASE=http://127.0.0.1:23750
DOCKER_LOCAL_IMAGE=kiloclaw:local
DOCKER_LOCAL_PORT_RANGE=45000-45999
```
2. run `./scripts/build-local-image.sh`
3. expose the docker API some how e.g. `socat TCP-LISTEN:23750,bind=127.0.0.1,reuseaddr,fork UNIX-CONNECT:/var/run/docker.sock`
4. start kiloclaw worker
5. Deploy a new instance, basically instantly 😁 

## Visual Changes

N/A - UI changes are control gating and provider/runtime labels in the existing KiloClaw admin surfaces.

## Reviewer Notes

Fly remains the production/default provider. The start failure path intentionally preserves the old Fly invariant: once a Fly machine ID is durably persisted, the DO leaves the instance in `starting` and lets Fly alarm reconciliation decide instead of marking it `stopped` from the async catch handler.

The adapter seam is provider-neutral for lifecycle routing/results, but some internal Fly helper translation still exists under the Fly-only helper layer. 

Docker-local is development-only and expects a locally exposed Docker HTTP API.

## What doesn't work

openclaw doctor - we need to move that to executed via the controller rather than via fly exec

### DIFF
--- a/apps/web/src/routers/admin-kiloclaw-instances-router.ts
+++ b/apps/web/src/routers/admin-kiloclaw-instances-router.ts
@@ -897,6 +897,12 @@ export const adminKiloclawInstancesRouter = createTRPCRouter({
       } catch (err) {
         throwKiloclawAdminError(err, 'Failed to verify machine state before destroy');
       }
+      if (status.provider !== 'fly') {
+        throw new TRPCError({
+          code: 'BAD_REQUEST',
+          message: `Direct Fly machine destroy is not supported for provider ${status.provider}`,
+        });
+      }
       if (status.flyAppName !== input.appName || status.flyMachineId !== input.machineId) {
         throw new TRPCError({
           code: 'BAD_REQUEST',
--- a/services/kiloclaw/src/providers/docker-local/index.ts
+++ b/services/kiloclaw/src/providers/docker-local/index.ts
@@ -97,13 +97,14 @@ function ensureDockerLocalConfigured(env: KiloClawEnv): string {
 
 async function dockerFetch(env: KiloClawEnv, path: string, init?: RequestInit): Promise<Response> {
   const base = ensureDockerLocalConfigured(env);
+  const url = `${base.replace(/\/+$/, '')}${path.startsWith('/') ? path : `/${path}`}`;
   const headers = new Headers(init?.headers);
   if (init?.body && !headers.has('content-type')) {
     headers.set('content-type', 'application/json');
   }
 
   try {
-    return await fetch(new URL(path, `${base}/`).toString(), {
+    return await fetch(url, {
       ...init,
       headers,
     });
--- a/services/kiloclaw/src/providers/docker-local/index.ts
+++ b/services/kiloclaw/src/providers/docker-local/index.ts
@@ -476,6 +476,8 @@ export const dockerLocalProviderAdapter: InstanceProviderAdapter = {
       await createContainer(env, state, providerState, runtimeSpec);
       await startContainer(env, containerName);
     }
+    // If Docker and the DO both report running, leave the dev container intact.
+    // Restart/redeploy is the explicit path that reapplies image/env/config changes.
 
     return {
       providerState,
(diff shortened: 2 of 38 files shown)


services/kiloclaw/src/providers/docker-local/index.ts
```diff
@@ -0,0 +1,582 @@
...
+  for (let port = start; port <= end; port++) {
+    if (!used.has(port)) return port;
+  }
+  throw dockerConfigError('Provider docker-local has no free ports in the configured range');
+}
+
+function buildContainerEnv(runtimeSpec: RuntimeSpec): string[] {
+  return Object.entries({
+    ...runtimeSpec.env,
+    ...runtimeSpec.bootstrapEnv,
```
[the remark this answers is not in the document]
pandemicsyn: nope

services/kiloclaw/src/providers/docker-local/index.ts
```diff
@@ -0,0 +1,582 @@
...
+  const used = new Set<number>();
+  for (const container of containers) {
+    for (const port of container.Ports ?? []) {
+      if (typeof port.PublicPort === 'number') {
+        used.add(port.PublicPort);
+      }
+    }
+  }
+  for (let port = start; port <= end; port++) {
+    if (!used.has(port)) return port;
```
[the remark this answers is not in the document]
pandemicsyn: No i think thats fine for this version.

apps/web/src/routers/admin-kiloclaw-instances-router.ts
```diff
@@ -897,6 +897,12 @@ export const adminKiloclawInstancesRouter = createTRPCRouter({
       } catch (err) {
         throwKiloclawAdminError(err, 'Failed to verify machine state before destroy');
       }
+      if (status.provider !== 'fly') {
```
RSO, line 900: We probably need a similar if in the Admin overview, where I just added the "Copy SSH" button that copies the `fly console ssh` command for that instance.

services/kiloclaw/src/schemas/instance-config.ts
```diff
@@ -55,7 +55,7 @@ export const CustomSecretMetaSchema = z.object({
 
 export type CustomSecretMeta = z.infer<typeof CustomSecretMetaSchema>;
 
-export const ProviderIdSchema = z.enum(['fly', 'northflank', 'aws', 'k8s']);
+export const ProviderIdSchema = z.enum(['fly', 'docker-local', 'northflank', 'aws', 'k8s']);
```
RSO: We may want to add a comment to link this line with the one in `apps/web/src/lib/kiloclaw/types.ts`

RSO (approved):
Didn't try it out yet, but did go through the changes, looking good!

services/kiloclaw/src/schemas/instance-config.ts
pandemicsyn: done, also paired down the list to just the "real" current set. (fly, docker, northflank)

PR #2283, opened 2026-04-10
feat(claw): add typed confirmation dialog for instance destroy

pandemicsyn:
## Summary

Replace the inline confirm/cancel button toggle in the KiloClaw settings "Danger Zone" with a modal confirmation dialog for instance destruction.

- **Typed confirmation** — user must type the instance name or instance ID to proceed. In org context the phrase is prefixed with the org name (e.g. `myorg/my-instance`).
- **Irreversibility warning** — dialog states the action is irreversible and data is unrecoverable.
- **Context labeling** — dialog title and body distinguish "Organization Instance" vs "Personal Instance".
- **Dialog locked during mutation** — Escape, overlay click, and close button are prevented while the destroy mutation is in flight.
- **New `instanceRecordId` field** on `KiloClawDashboardStatus` — always returns the Postgres row ID, even for legacy personal instances where `instanceId` is intentionally null (to avoid resolving the wrong DO). Used as the confirmation identifier.
- **`organizationName` threaded** from the org settings server page through `ClawSettingsPage` → `SettingsTab` → `DestroyInstanceDialog`.

## Verification

- [x] `pnpm typecheck` — passes (only pre-existing errors from missing `@kilocode/kiloclaw-secret-catalog`, `croner`, etc.)
- [x] Verified no dead/duplicated code blocks in `SettingsTab.tsx`
- [x] Confirmed `DialogContent` supports `onInteractOutside`/`onEscapeKeyDown` via Radix spread props
- [ ] Manual browser testing of destroy flow in personal and org contexts

## Visual Changes

| Before | After |
| ------ | ----- |
| Inline "Destroy Instance" / "Yes, destroy" / "Cancel" button toggle | Modal dialog with instance details, irreversibility warning, typed confirmation input, and Cancel/Destroy buttons |

## Reviewer Notes

- `instanceRecordId` is additive — does not change existing `instanceId` semantics (which remain null for legacy instances to avoid wrong-DO resolution).
- The billing lock dialog (`AccessLockedDialog.tsx`) has its own separate destroy flow that was intentionally not changed — it operates in a different context (subscription expired/locked).


pandemicsyn:
Screenshot 2026-04-09 at 9 15 30 PM

jeanduplessis (approved):

PR #2302, opened 2026-04-10
feat(admin): add volume bump to 15 GB workaround for KiloClaw instances

evanjacobson:
## Summary

This PR adds a "Bump Volume to 15GB" button to streamline Lígia's support workflow due to a recent up-tick in full user volumes. The scope is intentionally restricted to a static bump to 15 GB, as this button will be removed once we implement functionality for users to delete files by themselves. 

Planned follow-up: Refactor FLY_API_BASE as an export instead of redefining it across files. This keeps the scope focused on the volume bump changes.

Ligia has already confirm that the UX works for her, and Florian has ok'd this scoped change. 

<details><summary>Details</summary>
<p>

* The flow includes a confirmation dialog with an acknowledgment checkbox, input validation against the live Fly resource IDs (to prevent mismatches), audit logging, and a redeploy warning when Fly reports the machine needs a redeploy.

* The implementation follows the same patterns as the existing `destroy-fly-machine` platform route and the snapshot restore admin flow: tRPC admin procedure → `KiloClawInternalClient` → kiloclaw platform route → Fly Machines API.

* The Fly extend-volume API may return `needs_restart: true`, and the admin sees a warning toast in that case. The code defaults to true when Fly omits the flag, out of an abundance of caution. User machines are unusable when at full capacity, so at the time of expansion, a redeploy will not impact the user, regardless.

## Verification

- [x] `pnpm typecheck`
- [x] `pnpm lint`
- [x] Manual test: confirmed dialog opens, acknowledgment gate works, volume extends, toast reflects `needsRestart` status, audit log is written

</p>
</details> 

## Visual Changes

https://www.loom.com/share/96f9bf3840004f43ae3eebbbbee07801

## Reviewer Notes

- This is explicitly a temporary workaround, so the target size is hard-coded to 15 GB and the dialog copy says as much. No generalized volume resize UI is intended.
- The new `kiloclaw.volume.extend` audit action is added.  Let me know if you prefer to remove this detail, as this button will be removed later.


apps/web/src/app/admin/components/KiloclawInstances/BumpVolumeTo15GbDialog.tsx
evanjacobson, line 1: This is intentionally named and intentionally as containerized as possible, to simplify deletion later.

services/kiloclaw/src/routes/platform.ts
```diff
@@ -2126,7 +2002,7 @@ platform.post('/destroy-fly-machine', async c => {
     return c.json({ error: 'FLY_API_TOKEN is not configured' }, 503);
   }
 
-  const url = `https://api.machines.dev/v1/apps/${appName}/machines/${machineId}?force=true`;
+  const url = `${FLY_API_BASE}/v1/apps/${appName}/machines/${machineId}?force=true`;
```
evanjacobson: not directly in-scope but directly relevant to the now-exported FLY_API_BASE in this file

services/kiloclaw/src/routes/platform.ts
```diff
@@ -2167,4 +2041,74 @@ platform.post('/destroy-fly-machine', async c => {
   }
 });
 
+// POST /api/platform/extend-volume
+// Temporary workaround: extend a Fly volume to exactly 15 GB.
+const EXTEND_VOLUME_TARGET_SIZE_GB = 15;
```
evanjacobson, line 2270: while we're intentionally hardcoding this elsewhere, this is a shared file and it makes sense to design this more maintainably here.

apps/web/src/app/admin/components/KiloclawInstances/BumpVolumeTo15GbDialog.tsx
```diff
@@ -0,0 +1,159 @@
+'use client';
```
evanjacobson, line 9:
```suggestion
/**
 *
 * This is a temporary feature to assist Customer Support.
 * To be deleted after user-facing file deletion is launched. 
 * 
 * All changes for this feature are in PR #2302.
 * 
 */
'use client';
```

pandemicsyn (approved):

PR #2304, opened 2026-04-10
feat(claw): fire claw_setup_start_viewed on CreateInstanceCard mount

pedroheyerdahl:
## Motivation

The KiloClaw setup funnel was missing a top-of-funnel event. The earliest instrumented moment was `claw_create_instance_clicked`, but that fires only when the user actively clicks the button. There was no way to tell how many users landed on the Claw page, saw the setup card, and left without clicking. That gap makes it impossible to measure awareness-to-intent conversion at the top of the funnel, which is where any acquisition campaign would show its effect first.

## Implementation

`CreateInstanceCard` is the component that renders when a logged-in user has no instance yet. Adding a `useEffect` on mount that calls `posthog.capture('claw_setup_start_viewed')` is sufficient because this component is only mounted when the user has no instance — users who already have one never see it, so the event will not contaminate retention or engagement analysis.

The event name follows the convention used throughout `ClawOnboardingFlow` (`claw_setup_identity_viewed`, `claw_setup_permissions_viewed`, etc.). `claw_setup_start_viewed` is the step immediately before all of those, making its funnel position unambiguous.

`usePostHog` was already available from `posthog-js/react`, which is already imported in the file, so no new dependency was needed.

A review comment flagged that the original guard set `hasCapturedPageView.current = true` before checking whether `posthog` was available. A first render with `posthog` still `undefined` would poison the ref and silently drop the event. The guard was updated to bail out before touching the ref when `posthog` is not yet ready.

## Testing

Typechecked against the web app.

Verified in local dev against a fresh account with no Claw instance. With PostHog debug mode active and capturing opted back in via `opt_in_capturing()`, navigating to `/claw/new` produces exactly one `claw_setup_start_viewed` event on mount, sourced from `CreateInstanceCard.tsx`. Clicking "Try For Free" produces `claw_create_instance_clicked` with the expected `selected_model` property. Screenshot below.
Screenshot 2026-04-10 at 17 32 29

## Notes

The lint failures in `services/notifications/` were present on main before this branch was cut. They should be addressed separately.


pedroheyerdahl:
Fixed in 5551af5 — moved the `!posthog` guard before `hasCapturedPageView.current` so the ref is only set to `true` after PostHog is available and the event is actually captured. Also removed the optional chaining on `posthog.capture` now that it's guaranteed non-null at that point.

pandemicsyn (approved):

PR #2328, opened 2026-04-11
feat(kiloclaw): add chat platform guide links

emilieschario:
Adding links to the docs because they are more helpful than what's already there.


RSO:
CleanShot 2026-04-13 at 09 11 41@2x

Instead of rendering the links on the same line, I've opted to move them to a newline

RSO (approved):

PR #2335, opened 2026-04-12
feat(kiloclaw): add sidebar submenu

RSO:
## Summary
- Replaces the inline KiloClaw sidebar group with a top-level KiloClaw entry that opens a focused submenu when an instance exists.
- Links KiloClaw directly to the `/new` onboarding page when no instance exists, keeping KiloClaw active for `/claw/*` routes.
- Extends shared sidebar menu rendering to support action rows, active overrides, label-less groups, suffix icons, and pointer cursors.

## Visual Changes

https://github.com/user-attachments/assets/40c30ebd-c196-4b55-91ec-df2e14f3cf8f

https://github.com/user-attachments/assets/446375e6-5866-4834-957d-1f5a8e0a8d18

### DIFF
--- a/apps/web/src/app/(app)/components/AppSidebar.tsx
+++ b/apps/web/src/app/(app)/components/AppSidebar.tsx
@@ -29,7 +29,7 @@ function isKiloClawNewPath(pathname: string): boolean {
 export default function AppSidebar(props: React.ComponentProps<typeof Sidebar>) {
   const currentOrgId = useUrlOrganizationId();
   const pathname = usePathname();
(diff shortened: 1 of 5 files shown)


jeanduplessis (approved):

RSO:
Addressed the review feedback in c6012f2d1. The /claw/new auto-collapse now uses a non-persisting sidebar setter so it does not overwrite the sidebar cookie, and the personal/org KiloClaw entries now treat status as present/absent/unknown so loading state links to the base redirect route instead of prematurely linking to /new.

PR #2376, opened 2026-04-13
Feature/kiloclaw admin cli admin view and schema

evanjacobson:
## Summary

Admins can not currently initiate CLI recovery runs for users, which makes customer support a time-intensive operation for certain workflows. This PR adds the recovery CLI to the admin instance view, and adds protection against concurrent CLI runs. It also introduces some small UX improvements for users that are directly related to the other changes.

Although this is a large PR, it is designed to follow existing patterns for routing, DB ops, auditing, etc.

<details><summary>More Details</summary>
<p>

- Concurrent start requests on a controller are now serialized via a promise queue (runStartExclusive) to prevent race conditions on the activeRun check.
- The controller's 409 response for an already-running CLI run is now mapped to a user-facing TRPC CONFLICT error with a clear message, and the prefix is added to the platform's safe-error allowlist.
- The "Back to Dashboard" link on the CLI run detail page now routes to /claw/settings instead of /claw.
- If a CLI run is in progress, a user clicking the Recover with Kilo button will be routed into the CLI run view page for the current operation. 
- Users can cancel admin runs, and vice versa
- stripAnsi, timestamp formatting helpers, and DetailField are extracted into shared modules to remove duplication across admin components.

</p>
</details> 

## Verification

- [x] tests pass
- [x] manually tested

## Reviewer Notes

- A follow-up PR will address the same behaviors for org instances, in addition to other fixes for orgs. Orgs are currently non functional with this flow regardless, and this PR was large as-is, so I am scoping that work separately.

<details><summary>Known Limitations</summary>
<p>

- A CLI run can be "orphaned" if the DB insert fails, which will block the user from starting a run until the one active on the machine clears. I chose not to scope this into this PR as it is a very rare edge case, and the CLI runs are quick.
- There are known issues with orgs that I will address in a follow up

</p>
</details>

> [!IMPORTANT]
>  I encourage you to enabled "hide whitespace" when reviewing the diff. It will make things easier. 

> [!NOTE]
> I have squashed the commits to avoid a wall of text

### DIFF
--- a/apps/web/src/app/(app)/claw/components/InstanceControls.tsx
+++ b/apps/web/src/app/(app)/claw/components/InstanceControls.tsx
@@ -1,6 +1,7 @@
 'use client';
 
 import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useRouter } from 'next/navigation';
 import {
   ArrowUpCircle,
   Check,
@@ -32,6 +33,8 @@ import {
 import { RadioGroup, RadioGroupItem } from '@/components/ui/radio-group';
 import { Label } from '@/components/ui/label';
 import type { useKiloClawMutations } from '@/hooks/useKiloClaw';
+import { selectCurrentCliRun } from '@/lib/kiloclaw/cli-run-selection';
+import { useClawKiloCliRunHistory } from '../hooks/useClawHooks';
 import { useClawUpdateAvailable } from '../hooks/useClawUpdateAvailable';
 import { useGatewayUrl } from '../hooks/useGatewayUrl';
 import { ConfirmActionDialog } from './ConfirmActionDialog';
@@ -66,6 +69,7 @@ export function InstanceControls({
   onUpgradeHandled?: () => void;
   gatewayReady?: boolean;
 }) {
+  const router = useRouter();
   const posthog = usePostHog();
   const gatewayUrl = useGatewayUrl(status);
   const isRunning = status.status === 'running';
@@ -86,7 +90,7 @@ export function InstanceControls({
   const [confirmRestart, setConfirmRestart] = useState(false);
   const [confirmRedeploy, setConfirmRedeploy] = useState(false);
   const [redeployMode, setRedeployMode] = useState<'redeploy' | 'upgrade'>('redeploy');
-
+  const kiloCliRunHistory = useClawKiloCliRunHistory(isRunning);
   const { updateAvailable, catalogNewerThanImage, latestAvailableVersion, latestVersion } =
     useClawUpdateAvailable(status);
 
@@ -213,41 +217,41 @@ export function InstanceControls({
           </Badge>
         </div>
       </div>
-      {showUpgradeBanner && (
-        <Banner color="amber" className="mb-4">
-          <Banner.Icon>
-            <ArrowUpCircle />
-          </Banner.Icon>
-          <Banner.Content>
-            <Banner.Title>
-              {catalogNewerThanImage
-                ? `A newer OpenClaw version (${latestAvailableVersion}) is available`
-                : `A newer image (${latestVersion?.imageTag ?? 'unknown'}) is available`}
-            </Banner.Title>
-            <Banner.Description>
-              Upgrade your instance to get the latest features and fixes.
-            </Banner.Description>
-          </Banner.Content>
-          <Banner.Button
-            className="text-white"
-            onClick={() => {
-              setRedeployMode('upgrade');
-              setConfirmRedeploy(true);
-            }}
-          >
-            Upgrade now
-          </Banner.Button>
-          <button
-            type="button"
-            onClick={dismissBanner}
-            className="text-amber-400/60 hover:text-amber-400 transition-colors"
-            aria-label="Dismiss upgrade banner"
-          >
-            <X className="h-4 w-4" />
-          </button>
-        </Banner>
-      )}
-      <div className="flex flex-col gap-2 sm:flex-row sm:flex-wrap sm:items-center">
+      <div className="mt-4 flex flex-wrap gap-2">
+        {showUpgradeBanner && (
+          <Banner color="amber" className="mb-2 w-full">
+            <Banner.Icon>
+              <ArrowUpCircle />
+            </Banner.Icon>
+            <Banner.Content>
+              <Banner.Title>
+                {catalogNewerThanImage
+                  ? `A newer OpenClaw version (${latestAvailableVersion}) is available`
+                  : `A newer image (${latestVersion?.imageTag ?? 'unknown'}) is available`}
+              </Banner.Title>
+              <Banner.Description>
+                Upgrade your instance to get the latest features and fixes.
+              </Banner.Description>
+            </Banner.Content>
+            <Banner.Button
+              className="text-white"
+              onClick={() => {
+                setRedeployMode('upgrade');
+                setConfirmRedeploy(true);
+              }}
+            >
+              Upgrade now
+            </Banner.Button>
+            <button
+              type="button"
+              onClick={dismissBanner}
+              className="transition-colors text-amber-400/60 hover:text-amber-400"
+              aria-label="Dismiss upgrade banner"
+            >
+              <X className="h-4 w-4" />
+            </button>
+          </Banner>
+        )}
         <OpenClawButton
           canShow={isRunning && !!gatewayReady}
           gatewayUrl={gatewayUrl}
@@ -261,8 +265,8 @@ export function InstanceControls({
           disabled={
             !isStartable ||
             mutations.start.isPending ||
-            isAutoStarting ||
             isDestroying ||
+            isAutoStarting ||
             isStarting ||
             isRestarting ||
             isRecovering
@@ -353,8 +357,18 @@ export function InstanceControls({
           variant="outline"
           className="border-emerald-500/30 text-emerald-400 hover:bg-emerald-500/10 hover:text-emerald-300"
           disabled={!isRunning || isDestroying || isStarting || isRestarting || isRecovering}
-          onClick={() => {
+          onClick={async () => {
             posthog?.capture('claw_kilo_run_clicked', { instance_status: status.status });
+            try {
+              const { data } = await kiloCliRunHistory.refetch();
+              const latestRun = selectCurrentCliRun(data?.runs, status.instanceId);
+              if (latestRun?.status === 'running') {
+                router.push(`/claw/kilo-cli-run/${latestRun.id}`);
+                return;
+              }
+            } catch {
+              // Best-effort pre-check; fall through to open the dialog
+            }
             setKiloRunOpen(true);
           }}
         >
--- a/apps/web/src/app/(app)/claw/kilo-cli-run/[id]/page.tsx
+++ b/apps/web/src/app/(app)/claw/kilo-cli-run/[id]/page.tsx
@@ -70,7 +70,7 @@ export default function KiloCliRunPage() {
         icon={<KiloCrabIcon className="text-muted-foreground h-4 w-4" />}
       />
       <div className="flex items-center gap-3">
-        <Button variant="ghost" size="sm" onClick={() => router.push('/claw')}>
+        <Button variant="ghost" size="sm" onClick={() => router.push('/claw/settings')}>
           <ArrowLeft className="h-4 w-4" />
           Back to Dashboard
         </Button>
@@ -123,7 +123,7 @@ export default function KiloCliRunPage() {
             <p className="text-sm text-red-400">
               {statusQuery.error?.message ?? 'Failed to load run status'}
             </p>
-            <Button variant="outline" size="sm" onClick={() => router.push('/claw')}>
+            <Button variant="outline" size="sm" onClick={() => router.push('/claw/settings')}>
               Back to Dashboard
             </Button>
           </div>
--- a/apps/web/src/app/admin/components/KiloclawCliRuns/KiloclawCliRunsTab.tsx
+++ b/apps/web/src/app/admin/components/KiloclawCliRuns/KiloclawCliRunsTab.tsx
@@ -30,6 +30,7 @@ import { stripAnsi } from '@/lib/stripAnsi';
 const PAGE_SIZE = 25;
 
 type RunStatus = 'all' | 'running' | 'completed' | 'failed' | 'cancelled';
+type InitiatedBy = 'all' | 'admin' | 'user';
 
 function StatusBadge({ status }: { status: string }) {
   switch (status) {
@@ -82,6 +83,7 @@ export function CliRunsTab() {
   const [search, setSearch] = useState('');
   const [debouncedSearch, setDebouncedSearch] = useState('');
   const [statusFilter, setStatusFilter] = useState<RunStatus>('all');
+  const [initiatedByFilter, setInitiatedByFilter] = useState<InitiatedBy>('all');
   const [selectedRunId, setSelectedRunId] = useState<string | null>(null);
 
   const [debounceTimer, setDebounceTimer] = useState<ReturnType<typeof setTimeout> | null>(null);
@@ -103,6 +105,7 @@ export function CliRunsTab() {
         limit: PAGE_SIZE,
         search: debouncedSearch || undefined,
         status: statusFilter,
+        initiatedBy: initiatedByFilter,
       },
       { staleTime: 10_000 }
     )
@@ -140,6 +143,22 @@ export function CliRunsTab() {
             <SelectItem value="cancelled">Cancelled</SelectItem>
           </SelectContent>
         </Select>
+        <Select
+          value={initiatedByFilter}
+          onValueChange={v => {
+            setInitiatedByFilter(v as InitiatedBy);
+            setPage(0);
+          }}
+        >
+          <SelectTrigger className="w-[150px]">
+            <SelectValue />
+          </SelectTrigger>
+          <SelectContent>
+            <SelectItem value="all">All users</SelectItem>
+            <SelectItem value="admin">Admin</SelectItem>
+            <SelectItem value="user">User</SelectItem>
+          </SelectContent>
+        </Select>
         {isLoading && <Loader2 className="h-4 w-4 animate-spin" />}
         {pagination && (
           <span className="text-muted-foreground ml-auto text-sm">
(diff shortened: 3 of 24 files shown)


apps/web/src/app/(app)/claw/components/InstanceControls.tsx
```diff
@@ -350,8 +354,14 @@ export function InstanceControls({
           variant="outline"
           className="border-emerald-500/30 text-emerald-400 hover:bg-emerald-500/10 hover:text-emerald-300"
           disabled={!isRunning || isDestroying || isStarting || isRestarting || isRecovering}
-          onClick={() => {
+          onClick={async () => {
             posthog?.capture('claw_kilo_run_clicked', { instance_status: status.status });
+            const { data } = await kiloCliRunHistory.refetch();
+            const latestRun = selectCurrentCliRun(data?.runs, status.instanceId);
+            if (latestRun?.status === 'running') {
+              router.push(`/claw/kilo-cli-run/${latestRun.id}`);
```
evanjacobson: If there's a run in progress, clicking the `Recover with Kilo` button will directly route into the run view. 

This UX will be improved in a follow up soon after this PR, so this is the MVP as far as providing a decent UX with the enforcement of "one run at a time"

apps/web/src/app/(app)/claw/kilo-cli-run/[id]/page.tsx
```diff
@@ -75,7 +75,7 @@ export default function KiloCliRunPage() {
         icon={<KiloCrabIcon className="text-muted-foreground h-4 w-4" />}
       />
       <div className="flex items-center gap-3">
-        <Button variant="ghost" size="sm" onClick={() => router.push('/claw')}>
+        <Button variant="ghost" size="sm" onClick={() => router.push('/claw/settings')}>
```
Comment (evanjacobson) [inline apps/web/src/app/(app)/claw/kilo-cli-run/[id]/page.tsx:73]: Routes the user back to the settings page instead of the chat page

apps/web/src/app/admin/components/KiloclawInstances/KiloclawInstanceDetail.tsx
```diff
@@ -411,6 +414,239 @@ function VersionPinCard({ userId, instanceId }: { userId: string; instanceId: st
...
+          </div>
+        )}
+
+        {runStatus?.hasRun && (
+          <div className="space-y-3 rounded border p-3">
+            <div className="grid grid-cols-2 gap-4 md:grid-cols-5">
+              <DetailField label="Status">
+                <Badge
+                  className={
+                    runStatus.status === 'running'
```
evanjacobson: 4x nested ternary is the most I've personally witnessed.

packages/db/src/migrations/0086_crazy_anthem.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "kiloclaw_cli_runs" ADD COLUMN "initiated_by_admin_id" text;--> statement-breakpoint
+ALTER TABLE "kiloclaw_cli_runs" ADD CONSTRAINT "kiloclaw_cli_runs_initiated_by_admin_id_kilocode_users_id_fk" FOREIGN KEY ("initiated_by_admin_id") REFERENCES "public"."kilocode_users"("id") ON DELETE set null ON UPDATE no action;--> statement-breakpoint
+CREATE INDEX "IDX_kiloclaw_cli_runs_initiated_by_admin_id" ON "kiloclaw_cli_runs" USING btree ("initiated_by_admin_id");
```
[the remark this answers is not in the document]
evanjacobson: will create this manually instead. 

CREATE INDEX CONCURRENTLY "IDX_kiloclaw_cli_runs_initiated_by_admin_id" ON "kiloclaw_cli_runs" USING btree ("initiated_by_admin_id");

apps/web/src/app/admin/components/KiloclawInstances/KiloCliRunCard.tsx
```diff
@@ -0,0 +1,270 @@
...
+    trpc.admin.kiloclawInstances.startKiloCliRun.mutationOptions({
+      onSuccess: result => {
+        toast.success('CLI run started');
+        setRunId(result.id);
+        setShowOutput(true);
+        void refetchRuns();
+        void refetchStatus();
+      },
+      onError: err => {
+        toast.error(`Failed to start CLI run: ${err.message}`);
```
evanjacobson, line 78: can show raw errors — this is admin only view

apps/web/src/hooks/useKiloClaw.ts
```diff
@@ -121,6 +121,7 @@ export function useKiloCliRunHistory(enabled: boolean) {
     trpc.kiloclaw.listKiloCliRuns.queryOptions(undefined, {
       enabled,
       staleTime: 30_000,
+      refetchInterval: enabled ? 3_000 : false,
```
evanjacobson: re-fetch every 3s if active only

apps/web/src/routers/admin-kiloclaw-instances-router.ts
```diff
@@ -556,29 +565,217 @@ export const adminKiloclawInstancesRouter = createTRPCRouter({
...
+          prompt: row.prompt,
+          initiatedBy: getCliRunInitiatedBy(row),
+        };
+      } catch (err) {
+        console.error('Failed to get kilo CLI run status for user:', input.userId, err);
+        throwKiloclawAdminError(err, fallbackMessage);
+      }
+    }),
+
+  cancelKiloCliRun: adminProcedure
```
evanjacobson: [bot] should this be moved into cli-runs.ts?

packages/db/src/migrations/0086_misty_beast.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "kiloclaw_cli_runs" ADD COLUMN "initiated_by_admin_id" text;--> statement-breakpoint
+ALTER TABLE "kiloclaw_cli_runs" ADD CONSTRAINT "kiloclaw_cli_runs_initiated_by_admin_id_kilocode_users_id_fk" FOREIGN KEY ("initiated_by_admin_id") REFERENCES "public"."kilocode_users"("id") ON DELETE set null ON UPDATE no action;
```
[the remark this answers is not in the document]
evanjacobson: [bot] verify/fix in schema.ts and I will regenerate.

packages/db/src/schema.ts
```diff
@@ -3999,6 +3999,7 @@ export const kiloclaw_cli_runs = pgTable(
       .notNull()
       .references(() => kilocode_users.id, { onDelete: 'cascade' }),
     instance_id: uuid().references(() => kiloclaw_instances.id),
+    initiated_by_admin_id: text().references(() => kilocode_users.id, { onDelete: 'set null' }),
```
evanjacobson, line 4002: added to kiloclaw_cli_runs

services/kiloclaw/controller/src/routes/kilo-cli-run.ts
```diff
@@ -119,76 +129,80 @@ export function registerKiloCliRunRoutes(app: Hono, expectedToken: string): void
...
+      }
 
-    const parsed = StartRunBodySchema.safeParse(body);
-    if (!parsed.success) {
-      return c.json({ error: 'Invalid request body', details: parsed.error.flatten() }, 400);
-    }
+      const parsed = StartRunBodySchema.safeParse(body);
+      if (!parsed.success) {
+        return c.json(
+          { error: 'Invalid request body', details: z.treeifyError(parsed.error) },
```
evanjacobson, line 153: error.flatten() is deprecated in favor of z.treeifyError(error)

services/kiloclaw/src/routes/platform.ts
```diff
@@ -295,15 +295,21 @@ async function requireProviderCapability(
...
-  if (
-    typeof err === 'object' &&
-    err !== null &&
-    'status' in err &&
-    typeof (err as { status: unknown }).status === 'number'
-  ) {
-    const status = (err as { status: number }).status;
-    if (status >= 400 && status < 600) return status;
+  // Extract a valid HTTP status from the error or its cause, defaulting to 500.
+  for (const candidate of [err, err instanceof Error ? err.cause : undefined]) {
```
evanjacobson, line 325: this is just a refactor for the sake of clarity

services/kiloclaw/src/routes/platform.ts
```diff
@@ -398,7 +405,15 @@ function sanitizeOpenclawConfigError(
     'code' in err &&
     typeof (err as { code?: unknown }).code === 'string'
       ? (err as { code: string }).code
-      : undefined;
+      : typeof err === 'object' &&
```
evanjacobson: [bot] can we leverage a helper method here? Or at least explain to me what's happening

apps/web/src/routers/kiloclaw-router.ts
```diff
@@ -2227,41 +2237,51 @@ export const kiloclawRouter = createTRPCRouter({
       return {
         ...controllerStatus,
```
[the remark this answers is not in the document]
evanjacobson: This is not a concern in practice, as there is now a control in place to ensure one run at a time.

evanjacobson (commented):
Submitting pending review to unblock replies.

apps/web/src/routers/kiloclaw-router.ts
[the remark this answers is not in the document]
evanjacobson: This is already guarded by the `isControllerStatusForRun()` check on line 2221:

```typescript
if (!isControllerStatusForRun(row, controllerStatus)) {
  return { /* DB-only data */ };
}
```

If the controller is tracking a newer run, `isControllerStatusForRun` returns `false` (timestamps do not match) and we hit the early return with DB-only data. The code at line 2254 that spreads `controllerStatus` is only reachable when the controller IS tracking the requested run. The same guard exists in the admin router at line 683.

apps/web/src/app/admin/components/KiloclawCliRuns/KiloclawCliRunsTab.tsx
```diff
@@ -273,6 +299,11 @@ function RunDetail({
           <span className="truncate font-mono text-xs">{run.user_email ?? run.user_id}</span>
           <StatusBadge status={run.status} />
         </div>
+        {run.initiated_by_admin_id && (
+          <p className="text-muted-foreground text-xs">
+            Initiated by {run.initiated_by_admin_email ?? 'Admin'}
```
evanjacobson, line 304: This is useful info at a glance for us internally. Will not be available for users.

apps/web/src/lib/kiloclaw/cli-run-selection.ts
evanjacobson, line 1: picks the “best” CLI run for an instance like this:

1. use the one you explicitly asked for
2. else use the currently running one
3. else use any run for that instance
4. else return null

apps/web/src/routers/admin-kiloclaw-instances-router.ts
```diff
@@ -556,29 +565,217 @@ export const adminKiloclawInstancesRouter = createTRPCRouter({
...
+          await createKiloClawAdminAuditLog({
+            action: 'kiloclaw.cli_run.start',
+            actor_id: ctx.user.id,
+            actor_email: ctx.user.google_user_email,
+            actor_name: ctx.user.google_user_name,
+            target_user_id: input.userId,
+            message: `CLI run started on instance ${instance?.id}`,
+            metadata: {
+              runId,
+              instanceId: instance?.id,
```
evanjacobson: Follow up: add instance ID as a primary audit field

packages/db/src/migrations/0086_misty_beast.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "kiloclaw_cli_runs" ADD COLUMN "initiated_by_admin_id" text;--> statement-breakpoint
+ALTER TABLE "kiloclaw_cli_runs" ADD CONSTRAINT "kiloclaw_cli_runs_initiated_by_admin_id_kilocode_users_id_fk" FOREIGN KEY ("initiated_by_admin_id") REFERENCES "public"."kilocode_users"("id") ON DELETE set null ON UPDATE no action NOT VALID;--> statement-breakpoint
```
evanjacobson, line 2: Need to validate the `NOT VALID`...`VALIDATE` block works as expected & won't cause problems.

apps/web/src/routers/admin-kiloclaw-instances-router.ts
```diff
@@ -150,17 +159,23 @@ function throwKiloclawAdminError(
...
         code: 'NOT_FOUND',
         message: `Instance ${instanceId} not found`,
       });
     }
-    return instance;
+    return null;
+  }
+
+  if (instance.userId !== userId) {
+    throw new TRPCError({ code: 'NOT_FOUND', message: 'Instance not found' });
```
evanjacobson, line 184: This ensures the instanceId actually belongs to the userId in the request. Admins can still operate on any user's instance, they just have to pass the correct userId that matches the instance owner. This prevents auditing/scoping bugs.

evanjacobson (commented):
Submitting pending review.

services/kiloclaw/src/routes/platform.ts
evanjacobson: The diff already leverages helper methods — that is the refactor:

1. **`isRecord`** / **`isHttpStatus`** / **`hasStringCode`**: type guards replacing inline `as` casts (banned by project coding standards)
2. **`getErrorCode(err)`**: extracts a string `code` from `err` or `err.cause`, replacing the 7-line inline narrowing block in `sanitizeOpenclawConfigError`
3. **`statusCodeFromError`** now also walks `err.cause` so status codes survive error-wrapping boundaries (e.g. DO RPC). Before, a wrapped error with `.cause.status = 404` would default to 500

In short: the inline `typeof`/`as` narrowing was duplicated in `statusCodeFromError` and `sanitizeOpenclawConfigError`. The helpers deduplicate it and drop the `as` casts.

packages/db/src/migrations/0086_misty_beast.sql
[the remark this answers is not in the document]
evanjacobson: Verified — the schema and generated migration are already correct.

The migration (`0086_misty_beast.sql`) already uses the safe pattern:

```sql
ALTER TABLE ... ADD CONSTRAINT ... NOT VALID;
ALTER TABLE ... VALIDATE CONSTRAINT ...;
```

`VALIDATE CONSTRAINT` acquires a `SHARE UPDATE EXCLUSIVE` lock, which does **not** conflict with `ROW EXCLUSIVE` (INSERT/UPDATE/DELETE). Normal DML continues during validation. No changes needed in `schema.ts`.

apps/web/src/lib/kiloclaw/cli-runs.ts
```diff
@@ -0,0 +1,269 @@
...
+
+function getCliRunScopeConditions(params: {
+  runId: string;
+  userId: string;
+  instanceId: string | null;
+}): SQL[] {
+  return [
+    eq(kiloclaw_cli_runs.id, params.runId),
+    eq(kiloclaw_cli_runs.user_id, params.userId),
+    params.instanceId === null
```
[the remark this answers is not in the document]
evanjacobson: This concern is incorrect. `getCliRunScopeConditions` does **not** treat `instanceId: null` as `instance_id IS NULL`. When `params.instanceId` is null, the instance filter is **omitted entirely** (line 130–132), so the query is `id = ? AND user_id = ?` — this is intentionally broader and finds runs from any instance, including destroyed ones.

The UPDATE helpers (`markCliRunCancelled`, `persistCliRunControllerStatus`) are called with `row.instance_id` (the actual value from the DB row), not `params.instanceId`. See lines 180 and 244.

apps/web/src/lib/kiloclaw/cli-runs.ts
```diff
@@ -0,0 +1,272 @@
...
+    .where(and(...getCliRunScopeConditions(params)))
+    .limit(1);
+
+  if (!row) {
+    return { ok: false, runFound: false, cancelled: false };
+  }
+
+  if (row.status !== 'running') {
+    return { ok: true, runFound: true, cancelled: false };
+  }
```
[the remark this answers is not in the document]
evanjacobson: Already handled. Lines 170–172 resolve the worker instance ID from `row.instance_id` when `params.workerInstanceId` is undefined:

```typescript
const effectiveWorkerInstanceId =
  params.workerInstanceId ??
  (row.instance_id ? await resolveWorkerInstanceId(row.instance_id) : undefined);
```

If the DB row has an `instance_id` (destroyed instance), it resolves the worker instance ID from that row value — not from the null `params.workerInstanceId`.

apps/web/src/lib/kiloclaw/cli-runs.ts
```diff
@@ -0,0 +1,284 @@
...
+    (row.instance_id ? await resolveWorkerInstanceId(row.instance_id) : undefined);
+
+  const client = new KiloClawInternalClient();
+  const result = await client.cancelKiloCliRun(params.userId, effectiveWorkerInstanceId);
+
+  if (result.ok) {
+    await markCliRunCancelled({
+      runId: params.runId,
+      userId: params.userId,
+      instanceId: params.instanceId,
```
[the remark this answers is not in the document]
evanjacobson: This is incorrect. Line 180 passes `row.instance_id`, not `params.instanceId`:

```typescript
await markCliRunCancelled({
  runId: params.runId,
  userId: params.userId,
  instanceId: row.instance_id,  // from the DB row, not params
});
```

If the row was from a destroyed instance with `instance_id = "abc-123"`, then `markCliRunCancelled` receives `"abc-123"` and the WHERE clause correctly matches `instance_id = "abc-123"`.

apps/web/src/lib/kiloclaw/cli-runs.ts
```diff
@@ -0,0 +1,284 @@
...
+      prompt: row.prompt,
+      initiatedBy: getCliRunInitiatedBy(row),
+    };
+  }
+
+  if (shouldPersistCliRunControllerStatus(row, controllerStatus)) {
+    await persistCliRunControllerStatus({
+      runId: params.runId,
+      userId: params.userId,
+      instanceId: params.instanceId,
```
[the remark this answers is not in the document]
evanjacobson: Same as the cancel case — this was already fixed. Line 244 passes `row.instance_id`:

```typescript
await persistCliRunControllerStatus({
  runId: params.runId,
  userId: params.userId,
  instanceId: row.instance_id,  // from the DB row, not params
  controllerStatus,
});
```

The bot flagged the old code (commit `2e08be91c`) which used `params.instanceId`. The subsequent commits fixed both the cancel and persist paths to use `row.instance_id`.

packages/db/src/migrations/0086_misty_beast.sql
evanjacobson: Based on my research, this will be good to go.

PR #2386, opened 2026-04-13
feat(kiloclaw): enable Exa provider defaults in runtime image

joshavant:
## Summary
- Add KiloClawCustomizer runtime support for Exa web search gating, including plugin config schema and `createTool()` soft-disable behavior.
- Update controller config writer to handle Exa mode as tri-state (`kilo-proxy` / `disabled` / unset), preserve explicit providers, and auto-assign `kilo-exa` when provider is missing on upgraded images.
- Extend controller tests for provider auto-assignment and preserve/disable paths.

## Verification
- `pnpm --filter kiloclaw test -- controller/src/config-writer.test.ts`
- `pnpm test -- src/kilo-exa-web-search-provider.test.ts` (in `services/kiloclaw/plugins/kiloclaw-customizer`)
- `git push -u origin feat/kiloclaw-exa-image-runtime` (pre-push checks passed: `oxfmt --list-different .` and affected-package typecheck)

## Visual Changes
- N/A

## Reviewer Notes
- This PR is intentionally image/runtime-only so it can deploy before dashboard exposure.
- Follow-up dashboard PR will gate Exa UI by controller calver and wire user-facing controls.
- Migration default is intentional: when `tools.web.search.provider` is unset, the image auto-assigns `kilo-exa` even if `tools.web.search.enabled` had been `false`; explicit provider selections are still preserved.

### DIFF
--- a/services/kiloclaw/controller/src/config-writer.ts
+++ b/services/kiloclaw/controller/src/config-writer.ts
@@ -321,7 +321,10 @@ export function generateBaseConfig(
     customizerWebSearchConfig.enabled = false;
 
     const braveConfigured = Boolean(env.BRAVE_API_KEY?.trim());
-    if (braveConfigured) {
+    if (
+      braveConfigured &&
+      (!hasExplicitSearchProvider || config.tools?.web?.search?.provider === KILO_EXA_PROVIDER_ID)
+    ) {
       config.tools = config.tools ?? {};
       config.tools.web = config.tools.web ?? {};
       config.tools.web.search = config.tools.web.search ?? {};
--- a/services/kiloclaw/controller/src/config-writer.ts
+++ b/services/kiloclaw/controller/src/config-writer.ts
@@ -55,6 +55,24 @@ const ONBOARD_FLAGS = [
   '--skip-health',
 ] as const;
 
+const KILOCLAW_CUSTOMIZER_PLUGIN_ID = 'kiloclaw-customizer';
+const KILOCLAW_CUSTOMIZER_PLUGIN_PATH = '/usr/local/lib/node_modules/@kiloclaw/kiloclaw-customizer';
+const KILO_EXA_PROVIDER_ID = 'kilo-exa';
+
+type KiloExaSearchMode = 'kilo-proxy' | 'disabled';
+
+function resolveKiloExaSearchMode(value: string | undefined): KiloExaSearchMode {
+  const normalized = value?.trim().toLowerCase();
+  if (normalized === 'kilo-proxy') {
+    return 'kilo-proxy';
+  }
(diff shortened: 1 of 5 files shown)


services/kiloclaw/controller/src/config-writer.ts
```diff
@@ -261,14 +284,59 @@ export function generateBaseConfig(
...
+  } else if (hasExplicitSearchProvider) {
+    customizerWebSearchConfig.enabled =
+      config.tools?.web?.search?.provider === KILO_EXA_PROVIDER_ID;
+  } else {
+    customizerWebSearchConfig.enabled = true;
+    config.tools = config.tools ?? {};
+    config.tools.web = config.tools.web ?? {};
+    config.tools.web.search = config.tools.web.search ?? {};
+    config.tools.web.search.enabled = true;
+    config.tools.web.search.provider = KILO_EXA_PROVIDER_ID;
```
RSO: I think this is the same as the branch on line 302? If so, can't we just write `if (kiloExaSearchMode === "kilo-proxy" || kiloExaSearchmode === "unset")`?

RSO (approved):

services/kiloclaw/controller/src/config-writer.ts
joshavant: Good call. Consolidated this a little better.

services/kiloclaw/controller/src/config-writer.ts
```diff
@@ -272,14 +295,56 @@ export function generateBaseConfig(
...
-    (config.plugins.load.paths as string[]).push(customizerPluginPath);
+  if (!(config.plugins.load.paths as string[]).includes(KILOCLAW_CUSTOMIZER_PLUGIN_PATH)) {
+    (config.plugins.load.paths as string[]).push(KILOCLAW_CUSTOMIZER_PLUGIN_PATH);
   }
   config.plugins.entries = config.plugins.entries ?? {};
-  const customizerEntry = 'kiloclaw-customizer';
-  config.plugins.entries[customizerEntry] = config.plugins.entries[customizerEntry] ?? {};
-  config.plugins.entries[customizerEntry].enabled = true;
+  config.plugins.entries[KILOCLAW_CUSTOMIZER_PLUGIN_ID] =
+    config.plugins.entries[KILOCLAW_CUSTOMIZER_PLUGIN_ID] ?? {};
```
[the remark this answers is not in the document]
joshavant: This is intentional for this migration path: when provider is unset, we explicitly default to `kilo-exa` on upgraded images, even if `tools.web.search.enabled` was previously `false`. The preserve rule applies to explicit provider selections, which remain untouched. I added a regression test (`auto-assigns kilo-exa even when web search was explicitly disabled but provider is missing`) to lock this behavior.

PR #2389, opened 2026-04-14
fix(kiloclaw): preserve user-edited IDENTITY.md across redeploys

### LINKED ISSUE #2387: IDENTITY.md is overwritten on every redeploy, clobbering user edits
## Summary

`/root/.openclaw/workspace/IDENTITY.md` is regenerated from env vars on every controller boot, so any edits a user makes to the file are lost on the next redeploy.

## Expected

If a user edits `IDENTITY.md`, the edits should persist across redeploys — matching the behavior of `TOOLS.md`, which is only seeded on first provision.

## Actual

Every boot overwrites `IDENTITY.md` with content freshly generated from `KILOCLAW_BOT_NAME` / `KILOCLAW_BOT_NATURE` / `KILOCLAW_BOT_VIBE` / `KILOCLAW_BOT_EMOJI`.

## Code path

In `services/kiloclaw/controller/src/bootstrap.ts`:

1. `bootstrap()` → `bootstrapNonCritical()` runs the `onboard`/`doctor` step on every start.
2. `runOnboardOrDoctor()` ends with an unconditional call to `writeBotIdentityFile(env, deps)` — it runs in *both* the fresh-install branch and the existing-config branch.
3. `writeBotIdentityFile()` calls `atomicWrite(IDENTITY_MD_DEST, formatBotIdentityMarkdown(env), ...)`, unconditionally replacing the file.

## Asymmetry with TOOLS.md

`TOOLS.md` is seeded only inside the `if (!configExists)` branch, so user edits there survive redeploys. `IDENTITY.md` is written *outside* that guard, so it doesn't.

```ts
if (!configExists) {
  // ...
  if (deps.existsSync(TOOLS_MD_SOURCE)) {
    deps.mkdirSync(path.dirname(TOOLS_MD_DEST), { recursive: true });
    deps.copyFileSync(TOOLS_MD_SOURCE, TOOLS_MD_DEST);
  }
} else {
  // doctor path
}

writeBotIdentityFile(env, deps); // ← runs every boot
```

## Suggested fix

Either:

- **(a)** Move `writeBotIdentityFile(env, deps)` inside the `!configExists` branch so it only runs on first provision (mirrors `TOOLS.md`), or
- **(b)** Inside `writeBotIdentityFile`, skip the write if `IDENTITY_MD_DEST` already exists — preserves any existing file (including user edits) while still seeding on first boot.

(b) is slightly friendlier if the env-var-driven identity should still be able to seed a fresh container even when a config already exists for other reasons.

kvnkho:
## Summary

Fixes #2387.

`writeBotIdentityFile` runs on every controller boot (via `runOnboardOrDoctor`), and previously always overwrote `/root/.openclaw/workspace/IDENTITY.md` with content regenerated from `KILOCLAW_BOT_*` env vars. Any edits a user made to `IDENTITY.md` were lost on the next redeploy.

This PR mirrors the `TOOLS.md` behavior: seed on first boot, but skip the write when the file already exists. Legacy-identity-file cleanup still runs unconditionally.

## Verification

- [x] `pnpm exec vitest run controller/src/bootstrap.test.ts` in `services/kiloclaw` — 63 passed (including a new test asserting `IDENTITY.md` is not rewritten when it already exists).
- [x] `pnpm run typecheck` in `services/kiloclaw` — clean.
- [x] `scripts/lint-all.sh` — clean.
- [x] `pnpm format:check` — clean.

## Visual Changes

N/A

## Reviewer Notes

Considered an alternative (guarding the call site inside `runOnboardOrDoctor` to only run on fresh installs), but guarding inside `writeBotIdentityFile` itself keeps the seeding path working even if a config already exists for other reasons, and keeps the rule in one place. Open to flipping if reviewers prefer the other shape.

### DIFF
--- a/services/kiloclaw/controller/src/bootstrap.test.ts
+++ b/services/kiloclaw/controller/src/bootstrap.test.ts
@@ -587,6 +587,25 @@ describe('writeBotIdentityFile', () => {
       ['/root/.openclaw/workspace/BOOTSTRAP.md'],
     ]);
   });
+
+  it('preserves existing IDENTITY.md across redeploys', () => {
+    const harness = fakeDeps();
+    (harness.deps.existsSync as ReturnType<typeof vi.fn>).mockImplementation(
+      (p: string) => p === '/root/.openclaw/workspace/IDENTITY.md'
+    );
+
+    writeBotIdentityFile(
+      { KILOCLAW_BOT_NAME: 'Milo', KILOCLAW_BOT_NATURE: 'Operator' },
+      harness.deps
+    );
+
+    expect(
+      harness.renameCalls.some(call => call.to === '/root/.openclaw/workspace/IDENTITY.md')
+    ).toBe(false);
+    expect(harness.writeCalls.some(call => call.path.includes('/workspace/IDENTITY.md'))).toBe(
+      false
+    );
+  });
 });
 
 // ---- runOnboardOrDoctor ----
--- a/services/kiloclaw/controller/src/bootstrap.ts
+++ b/services/kiloclaw/controller/src/bootstrap.ts
@@ -285,12 +285,17 @@ export function writeBotIdentityFile(
     'mkdirSync' | 'writeFileSync' | 'renameSync' | 'unlinkSync' | 'existsSync'
   > = defaultDeps
 ): void {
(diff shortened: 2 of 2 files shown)


pandemicsyn:
Apologies we didn't see this sooner, we actually fixed this as well earlier this week. https://github.com/Kilo-Org/cloud/pull/2836

PR #2408, opened 2026-04-14
feat(models): include non-monitored models in health check response

markijbema:
## Summary

- Instead of excluding Gemini from the `/api/models/up` endpoint entirely, it is now included with a `monitored: false` flag. Non-monitored models still get traffic data queried but cannot trigger unhealthy status or 503 responses — they're always reported as `healthy: true`.
- The admin model status page (`/admin/alerting?tab=model-status`) now shows non-monitored models in a clearly separated section below the monitored ones, at reduced opacity with an "informational only" label.
- The `HEALTH_CHECK_EXCLUSIONS` set remains the single source of truth for which models are non-monitored.

## Verification

- [x] `pnpm typecheck` — no new errors (pre-existing `linkify-it` error unrelated)
- [x] `pnpm format` — clean

## Visual Changes

The admin model status table now has a second section below the main table for non-monitored models, rendered at 75% opacity with "Non-Monitored (informational only)" header and subtitle "Traffic data shown but excluded from health alerting".

## Reviewer Notes

- The `ModelHealthMetrics` type gains a `monitored: boolean` field. This is additive and backward-compatible — the admin page uses `metrics.monitored === false` to classify, so older snapshots without the field default to monitored.
- The SQL query now includes all preferred models (monitored + non-monitored) in its `IN (...)` clause, so we get real traffic data for Gemini.

### DIFF
--- a/apps/web/src/app/admin/alerting/ModelStatusContent.tsx
+++ b/apps/web/src/app/admin/alerting/ModelStatusContent.tsx
@@ -193,86 +193,70 @@ export function ModelStatusContent() {
       ) : monitoredModels.length === 0 && nonMonitoredModels.length === 0 ? (
         <div className="text-muted-foreground py-8 text-center">No monitored models found.</div>
       ) : (
-        <div className="flex flex-col gap-y-6">
-          <div className="overflow-x-auto">
-            <Table>
-              <TableHeader>
-                <TableRow>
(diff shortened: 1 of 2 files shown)


markijbema:
CleanShot 2026-04-14 at 15 33 36

chrarnoldus (approved):

PR #2417, opened 2026-04-14
feat(mobile): add cloud agent chat experience

eshurakov:
## Summary

- Adds the mobile Cloud Agent chat experience, including the Agents tab session list, chat detail flow, new-session screen, composer, tool/message renderers, and model/mode/repository pickers.
- Extends the shared Cloud Agent SDK so mobile can configure runtime dependencies, preserve message IDs across mobile/backend sends, reconcile optimistic user text with authoritative backend parts, and keep message ordering consistent with millisecond timestamps.
- Updates CLI session APIs and live transport behavior to support mobile session listing, recent repository filters, remote session startup, websocket headers, and custom agent question answers.
- Adjusts mobile app configuration for the new agent flow, including Metro/Sentry setup, environment keys, tRPC client headers, Expo secure-store persistence for filters, and navigation layout changes.
- Adds regression coverage for SDK runtime configuration, connection lifecycle idempotency, optimistic-message reconciliation, storage helpers, and fixture-runner scenarios.

## Verification

- [x] Manually checked in iOS simulator and iOS device


iscekic (commented):
LLM review, feel free to ignore if not valid.

---

**1. `OrganizationProvider` — stale org ID after sign-out** (`apps/mobile/src/lib/organization-context.tsx`)

`const preloaded = SecureStore.getItemAsync(STORAGE_KEY)` fires once at module load. The `useEffect` does `await preloaded`, which returns the already-resolved value forever — even after `setOrganizationId(null)` deletes the key from SecureStore. If the provider is unmounted and remounted without a full app restart (e.g. sign-out → sign-in flow), the old user's org ID leaks through.

**Suggestion:** Move the `SecureStore.getItemAsync` call inside the `useEffect` so each mount reads fresh data.

---

**2. `withCloudAgentDiagnostics` is a no-op** (`apps/mobile/src/components/agents/mobile-session-diagnostics.ts`)

This wrapper is called for every API operation (`send`, `interrupt`, `answer`, `reject`, `permission`, `prepare`, `initiate`, `getTicket`) but ignores its `action` and `organizationId` parameters (underscored), adds a gratuitous `await Promise.resolve()`, and just calls `run()` directly. The companion function `formatSafeCloudAgentFailureDiagnostic` exists in the same file and is clearly intended to be used here but was never wired up.

Errors DO propagate normally (no try/catch), so this isn't an error suppressor — it's dead infrastructure. Either wire up the intended diagnostics or remove the wrapper.

---

**3. `SessionDetailScreen` — query errors silently treated as personal session** (`apps/mobile/src/app/(app)/agent-chat/[session-id].tsx`)

When `routeOrganizationId` is undefined and the `cliSessionsV2.get` query errors, TanStack Query v5 sets `status = 'error'` so `isPending = false`. The guard falls through, `sessionQuery.data` is undefined, and `organizationId` resolves to `undefined` — rendering the session as personal with no error indication.

**Suggestion:** Add an `isError` guard before rendering `AgentSessionProvider`.

---

**4. `useAvailableModels` bypasses tRPC** (`apps/mobile/src/lib/hooks/use-available-models.ts`)

This hook reads `'auth-token'` directly from SecureStore and does a raw `fetch` — duplicating auth logic, hardcoding the token key (not imported from `auth-context.tsx`), and bypassing tRPC batching/refresh mechanisms. If the token key name changes, this silently stops authenticating.

---

**5. `tsconfig.json` path aliases shadow mobile paths** (`apps/mobile/tsconfig.json`)

`@/types/*` and `@/lib/cloud-agent-next/*` resolve to `apps/web/src/` before the catch-all `@/*` → `./src/*`. This is intentional for the SDK's cross-package imports, but any future mobile file at `src/types/` or `src/lib/cloud-agent-next/` will be silently resolved from the web app instead. Minor footgun worth being aware of.

---

**6. `upsertInitialMessageFromFetchedMetadata` uses `Date.now()`** (`apps/web/src/lib/cloud-agent-sdk/session-manager.ts`)

The optimistic initial message gets the client's current time (`Date.now()`), which may differ from the server's actual `time.created` that arrives later via `message.updated`. Self-correcting when the server event replaces it, but could briefly cause an ordering glitch. Minor.

---

**7. Android rename is a dead-end** (`apps/mobile/src/components/agents/session-row.tsx`)

The "Rename" option appears in the Android action sheet but shows `Alert.alert('Rename Session', 'Rename is not supported on this device.')`. Should either hide the option on Android or show a modal with a `TextInput`.

apps/mobile/src/components/agents/mobile-session-manager.ts
```diff
@@ -0,0 +1,277 @@
...
+              skipBatchOptions
+            );
+            return;
+          }
+          await trpcClient.cloudAgentNext.answerPermission.mutate(input, skipBatchOptions);
+        });
+      },
+    },
+    prepare: async input => {
+      await Promise.resolve();
```
iscekic: Why do we need this?

apps/mobile/src/components/agents/session-row.tsx
```diff
@@ -0,0 +1,204 @@
...
+            }
+          },
+        },
+      ],
+      'plain-text',
+      currentTitle
+    );
+  } else {
+    // Android doesn't support Alert.prompt — use a simple Alert as fallback
+    Alert.alert('Rename Session', 'Rename is not supported on this device.');
```
iscekic: We should have a proper fallback for Android, I think the bot flagged this too.

apps/mobile/src/components/agents/use-session-auto-scroll.ts
```diff
@@ -0,0 +1,124 @@
...
+  );
+
+  const handleListLayout = useCallback(() => {
+    if (shouldAutoScrollRef.current) {
+      scheduleScrollToLatestMessage();
+    }
+  }, [scheduleScrollToLatestMessage]);
+
+  return {
+    flatListRef,
```
iscekic, line 118: I'm assuming perf is fine, in the past I've used https://shopify.github.io/flash-list/ for large virtual lists and it worked well.

apps/mobile/src/lib/auth/auth-context.tsx
```diff
@@ -50,6 +50,9 @@ export function AuthProvider({ children }: { readonly children: ReactNode }) {
 
   const signOut = useCallback(async () => {
     await SecureStore.deleteItemAsync(TOKEN_KEY);
+    // Clear per-user preferences so they don't leak to the next signed-in account
+    await SecureStore.deleteItemAsync('selected-organization');
+    await SecureStore.deleteItemAsync('agent-session-filters');
```
iscekic: I guess we should use constants for these (export/import them)

apps/mobile/src/components/agents/mobile-session-manager.ts
```diff
@@ -0,0 +1,279 @@
...
+        }
+        if (isRemote) {
+          return { type: 'remote', kiloSessionId };
+        }
+        return { type: 'read-only', kiloSessionId };
+      } catch {
+        return {
+          type: 'cloud-agent',
+          kiloSessionId,
+          cloudAgentSessionId: kiloSessionId as unknown as CloudAgentSessionId,
```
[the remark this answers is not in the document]
eshurakov: Already addressed — the catch block now returns `{ type: 'read-only', kiloSessionId }` (line 66), not `cloud-agent`. This was fixed in a previous revision.

apps/mobile/src/components/agents/mobile-session-manager.ts
```diff
@@ -0,0 +1,279 @@
...
+            method: 'POST',
+            headers: {
+              'Content-Type': 'application/json',
+              ...(token ? { Authorization: `Bearer ${token}` } : {}),
+            },
+            body: JSON.stringify(body),
+          }
+        );
+        if (!response.ok) {
+          const errorData = (await response.json()) as { error?: string };
```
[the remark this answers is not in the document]
eshurakov: Already addressed — the body is now parsed once at line 88 (`const data = (await response.json())`), then `response.ok` is checked afterward. No double-read.

apps/mobile/src/app/(app)/(tabs)/_layout.tsx
```diff
@@ -45,18 +46,12 @@ export default function TabsLayout() {
         name="(2_agents)"
         options={{
           title: 'Agents',
+          href: hasCloudAgentAccess ? undefined : null,
```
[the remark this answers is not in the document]
eshurakov: Already addressed — the tab no longer uses `href: null`. Instead, `tabPress` listener calls `e.preventDefault()` when `!hasCloudAgentAccess` (lines 62-69). Deep link route guards are a valid follow-up concern but beyond the scope of tab layout gating.

apps/mobile/src/components/agents/session-list-screen.tsx
```diff
@@ -0,0 +1,266 @@
...
+    isError,
+    refetch,
+  } = useAgentSessions({
+    createdOnPlatform,
+    gitUrl,
+    enabled: filtersLoaded,
+  });
+  const { data: recentRepositories } = useRecentAgentRepositories({ enabled: filtersLoaded });
+
+  useFocusEffect(
```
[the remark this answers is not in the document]
eshurakov: Already addressed — `refetch` is stored in a ref (lines 85-88) and the `useFocusEffect` callback has empty deps (`useCallback(() => { void refetchRef.current(); }, [])`), so it only fires on focus change, not on every render.

apps/mobile/src/lib/organization-context.tsx
```diff
@@ -0,0 +1,64 @@
...
+  createContext,
+  type ReactNode,
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+
+const STORAGE_KEY = 'selected-organization';
```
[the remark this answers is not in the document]
eshurakov: Fixed — moved `SecureStore.getItemAsync` from module-level `preloaded` promise into the `useEffect` body so each mount reads fresh data. Also added cleanup flag to prevent state updates on unmounted components. If sign-out calls `setOrganizationId(null)` (which deletes the key), the next mount starts clean for the new user.

eshurakov:
Addressing iscekic's review points:

**1. OrganizationProvider stale preloaded promise** — Fixed. Moved `SecureStore.getItemAsync` from module-level into the `useEffect` body with a cancellation flag. Each mount now reads fresh data from SecureStore.

**2. `withCloudAgentDiagnostics` is a no-op** — Acknowledged. This is scaffolding left for future diagnostic instrumentation (e.g. structured logging, Sentry breadcrumbs). `formatSafeCloudAgentFailureDiagnostic` is actively used in `onSendFailed`. The wrapper is intentionally a passthrough for now — wiring it up requires deciding on the diagnostic backend, which is out of scope here.

**3. SessionDetailScreen error handling** — Fixed. Added an `isError` guard that shows an error screen with ScreenHeader + retry button, instead of silently falling through to personal context.

**4. `useAvailableModels` bypasses tRPC** — Intentional. The endpoints (`/api/openrouter/models`, `/api/organizations/:id/models`) are existing Next.js API routes that return a different shape than what a tRPC procedure would provide. This was a deliberate decision to avoid adding new tRPC procedures just for mobile model listing when the API routes already exist and work.

**5. `tsconfig.json` path aliases** — Noted. The `@/types/*` and `@/lib/cloud-agent-next/*` aliases resolve to `apps/web/src/` intentionally for cross-package SDK imports. No mobile files exist at those paths today and we would catch a conflict via typecheck if one were added.

**6. `Date.now()` in optimistic messages** — Acknowledged. Self-correcting when the server event replaces the optimistic message. No change needed.

**7. Android rename dead-end** — Fixed. Removed the rename option from the Android action sheet entirely (Android lacks `Alert.prompt`). The iOS action sheet and `Alert.prompt` flow are unchanged.

apps/mobile/src/lib/hooks/use-app-lifecycle.ts
```diff
@@ -0,0 +1,28 @@
...
+      setIsActive(nextState === 'active');
+    });
+    return () => {
+      subscription.remove();
+    };
+  }, []);
+
+  useEffect(() => {
+    const unsubscribe = addEventListener(state => {
+      setIsConnected(state.isConnected ?? true);
```
iscekic: I can't remember if this is reliable for networks with no internet connectivity - worth doing a small ping every x seconds instead? Your call.

apps/mobile/src/lib/hooks/use-app-lifecycle.ts
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useState } from 'react';
+import { AppState } from 'react-native';
+import { addEventListener } from '@react-native-community/netinfo';
+
+export function useAppLifecycle() {
```
iscekic, line 5: We could reuse this hook in kiloclaw

apps/mobile/src/lib/hooks/use-cloud-agent-access.ts
```diff
@@ -0,0 +1,17 @@
+import { useQuery } from '@tanstack/react-query';
+
+import { useTRPC } from '@/lib/trpc';
+
+/**
+ * Cloud agents are currently limited to internal kilocode.ai users.
```
iscekic: This file will be removed

apps/mobile/src/lib/config.ts
```diff
@@ -15,3 +15,9 @@ export const API_BASE_URL: string = required('apiBaseUrl');
 export const WEB_BASE_URL: string = required('webBaseUrl');
 export const APPSFLYER_DEV_KEY: string = required('appsFlyerDevKey');
 export const APPSFLYER_APP_ID: string = required('appsFlyerAppId');
+
+export const CLOUD_AGENT_WS_URL: string =
+  (extra?.cloudAgentWsUrl as string | undefined) ?? API_BASE_URL.replace(/^http/, 'ws');
+
+export const SESSION_INGEST_WS_URL: string =
```
iscekic: I think we should always make env vars `required` (don't forget to add them in expo)

apps/mobile/metro.config.js
iscekic, line 1: Lots of changes in this file, I'd double-check if they're really needed. Biggest one I see is, if we're importing from `cloud-agent-sdk`, I expect pnpm to resolve the deps (or for the imported package to have peer deps we provide).

apps/mobile/package.json
```diff
@@ -4,8 +4,8 @@
   "version": "1.0.0",
   "scripts": {
     "start": "expo start --dev-client",
-    "android": "expo start --android",
-    "ios": "expo start --ios",
+    "android": "expo run:android",
+    "ios": "expo run:ios",
```
iscekic: Double-check

iscekic (commented):
Nothing too major.

apps/mobile/src/components/agents/mobile-session-manager.ts
eshurakov: Removed. The `await Promise.resolve()` calls were workarounds for conflicting lint rules (`require-await` vs `promise-function-async` vs `arrow-body-style`). Refactored all API callbacks to use `async` block bodies with `await withCloudAgentDiagnostics(...)` directly, which satisfies all three rules without a no-op yield.

apps/mobile/src/components/agents/session-row.tsx
eshurakov: Added a `Modal` + `TextInput` rename flow for Android. Tapping Rename in the Android Alert opens a modal with `defaultValue`, ref-based text tracking, and Rename/Cancel buttons. `onSubmitEditing` also confirms. iOS still uses `Alert.prompt`.

apps/mobile/src/lib/auth/auth-context.tsx
eshurakov: Fixed. Exported `ORGANIZATION_STORAGE_KEY` from `organization-context.tsx` and `SESSION_FILTERS_KEY` from `use-persisted-agent-session-filters.ts`, then imported both in `auth-context.tsx` signOut.

apps/mobile/src/components/agents/use-session-auto-scroll.ts
eshurakov: Noted. FlatList is fine for message lists at current scale. FlashList is a reasonable optimization if we hit perf issues with large session histories — will keep it in mind.

apps/mobile/src/lib/hooks/use-app-lifecycle.ts
eshurakov: Changed to `state.isInternetReachable ?? state.isConnected ?? true`. `isInternetReachable` checks actual internet connectivity (not just interface presence), falling back to `isConnected` when netinfo returns `null` during initial fetch.

apps/mobile/src/lib/hooks/use-app-lifecycle.ts
eshurakov: Checked — kiloclaw is a Cloudflare Worker (server-side), so it has no use for AppState/netinfo. However, `apps/mobile/src/components/kiloclaw/chat.tsx` has 3 independent `AppState.addEventListener` calls that duplicate what `useAppLifecycle` provides. Worth consolidating those as a follow-up. The cloud-agent-sdk also exports `ConnectionLifecycleHooks` for WS reconnection on app lifecycle events — the mobile session manager does not pass a mobile implementation yet, so there is a gap for automatic WS reconnection on foreground/network recovery.

apps/mobile/src/lib/hooks/use-cloud-agent-access.ts
eshurakov: Noted — will leave as-is for this PR.

apps/mobile/src/lib/config.ts
eshurakov: Changed both to `required()`. The fallback `API_BASE_URL.replace(/^http/, 'ws')` would produce `wss://api.kilo.ai` which is wrong — the actual WS services are on different subdomains (`cloud-agent-next.kilosessions.ai`, `ingest.kilosessions.ai`). Both values are already in `apps/mobile/.env` and `env-keys.js`.

apps/mobile/metro.config.js
eshurakov: All changes are needed because `cloud-agent-sdk` is not a proper workspace package — it is a directory alias pointing to `apps/web/src/lib/cloud-agent-sdk/`. Each config serves a specific purpose:

- `watchFolders = [monorepoRoot]` — pnpm hoists real package paths outside `apps/mobile/`, so Metro needs the monorepo root to resolve them
- `nodeModulesPaths` — SDK deps (jotai, zod) live in root `node_modules`, not `apps/mobile/node_modules`
- `extraNodeModules` — maps bare `cloud-agent-sdk` imports to the web app directory
- `resolveRequest` — remaps `@/` imports *originating from within the SDK* back to `apps/web/src/`, since the SDK uses the web app's `@/` alias internally

Making `cloud-agent-sdk` a proper workspace package would eliminate all of this, but that is a larger refactor. For now these are all required for Metro to bundle the SDK.

apps/mobile/package.json
eshurakov: Verified. `expo start --ios/--android` targets Expo Go; `expo run:ios/android` builds and runs a dev client. Since we use dev builds (per AGENTS.md), this change is correct. The PR also adds `expo-crypto` (for `generateMessageId` UUID generation) and `jotai` (SDK's session state store) as direct dependencies.

apps/mobile/src/lib/auth/auth-context.tsx
iscekic: We can consider centralizing keys in a single file - up to you.

apps/mobile/src/lib/hooks/use-app-lifecycle.ts
iscekic: Yeah that's what I meant - we could replace the existing event listeners with this reusable hook. Small change, could be done in this PR.

apps/mobile/src/lib/hooks/use-cloud-agent-access.ts
iscekic: Not will - should remove. 😄

apps/mobile/package.json
iscekic: But isn't start accurate? We build a dev client on expo, download it to the device, which then talks to the dev server running on your pc.

eshurakov: I've reverted it

apps/mobile/src/lib/hooks/use-cloud-agent-access.ts
eshurakov: removed!

apps/mobile/src/app/(app)/(tabs)/_layout.tsx
```diff
@@ -46,17 +45,10 @@ export default function TabsLayout() {
...
-            color: colors.mutedForeground,
-            fontSize: 9,
-          },
         }}
         listeners={{
-          tabPress: e => {
-            e.preventDefault();
-            toast('Agents is coming soon');
+          tabPress: () => {
+            void Haptics.selectionAsync();
```
iscekic, line 51: Let's add a directive to `AGENTS.md` for the project - haptics should be used where appropriate.

apps/mobile/src/lib/config.ts
```diff
@@ -15,3 +15,6 @@ export const API_BASE_URL: string = required('apiBaseUrl');
 export const WEB_BASE_URL: string = required('webBaseUrl');
 export const APPSFLYER_DEV_KEY: string = required('appsFlyerDevKey');
 export const APPSFLYER_APP_ID: string = required('appsFlyerAppId');
+
+export const CLOUD_AGENT_WS_URL: string = required('cloudAgentWsUrl');
+export const SESSION_INGEST_WS_URL: string = required('sessionIngestWsUrl');
```
iscekic, line 20: Double-check - did you add these in expo?

apps/mobile/src/lib/storage-keys.ts
iscekic, line 1: Double-check - are all keys in here?

apps/mobile/src/components/notifications-card.tsx
iscekic, line 1: Follow-up (and I think this will be pretty cool):
1. add a "cloud agent" toggle for notifications in the profile settings panel
2. integrate cloud-agent -> notifications service (send notification when agent turn is over)
3. handle the notification in-app
4. obtain user permission for notifications (see pattern I used for kiloclaw or do something better, your call)

iscekic (approved):
Small remarks, approving to unblock.

apps/mobile/src/lib/config.ts
eshurakov: Just added!

PR #2421, opened 2026-04-14
Fix org display of volume usage

evanjacobson:
## Summary

Tl;dr: This PR fixes a bug where org instances were displaying personal instance's volume usage data due to the handling of orgs in getUserFromAuth (orgs are not retained by that method in web).

The Instance tab previously fetched volume usage via a raw API route. This route relied on `getUserFromAuth` to determine the active organization, but browser session auth does not automatically populate the organization context. As a result, viewing an organization's instance would incorrectly display the user's personal instance disk usage instead.

This PR resolves the issue by migrating the disk usage fetch to tRPC. It adds `getDiskUsage` endpoints to both the personal and organization KiloClaw routers and replaces the raw fetch with a new `useClawDiskUsage` hook. By passing the explicit organizationId from the frontend, we guarantee the correct instance data is loaded. 

<details><summary>More Info</summary>
<p>

`getUserFromAuth` has two paths: extension requests read `organizationId` from the `X-KiloCode-OrganizationId` header, but browser session requests always pass `undefined` (line 863 of `user.server.ts` even has a comment about this). The dashboard always takes the browser path, so the API route always resolved the personal instance regardless of org context. The component had the correct org ID from React context but never sent it to the server.

The fix moves disk usage fetching from a raw `fetch` call to tRPC procedures on both the personal and org kiloclaw routers, mirroring how every other claw query works. The org procedure goes through `organizationMemberProcedure` which enforces membership. On the client side, a new `useClawDiskUsage` dispatcher hook in `useClawHooks.ts` follows the same dual-`useQuery`-with-`enabled` pattern used by the other 14 dispatcher hooks, routing to the correct router based on `ClawContext`.
</p>
</details> 

## Reviewer notes
AE cannot be tested locally, which makes verifying all behavior difficult for PRs that handle it. We should scope a local AE situation into future work if possible.


apps/web/src/routers/organizations/organization-kiloclaw-router.ts
```diff
@@ -302,6 +303,18 @@ export const organizationKiloclawRouter = createTRPCRouter({
     } satisfies KiloClawDashboardStatus;
   }),
 
+  getDiskUsage: organizationMemberProcedure.query(async ({ ctx, input }) => {
+    const instance = await requireOrgInstance(ctx.user.id, input.organizationId);
```
evanjacobson, line 310: This method throws NOT_FOUND just like what happens in kiloclaw-router.

apps/web/src/routers/kiloclaw-router.ts
```diff
@@ -1528,6 +1529,21 @@ export const kiloclawRouter = createTRPCRouter({
...
   }),
 
+  getDiskUsage: baseProcedure.query(async ({ ctx }) => {
+    const instance = await getActiveInstance(ctx.user.id);
+    if (!instance) {
+      throw new TRPCError({ code: 'NOT_FOUND', message: 'No active instance' });
+    }
+    try {
+      return await queryDiskUsage(instance.sandboxId);
+    } catch {
```
St0rmz1: **Low:** The bare `catch {}` discards the original error, which is a small regression vs. the deleted API route (which did `console.error('[api/kiloclaw/disk-usage] error:', error)` before returning 502).

When Analytics Engine config or upstream calls break in prod, Sentry will capture the generic `TRPCError` but not the underlying cause, so there's no stack trace to investigate from. Consider capturing the error and passing it through as `cause` (and/or logging via `sentryLogger`) before rethrowing:

```ts
} catch (error) {
  sentryLogger.error('[kiloclaw.getDiskUsage] failed', { error });
  throw new TRPCError({
    code: 'INTERNAL_SERVER_ERROR',
    message: 'Failed to fetch disk usage',
    cause: error,
  });
}
```

Same applies to the org router's `getDiskUsage` handler.

apps/web/src/app/(app)/claw/components/InstanceTab.test.ts
```diff
@@ -6,7 +6,7 @@ import {
   getVolumeUsagePercent,
   getVolumeBarColor,
 } from '@/lib/kiloclaw/instance-display';
-import { diskUsageQueryKey, hasVolumeUsageData } from './InstanceTab';
+import { hasVolumeUsageData } from './InstanceTab';
```
St0rmz1, line 9: **Low:** This PR removes the three `diskUsageQueryKey` tests and doesn't add a replacement test covering the actual bug that motivated the fix (org users seeing personal-instance data because the org context wasn't forwarded).

A focused test that exercises `useClawDiskUsage` and asserts it routes to the org query with the expected `organizationId` when `ClawContext` has one, and to the personal query otherwise, would lock the fix in and prevent a future refactor from re-introducing the same class of bug.

I realize the other dispatcher hooks in `useClawHooks.ts` may not have this coverage either, so totally fine to punt; just flagging because this PR exists specifically because the routing was wrong, which makes it the natural place to start the pattern.

St0rmz1 (approved):

PR #2423, opened 2026-04-14
fix(openclaw security advisor): schema + CTA + rate limit + version foundation

St0rmz1:
## Summary

Batched fixes and hardening for the Security Advisor API endpoint that was merged in PR #2201. Everything verified end to end against local OpenClaw plus local Next.js dev.

**Correctness**
- `finding.remediation` is now `.nullable().optional()` (was just `.nullable()`). OpenClaw audit output can omit the field entirely on some findings rather than setting it to null, and the previous schema 400'd every real audit submission. We also coerce `undefined` to `null` in the `ReportFinding.fix` mapping to keep downstream types honest.
- `invalid_payload` error message is now `JSON.stringify(z.treeifyError(...))` (was `${z.treeifyError(...)}`, which produced the infamous `Invalid request body: [object Object]` we hit in production).
- CTA no longer leaks an HTML comment directive into chat output. Small summarizing models (gpt-4.1-nano, etc.) ignored the `<!-- display-verbatim -->` hint and rendered it literally into their paraphrase. Replaced with a heading and bold CTA that survive capable model reformatting on their own. Users on small models should use the plugin's `/security-checkup` slash command for deterministic verbatim output.

**Version foundation (kilo-code-bot review incorporated)**
- `source.pluginVersion` validated against a semver regex matching the `@kilocode/cli` format. Plugin callers must send it in semver form (enforced via `superRefine` when `source.method === 'plugin'`).
- Other methods (`api`, `webhook`, `cloud-agent`) are not required to send `pluginVersion`, so this is not a breaking change for integrations that are not plugins. The previous contract is preserved.
- Structured log line on every scan (`[SecurityAdvisor] scan { userId, pluginVersion, openclawVersion, sourcePlatform, sourceMethod }`) for observability from day one.

**Rate limit**
- `RATE_LIMIT_PER_DAY` raised from 5 to 50. Report generation is template based with zero LLM cost, so 50 per day is still trivially abuse proof, and 5 was too restrictive for real users iterating on config hardening.

## Verification

- [x] `pnpm --filter web test -- security-advisor` passes 60 of 60 tests (includes 5 new cases plus 2 new regression guards for the missing `remediation` and `[object Object]` error message bugs from the earlier WIP commits, which had no test coverage before this PR)
- [x] `pnpm --filter web lint` reports 0 warnings and 0 errors
- [x] `pnpm --filter web typecheck` is clean
- [x] `oxfmt --list-different` on all touched files is clean
- [x] Live smoke tests against local dev server (`valid`, `kiloclaw`, `invalid-ip`, `ipv6`, `bad-api-version`, `malformed-json`, `missing-fields`, `rate-limit`) all pass
- [x] Targeted POST with a finding that omits the `remediation` field entirely returns 200 success (regression guard for the `.optional()` schema change)
- [x] Targeted POST with an invalid severity returns an error response containing a readable zod tree, not `[object Object]` (regression guard for the `JSON.stringify` fix)
- [x] Real OpenClaw plugin exercised the endpoint end to end on local docker OpenClaw plus fresh user signup, and got a full report back with the new CTA shape rendering correctly

## Visual Changes

N/A, server only.

## Reviewer Notes

- `source.pluginVersion` is conditionally required via `superRefine`, not at the field level. This is the fix for [kilo-code-bot's review comment](https://github.com/Kilo-Org/cloud/pull/2423#discussion_r3081688628). An earlier revision required the field for every caller, which would have silently broken `api`, `webhook`, and `cloud-agent` methods. The current implementation enforces version presence only when `source.method === 'plugin'` and also lets `SemverString` validate the format whenever the field is present. A new test (`"accepts non-plugin methods without pluginVersion"`) locks in the behavior for all three methods that are not plugins, so a future tightening cannot re-break it without a test failure.
- Two regression guards added for bugs that were previously shipped without coverage:
  - `"accepts a finding with remediation field entirely omitted"` (schemas.test.ts) guards the `.nullable().optional()` change against reverts.
  - `"returns a readable zod error in the invalid_payload message, not [object Object]"` (route.test.ts) guards the `JSON.stringify(z.treeifyError(...))` change. The existing `invalid_payload` test only asserted on `error.code`, so the format of the string shown to users could regress silently.
- HTML comment removal: earlier CTA tests positively asserted the `<!-- display-verbatim -->` directive was present. I inverted those assertions to now verify no `<!--` appears in the rendered markdown at all. If anyone tries to re-add an agent directive as an HTML comment, the test fails.
- Rate limit bump is a trivial one line change (`schemas.ts:137`). Existing rate-limiter tests reference the constant by name so they scale automatically.
- No DB schema changes. The `security_advisor_scans.plugin_version` column already exists from PR #2201, so `recordSecurityAdvisorScan` just starts writing values that were previously always `null` or `undefined`.
- Out of scope for this PR (deferred): plugin changes live in `Kilo-Org/kilocode` on `feat/openclaw-security-advisor` (slash command, device auth, README, LICENSE). The server changes here ship first per the dependency order in the plan doc. The plugin PR consumes this server contract.


apps/web/src/lib/security-advisor/report-generator.ts
```diff
@@ -170,7 +170,7 @@ function mapFinding(finding: AuditFinding, isKiloClaw: boolean): ReportFinding {
     risk:
       template?.risk ??
       `Your OpenClaw instance reports this finding and should be reviewed: ${finding.detail}`,
-    fix: finding.remediation,
+    fix: finding.remediation ?? null,
```
evanjacobson, line 173: is it better to make this optional in the schema?

evanjacobson (approved):

apps/web/src/lib/security-advisor/report-generator.ts
St0rmz1: finding.remediation was changed to .nullable().optional() on the input schema (what OpenClaw sends in), but the output type ReportFinding.fix is kept as string | null — making ?? null the correct coercion at the boundary. This keeps the report output type strict and consistent, rather than propagating undefined into report generation and rendering logic.

PR #2426, opened 2026-04-14
(admin) Recover with Kilo CLI: Phase 1 changes — utilities

evanjacobson:
## Summary

This is the first split-out phase from the larger KiloClaw admin CLI run work. This PR adds shared utilities up front so follow-up PRs can stay focused on behavior instead of mixing feature changes with extraction.

Callers have been updated where safe. `apps/web/src/lib/kiloclaw/cli-run-selection.ts` is not ready to be used yet.

## Reviewer Notes
This PR is intentionally additive and does not wire the helpers into existing UI yet. The consumer wiring was deferred to the later UI phase because those edits depend on the admin/router/schema work from subsequent split PRs.

Phase 1 for #2376

### DIFF
--- a/apps/web/src/app/admin/components/KiloclawInstances/shared.tsx
+++ b/apps/web/src/app/admin/components/KiloclawInstances/shared.tsx
@@ -18,7 +18,7 @@ export function formatRelativeTime(timestamp: string | null): string {
 }
 
 export function formatAbsoluteTime(timestamp: string): string {
-  return parseTimestamp(timestamp).toLocaleString();
+  return parseTimestamp(timestamp).toLocaleString('en-US');
 }
 
 export function DetailField({ label, children }: { label: string; children: React.ReactNode }) {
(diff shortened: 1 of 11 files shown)


apps/web/src/app/admin/components/KiloclawInstances/shared.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  return new Date(timestamp);
+}
+
+export function formatRelativeTime(timestamp: string | null): string {
+  if (!timestamp) return '—';
+  return formatDistanceToNow(parseTimestamp(timestamp), { addSuffix: true });
+}
+
+export function formatAbsoluteTime(timestamp: string): string {
```
St0rmz1, line 20: formatAbsoluteTime uses toLocaleString() — This produces output dependent on the browser/server locale, which can cause hydration mismatches in Next.js if called during SSR. It was already this way before extraction, but now that it's centralized it's a good time to consider passing an explicit locale option (toLocaleString('en-US', ...)) or using date-fns/format for deterministic output.


Consider. Next.js renders the page HTML on the server first, then ships that HTML + the React JS bundle to the browser. React "hydrates".  It re runs the component tree client side and compares the output to the server rendered HTML. If they don't match, React logs a hydration error and throws away the server HTML, re-rendering from scratch.

If you're feeling adventurous, nows the time to put in the protection since its now a single location.

St0rmz1 (approved):

PR #2428, opened 2026-04-14
(admin) Recover with Kilo CLI: Phase 2 changes —  add CLI run persistence helpers

evanjacobson:
## Summary

KiloClaw CLI runs need shared persistence logic before admin-triggered runs can be wired safely. This adds the schema support and central CLI run helpers for creating, cancelling, polling, and persisting run state, following the existing Drizzle query/helper patterns.

## Reviewer Notes

Phase 2 only adds schema/library support; router/UI wiring is intentionally left for later phases.


packages/db/src/schema.ts
```diff
@@ -3999,6 +3999,7 @@ export const kiloclaw_cli_runs = pgTable(
       .notNull()
       .references(() => kilocode_users.id, { onDelete: 'cascade' }),
     instance_id: uuid().references(() => kiloclaw_instances.id),
+    initiated_by_admin_id: text().references(() => kilocode_users.id, { onDelete: 'set null' }),
```
evanjacobson, line 4016: is GDPR soft delete a concern with admins?

apps/web/src/lib/kiloclaw/cli-runs.test.ts
```diff
@@ -0,0 +1,289 @@
+process.env.KILOCLAW_API_URL ||= 'http://localhost:8795';
```
evanjacobson, line 1: the env vars must be populated before the ES module graph resolves config.server.ts

apps/web/src/lib/user.ts
```diff
@@ -647,6 +647,10 @@ export async function softDeleteUser(userId: string) {
       .delete(kiloclaw_earlybird_purchases)
       .where(eq(kiloclaw_earlybird_purchases.user_id, userId));
     await tx.delete(kiloclaw_email_log).where(eq(kiloclaw_email_log.user_id, userId));
+    await tx
```
evanjacobson, line 651: gdpr soft delete

packages/db/src/migrations/0086_misty_beast.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "kiloclaw_cli_runs" ADD COLUMN "initiated_by_admin_id" text;--> statement-breakpoint
+ALTER TABLE "kiloclaw_cli_runs" ADD CONSTRAINT "kiloclaw_cli_runs_initiated_by_admin_id_kilocode_users_id_fk" FOREIGN KEY ("initiated_by_admin_id") REFERENCES "public"."kilocode_users"("id") ON DELETE set null ON UPDATE no action NOT VALID;--> statement-breakpoint
+ALTER TABLE "kiloclaw_cli_runs" VALIDATE CONSTRAINT "kiloclaw_cli_runs_initiated_by_admin_id_kilocode_users_id_fk";
```
evanjacobson: I have repeatedly validated that this is the best way to proceed with this FK. 

This will prevent the table from fully locking during the transaction.

apps/web/src/lib/kiloclaw/cli-runs.ts
```diff
@@ -0,0 +1,361 @@
...
+      instanceId: row.instance_id,
+    });
+  }
+
+  // When result.ok is false the controller rejected the cancel (e.g. the process
+  // had already exited between our DB check and the cancel request). The DB row
+  // intentionally stays 'running' so the caller can retry or poll again — the
+  // next getKiloCliRunStatus call will pick up the terminal state from the
+  // controller and persist it.
+  return { ok: result.ok, runFound: true, cancelled: result.ok };
```
St0rmz1: Small correctness thing worth tightening up here: this line reports `cancelled: result.ok` based purely on the controller's response, but the DB UPDATE inside `markCliRunCancelled` might have matched zero rows. That happens if the row flipped off `'running'` between the `SELECT` at the top of this function and the UPDATE below (for example, a concurrent `getCliRunStatus` poll just persisted a terminal state, or another cancel request won the race). The WHERE clause in `markCliRunCancelled` includes `status = 'running'`, which is the right guard, but it means the update silently becomes a no-op in that race, and we still tell the caller the run was cancelled.

A clean fix is to have `markCliRunCancelled` do `.returning({ id: kiloclaw_cli_runs.id })` and return whether a row came back, then compute `cancelled = result.ok && didUpdate` here. That way the return value matches reality.

Worth adding a test for this too: create a run, flip it to `completed` directly in the DB, then call `cancelCliRun` with a mocked controller that says `running` and returns `ok: true`. Today that would assert `cancelled: true`, which is the bug. After the fix it should assert `cancelled: false`.

Not a big deal for Phase 2 since nothing consumes this yet, but Phase 3 will likely branch on `cancelled` for UI state and audit logging, and it's easier to fix now than to debug a phantom cancel later.

apps/web/src/lib/kiloclaw/cli-runs.ts
```diff
@@ -0,0 +1,361 @@
...
+      output,
+      exitCode: row.exit_code,
+      startedAt: row.started_at,
+      completedAt,
+      prompt: row.prompt,
+      initiatedBy: getCliRunInitiatedBy(row),
+    };
+  }
+
+  if (!isControllerStatusForRun(row, controllerStatus)) {
```
St0rmz1, line 333: This branch has a subtle lost-row problem worth fixing. If our DB row is still `'running'` and the controller reports `hasRun: true` but with a different `startedAt` (meaning a newer CLI run has taken over on the same instance), we return `status: 'running'` and persist nothing. On the next poll we hit the same branch again, and the same one after that. There is no code path that ever resolves this row, so it stays `'running'` in Postgres forever.

Compare this to the `!controllerStatus.hasRun` branch right above, which correctly handles "controller has no run at all" by persisting `'failed'` with `LOST_CONTROLLER_RUN_OUTPUT`. The "controller has a newer run" case is really the same shape of problem: our row will never get a real terminal state from the controller because the controller has already moved on. It needs the same treatment.

Suggested fix: in this branch, call `persistCliRunControllerStatus` with `status: 'failed'` and a distinct message like `'[run state unavailable: controller has moved on to a newer run]'`, then return the terminal result. That way the row gets cleaned up and any UI or admin tooling filtering on `status = 'running'` will not see ghosts.

Also worth adding a test for it: create a running row with one `startedAt`, mock the controller to return `hasRun: true` with a different `startedAt`, call `getCliRunStatus`, and assert the row ends up `'failed'` with the superseded message.

Why this matters in practice: stale `'running'` rows will quietly accumulate every time an instance gets reused, and any dashboard or admin query looking at in-progress runs will show counts that never go down.

apps/web/src/lib/kiloclaw/cli-runs.test.ts
```diff
@@ -0,0 +1,289 @@
...
+
+describe('getCliRunInitiatedBy', () => {
+  it('returns user when admin id is null', () => {
+    expect(getCliRunInitiatedBy({ initiated_by_admin_id: null })).toBe('user');
+  });
+
+  it('returns admin when admin id is present', () => {
+    expect(getCliRunInitiatedBy({ initiated_by_admin_id: 'admin-user-id' })).toBe('admin');
+  });
+});
```
St0rmz1, line 659: Nice coverage on `cancelCliRun` and `markCliRunCancelled`, including the tricky scoping cases with null instance IDs. One gap worth filling before Phase 3 lands: there are no direct tests for `getCliRunStatus` itself, which is going to be the hot polling path once the UI is wired up.

The branches in `getCliRunStatus` that are currently untested:

1. Row is already terminal (`status !== 'running'`). Should return the row state without calling the controller at all. Worth asserting the controller mock is not called in this case.
2. Controller returns `hasRun: false`. Should persist `'failed'` with `LOST_CONTROLLER_RUN_OUTPUT` and return the terminal result. Assert both the returned value and the row in the DB.
3. Controller timestamp mismatches the row (see the other comment on M2). Currently returns `'running'` and persists nothing; once that branch is fixed, assert it persists the superseded terminal state.
4. Controller says running and timestamps match. Should return a pass-through of controller state with `prompt` and `initiatedBy` layered on.
5. `shouldPersistCliRunControllerStatus` returns true (controller reached a terminal state for this exact run). Should persist and return.

To make these testable cleanly, you will probably want to add the same injection pattern you used in `cancelCliRun` (the optional `getControllerStatus` and `cancelControllerRun` params) to `getCliRunStatus` so tests can mock the controller without spinning up the real `KiloClawInternalClient`. That keeps the test style consistent with what you already have here.

Why this matters: `getCliRunStatus` carries most of the consistency logic in this file (timestamp matching, lost-run detection, terminal-state persistence). Every one of those branches is something that will either cause stale rows or misreport run state if it regresses, and the existing tests would not catch any of it. The project convention is that every consistency branch gets a direct test, so adding these now saves debugging time later when Phase 3 starts exercising all this code for real.

St0rmz1 (approved):

apps/web/src/lib/kiloclaw/cli-runs.ts
evanjacobson: This was very helpful, thank you!

PR #2429, opened 2026-04-14
(admin) Recover with Kilo CLI: Enforce single recovery run

evanjacobson:
## Summary

Kilo CLI run conflicts were surfacing as generic failures because expected 409s lost their status/code while crossing the Durable Object RPC boundary. This PR preserves those conflicts end-to-end, prevents concurrent recovery runs from racing on the controller, and deduplicates cancel/status logic into shared helpers.

<details><summary>Details</summary>
<p>

- Return a serializable `{ conflict: { code, error } }` shape from the DO for expected CLI run conflicts, then map it back to HTTP 409 and tRPC CONFLICT in the platform, personal, org, and admin paths.
- Serialize controller start requests via a promise queue (`runStartExclusive`) so only one recovery run can start on a machine at a time.
- Change the controller cancel endpoint from 404 to 409 with a structured `{ code, error }` response when there is no active run, matching the start conflict shape.
- Translate controller 409 on cancel to `{ ok: false }` in the shared `cancelCliRun` helper so late-arriving cancels don't surface as unhandled errors.
- Reconcile stale running DB rows when cancel discovers the controller has no active run (marks them `failed` with a descriptive output message).
- Extract duplicated `getKiloCliRunStatus` and `cancelKiloCliRun` DB + controller polling logic from the personal and org routers into shared helpers in `cli-runs.ts`.
- Add `controller_route_unavailable` → `PRECONDITION_FAILED` handling to the org router's `startKiloCliRun` (previously only in the personal router).
- Add test coverage across the controller, DO, platform, and tRPC boundaries using the same conflict shapes production sees.

</p>
</details> 

## Verification

<details><summary>Details</summary>
<p>

- [x] `pnpm format` — applied
- [x] `pnpm test -- apps/web/src/routers/kiloclaw-start-kilo-cli-run.test.ts` — all pass
- [x] `pnpm test -- apps/web/src/routers/admin-kiloclaw-instances-router.test.ts` — all pass
- [x] `pnpm test -- apps/web/src/lib/kiloclaw/cli-runs.test.ts` — all pass
- [x] `pnpm test -- services/kiloclaw/controller/src/routes/kilo-cli-run.test.ts` — all pass (via `pnpm --filter kiloclaw-controller test`)
- [x] `pnpm test -- services/kiloclaw/src/durable-objects/kiloclaw-instance.test.ts` — all pass
- [x] `pnpm test -- services/kiloclaw/src/routes/platform-sanitize-error.test.ts` — all pass

</p>
</details> 

## Visual Changes

N/A

## Reviewer Notes

- The DO returns conflicts as values instead of throwing because Cloudflare RPC strips custom error properties. Only `.message` survives across the boundary. 
  - The platform route validates the conflict shape with `KiloCliRunConflictSchema` and returns 502 if the shape is malformed.
- Start and cancel intentionally keep different fallback conflict codes. This preserves operation-specific meaning even when older controllers return 409 without a structured code field.
  - `kilo_cli_run_already_active` for start
  - `kilo_cli_run_no_active_run` for cancel. 
- The controller's cancel endpoint changed from 404 to 409 (the resource exists, it's just in the wrong state for the operation). The web app's `cancelCliRun` helper catches this 409 and translates it to `{ ok: false }` so the caller can retry or poll.
- Cancel/status logic was previously duplicated across `kiloclaw-router.ts` and `organization-kiloclaw-router.ts`. Both now delegate to the shared `cancelCliRun` / `getCliRunStatus` helpers in `cli-runs.ts`, which also handle the new conflict translation.

### DIFF
--- a/apps/web/src/lib/kiloclaw/cli-runs.ts
+++ b/apps/web/src/lib/kiloclaw/cli-runs.ts
@@ -156,6 +156,19 @@ export function getEmptyCliRunStatus(): CliRunStatusResult {
   };
 }
 
+/**
+ * Cancel a running CLI run, reconciling controller and DB state.
+ *
+ * Limitation: when the controller reports `hasRun: false`, the run's actual
+ * terminal state is already evicted from controller memory. We record the row
+ * as `'failed'` with a sentinel output, even though the run may have completed
+ * successfully just before the cancel arrived. A re-poll cannot recover the
+ * real outcome because the controller has already discarded it. This is a
+ * strict improvement over leaving the row stuck in `'running'` forever, but
+ * callers should be aware that a `'failed'` status with the
+ * `LOST_CONTROLLER_RUN_OUTPUT` sentinel does not necessarily mean the run
+ * actually failed — it means the outcome is unknown.
+ */
 export async function cancelCliRun(params: {
   runId: string;
   userId: string;
@@ -283,6 +296,14 @@ export async function cancelCliRun(params: {
   return { ok: true, runFound: true, cancelled: didUpdate };
 }
 
+/**
+ * Poll the status of a CLI run, reconciling controller and DB state.
+ *
+ * Limitation: when the controller reports `hasRun: false`, the run's actual
+ * terminal state is already evicted from controller memory. We record the row
+ * as `'failed'` with a sentinel output, even though the run may have completed
+ * successfully. See {@link cancelCliRun} for the same limitation and rationale.
+ */
 export async function getCliRunStatus(params: {
   runId: string;
   userId: string;
--- a/apps/web/src/lib/kiloclaw/cli-runs.ts
+++ b/apps/web/src/lib/kiloclaw/cli-runs.ts
@@ -206,6 +206,21 @@ export async function cancelCliRun(params: {
     });
 
   const controllerStatus = await getControllerStatus(params.userId, effectiveWorkerInstanceId);
+  if (!controllerStatus.hasRun) {
+    await persistCliRunControllerStatus({
+      runId: params.runId,
+      userId: params.userId,
+      instanceId: row.instance_id,
+      controllerStatus: {
+        status: 'failed',
+        exitCode: row.exit_code,
+        output: row.output ?? LOST_CONTROLLER_RUN_OUTPUT,
+        completedAt: new Date().toISOString(),
+      },
+    });
+    return { ok: true, runFound: true, cancelled: false };
+  }
+
   if (!isControllerStatusForRun(row, controllerStatus)) {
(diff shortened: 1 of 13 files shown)


apps/web/src/lib/kiloclaw/cli-runs.ts
```diff
@@ -200,12 +200,37 @@ export async function cancelCliRun(params: {
...
+        // caller can retry or poll for the terminal state.
+        if (err instanceof KiloClawApiError && err.statusCode === 409) {
+          return { ok: false };
+        }
+        throw err;
+      }
     });
 
   const controllerStatus = await getControllerStatus(params.userId, effectiveWorkerInstanceId);
+  if (!controllerStatus.hasRun) {
```
St0rmz1, line 232: **Functional: Cancel helper records 'failed' for a possibly successful run**

**What**: When the user clicks "cancel" but the controller no longer has an active run, the helper writes the local row as `failed`. In the common case the run actually completed naturally (success or failure) just before the cancel arrived, and no status poll caught the natural completion. We now permanently record it as `failed` even if it actually finished as `completed`.

**Why it matters**: This is real data loss for the run history. A user who clicked cancel on a run that was about to finish will see "failed" with a "controller no longer has an active CLI run" output blob, instead of the actual result. This is a strict improvement over the prior code (which would have written `cancelled`), and the `LOST_CONTROLLER_RUN_OUTPUT` sentinel is honest about the situation, so it is not blocking. Worth being aware of.

**Suggestion**: Before writing `failed`, you could re-poll `getKiloCliRunStatus` once to see if the controller now reports a terminal state for this run, and prefer that. Alternatively, document this limitation in the helper's docstring so the next reader does not have to derive it from the call sites. A targeted test that asserts "if the controller's status changes between the helper's poll and its write, the real terminal state wins" would also catch any future regression here.

St0rmz1 (approved):

apps/web/src/lib/kiloclaw/cli-runs.ts
evanjacobson: I investigated the re-poll suggestion. Tl;dr: hasRun: false is always permanent, so a re-poll is extraneous for this check.

<details><summary>Details</summary>
<p>

The controller stores run state in a module-level let activeRun variable (pure in-memory). It's null on process start and only set when a new run begins. When activeRun is null, that data is gone. There's no disk persistence or recovery path. 

At the DO layer, hasRun: false is returned when state.status !== 'running', which means the machine isn't running. If the machine restarts, the controller process also restarts with activeRun = null.

</p>
</details> 

There is also a guard for row updates to only touch the 'running' status. If the row has already been updated out of 'running', it will not be updated by this call. 


Added docstrings & the test per your suggestion.

PR #2482, opened 2026-04-15
fix(kiloclaw): move instance and subscription creation to worker runtime

jeanduplessis:
## Summary

Moves KiloClaw instance row creation (`kiloclaw_instances`) and subscription row creation (`kiloclaw_subscriptions`) from Next.js to the worker runtime, establishing the kiloclaw worker as the single owner of instance persistence (datamodel rule 21) and kiloclaw-billing as the single owner of subscription bootstrap. Also introduces **successor subscription rows** to preserve billing continuity across personal reprovision.

### Why this change is needed

Instance and subscription creation was spread across Next.js (`ensureActiveInstance`, `getOrCreateInstanceForBilling`, inline subscription inserts in checkout/webhook handlers) and the worker provision endpoint, creating race conditions between concurrent provisions, orphaned DB rows on provision failure, and coupling between the billing layer and the instance lifecycle. Stripe webhook handlers used `user_id`-scoped queries that could bleed data across personal and org instances. When users destroyed and reprovisioned, their subscription row was updated in-place, losing history and creating ambiguous Stripe routing.

### How this is addressed

- **Worker-owned instance persistence** — Worker provision endpoint now inserts `kiloclaw_instances` rows. `ensureActiveInstance` converted to read-only.
- **Worker-owned subscription bootstrap** — New `/bootstrap-subscription` endpoint in `kiloclaw-billing` creates trial, earlybird, and org subscription rows. Called by the kiloclaw worker via service binding after instance insertion. Next.js no longer creates subscription rows during provision or checkout — it only updates existing rows.
- **Successor subscription transfer** — On reprovision, the bootstrap service creates a new subscription row on the new instance, copying entitlement and Stripe ownership from the predecessor. The predecessor is marked `status='canceled'` and linked via `transferred_to_subscription_id`. This preserves a complete chain of billing history per user.
- **Stripe routing by subscription ID** — Webhook handlers key on `stripe_subscription_id` (not `user_id`), following the `transferred_to_subscription_id` chain to the current row. If resolution is ambiguous, cyclic, crosses org boundaries, or the target is missing, the event is quarantined with a logged warning rather than guessed.
- **Provision serialization** — `withKiloclawProvisionContextLock()` acquires a PostgreSQL advisory lock per user/org to serialize concurrent provisions.
- **Current personal subscription resolution** — New `resolveCurrentPersonalSubscriptionRow()` returns the single non-transferred, instance-bound personal subscription. All billing flows (checkout, credit enrollment, Kilo Pass upsell, settlement) route through this function.
- **Subscription change audit log** — `kiloclaw_subscription_change_log` table records before/after JSONB snapshots for all subscription mutations.
- **Migration backfill** — `0091_living_vulture.sql` adds the `transferred_to_subscription_id` column, reconstructs historical predecessor rows from the change log, attaches detached subscription rows to orphan instances, and merges remaining detached rows into active trial subscriptions.
- **GDPR soft-delete update** — KiloClaw FKs changed to `ON DELETE NO ACTION`; precondition blocks user deletion while active instances exist.

### Architecture: before and after

<details>
<summary>Before: Next.js-owned provision</summary>

```mermaid
sequenceDiagram
    participant User
    participant NextJS as Next.js
    participant DB as PostgreSQL
    participant Worker as kiloclaw worker
    participant Fly

    User->>NextJS: provision mutation
    NextJS->>DB: INSERT kiloclaw_subscriptions (trial)
    Note over NextJS,DB: ensureProvisionAccess()
    NextJS->>DB: INSERT kiloclaw_instances
    NextJS->>DB: INSERT kiloclaw_inbound_email_aliases
    Note over NextJS,DB: ensureActiveInstance()
    NextJS->>Worker: POST /api/platform/provision
    Worker->>Fly: create machine + volume
    Fly-->>Worker: sandboxId
    Worker-->>NextJS: { sandboxId, instanceId }
    NextJS-->>User: provisioned

    Note over User,Fly: Later: Stripe checkout
    User->>NextJS: checkout mutation
    Note over NextJS: Stripe creates subscription
    Note over NextJS,DB: Webhook: INSERT kiloclaw_subscriptions (paid)<br/>instance_id may be NULL if instance not yet provisioned
```

On **reprovision** (destroy → provision again), `adoptOrphanedSubscription()` finds the old subscription row on the destroyed instance and updates `instance_id` in-place to point to the new instance. History is lost.

</details>

<details>
<summary>After: Worker-owned provision</summary>

```mermaid
sequenceDiagram
    participant User
    participant NextJS as Next.js
    participant DB as PostgreSQL
    participant Worker as kiloclaw worker
    participant Billing as kiloclaw-billing worker
    participant Fly

    User->>NextJS: provision mutation
    NextJS->>DB: pg_advisory_xact_lock (provision lock)
    NextJS->>DB: SELECT (read-only access check)
    Note over NextJS,DB: ensureProvisionAccess()<br/>returns bootstrapSubscription flag
    NextJS->>Worker: POST /api/platform/provision
    Worker->>DB: INSERT kiloclaw_instances
    Worker->>DB: INSERT kiloclaw_inbound_email_aliases
    Worker->>Fly: create machine + volume
    Fly-->>Worker: sandboxId
    Worker->>Billing: POST /bootstrap-subscription
    Billing->>DB: INSERT kiloclaw_subscriptions (trial)
    Note over Billing,DB: or: create successor row on reprovision
    Billing->>DB: INSERT kiloclaw_subscription_change_log
    Billing-->>Worker: { subscriptionId }
    Worker-->>NextJS: { sandboxId, instanceId }
    NextJS-->>User: provisioned

    Note over User,Fly: Later: Stripe checkout
    User->>NextJS: checkout mutation
    Note over NextJS: Stripe creates subscription
    Note over NextJS,DB: Webhook: UPDATE existing subscription row<br/>follows transferred_to chain to current row
```

On **reprovision**, the billing worker creates a **successor subscription row** linked to the new instance. The predecessor row is marked `status='canceled'` with `transferred_to_subscription_id` pointing to the successor. Stripe webhook handlers follow the transfer chain to route events to the current row.

```mermaid
graph LR
    A["Sub A (instance 1)<br/>status: canceled"] -->|transferred_to| B["Sub B (instance 2)<br/>status: trialing"]
    B -->|transferred_to| C["Sub C (instance 3)<br/>status: active"]
    style A fill:#fee,stroke:#c33
    style B fill:#fee,stroke:#c33
    style C fill:#efe,stroke:#3a3
```

</details>

## Verification

- No manual testing performed; changes are covered by existing and new automated tests in `kiloclaw-billing-router.test.ts` and `bootstrap.test.ts`. CI runs the full suite.

## Visual Changes

N/A

### Human Reviewer

- **Architectural change**: Instance creation moved to kiloclaw worker, subscription creation moved to kiloclaw-billing worker. `ensureActiveInstance` is now read-only. Next.js checkout/webhook handlers no longer insert subscription rows — they only update existing rows that the bootstrap service created.
- **Successor subscription model**: On reprovision, the old subscription becomes a predecessor (canceled, linked via `transferred_to_subscription_id`). The new subscription inherits Stripe ownership and entitlement. All live queries filter to `transferred_to_subscription_id IS NULL` to see only current rows.
- **FK behavior change**: KiloClaw tables changed from `ON DELETE CASCADE` to `ON DELETE NO ACTION` on `kilocode_users` FK. Soft-delete now requires instances to be destroyed first (precondition check added).
- **Stripe metadata dependency**: New checkout sessions include `instanceId` and `billingContext` in metadata. Existing in-flight checkouts without these fall back to legacy adoption logic.
- **Service binding**: kiloclaw worker calls kiloclaw-billing via service binding for subscription bootstrap — binding configured in `wrangler.jsonc`.
- **Migration complexity**: `0091_living_vulture.sql` includes a multi-step backfill that reconstructs predecessor chains from the change log, attaches detached rows, and merges duplicates. Review the SQL carefully.
- **Quarantine behavior**: Stripe events and credit settlements that cannot resolve to a single current subscription are silently quarantined (logged as warnings, not errors). This is intentional — it prevents guessing and lets operators investigate.

### Code Reviewer Agent
<details>

- `bootstrapPersonalSubscription` in `services/kiloclaw-billing/src/bootstrap.ts` handles personal paths: existing-on-instance (no-op), successor transfer (copy + link predecessor), and new trial. Org path creates managed or trial subscription based on seat license.
- `createSuccessorPersonalSubscription` copies all fields from predecessor except Stripe ownership, marks predecessor canceled with `transferred_to_subscription_id` pointing to successor, then transfers Stripe IDs in a separate update.
- `applyStripeFundedKiloClawPeriod` follows the `transferred_to_subscription_id` chain (max 8 hops) to find the current target. Returns `false` if unresolvable (quarantines the charge). `PersonalStripeResolutionError` and `CreditSettlementResolutionError` track resolution failure reasons.
- Provision lock uses `pg_advisory_xact_lock(hashtext(key))` — transaction-scoped, auto-released on commit/rollback.
- `resolveCurrentPersonalSubscriptionRow` enforces at-most-one invariant: throws `CurrentPersonalSubscriptionResolutionError` if multiple non-transferred personal rows found.
- Analytics blob layout extended: blob14=orgId, blob15=instanceId. `safeInstanceIdFromSandboxId` never throws (analytics is fire-and-forget).
- All `handleKiloClaw*` webhook handlers refactored from `(user_id, stripe_subscription_id)` lookups to `stripe_subscription_id`-only with lineage following.

</details>


packages/db/src/migrations/0091_living_vulture.sql
```diff
@@ -0,0 +1,164 @@
+ALTER TABLE "kiloclaw_subscriptions" ADD COLUMN "transferred_to_subscription_id" uuid;--> statement-breakpoint
+ALTER TABLE "kiloclaw_subscriptions" ADD CONSTRAINT "kiloclaw_subscriptions_transferred_to_subscription_id_kiloclaw_subscriptions_id_fk" FOREIGN KEY ("transferred_to_subscription_id") REFERENCES "public"."kiloclaw_subscriptions"("id") ON DELETE no action ON UPDATE no action;--> statement-breakpoint
+CREATE INDEX "IDX_kiloclaw_subscriptions_transferred_to" ON "kiloclaw_subscriptions" USING btree ("transferred_to_subscription_id");--> statement-breakpoint
```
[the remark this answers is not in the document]
jeanduplessis: Acceptable - these are not high write hot spots

services/kiloclaw-billing/src/bootstrap.ts
```diff
@@ -0,0 +1,501 @@
...
+): KiloClawSubscription[] {
+  return subscriptions.filter(subscription => {
+    if (subscription.transferred_to_subscription_id) {
+      return false;
+    }
+    if (!subscription.instance_id) {
+      return false;
+    }
+    const instance = instancesById.get(subscription.instance_id);
+    return !instance || instance.organizationId === null;
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 94bb37e67. Bootstrap current-row resolution now ignores subscription rows whose anchor  row is missing, so phantom orphan rows no longer participate in current-row selection. Added regression coverage in  to prove missing-anchor rows are ignored while valid destroyed-instance successor transfer still works.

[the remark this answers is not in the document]
jeanduplessis: Fixed in 94bb37e67. Bootstrap current-row resolution now ignores subscription rows whose anchor kiloclaw_instances row is missing, so phantom orphan rows no longer participate in current-row selection. Added regression coverage in services/kiloclaw-billing/src/bootstrap.test.ts to prove missing-anchor rows are ignored while valid destroyed-instance successor transfer still works.

services/kiloclaw/src/routes/platform.ts
```diff
@@ -261,6 +345,150 @@ async function withResolvedDORetry<TResult>(
...
+async function bootstrapProvisionedSubscription(params: {
+  env: AppEnv['Bindings'];
+  userId: string;
+  instanceId: string;
+  orgId: string | null;
+}): Promise<void> {
+  if (!params.env.KILOCLAW_BILLING) {
+    throw new Error('KILOCLAW_BILLING service binding is not configured');
+  }
+  if (!params.env.KILOCLAW_INTERNAL_API_SECRET) {
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in ac1d66696. Platform routes now require and validate KILOCLAW_INTERNAL_API_SECRET end-to-end: env gate in services/kiloclaw/src/index.ts switched to KILOCLAW_INTERNAL_API_SECRET, internalApiMiddleware now authenticates against same secret, and regression tests cover both missing KILOCLAW_INTERNAL_API_SECRET and distinct next/claw secret values.

services/kiloclaw/src/routes/platform.ts
```diff
@@ -78,6 +82,42 @@ const KiloCliRunConflictSchema = z.object({
...
 
+const EXPLICIT_INSTANCE_DO_KEY_CACHE_TTL_MS = 5 * 60 * 1000;
+const EXPLICIT_INSTANCE_DO_KEY_CACHE_MAX = 1000;
+const explicitInstanceDoKeyCache = new Map<
+  string,
+  {
+    doKey: string;
+    expiresAt: number;
+  }
+>();
```
pandemicsyn: Should this maybe just use CF KV? Not sure build in a in worker memache will work out that well ?

services/kiloclaw/src/routes/platform.ts
```diff
@@ -558,28 +960,112 @@ platform.post('/provision', async c => {
...
+          'destroy'
+        ).catch(destroyErr => {
+          console.error(
+            '[platform] Failed to destroy provisioned instance after bootstrap error:',
+            destroyErr
+          );
+        });
+        await markProvisionedInstanceDestroyed({
+          env: c.env,
+          instanceId: provisionedInstanceId,
```
pandemicsyn, line 1005: as far as lifecycle events go. Only new ones I thought might be useful are related to thes, like a:

```
writeEvent(c.env, {
        event: 'instance.subscription_bootstrapped',
        delivery: 'http',
        userId,
        instanceId: provisionedInstanceId,
        orgId: orgId ?? undefined,
      });
```

?

If you're looking at an instance in the Admin UI and troubleshooting things - it could (maybe?) be useful to have events in there if subscription succeeded/failed ?

pandemicsyn: @jeanduplessis totally up to you on this one. Also fine shipping without.

services/kiloclaw/src/index.test.ts
```diff
@@ -79,7 +108,8 @@ describe('platform route env validation', () => {
         body: JSON.stringify({ userId: 'user-1' }),
       }),
       {
-        INTERNAL_API_SECRET: 'secret-123',
+        INTERNAL_API_SECRET: 'next-secret',
+        KILOCLAW_INTERNAL_API_SECRET: 'claw-secret',
```
pandemicsyn: I think `next-secret` is confusing here, this isn't the next auth secret or the like.

services/kiloclaw/.dev.vars.example
```diff
@@ -17,8 +17,8 @@
 # @from NEXTAUTH_SECRET
 NEXTAUTH_SECRET=dev-secret-change-me
 # Platform API authentication key (for internal service-to-service calls)
-# @from INTERNAL_API_SECRET
-INTERNAL_API_SECRET=dev-internal-secret
```
pandemicsyn, line 21: This is fine, but when this merges can we let folks know we're breaking their .dev.vars ? 

We also need to add this new secret in cloudflare otherwise this will break prod?

pandemicsyn (changes requested):

services/kiloclaw-billing/src/bootstrap.ts
```diff
@@ -0,0 +1,644 @@
...
+            plan: 'trial',
+            status: new Date(trialEndsAt).getTime() > now.getTime() ? 'trialing' : 'canceled',
+            access_origin: null,
+            payment_source: null,
+            cancel_at_period_end: false,
+            trial_started_at: organization.createdAt,
+            trial_ends_at: trialEndsAt,
+          }
+    )
+    .returning();
```
pandemicsyn: The personal bootstrap path has an advisory lock in the mix right ? But the org version doesn't. Does this need a onConflictDoNothing or is a violation bubbling up here ok?

services/kiloclaw-billing/src/index.ts
```diff
@@ -36,7 +43,119 @@ function log(level: 'info' | 'warn' | 'error', message: string, fields: BillingL
...
+          statusCode: 500,
+        });
+        return Response.json(
+          { error: 'KILOCLAW_INTERNAL_API_SECRET is not configured' },
+          { status: 500 }
+        );
+      }
+
+      const providedSecret = request.headers.get('x-internal-api-key');
+      if (!providedSecret || providedSecret !== env.KILOCLAW_INTERNAL_API_SECRET) {
```
pandemicsyn: ~~Spirit of Storms compels me to tell you that this should use timingSafeEqual. There might be a utility func already.~~

services/kiloclaw/src/routes/platform.ts
jeanduplessis: Leaning toward keeping the in-isolate Map here:

- TTL is 5 min with a 1000-entry cap; per-isolate is fine because the miss path (Postgres lookup via Hyperdrive on indexed `kiloclaw_instances.id`) is already cheap.
- KV has ~60s minimum edge TTL and eventual consistency across colos, so it doesn't meaningfully beat a 5-minute in-memory cache for this lookup shape, and it adds a request per miss.
- Deploys flushing the map is acceptable — Postgres is the source of truth, the cache is best-effort.

Knowing that, do you still think we'd be better off with KV (or something else) here, or fine to keep the in-memory cache with a clarifying comment about per-isolate/bounded/best-effort semantics?

services/kiloclaw-billing/src/index.ts
pandemicsyn: actually, should have thought of this when i was reviewing the kiloclaw worker side, but service bindings mean we can call RPC's directly. 

So kiloclaw worker can just call `env.KILOCLAW_BILLING.bootstrapProvisionSubscription(userId, instanceId, orgId)` 

Which will let ya get rid of this stuff. https://developers.cloudflare.com/workers/runtime-apis/bindings/service-bindings/rpc/

pandemicsyn (changes requested):

services/kiloclaw/.dev.vars.example
jeanduplessis: Context on the rename: `INTERNAL_API_SECRET` was doing double duty — it was both (a) the Next.js → kiloclaw inbound auth secret and (b) a shared name in the kiloclaw-billing worker for bootstrap auth. That ambiguity was flagged in 9255ab8f0 (`fix(kiloclaw-billing): secure bootstrap auth and quarantine ambiguity`). Splitting them into `INTERNAL_API_SECRET` (Next-side, unchanged everywhere else) and `KILOCLAW_INTERNAL_API_SECRET` (kiloclaw-side) lets the billing worker authenticate to kiloclaw with its own distinct secret and makes greps/audits unambiguous.

Re your two concerns:

1. **Prod rollout** — you're right, this will break prod if we don't set the new secret first. Plan: `wrangler secret put KILOCLAW_INTERNAL_API_SECRET` on both kiloclaw and kiloclaw-billing (same value as current `INTERNAL_API_SECRET`) **before** this merges. Will verify via `wrangler secret list` pre-merge.
2. **`.dev.vars` break** — the dev-start script (fe9cea1b9 `fix(dev): sync kiloclaw internal dev secret`) already syncs `KILOCLAW_INTERNAL_API_SECRET` from Vercel, so devs who rerun `pnpm dev:start` will pick it up automatically. I'll post a team heads-up in #dev when this merges with the "delete `.dev.vars` and rerun dev:start" instruction for anyone whose file is stale.

services/kiloclaw/src/routes/platform.ts
jeanduplessis: Dropped the in-isolate cache entirely in favor of just hitting Postgres (via Hyperdrive) every time — agreed it's not worth carrying a speculative optimization into this PR when Hyperdrive already caches hot lookups. If we see this lookup show up in traces or DB load, we can revisit (KV, DO-side memoization, or a bounded Map with real numbers behind it). Test that asserted the cache behavior was removed along with it.

services/kiloclaw-billing/src/index.ts
jeanduplessis: Good call, done. Converted kiloclaw-billing to a `WorkerEntrypoint` and the kiloclaw worker now calls `env.KILOCLAW_BILLING.bootstrapProvisionSubscription({ userId, instanceId, orgId })` directly.

Changes:
- `services/kiloclaw-billing/src/index.ts`: new `KiloClawBillingService extends WorkerEntrypoint` class; the old `/bootstrap-subscription` HTTP handler (and its secret/Zod/JSON framing) is gone.
- `services/kiloclaw/wrangler.jsonc`: service binding now specifies `entrypoint: "KiloClawBillingService"`.
- `services/kiloclaw/src/kiloclaw-billing-binding.ts`: typed binding shim (mirrors the session-ingest pattern already in this repo) — `wrangler types` still only sees `Fetcher`, so the RPC method signature is declared here and `AppEnv['Bindings']['KILOCLAW_BILLING']` points at it.
- `services/kiloclaw/src/routes/platform.ts:580`: `.fetch(...)` → `.bootstrapProvisionSubscription(...)`, and the `KILOCLAW_INTERNAL_API_SECRET` check on this path is gone.
- Test coverage: HTTP-based test in `kiloclaw-billing/src/index.test.ts` replaced with RPC-direct tests (success + Zod rejection).

On the secret: `KILOCLAW_INTERNAL_API_SECRET` still exists, because the **other** direction (kiloclaw-billing → kiloclaw `/api/platform/*` in `lifecycle.ts:468`) and Next.js → kiloclaw `/api/platform/*` both still need it. The rename (`INTERNAL_API_SECRET` → `KILOCLAW_INTERNAL_API_SECRET`) from 9255ab8f0 is still load-bearing there — it's what separates kiloclaw-billing's outbound-to-Next secret from its outbound-to-kiloclaw secret. RPC-ifying that direction would require kiloclaw's Hono API to also go through `WorkerEntrypoint`, which is a much bigger refactor since those routes are also called by Next.js (which can't use service bindings).

services/kiloclaw-billing/src/bootstrap.ts
jeanduplessis: Good catch — fixed. Two things to clarify first:

1. No actual **advisory lock** in play anywhere; the "personal path" protection you saw was `SELECT ... FOR UPDATE` row locks, and those only run inside `createSuccessorPersonalSubscription` (the transfer path). The fresh-insert personal path was just as racy as the org path.
2. The partial unique index `UQ_kiloclaw_subscriptions_instance` (on `instance_id WHERE instance_id IS NOT NULL`) already kept a duplicate row from landing, but the violation bubbling up from the bootstrap path into `platform.ts` triggered the destroy-DO + mark-instance-destroyed cleanup — so the loser of a benign race (retry, double-submit) would tear down an instance the winner had just finished bootstrapping. That was the real bug.

Fix: added an `insertSubscriptionIdempotent` helper that does `onConflictDoNothing({ target: instance_id }).returning()` and reselects on empty return. Applied to both `bootstrapOrganizationSubscription` and `bootstrapPersonalSubscription`'s fresh-insert paths. Race loser now returns the winner's row, logs `decision: existing_for_instance_race`, and skips the change-log write (winner already emitted `created`).

Tests: two new cases in `bootstrap.test.ts` — `concurrent insert race` describe block — covering both personal and org races, asserting race loser returns winner's row and emits no duplicate change-log entry.

jeanduplessis: Follow-up from self-review: the first pass had a latent bug. `onConflictDoNothing({ target: kiloclaw_subscriptions.instance_id })` alone would have failed at runtime against the partial unique index with `there is no unique or exclusion constraint matching the ON CONFLICT specification` — Postgres requires the index predicate to be restated on the ON CONFLICT clause for partial-index arbiter inference.

Verified against the dev DB with both forms:

```
-- Without predicate:
INSERT ... ON CONFLICT (instance_id) DO NOTHING;
ERROR:  there is no unique or exclusion constraint matching the ON CONFLICT specification

-- With predicate (what we now emit):
INSERT ... ON CONFLICT ("instance_id") WHERE "_t"."instance_id" IS NOT NULL DO NOTHING;
INSERT 0 0   -- success, conflict swallowed
```

Fix: added `where: isNotNull(kiloclaw_subscriptions.instance_id)` to the onConflictDoNothing call. Drizzle emits the predicate table-qualified (`where "kiloclaw_subscriptions"."instance_id" is not null`), which Postgres accepts and correctly matches `UQ_kiloclaw_subscriptions_instance`.

Added `services/kiloclaw-billing/src/bootstrap-sql-shape.test.ts` as a regression guard that asserts on the emitted SQL shape, so this class of bug fails at unit-test time rather than in production.

services/kiloclaw/src/routes/platform.ts
jeanduplessis: Already in — landed as part of a9369bb50 (`fix(kiloclaw): add structured logging for provision write paths`). Same event name you suggested:

- `services/kiloclaw/src/routes/platform.ts:967` — `instance.subscription_bootstrapped` on success
- `services/kiloclaw/src/routes/platform.ts:980` — `instance.subscription_bootstrap_failed` on failure (with sanitized error + durationMs)

Also got symmetric coverage on the instance-record insert:

- `services/kiloclaw/src/routes/platform.ts:909` — `instance.record_inserted`
- `services/kiloclaw/src/routes/platform.ts:922` — `instance.record_insert_failed`

All four carry `userId`, `instanceId`, `sandboxId`, `orgId`, and `durationMs`, so the admin UI has the visibility you were thinking of.

pandemicsyn (approved):

PR #2487, opened 2026-04-15
fix(admin): make admin CLI runs manageable and attributable

evanjacobson:
## Summary

Admin-initiated CLI runs were previously fire-and-forget. There was no database record, no way to check status or cancel, and no attribution back to the admin who started them. If something went wrong, the only option was to destroy the instance.

This PR gives admin CLI runs the same lifecycle as user-started runs by routing them through the shared `createCliRun` / `getCliRunStatus` / `cancelCliRun` helpers. Each run records which admin started it, and the admin list view can now filter by initiator. Start and cancel actions write audit logs with instance and run metadata.

Ownership checks prevent admins from accidentally operating on an instance that belongs to a different user than the one selected.

## Verification

- Not manually tested

## Visual Changes

N/A

## Reviewer Notes

- The reconciliation paths for stale runs (controller has moved on, instance destroyed) mirror the user-facing `cancelCliRun` fallback.
- Left join is intentionally omitted from the `listAllCliRuns` query on the CLI runs tab, as we do not support searching by admin initiator. 
- startKiloCliRun in the admin router creates the DB row after the controller run starts. This mirrors the other existing audit patterns in the file

### DIFF
--- a/apps/web/src/app/admin/components/KiloclawCliRuns/KiloclawCliRunsTab.tsx
+++ b/apps/web/src/app/admin/components/KiloclawCliRuns/KiloclawCliRunsTab.tsx
@@ -256,7 +256,10 @@ function RunDetail({
 }) {
   const trpc = useTRPC();
   const { data, isLoading } = useQuery(
-    trpc.admin.kiloclawInstances.getCliRunOutput.queryOptions({ runId: run.id })
+    trpc.admin.kiloclawInstances.getCliRunOutput.queryOptions({
+      userId: run.user_id,
+      runId: run.id,
+    })
   );
   const output = data?.output ?? null;
 
--- a/apps/web/src/routers/admin-kiloclaw-instances-router.ts
+++ b/apps/web/src/routers/admin-kiloclaw-instances-router.ts
@@ -26,6 +26,7 @@ import {
   createKiloClawAdminAuditLog,
   listKiloClawAdminAuditLogs,
 } from '@/lib/kiloclaw/admin-audit-log';
+import { cancelCliRun, createCliRun, getCliRunStatus } from '@/lib/kiloclaw/cli-runs';
 import type {
   PlatformDebugStatusResponse,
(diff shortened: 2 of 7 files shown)


apps/web/src/routers/admin-kiloclaw-instances-router.ts
```diff
@@ -643,29 +666,136 @@ export const adminKiloclawInstancesRouter = createTRPCRouter({
...
+          const resolvedInstanceId = instance?.id ?? null;
+          const usedFallback = !resolvedInstanceId;
+          try {
+            await createKiloClawAdminAuditLog({
+              action: 'kiloclaw.cli_run.cancel',
+              actor_id: ctx.user.id,
+              actor_email: ctx.user.google_user_email,
+              actor_name: ctx.user.google_user_name,
+              target_user_id: input.userId,
+              message: 'CLI run cancelled',
```
evanjacobson, line 781: cleaner message if we move the instance ID to metadata due to the fallback

apps/web/src/routers/admin-kiloclaw-instances-router.ts
evanjacobson, line 1: the count query intentionally does not do the leftJoin because it's not searching on the initiating admin's email.

apps/web/src/app/admin/components/KiloclawCliRuns/KiloclawCliRunsTab.tsx
```diff
@@ -256,7 +256,10 @@ function RunDetail({
 }) {
   const trpc = useTRPC();
   const { data, isLoading } = useQuery(
-    trpc.admin.kiloclawInstances.getCliRunOutput.queryOptions({ runId: run.id })
+    trpc.admin.kiloclawInstances.getCliRunOutput.queryOptions({
```
evanjacobson, line 259: required for typecheck

jeanduplessis (approved):

PR #2488, opened 2026-04-16
Feature/kiloclaw admin cli admin UI

evanjacobson:
## Summary

Admins previously had no way to start, watch, or cancel a KiloClaw CLI run against a specific instance from the admin UI, and the global CLI runs tab made it hard to tell admin-initiated runs apart from user-initiated ones or to search by instance. 

This PR adds a `KiloCliRunCard` to the admin instance detail page for starting, polling, and cancelling a CLI run scoped to that instance, and extends the global CLI runs tab with an admin/user filter, an admin badge, instance ID display and search, and a redesigned single-run view. The admin card follows the existing admin card conventions already used by adjacent instance-detail widgets, so it should feel uniform with the rest of the admin UI.

<details><summary>More Details</summary>
<p>

- The `listKiloCliRuns` tRPC query now accepts an optional `instanceId` so the per-instance card only sees runs for that instance. 
- `cancelCliRun` now returns the resolved `instanceId` so audit logs can record the run's actual instance rather than re-resolving from the user.

</p>
</details> 

## Verification

- [x] Started, polled, and cancelled a CLI run from the admin instance detail page
- [x] Confirmed admin-initiated runs show the Admin badge in the CLI runs tab and can be filtered by source
- [x] Searched the CLI runs tab by partial instance ID
- [x] `pnpm test -- apps/web/src/routers/admin-kiloclaw-instances-router.test.ts apps/web/src/lib/kiloclaw/cli-runs.test.ts`

## Reviewer Notes

- Cancellation audit metadata now uses the instanceId returned from `cancelCliRun` instead of the router's resolved instance. This is intentional — the run row is the source of truth for which instance the run targeted.
- `listKiloCliRuns` gained an optional `instanceId` filter; existing callers without it are unaffected.

### DIFF
--- a/apps/web/src/lib/kiloclaw/cli-runs.ts
+++ b/apps/web/src/lib/kiloclaw/cli-runs.ts
@@ -79,7 +79,7 @@ export type CancelCliRunParams = {
 };
 
 export type CancelCliRunResult =
-  | { ok: boolean; runFound: true; cancelled: boolean }
+  | { ok: boolean; runFound: true; cancelled: boolean; instanceId: string | null }
   | { ok: false; runFound: false; cancelled: false };
 
 export type CliRunStatusResult = KiloCliRunStatusResponse & {
@@ -194,7 +194,7 @@ export async function cancelCliRun(params: {
   }
 
   if (row.status !== 'running') {
-    return { ok: true, runFound: true, cancelled: false };
+    return { ok: true, runFound: true, cancelled: false, instanceId: row.instance_id };
   }
 
   const effectiveWorkerInstanceId =
@@ -241,7 +241,7 @@ export async function cancelCliRun(params: {
         completedAt: new Date().toISOString(),
       },
     });
-    return { ok: true, runFound: true, cancelled: false };
+    return { ok: true, runFound: true, cancelled: false, instanceId: row.instance_id };
   }
 
   if (!isControllerStatusForRun(row, controllerStatus)) {
@@ -258,7 +258,7 @@ export async function cancelCliRun(params: {
         completedAt: new Date().toISOString(),
       },
     });
-    return { ok: true, runFound: true, cancelled: false };
+    return { ok: true, runFound: true, cancelled: false, instanceId: row.instance_id };
   }
 
   // The controller already reached a terminal state for this run — persist it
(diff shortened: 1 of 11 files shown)


apps/web/src/routers/admin-kiloclaw-instances-router.ts
```diff
@@ -802,12 +802,24 @@ export const adminKiloclawInstancesRouter = createTRPCRouter({
     }),
 
   listKiloCliRuns: adminProcedure
-    .input(z.object({ userId: z.string().min(1), limit: z.number().min(1).max(50).default(20) }))
+    .input(
+      z.object({
+        userId: z.string().min(1),
+        instanceId: z.uuid().optional(),
```
evanjacobson, line 804: adding instanceId here is a allows filter by instance

apps/web/src/lib/kiloclaw/cli-runs.ts
```diff
@@ -79,7 +79,7 @@ export type CancelCliRunParams = {
 };
 
 export type CancelCliRunResult =
-  | { ok: boolean; runFound: true; cancelled: boolean }
+  | { ok: boolean; runFound: true; cancelled: boolean; instanceId: string | null }
```
evanjacobson, line 82: Threading instanceId through here ensures we show it on the correct page for users with multiple instances.

apps/web/src/routers/admin-kiloclaw-instances-router.test.ts
```diff
@@ -137,18 +136,77 @@ beforeEach(async () => {
...
-  await db.delete(kiloclaw_cli_runs).where(inArray(kiloclaw_cli_runs.user_id, userIds));
-  // Deleting instances cascades to inbound email aliases
-  await db.delete(kiloclaw_instances).where(inArray(kiloclaw_instances.user_id, userIds));
-  await db.delete(kilocode_users).where(inArray(kilocode_users.id, userIds));
+  await db.delete(kiloclaw_cli_runs).where(eq(kiloclaw_cli_runs.id, cliRunId));
+  await db.delete(kiloclaw_instances).where(eq(kiloclaw_instances.id, cliRunInstanceId));
 });
 /* eslint-enable drizzle/enforce-delete-with-where */
 
+describe('admin.kiloclawInstances.listKiloCliRuns', () => {
```
evanjacobson, line 152: added a missing test

evanjacobson:
> The previous auto-scroll bug is still unresolved: the ref/effect were removed, but the output panel still does not follow live log growth.

There is a scrollbar in the panel, and this is an admin feature. This is a minor UX detail that I think is at most suited for a follow-up instead of blocking merge.

jeanduplessis (approved):

PR #2538, opened 2026-04-17
fix(billing): disable first top-up bonus

emilieschario:
*  Set the first top-up bonus amount to $0 and skip granting the first-top-up promo when the amount is disabled.
* Remove first-top-up bonus upsell copy from checkout, onboarding, low-credit notifications, and paid-model credit errors.
* Update affected billing and code-review status tests for the no-bonus messaging and balance behavior.


apps/web/src/components/payment/FirstTopup24for24CountdownTimer.tsx
```diff
@@ -1,33 +0,0 @@
-'use client';
```
emilieschario, line 1: I can be convinced not to remove this file but I think this was a one-off promo so

apps/web/src/lib/firstTopupBonus.ts
```diff
@@ -5,14 +5,13 @@ import { db } from '@/lib/drizzle';
 import { FIRST_TOPUP_BONUS_AMOUNT } from '@/lib/constants';
 
 export async function processFirstTopupBonus(user: User) {
-  // this is run after topping up, so a user which has done their first
-  // topup will have exactly one topup
-  // Uses primary db for read-after-write consistency (payment was just inserted)
+  if (FIRST_TOPUP_BONUS_AMOUNT <= 0) return;
```
markijbema, line 8: i was afraid this would mean people would get the topup with the second topup if we ever make it non-zero, but there is another guard preventing this

apps/web/src/components/payment/FirstTopup24for24CountdownTimer.tsx
markijbema: Don't you remember? This is for grokpocalypse

markijbema (approved):

PR #2539, opened 2026-04-17
feat(admin): add backfill to block users on blacklisted domains

markijbema:
## Summary

- Adds an admin backfill API route at `/admin/api/backfills/block-blacklisted-domains` that sets `blocked_reason = 'domainblocked'` on users whose email matches a `BLACKLIST_DOMAINS` entry but who don't already have a `blocked_reason` set.
- GET returns the count of unblocked users on blacklisted domains; POST processes them in batches (1000 users × 50 batches per request).
- Follows the same pattern as the existing `normalized-email` backfill route.

## Verification

- [x] Typecheck passes (pre-existing errors only, none from this change)
- [x] Formatted with `pnpm format`
- [ ] Not manually tested against a live database — the route is admin-only and idempotent

## Visual Changes

N/A

## Reviewer Notes

- The `blocked_reason` value `'domainblocked'` matches the convention used by the prior backfill script (`d2025-08-29-backfill-domainblocked-users.ts`).
- The UPDATE includes a redundant `isNull(blocked_reason)` guard to avoid race conditions with concurrent blocks.
- Uses `or()` with `!` on each condition — this is safe because `or()` with two SQL conditions always returns a defined value; however if the linter flags it, it can be restructured.

### DIFF
--- a/apps/web/src/app/admin/api/backfills/block-blacklisted-domains/route.ts
+++ b/apps/web/src/app/admin/api/backfills/block-blacklisted-domains/route.ts
@@ -11,13 +11,10 @@ const BLACKLIST_DOMAINS = blacklistDomainsEnv
   : [];
 
 function blacklistedDomainConditions() {
-  const conditions = BLACKLIST_DOMAINS.map(
-    domain =>
-      or(
-        sql`lower(${kilocode_users.google_user_email}) LIKE ${`%@${domain.toLowerCase()}`}`,
-        sql`lower(${kilocode_users.google_user_email}) LIKE ${`%.${domain.toLowerCase()}`}`
-      )!
-  );
+  const conditions = BLACKLIST_DOMAINS.flatMap(domain => [
+    sql`lower(${kilocode_users.google_user_email}) LIKE ${`%@${domain.toLowerCase()}`}`,
+    sql`lower(${kilocode_users.google_user_email}) LIKE ${`%.${domain.toLowerCase()}`}`,
+  ]);
   return or(...conditions);
 }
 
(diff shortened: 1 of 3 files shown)


RSO (approved):

apps/web/src/app/admin/api/backfills/block-blacklisted-domains/route.ts
```diff
@@ -0,0 +1,95 @@
+import { NextResponse } from 'next/server';
+import { getUserFromAuth } from '@/lib/user.server';
+import { db } from '@/lib/drizzle';
+import { kilocode_users } from '@kilocode/db';
+import { and, count, isNull, or, sql } from 'drizzle-orm';
+import { getEnvVariable } from '@/lib/dotenvx';
+
+const blacklistDomainsEnv = getEnvVariable('BLACKLIST_DOMAINS');
```
chrarnoldus, line 8: why is this an env var?

chrarnoldus (approved):

apps/web/src/app/admin/api/backfills/block-blacklisted-domains/route.ts
```diff
@@ -0,0 +1,95 @@
...
+  if (authFailedResponse) return authFailedResponse;
+
+  if (BLACKLIST_DOMAINS.length === 0) {
+    return NextResponse.json({ unblocked: 0 });
+  }
+
+  const [result] = await db
+    .select({ count: count() })
+    .from(kilocode_users)
+    .where(and(isNull(kilocode_users.blocked_reason), blacklistedDomainConditions()));
```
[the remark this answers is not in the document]
markijbema: this is the empty set

apps/web/src/app/admin/api/backfills/block-blacklisted-domains/route.ts
markijbema: because we want to read it quicly, and don't want to share it (it was in code)

PR #2562, opened 2026-04-17
feat(kiloclaw): add Northflank provider

pandemicsyn:
## Summary

- Adds the Northflank KiloClaw provider implementation behind runtime provider rollout config, including provider state schema, registry wiring, SDK-backed Northflank client/config helpers, deterministic resource naming, and lifecycle support for provision/start/stop/restart/inspect/destroy.
- Implements the Northflank architecture as one deterministic Northflank project per KiloClaw instance, with a persistent `/root` volume, service creation at `instances: 0`, restricted project secrets for `KILOCLAW_ENV_KEY`, trusted edge-header routing, and Postgres restore recovery from deterministic names.
- Updates dev image publishing so the on-demand KiloClaw dev workflow pushes the same image tag to Fly Registry and GHCR for Northflank pulls, and documents/configures the required `NF_*` Worker settings.
- Makes Northflank available for rollout selection while keeping actual traffic gated by runtime rollout percentages and enabled organization IDs; existing Fly instances remain sticky to Fly and `docker-local` remains development-only.
`OPENCLAW_DISABLE_BONJOUR=1` overlayed on Northflank's `runtimeEnvironment` so the `@homebridge/ciao` 63-byte DNS label assertion stops crashing the controller inside Kubernetes pods.
- **`ensureSecret` content-hashes `bootstrapEnv`** (SHA-256 of canonicalized JSON) and persists the hash in the Northflank provider state. Subsequent start/restart skips the secret write when the hash matches, avoiding Northflank's secret-update-triggered service redeploy on every alarm-driven restart. Restore from Postgres leaves the hash null, so the next `ensureSecret` writes and persists a fresh one.
- **Guards against a duplicate-provision race** (belt + suspenders). UI: `BotIdentityStep` adds a synchronous `useRef` re-entry guard so rapid double-clicks on the Continue button don't slip past `isContinuing` before React's next render applies `disabled={isContinuing}`. Worker DO: `KiloClawInstance.startAsync` short-circuits when status is already `starting` with a fresh `startingAt`, preventing a duplicate `waitUntil(start())` from re-entering `provider.startRuntime` sequentially after the first completes (the in-memory `startInProgress` flag only guarded concurrent re-entry).
- **Redacts the edge-header secret in Northflank API errors.** `northflankClientConfig` now threads `edgeHeaderValue` into the client's `redactValues`, so Northflank 4xx/5xx responses that echo the submitted payload (which carries the edge-header value under the non-sensitive key `value` in `buildPortSecurity`) are stripped from `[northflank] api_request_failed` logs and `NorthflankApiError` messages. `redactUnknown`'s key-based heuristic does not catch `value`, so explicit `redactValues` entry is required.

## Verification

- [x] `pnpm format` — passed
- [x] `pnpm --filter kiloclaw typecheck` — passed
- [x] `pnpm --filter web typecheck` — passed
- [x] `pnpm lint` — 0 warnings, 0 errors
- [x] `pnpm --filter kiloclaw test` — 85 files / 1675 tests passed
- [x] `pnpm --filter web exec jest ClawOnboardingFlow` — 21/21 passed
- [x] Live provision against Northflank (`kilo-dev` team, `us-central`) reaches `state: ready` and routes through envoy end-to-end with the current code
- [x] `git push` pre-push checks (typecheck / format / lint across 5150 files) — passed

## Visual Changes

N/A

## Reviewer Notes

- `NORTHFLANK_ROLLOUT_AVAILABLE` is enabled in this PR, but new Northflank selection still requires runtime rollout config to set percentages and enabled org IDs.
- `@northflank/js-client` currently emits a Node `punycode` deprecation warning in tests from its dependency tree; tests and Wrangler dry-run pass.
- `NorthflankProviderStateSchema` grows a new `secretContentHash` field (nullable, defaults to `null`). Existing persisted DO state hydrates with `null`, so the first start/restart after deploy will write the secret once and persist a fresh hash — subsequent restarts with unchanged `bootstrapEnv` become no-ops on the secret side.
- Pairing still works on Northflank via the controller HTTP primary path (`/_kilo/pairing/*`)
- `runDoctor` on Northflank remains unavailable until a controller-side `POST /_kilo/doctor` route ships (actually a pr for that)
- Production Northflank config landed: `wrangler.jsonc` now has `NF_TEAM_ID=kilo-prod`, `NF_REGION`, `NF_DEPLOYMENT_PLAN`, `NF_EDGE_HEADER_NAME`, `NF_IMAGE_PATH_TEMPLATE`, `NF_IMAGE_CREDENTIALS_ID`. Before merging, I'll set `NF_API_TOKEN` and `NF_EDGE_HEADER_VALUE` as Worker secrets (Cloudflare dashboard → Workers → kiloclaw → Secrets).
- Prod `deploy-kiloclaw.yml` now dual-pushes to `registry.fly.io/kiloclaw-machines` and `ghcr.io/kilo-org/kiloclaw` so every content-hash tag is reachable by both Fly and Northflank. When a deploy reuses a cached Fly image, the workflow copies it to GHCR if missing.


services/kiloclaw/src/durable-objects/kiloclaw-instance/postgres.ts
```diff
@@ -12,13 +17,80 @@ import { getAppKey, getFlyConfig } from './types';
...
+  }
+  const personal = await getActivePersonalInstance(db, userId);
+  return personal ? await getInstanceBySandboxId(db, personal.sandboxId) : null;
+}
+
+async function recoverNorthflankProviderState(
+  env: KiloClawEnv,
+  sandboxId: string
+): Promise<NorthflankProviderState> {
+  const config = getNorthflankConfig(env);
```
St0rmz1: The other Northflank call sites all go through `northflankClientConfig` so the edge header value lands in `redactValues` before any request goes out. This recovery path uses `getNorthflankConfig` directly. Because every call here is a GET (`findProjectByName`, `findVolumeByName`, `findServiceByName`, `findProjectSecretByName`) the edge header value never rides in a request body today, so there is no live leak. The concern is drift: the next person who adds a POST or PATCH on this path, or who shares a helper between recovery and provisioning, will silently drop edge header redaction.

Suggestion: export `northflankClientConfig` (or move it next to `getNorthflankConfig` in `northflank/config.ts`) and use it everywhere a Northflank request is made.

services/kiloclaw/src/durable-objects/kiloclaw-instance/postgres.ts
```diff
@@ -12,13 +17,80 @@ import { getAppKey, getFlyConfig } from './types';
...
+  const names = await northflankResourceNames(sandboxId);
+  const project = await findProjectByName(config, names.projectName);
+  const volume = project ? await findVolumeByName(config, project.id, names.volumeName) : null;
+  const service = project ? await findServiceByName(config, project.id, names.serviceName) : null;
+  const secret = project
+    ? await findProjectSecretByName(config, project.id, names.secretName)
+    : null;
+
```
St0rmz1, line 74: After the project is found, the volume, service, and secret lookups have no ordering dependency on each other but run serially. With `findProjectByName` now paginating up to 50 pages, the project lookup itself can already be slow on a saturated team. Stacking three more sequential round trips behind it directly extends how long a recovered instance sits in `provisioned` before it can serve traffic.

Suggestion: once `project` is known, run the three remaining lookups with `Promise.all`.

services/kiloclaw/src/northflank/client.ts
```diff
@@ -0,0 +1,696 @@
...
+  return response.data;
+}
+
+export async function findVolumeByName(
+  config: NorthflankClientConfig,
+  projectId: string,
+  name: string
+): Promise<NorthflankVolume | null> {
+  const volumes = await listVolumes(config, projectId);
+  return volumes.find(volume => volume.name === name) ?? null;
```
St0rmz1: The pagination fix that landed for `findProjectByName` did not extend to `findVolumeByName`, `findServiceByName`, or `findProjectSecretByName`. They still inspect only the first page of their list endpoints. With the current "one project per instance" architecture each project holds at most one of each resource, so this is not exploitable today. Worth flagging because the project list bug had the same shape and the same mitigations (deterministic names plus 409 conflict recovery), so it is easy to forget if the project layout ever grows or if leftover resources ever accumulate from a failed cleanup.

Suggestion: either paginate the same way `findProjectByName` does, or leave a comment on each helper noting the assumption that only one of each resource exists in a project, so reading just the first page is sufficient.

St0rmz1 (approved):

PR #2566, opened 2026-04-17
feat(kiloclaw): ship brokered Google OAuth calendar integration

joshavant:
## Summary
- Add brokered Google OAuth support for KiloClaw with a calendar-first rollout: web connect/callback/disconnect flows, worker/controller token/status/migrate endpoints, controller token provider, and gog shim integration.
- Add persistent Google OAuth connection storage with source-aware grants (`grants_by_source`) and effective capability handling.
- Preserve legacy continuity with best-effort migration from legacy gog credentials into broker storage plus controlled `gog.real` fallback when migration fails.
- Harden auth/security boundaries and close review findings across web/controller/worker routes.

### Web (`apps/web`)
- Added Google connect/callback/disconnect API routes + tests.
- Added signed OAuth state payload handling and capability/scope helpers.
- Added broker connection upsert/clear integration via KiloClaw internal client.
- Hardened callback/connect/disconnect behavior:
  - callback state/user/owner validation tightened
  - callback input sanitization for CodeQL-sensitive branches
  - callback telemetry redaction (no raw OAuth code/state query payload)
  - disconnect is POST-only + same-origin guard
  - POST redirects use **303** (See Other) to avoid method-preserving redirect issues
- Restricted Google settings UI visibility to admins.
- Updated soft-delete to remove per-instance Google OAuth credentials while preserving retained billing/audit records.

### Worker + Controller (`services/kiloclaw`)
- Added `/google/token`, `/google/status`, `/google/migrate-legacy` routes + expanded auth/error-path coverage.
- Added legacy migration utility from gog keyring exports to broker storage.
- Added payload validation hardening for migration outbound body (`accountEmail`, secrets, scopes/services normalization).
- Added product telemetry extras for legacy migration attempt/success/failure reason.
- Updated gog shim behavior:
  - broker routing is calendar-only (`calendar|cal`)
  - `gmail|drive|docs|sheets` fall through to `gog.real` until broader consent rollout
- Fixed legacy migrate state consistency:
  - for existing `kilo_owned`, DO status/scopes/lastError now preserve DB truth (no forced active desync)
- Tightened token capability guard:
  - `/google/token` now requires non-empty capabilities list (empty list returns 400)

### Data layer (`apps/web` + DB)
- Upsert race hardening for callback concurrency:
  - conflict-safe insert path (`onConflictDoUpdate`) for same `instance_id`
  - follow-up winner-row merge to preserve legacy grants under races
- DB migration metadata corrected after rebase collisions:
  - mainline keeps `0097_fast_tempest`
  - OAuth migration moved to `0098_google_oauth_kiloclaw_broker`
  - added `meta/0098_snapshot.json`
  - updated `meta/_journal.json`

### Dev / Runtime guidance
- Updated seeded Google Workspace TOOLS section to match current gog syntax.
- Added guidance to prefer all-calendar retrieval and explicit date-window alignment for calendar questions.

## Security / Review Follow-ups Included
- Auth enforced for `/_kilo/google-oauth/*` controller routes.
- Extensive auth-boundary + branch coverage added for token/status/migrate endpoints.
- Disconnect GET regression test present (`method_not_allowed`, no mutation).
- Legacy migration CodeQL concern addressed with explicit validation + threat-model comment.
- Callback CodeQL bypass findings addressed by moving verified-state checks ahead of sensitive branches and sanitizing callback params.

## Rollout Notes
- Intentional calendar-first broker rollout.
- Fresh OAuth installs default to calendar broker capability.
- Non-calendar Google commands continue via `gog.real` until broader capability consent is enabled.

## Focused Verification
- `pnpm --filter kiloclaw test -- "src/routes/controller.test.ts"`
- `pnpm --filter kiloclaw test -- controller/src/gog-shim.test.ts`
- `pnpm --filter kiloclaw test -- controller/src/bootstrap.test.ts`
- `pnpm --filter kiloclaw typecheck`
- `pnpm --filter @kilocode/db typecheck`
- `pnpm drizzle check`
- Manual E2E smoke validation on fresh KiloClaw instances for OAuth calendar connect + broker-backed calendar reads.


apps/web/src/app/api/integrations/google/callback/route.ts
```diff
@@ -0,0 +1,196 @@
...
+      return NextResponse.redirect(
+        new URL(buildGoogleRedirectPath(state, `error=${errorCode}`), APP_URL)
+      );
+    }
+
+    if (!code) {
+      captureMessage('Google callback missing code', {
+        level: 'warning',
+        tags: { endpoint: 'google/callback', source: 'google_oauth' },
+        extra: { state, allParams: Object.fromEntries(searchParams.entries()) },
```
St0rmz1: The Sentry visibility on this callback is great for triage, but a couple of these `extra` payloads send sensitive OAuth values to Sentry that are worth tightening up.

**The concern:**

`Object.fromEntries(searchParams.entries())` serializes every query parameter, which on the success path includes Google's `code` (the OAuth authorization code). That code is short-lived but redeemable: anyone with read access to Sentry could POST it to `https://oauth2.googleapis.com/token` during its validity window and obtain access plus refresh tokens for the user's Google account. Codes are single-use, but the window between when Sentry receives the log and when our callback exchanges the code is enough for a race.

The signed `state` is much less sensitive (an attacker can't forge a new one without `NEXTAUTH_SECRET`), but it still embeds the originating user ID and instance ID, which is more identity context than Sentry needs for triage.

**Affected sites in this file:**

- Line ~50 (`Google OAuth error`): logs `state`
- Line ~64 (`Google callback missing code`): logs `state` and `allParams` (which can include `code`)
- Line ~76 (`Google callback invalid or tampered state`): same as above
- Line ~170 (outer catch): logs `state` (already uses `hasCode: !!...`, which is the right pattern)

**Suggested fix:**

Drop `code` from any Sentry payload, and either omit `state` or log a short hash for correlation. A small helper keeps it consistent:

```ts
function sentryOAuthContext(searchParams: URLSearchParams) {
  const state = searchParams.get('state');
  return {
    hasCode: !!searchParams.get('code'),
    hasState: !!state,
    stateHash: state ? createHash('sha256').update(state).digest('hex').slice(0, 8) : null,
    error: searchParams.get('error'),
    errorDescription: searchParams.get('error_description'),
  };
}
```

Then each `captureMessage` extras becomes `extra: sentryOAuthContext(searchParams)`. Triage still gets the failure shape and a stable correlation ID across log lines, but no redeemable credential leaves the request.

Worth adding a quick test that asserts `code` never appears in the captured Sentry payload, so future edits don't regress.

services/kiloclaw/src/routes/controller.ts
```diff
@@ -56,6 +67,30 @@ const CheckinSchema = z.object({
...
+  refreshToken: z.string().min(1),
+  oauthClientId: z.string().min(1),
+  oauthClientSecret: z.string().min(1),
+  scopes: z.array(z.string().min(1)).default([]),
+  capabilities: z.array(z.string().min(1)).default([]),
+});
+
+const GOOGLE_CAPABILITY_SCOPES: Record<string, readonly string[]> = {
+  calendar_read: ['https://www.googleapis.com/auth/calendar.readonly'],
+};
```
St0rmz1, line 103: The web side accepts `gmail_read` as a valid capability (`apps/web/src/lib/integrations/google/capabilities.ts:11` enum, plus parsing in `parseGoogleCapabilities`), but the broker's scope map here only knows about `calendar_read`. That creates an inconsistency worth resolving before users run into it.

**The concrete failure mode:**

If a user (or a future caller) hits `/api/integrations/google/connect?capabilities=gmail_read`:

1. Web validates the capability fine (it's in the enum).
2. OAuth flow completes; `upsertKiloClawGoogleOAuthConnection` writes a row with `capabilities: ['gmail_read']`.
3. UI shows "Connected as foo@example.com".
4. Calendar/Gmail tooling in the container calls `/_kilo/google-oauth/token` → `/api/controller/google/token` with `capabilities: ['gmail_read']`.
5. The broker rejects at the `unsupportedCapabilities.length > 0` check on line ~447 with HTTP 400 "Unsupported capabilities: gmail_read".

The user sees a "Connected" UI but every call from the gog shim fails with no obvious recovery path. They'd need to disconnect and reconnect with the right capability set, which they have no way to know.

**Two reasonable fixes, depending on intent:**

Option A (recommended for this PR's scope): tighten the web schema so only shippable capabilities are accepted.

```ts
// apps/web/src/lib/integrations/google/capabilities.ts
export const GoogleCapabilitySchema = z.enum([
  GOOGLE_CAPABILITY.CALENDAR_READ,
  // GOOGLE_CAPABILITY.GMAIL_READ — re-add when broker scopes ship
]);
```

Option B (if Gmail is meant to ship in this PR): add the Gmail scopes to the broker.

```ts
// services/kiloclaw/src/routes/controller.ts
const GOOGLE_CAPABILITY_SCOPES: Record<string, readonly string[]> = {
  calendar_read: ['https://www.googleapis.com/auth/calendar.readonly'],
  gmail_read: ['https://www.googleapis.com/auth/gmail.readonly'],
};
```

The PR title says calendar integration, so Option A is probably right.

**Longer-term thought:**

The capability enum and scope map exist in two places (web `capabilities.ts` and worker `controller.ts`). A single shared source (e.g., a small `@kilocode/google-capabilities` module, or moving it under `@kilocode/db` next to the schema enum) would prevent this kind of drift from recurring. Not blocking for this PR.

apps/web/src/app/api/integrations/google/callback/route.ts
```diff
@@ -0,0 +1,196 @@
...
+      return NextResponse.redirect(
+        new URL(buildGoogleRedirectPath(state, 'error=missing_instance'), APP_URL)
+      );
+    }
+
+    const isUserOwnerMatch =
+      verifiedState.owner.type === 'user' &&
+      instance.userId === user.id &&
+      instance.organizationId === null;
+
```
St0rmz1, line 168: The `isOrgOwnerMatch` check requires `instance.userId === user.id`, which means only the user who originally provisioned the instance can complete an OAuth flow against it. For org-scoped flows that's likely too tight.

**The scenario this breaks:**

- Alice provisions the org KiloClaw instance, so `instance.userId = alice`.
- Six months later Alice leaves, or the calendar refresh token hits `invalid_grant` and ends up `action_required`.
- Bob (also an org admin) tries to reconnect Google. He passes `ensureOrganizationAccess` because he's an org member, his session matches `verifiedState.userId`, and the state is signed and valid.
- The callback rejects because `instance.userId === bob` is false (Alice provisioned it). Redirects to `/claw/settings?error=unauthorized`.

The org has no path to recover the connection without standing up a new instance.

**Suggested fix:**

For the `org` branch, drop `instance.userId === user.id` and trust `ensureOrganizationAccess` plus the `instance.organizationId === verifiedState.owner.id` check that's already there:

```ts
const isUserOwnerMatch =
  verifiedState.owner.type === 'user' &&
  instance.userId === user.id &&
  instance.organizationId === null;

const isOrgOwnerMatch =
  verifiedState.owner.type === 'org' &&
  instance.organizationId === verifiedState.owner.id;
// ensureOrganizationAccess({ user }, verifiedState.owner.id) already ran above,
// so we know the current user is a member of this org.
```

For the `user` (personal) branch, the existing `instance.userId === user.id` check is correct and should stay.

**Alternative if this is intentional:**

If the design is "the OAuth identity must match the human who set up the instance" (e.g., calendar identity is per-user by design and not shareable), please add a code comment so future readers don't relax it accidentally. A test that exercises the "second org admin reconnects" scenario would also nail down the intent.

**Worth a quick test either way:**

- Org admin A provisions the instance; org admin B completes the OAuth callback. Expected: success (under the suggested fix) or explicit unauthorized (under the current behavior, with the rationale documented).

apps/web/src/app/api/integrations/google/disconnect/route.ts
```diff
@@ -0,0 +1,85 @@
...
+          organizationId,
+        },
+      });
+
+      return NextResponse.redirect(
+        new URL(buildDisconnectPath(organizationId, 'error=missing_instance'), APP_URL)
+      );
+    }
+
+    await clearKiloClawGoogleOAuthConnection(instance.id);
```
St0rmz1, line 95: Disconnect drops the row and clears DO state without first reading what was there. Two small concerns worth tightening up before this ships.

**No audit trail of the disconnect:**

After this runs, the row is gone and there's no record of which Google account was disconnected, who triggered it, or when. For an OAuth integration with refresh tokens, that's a question you'll want answered during incident review (compromise, support ticket, legal hold, "who unhooked our calendar last Tuesday?"). Even a single structured log line before the delete would be enough:

```ts
const existing = await getKiloClawGoogleOAuthConnection(instance.id);
if (existing) {
  console.log('[google-disconnect] removing connection', {
    instanceId: instance.id,
    userId: user.id,
    accountEmail: existing.account_email,
    accountSubject: existing.account_subject,
    organizationId: organizationId ?? null,
  });
}
await clearKiloClawGoogleOAuthConnection(instance.id);
```

Bonus: emit a PostHog event (`kc_google_oauth_disconnected`) so the connect/disconnect funnel is visible alongside the existing claw events.

**Two-step write can leave state diverged:**

The DB delete and the DO clear are sequential, both can throw, and there's no retry or compensating action. If `kiloclawClient.clearGoogleOAuthConnection(...)` fails after the row is already deleted, the DO still thinks Google is connected (`googleOAuthConnected: true` in status), but no connection actually exists in the DB. The user sees a stale "Connected" badge and disconnect appears to have failed (the catch redirects to `error=disconnect_failed`).

A few ways to handle this, in increasing order of effort:

1. Swap the order: clear the DO first (idempotent, no row mutation), then the DB delete. If the DB delete fails, the DO at least reflects "not connected" which is the safer drift direction for the user.
2. Wrap in a retry with backoff for the DO call, since it's just a state reconciliation and should be safe to repeat.
3. If you want strict atomicity, model the disconnect as an event the controller route consumes idempotently from the worker DB.

(1) is probably the right call for the scope of this PR.

**Worth a test:**

When the second clear call throws, the DB row should still be gone (or still present, depending on which order you pick), and the redirect should land on `error=disconnect_failed`. The existing test mocks both calls separately, so adding a divergence assertion is a small addition.

services/kiloclaw/controller/src/legacy-google-migration.ts
```diff
@@ -0,0 +1,193 @@
...
+  } catch {
+    return false;
+  }
+
+  const email = authList.accounts?.find(a => typeof a.email === 'string')?.email;
+  if (!email) {
+    return false;
+  }
+
+  const tmpPath = path.join(os.tmpdir(), `gog-legacy-token-${Date.now()}.json`);
```
St0rmz1: The migration writes the plain refresh token to a predictable temp path. A few small hardenings worth applying since refresh tokens are long-lived secrets.

**Three issues stacked here:**

1. **Predictable path.** `gog-legacy-token-${Date.now()}.json` is guessable to the millisecond. On a single-tenant container that's mostly theoretical, but on shared `/tmp` (e.g., a developer running this script locally during testing) a colocated process can predict the name and pre-create a symlink to redirect the write or the read. Standard TOCTOU on `/tmp`.

2. **No mode on the file.** `gog auth tokens export` writes with whatever default mode the binary uses, which is typically `0644`. Anything running as another UID on the host that can see the container's `/tmp` (rare in production, common in dev) can read the file while the migration is in flight.

3. **Silent unlink failure.** The `unlinkSync` in the `finally` block has `// ignore`, so if the unlink fails for any reason, the plain refresh token sits in `/tmp` for the rest of the container's life with no signal that cleanup didn't work.

**Suggested fix:**

Use a unique 0700 directory and clean it up unconditionally; tighten the file mode after export:

```ts
const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'gog-legacy-'));
const tmpPath = path.join(tmpDir, 'token.json');

try {
  execFileSync('/usr/local/bin/gog.real',
    ['auth', 'tokens', 'export', email, tmpPath, '--overwrite'],
    { env: gogEnv, encoding: 'utf8' }
  );
  // Tighten permissions immediately after the export.
  try { fs.chmodSync(tmpPath, 0o600); } catch { /* best effort */ }

  const exportPayload = JSON.parse(fs.readFileSync(tmpPath, 'utf8')) as GogTokenExportJson;
  // ... rest unchanged
} finally {
  try {
    fs.rmSync(tmpDir, { recursive: true, force: true });
  } catch (err) {
    console.error('[legacy-google-migration] failed to remove tmp dir', tmpDir, err);
  }
}
```

`mkdtempSync` creates the directory atomically with mode 0700 (subject to umask), so the symlink race is closed. Logging the rmSync error rather than swallowing it gives you a signal if cleanup ever silently breaks.

**Even better, if the upstream tool supports it:**

Check whether `gog auth tokens export` accepts `-` or `/dev/stdout` as the destination. If so, you can redirect via stdio and the secret never lands on disk:

```ts
const stdout = execFileSync('/usr/local/bin/gog.real',
  ['auth', 'tokens', 'export', email, '-', '--overwrite'],
  { env: gogEnv, encoding: 'utf8' }
);
const exportPayload = JSON.parse(stdout) as GogTokenExportJson;
```

If that's not supported by the gog binary version pinned for this PR, the mkdtemp approach is the next-best.

**No new test needed for this:**

The behavior is unchanged from a successful path; the hardening is about the on-disk surface area during the brief window the token exists. Verifying the cleanup path runs (e.g., a test that asserts `tmpDir` doesn't exist after migration completes) would still be useful but is optional.

services/kiloclaw/src/routes/controller.ts
```diff
@@ -249,4 +420,297 @@ controller.post('/checkin', async (c: Context<AppEnv>) => {
   return c.body(null, 204);
 });
 
+controller.post('/google/token', async (c: Context<AppEnv>) => {
```
St0rmz1, line 464: The new `/google/token`, `/google/status`, and `/google/migrate-legacy` routes are about 350 lines that handle decryption, Google token-endpoint calls, error mapping, and writes to both the worker DB and the DO state. None of that is exercised by tests in this PR. Per the team's TypeScript Quality Rules ("Security Boundary Test Coverage"), these endpoints need test coverage before merge.

**Why it's worth doing now rather than in a follow-up:**

This is the surface that decides whether a refresh token is still valid, whether to flip a connection to `action_required`, whether to trust a capability in a token request, and whether to mirror state changes into the DO. Each of those is a place where a silent regression would manifest as "calendar sometimes works, sometimes doesn't, customer can't tell why". Tests are the cheap insurance.

**Suggested minimum coverage:**

For `/google/token` (the highest-risk one):

- Returns 401 when the bearer or `x-kiloclaw-gateway-token` header is missing or malformed.
- Returns 403 when the gateway token doesn't match the expected derived value, and when the API key doesn't match the DO config.
- Returns 404 when no `kiloclaw_google_oauth_connections` row exists for the instance.
- Returns 409 when the connection's `status` is `action_required`.
- Returns 412 when a requested capability isn't in the connection's `capabilities` array.
- Returns 400 when an unsupported capability is requested (e.g., `gmail_read` today, see M2).
- Happy path: returns access token, expiry, account email; calls `updateGoogleOAuthConnectionTokenData` and the DO `updateGoogleOAuthConnection` with `status: 'active'`.
- `invalid_grant` from Google flips DB row to `action_required` AND mirrors that status to the DO, then returns 409 with `reason: 'invalid_grant'`.
- `deleted_client` from Google does the same.
- A non-mapped 5xx from Google returns 502 without flipping status (so the next call can retry).
- Decryption failure path flips status to `action_required` with `last_error: 'refresh_token_decryption_failed'`.

For `/google/status`:

- Auth boundary cases (same as above).
- Returns `{ connected: false, accounts: [] }` when no row exists.
- Returns `connected: true` only when `status === 'active'`; returns `connected: false` for other statuses while still listing the account.

For `/google/migrate-legacy`:

- Auth boundary cases.
- Returns `{ migrated: false, reason: 'kilo_owned_already_active' }` when a kilo-owned row already exists.
- Inserts a new row with `credential_profile: 'legacy'` when no row exists.
- Updates an existing legacy row in place; preserves the unique-on-instance_id constraint.
- Mirrors to the DO with `status: 'active'`.

**Practical scaffolding:**

The existing `services/kiloclaw/src/routes/controller.test.ts` already has the test setup for this file. The new tests can mock `getGoogleOAuthConnectionByInstanceId`, `updateGoogleOAuthConnectionTokenData`, the DO stub's `getConfig` / `updateGoogleOAuthConnection`, and `fetch` (for the Google token endpoint). The decrypt/encrypt helpers can be mocked to identity functions in tests so you assert the round-trip semantics rather than the crypto.

If the full set is too much for this PR, the highest-value subset to land here is: the auth boundaries on `/google/token`, the `invalid_grant → action_required` transition, and the capability-not-granted check. The rest can be a follow-up issue, but please don't merge with zero coverage on these branches.

apps/web/src/app/api/integrations/google/disconnect/route.ts
```diff
@@ -0,0 +1,85 @@
...
+
+function buildDisconnectPath(organizationId: string | undefined, queryParam: string): string {
+  if (organizationId) {
+    return `/organizations/${organizationId}/claw/settings?${queryParam}`;
+  }
+
+  return `/claw/settings?${queryParam}`;
+}
+
+export async function GET(request: NextRequest) {
```
St0rmz1: The disconnect endpoint mutates state on a `GET`, which is reachable via top-level CSRF even with NextAuth's defaults. Worth converting to `POST` before this ships.

**Why GET-mutates-state matters here, even with `SameSite=Lax`:**

NextAuth defaults the session cookie to `SameSite=Lax`, which is good and blocks the easy CSRF vectors:

- `` — cookies not sent.
- `<iframe src="...disconnect">` — cookies not sent.
- XHR / fetch from another origin — cookies not sent.
- Link previews from Slack, iMessage, etc. — those bots have no session at all.

What `SameSite=Lax` does NOT block is top-level GET navigation. Any of these in an attacker-controlled page will send the session cookie:

- `<a href="https://app.kilocode.com/api/integrations/google/disconnect">Click here for a coupon</a>`
- `<meta http-equiv="refresh" content="0;url=https://app.kilocode.com/api/integrations/google/disconnect">`
- `window.location = 'https://app.kilocode.com/api/integrations/google/disconnect'` (after a click handler)

A user who clicks any such link or hits a redirect chain ends up with a silently-deleted Google connection. Damage is bounded (they can reconnect), but it's annoying, support-visible, and the standard fix is small.

There's also a less adversarial case: misconfigured prefetchers, dev tools that follow links, or browser extensions that probe URLs can hit the endpoint with the user's cookies and silently disconnect them.

**Suggested fix:**

Convert the route to `POST` and submit from the settings UI as a form (or a fetch with the standard `__Host-next-auth.csrf-token` Double-Submit pattern this app already uses for NextAuth flows):

```ts
// apps/web/src/app/api/integrations/google/disconnect/route.ts
export async function POST(request: NextRequest) { /* ... existing body ... */ }
```

```tsx
// apps/web/src/app/(app)/claw/components/SettingsTab.tsx
<form action={googleCalendarDisconnectHref} method="POST">
  <Button type="submit" variant="outline" size="sm">Disconnect</Button>
</form>
```

If you want to keep the redirect-based UX (so success/error lands on the same `?success=...` query) the form-POST pattern handles that fine: NextResponse.redirect from the POST handler still works and the browser follows the 303/307.

Same treatment is worth considering for `/connect`, though connect is less urgent because the user has to go through Google's consent screen anyway, which gives them a chance to notice. Disconnect has no such interstitial.

**Test to add:**

- POST returns the existing redirect.
- GET returns 405 (or a redirect to settings with an error) so the old behavior can't accidentally come back.

apps/web/src/app/(app)/claw/components/SettingsTab.tsx
```diff
@@ -854,6 +857,26 @@ export function SettingsTab({
...
+    if (organizationId) {
+      params.set('organizationId', organizationId);
+    }
+
+    const qs = params.toString();
+    return qs.length > 0
+      ? `/api/integrations/google/disconnect?${qs}`
+      : '/api/integrations/google/disconnect';
+  }, [organizationId]);
+  const canSeeGoogleCalendar = !!user?.is_admin;
```
St0rmz1, line 906: The UI hides the Google Calendar card from non-admins, but the connect / callback / disconnect API routes don't enforce the same check. Worth mirroring the gate on the server so the rollout guarantee is real.

**What's there today:**

```ts
// SettingsTab.tsx
const canSeeGoogleCalendar = !!user?.is_admin;
{canSeeGoogleCalendar && ( /* Connect / Disconnect card */ )}
```

```ts
// All three routes: connect, callback, disconnect
const { user, authFailedResponse } = await getUserFromAuth({ adminOnly: false });
```

So a non-admin who knows the URLs (now public in this PR diff and this comment) can still hit:

- `GET /api/integrations/google/connect` → walks through Google consent against their own instance.
- `GET /api/integrations/google/callback?...` → completes the flow.
- `GET /api/integrations/google/disconnect` → tears it down.

**Why this is moderate not critical:**

`user.is_admin` here is the Kilocode staff admin flag, not an org admin or tenant role. The cross-tenant blast radius is zero: a non-admin can only act on their own personal instance or an org they're already a member of. So the impact is "non-admin can use a staff-gated feature on their own instance", not "non-admin can touch someone else's data".

That said, controls that live only in the UI tend to drift, and the rollout guarantee in the commit message ("only admin accounts can manage OAuth visibility and actions") isn't actually enforced server-side today. If the gate is meant as a staged rollout while you collect telemetry, an enthusiastic user finding the URL skews that data.

**Suggested fix (depends on intent, see Q1 in the review summary):**

If `is_admin` is the intended long-term gate:

```ts
// All three google routes
const { user, authFailedResponse } = await getUserFromAuth({ adminOnly: true });
if (authFailedResponse) {
  // Existing handling — redirect to sign-in or to the settings error page,
  // depending on whether the user is unauthenticated or just non-admin.
}
```

If `is_admin` is really standing in for a feature flag, swap to your feature-flag check on the server (and rename the UI variable so future readers don't conflate "admin" with "in the rollout"):

```ts
const isInGoogleCalendarRollout = await checkGrowthBookFlag(user.id, 'kc_google_calendar');
if (!isInGoogleCalendarRollout) return NextResponse.redirect(...);
```

**Tests worth adding (small, follow the existing connect/callback/disconnect patterns):**

- Admin user → flow proceeds.
- Non-admin user → redirected to settings with `error=forbidden` (or 403 if you prefer), and downstream side effects (`exchangeGoogleOAuthCode`, `clearKiloClawGoogleOAuthConnection`, etc.) are not called.

The TypeScript Quality Rules' "Security Boundary Test Coverage" applies here too: if you add the check, also add the negative-path test in the same PR.

apps/web/src/app/api/integrations/google/connect/route.ts
```diff
@@ -0,0 +1,105 @@
...
+export async function GET(request: NextRequest) {
+  let organizationId: string | undefined;
+
+  try {
+    const { user, authFailedResponse } = await getUserFromAuth({ adminOnly: false });
+    if (authFailedResponse) {
+      return NextResponse.redirect(new URL('/users/sign_in', APP_URL));
+    }
+
+    const capabilities = parseGoogleCapabilities(request.nextUrl.searchParams.get('capabilities'));
```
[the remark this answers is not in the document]
joshavant: Addressed in `323da4c90`: connect no longer accepts capability overrides from query params, and connect/disconnect routes are admin-only.

services/kiloclaw/controller/src/routes/google-oauth-token.ts
```diff
@@ -0,0 +1,80 @@
...
+    }>;
+  }>;
+};
+
+export function registerGoogleOAuthTokenRoutes(
+  app: Hono,
+  _expectedToken: string,
+  tokenProvider: TokenProvider
+): void {
+  app.post('/_kilo/google-oauth/token', async c => {
```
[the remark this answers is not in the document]
joshavant: Addressed in `8dc9b7525`: added bearer auth middleware for `/_kilo/google-oauth/*` routes, plus tests in `google-oauth-token.test.ts`.

apps/web/src/app/api/integrations/google/disconnect/route.ts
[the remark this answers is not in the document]
joshavant: Fixed in `323da4c90`: disconnect is now same-origin `POST` (not mutating `GET`), and the settings UI submits via POST.

services/kiloclaw/src/routes/controller.ts
joshavant: Resolved by `b44f55724`: broker capability scope mapping now includes `gmail_read` and `drive_read` (plus shim routing updates), so web/broker capability handling is consistent.

apps/web/src/app/api/integrations/google/callback/route.ts
joshavant: Addressed in `a8c5566c6`: callback Sentry payloads are now sanitized (no raw `code`, no raw `state`, no `allParams`). We now emit only `hasCode`, `hasState`, `stateHash`, and OAuth error fields, with test coverage.

packages/db/src/migrations/0089_gorgeous_bloodstorm.sql
```diff
@@ -0,0 +1,26 @@
...
+	"last_error" text,
+	"last_error_at" timestamp with time zone,
+	"connected_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"updated_at" timestamp with time zone DEFAULT now() NOT NULL,
+	CONSTRAINT "kiloclaw_google_oauth_connections_status_check" CHECK ("kiloclaw_google_oauth_connections"."status" IN ('active', 'action_required', 'disconnected')),
+	CONSTRAINT "kiloclaw_google_oauth_connections_credential_profile_check" CHECK ("kiloclaw_google_oauth_connections"."credential_profile" IN ('legacy', 'kilo_owned'))
+);
+--> statement-breakpoint
+ALTER TABLE "kiloclaw_google_oauth_connections" ADD CONSTRAINT "kiloclaw_google_oauth_connections_instance_id_kiloclaw_instances_id_fk" FOREIGN KEY ("instance_id") REFERENCES "public"."kiloclaw_instances"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
```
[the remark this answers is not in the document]
joshavant: Follow-up: agreed — we are switching this FK to `ON DELETE NO ACTION` (no cascade) in the unshipped baseline (`schema.ts`, `0089`, and snapshot) so it lands as the final first-ship shape.

apps/web/src/app/api/integrations/google/callback/route.ts
joshavant: Good catch — agreed and fixed in follow-up. For org-owned callbacks we no longer require `instance.userId === user.id`; we now rely on org membership (`ensureOrganizationAccess`) plus `instance.organizationId` match. Added callback test coverage for “admin B reconnects instance created by admin A.”

apps/web/src/app/api/integrations/google/disconnect/route.ts
joshavant: Good call — fixed in follow-up. Disconnect now clears DO state first, then deletes the DB row, so failure mode avoids stale connected UI state. Added disconnect audit logging context (`instanceId`, actor, org, account identifiers) and test coverage for the DO-failure/no-DB-delete path plus call ordering.

services/kiloclaw/controller/src/legacy-google-migration.ts
joshavant: Good catch — addressed in `2cb920642`. Legacy migration now uses `mkdtemp` + scoped token file path, applies best-effort `0600` chmod after export, and performs recursive temp-dir cleanup with error logging instead of silent unlink failures.

services/kiloclaw/src/routes/controller.ts
joshavant: Addressed in 9f32509bc. Added comprehensive tests for /google/token, /google/status, and /google/migrate-legacy auth boundaries plus state-transition and error branches, including the kilo_owned_already_active no-op path that returns migrated=false with reason kilo_owned_already_active. Verified with pnpm --filter kiloclaw test -- "src/routes/controller.test.ts" (66 files, 1360 tests passed).

apps/web/src/app/api/integrations/google/disconnect/route.ts
joshavant: Addressed. Added a regression test in b9a37df57 that asserts GET /api/integrations/google/disconnect redirects with error=method_not_allowed and does not clear DO/DB state. POST disconnect behavior remains covered and unchanged.

packages/db/src/schema.ts
```diff
@@ -3686,6 +3686,74 @@ export const kiloclaw_instances = pgTable(
...
+  oauth?: string[];
+};
+
+export const kiloclaw_google_oauth_connections = pgTable(
+  'kiloclaw_google_oauth_connections',
+  {
+    id: uuid()
+      .default(sql`gen_random_uuid()`)
+      .primaryKey()
+      .notNull(),
```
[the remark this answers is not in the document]
joshavant: Addressed in 1a23e9dea (and preserved through rebase). We now explicitly delete `kiloclaw_google_oauth_connections` rows for the user-owned instance IDs before deleting `kiloclaw_instances` in `apps/web/src/lib/user.ts`, so soft-delete no longer hits the `NO ACTION` FK.

apps/web/src/lib/user.ts
```diff
@@ -657,11 +660,30 @@ export async function softDeleteUser(userId: string) {
     await tx.delete(device_auth_requests).where(eq(device_auth_requests.kilo_user_id, userId));
     await tx.delete(auto_top_up_configs).where(eq(auto_top_up_configs.owned_by_user_id, userId));
     await tx.delete(kiloclaw_access_codes).where(eq(kiloclaw_access_codes.kilo_user_id, userId));
+    // kiloclaw_subscriptions, kiloclaw_cli_runs, kiloclaw_google_oauth_connections,
+    // and kiloclaw_inbound_email_aliases all reference kiloclaw_instances.
+    // Delete them before instances to satisfy FK constraints.
+    await tx.delete(kiloclaw_subscriptions).where(eq(kiloclaw_subscriptions.user_id, userId));
```
St0rmz1: Nice use of `inArray` + instance subquery for the OAuth connections delete below. Flagging the three new deletes above it though, because they silently change the soft-delete contract.

```ts
await tx.delete(kiloclaw_subscriptions).where(eq(kiloclaw_subscriptions.user_id, userId));
await tx
  .delete(kiloclaw_earlybird_purchases)
  .where(eq(kiloclaw_earlybird_purchases.user_id, userId));
await tx.delete(kiloclaw_email_log).where(eq(kiloclaw_email_log.user_id, userId));
```

**The contract this breaks:** on `main` these three tables are intentionally kept by `softDeleteUser` (see the "What is kept" section of the docstring). Three existing tests encode that:

- `user.test.ts:1139` asserts `should retain kiloclaw_earlybird_purchases`
- `user.test.ts:1204` asserts `should retain kiloclaw_subscriptions`
- `user.test.ts:1338` asserts `should retain kiloclaw_email_log rows`

CI `test` is currently red on the first two (`Expected: 1, Received: 0`).

**Why it matters:** subscriptions and earlybird purchases are financial records used for refunds, chargebacks, and tax questions, and the email log is relevant for retention and consent history. Dropping them on soft-delete removes the paper trail that the retain tests were asserting we would keep.

**Why the inline comment's rationale doesn't hold:** the comment says these need to go before an instance delete for FK reasons, but `softDeleteUser` never deletes `kiloclaw_instances`. The precondition blocks soft-delete when active instances exist, and destroyed ones stay in place. Only `kiloclaw_google_oauth_connections` needed FK aware cleanup, and that is already handled in the delete below.

**Suggested fix:** drop these three lines, and update the comment above the OAuth connections delete to explain the real reason (credentials cleanup, not FK ordering):

```ts
// kiloclaw_google_oauth_connections holds encrypted refresh tokens scoped
// to a user's instances. Delete those rows explicitly during soft-delete so
// we don't leave long lived Google credentials behind after anonymization.
```

A small positive test ("soft-delete removes OAuth connections for the user's instances") would also make that intent stick.

services/kiloclaw/controller/src/legacy-google-migration.ts
```diff
@@ -0,0 +1,227 @@
...
+
+    const endpoint = endpointFor(options.checkinUrl);
+    const response = await fetch(endpoint, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${options.apiKey}`,
+        'x-kiloclaw-gateway-token': options.gatewayToken,
+      },
+      body: JSON.stringify(body),
```
St0rmz1, line 243: The legacy migration flow is a nice, well scoped utility. One thing worth resolving before merge though: the CodeQL "file data in outbound network request" alert on this `fetch` is currently live and keeping the `CodeQL` check red.

**What CodeQL is flagging:**
The request body here includes `refreshToken`, `oauthClientId`, `oauthClientSecret`, `scopes`, and `capabilities`, all derived from two JSON files on disk (the temp file from `gog auth tokens export` and the client credentials file from `gog auth credentials list`). CodeQL treats "file content steers outbound HTTP body" as a data flow worth calling out because a tainted file can redirect what a trusted service sends off box.

**Why this is almost certainly a false positive in context:**
- The destination URL comes from `options.checkinUrl`, which is populated by the gateway/DO pair during startup, not from the filesystem.
- The outbound request is authenticated with both `Bearer ${apiKey}` and `x-kiloclaw-gateway-token`, so a rogue caller cannot just hit the broker.
- The on-disk sources are written by the same `gog` binary the container installs. The threat model would require an attacker who already has write access inside the sandbox, at which point they can do worse things directly.

Worth doing one of these anyway so the check goes green and future reviewers do not trip on it:

1. **Dismiss the alert in GitHub** with a short justification referencing the bullets above. Also drop a one line comment next to the `fetch` pointing to the alert ID so the rationale is discoverable in code too.
2. **Defend in depth** by narrowing the body to precomputed, validated fields before the fetch:

   ```ts
   const payload = {
     sandboxId: options.sandboxId,
     accountEmail: z.string().email().parse(exportPayload.email || email),
     accountSubject: z.string().min(1).parse(exportPayload.email || email),
     refreshToken: z.string().min(1).parse(exportPayload.refresh_token),
     oauthClientId: z.string().min(1).parse(credentials.clientId),
     oauthClientSecret: z.string().min(1).parse(credentials.clientSecret),
     scopes: z.array(z.string()).parse(exportPayload.scopes || []),
     capabilities: mapServicesToCapabilities(exportPayload.services || []),
   };
   ```

   That keeps CodeQL happy and protects us if the `gog` export format ever drifts and starts emitting unexpected types.

Option 1 is the lower effort path if the team is comfortable with the threat model. Option 2 is a small upgrade and also serves as documentation of the expected shape.

services/kiloclaw/controller/src/gog-shim.ts
```diff
@@ -0,0 +1,284 @@
...
+broker_capabilities=''
+case "$cmd" in
+  calendar|cal)
+    broker_capabilities='["calendar_read"]'
+    if [[ "\${KILOCLAW_GOOGLE_LEGACY_MIGRATION_FAILED:-}" == "1" ]]; then
+      exec "\${REAL_GOG}" "$@"
+    fi
+    ;;
+  gmail)
+    broker_capabilities='["gmail_read"]'
```
St0rmz1: Calendar routing here looks right. The `gmail` and `drive|docs|sheets` branches though will fail for any fresh OAuth user, because the broker never gets `gmail_read` or `drive_read` granted on their connection in the first place.

**Why that happens end to end:**

- The connect route uses `DEFAULT_GOOGLE_CAPABILITIES` (`apps/web/src/lib/integrations/google/capabilities.ts:33`), which is hardcoded to `[CALENDAR_READ]`, and the `capabilities` query override was intentionally removed in `323da4c90`. So Google consent for any new user grants only calendar scopes.
- `upsertKiloClawGoogleOAuthConnection` writes `capabilities: ['calendar_read']` into the row, plus `grants_by_source.oauth = ['calendar_read']`.
- Inside the container, an agent (or the user) runs something like `gog gmail messages list`. This shim matches the `gmail)` case and sends `{"capabilities":["gmail_read"]}` to `/_kilo/google-oauth/token`.
- On the worker side (`services/kiloclaw/src/routes/controller.ts:486`), `gmail_read` is now in `GOOGLE_CAPABILITY_SCOPES` so it passes the "unsupported capabilities" check, but it fails `hasRequiredCapabilities(capabilities, connection.capabilities)` at line 496 and the broker returns 412 `Requested capabilities are not granted for this instance`.
- The shim surfaces that as `[gog-wrapper] Requested capabilities are not granted for this instance` and exits 78.

**Legacy users are fine:**
`migrateLegacyGoogleCredentialsToBroker` calls `mapServicesToCapabilities` on the legacy `services` field, so if a user previously set up gog with `gmail` or `drive`, those capabilities land in `grants_by_source.legacy` during migration and the broker will grant them. This only breaks new OAuth only installs.

**KILOCLAW_GOOGLE_LEGACY_MIGRATION_FAILED does not save us here:**
That flag only flips to `1` when migration genuinely fails. For a fresh install (no legacy account), migration returns `no_legacy_account` and the flag stays `0`, so these branches still try the broker and still get 412.

**Suggested fix (smallest change, matches the PR title's Calendar scope):**
Fall through to the real binary for non calendar Google commands until the connect flow actually grants those scopes:

```bash
case "$cmd" in
  calendar|cal)
    broker_capabilities='["calendar_read"]'
    if [[ "${KILOCLAW_GOOGLE_LEGACY_MIGRATION_FAILED:-}" == "1" ]]; then
      exec "${REAL_GOG}" "$@"
    fi
    ;;
  gmail|drive|docs|sheets)
    # Broker does not yet cover these through OAuth consent. Legacy
    # migration may grant them via grants_by_source.legacy, but new
    # OAuth only installs will hit a 412 at the broker. Fall through
    # to the legacy binary until the connect flow requests the
    # corresponding scopes.
    exec "${REAL_GOG}" "$@"
    ;;
  *)
    exec "${REAL_GOG}" "$@"
    ;;
esac
```

**Alternative, if Gmail/Drive are meant to ship in this PR:**
Add `GMAIL_READ` (and `DRIVE_READ`) to `DEFAULT_GOOGLE_CAPABILITIES` so the consent screen actually asks for them. The PR title reads "Calendar integration" though, so I think option 1 matches the stated scope.

**Test worth adding either way:**
A broker test that asserts `hasRequiredCapabilities(['gmail_read'], ['calendar_read'])` returns 412, plus a shim test that asserts `gog gmail ...` does not call `/_kilo/google-oauth/token` (under option 1) or that `gog gmail ...` does succeed against a connection with only calendar grants (under option 2, with the consent change in place).

apps/web/src/lib/user.ts
joshavant: Addressed in 98af7df0c

services/kiloclaw/controller/src/legacy-google-migration.ts
joshavant: Follow-up: addressed in 38b6a9195 (+ formatter follow-up 0b35b2e0a). Added explicit payload validation/sanitization before fetch (email-like/non-empty checks, string[] scopes, capabilities only from normalized services) and documented the threat model inline. The CodeQL alert has now been dismissed with that rationale.

services/kiloclaw/controller/src/gog-shim.ts
joshavant: Addressed in 66cb4f391. The shim now routes only `calendar|cal` through the broker; `gmail|drive|docs|sheets` fall through to `gog.real` until OAuth consent includes those capabilities. Added/updated shim tests to assert this routing behavior.

apps/web/src/lib/kiloclaw/google-oauth-connections.ts
```diff
@@ -0,0 +1,216 @@
...
+      capabilities: effectiveCapabilities,
+    };
+  }
+
+  const status: KiloClawGoogleOAuthStatus = 'active';
+  const grantsBySource: GrantsBySource = {
+    oauth: oauthCapabilities,
+  };
+  const effectiveCapabilities = effectiveCapabilitiesFromGrants(grantsBySource);
+  await db.insert(kiloclaw_google_oauth_connections).values({
```
St0rmz1: Worth tightening the upsert flow here. The `db.select(...).limit(1)` above followed by this `db.insert(...)` isn't atomic, so two OAuth callbacks completing for the same `instance_id` in the same window will both see `existing === null`, both call `db.insert`, and the second one will throw a unique-constraint violation on `UQ_kiloclaw_google_oauth_connections_instance`.

**When this actually fires:**

It's narrow, since OAuth callbacks for the same instance usually serialize on a single user's browser. But it's reachable: a double-click on the Google consent screen's Continue button, a stale tab completing in parallel with a fresh one, or any retry scenario on the callback path. The outer `try/catch` in `callback/route.ts` turns the unique-violation into `error=connection_failed`, so the user sees a failure toast on what was effectively a successful retry.

**Fix:** use a single `INSERT ... ON CONFLICT (instance_id) DO UPDATE`. Drizzle supports this via `.onConflictDoUpdate({ target: kiloclaw_google_oauth_connections.instance_id, set: { ... } })`, which collapses the select/update/insert branches into one statement. The same table is already upserted that way in `services/kiloclaw/src/routes/controller.ts` (the `/google/migrate-legacy` path uses raw SQL with `ON CONFLICT (instance_id) DO UPDATE`), so there's a consistent pattern to follow.

**Suggested regression test:** a unit test that fires two `upsertKiloClawGoogleOAuthConnection` calls concurrently for the same `instance_id` and asserts both resolve successfully. Today that test would fail.

apps/web/src/app/(app)/claw/components/SettingsTab.tsx
joshavant: Not addressing this because scope changed and these endpoints are actually in use now.

apps/web/src/lib/kiloclaw/google-oauth-connections.ts
joshavant: Addressed in 2aa795312. The missing-row path now uses `insert(...).onConflictDoUpdate({ target: instance_id, set: ... })`, so concurrent callbacks for the same instance no longer fail on unique constraint races. Added regression coverage in `google-oauth-connections.test.ts` for concurrent upserts.

apps/web/src/lib/kiloclaw/google-oauth-connections.ts
```diff
@@ -0,0 +1,237 @@
...
+      target: kiloclaw_google_oauth_connections.instance_id,
+      set: {
+        account_email: input.accountEmail,
+        account_subject: input.accountSubject,
+        oauth_client_id: GOOGLE_WORKSPACE_OAUTH_CLIENT_ID,
+        oauth_client_secret_encrypted: null,
+        credential_profile: 'kilo_owned',
+        refresh_token_encrypted: encryptedRefreshToken,
+        scopes: nextScopes,
+        grants_by_source: grantsBySource,
```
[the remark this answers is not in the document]
joshavant: Good catch — addressed in a8d8e0a82 (plus formatter follow-up c671891f2). The conflict handler no longer writes `grants_by_source`/`capabilities` from insert-time payload. After insert/upsert, we re-read the winner row and run the normal merge update so legacy grants are preserved under callback races.

St0rmz1 (approved):

PR #2579, opened 2026-04-18
fix(kiloclaw): tighten checkout guardrails

jeanduplessis:
## Summary
Rejects invalid Kilo Pass upsell selections server-side while preserving standalone checkout promo-code support.

### Why this change is needed
The upsell flow allowed a monthly tier 19 selection to proceed with commit hosting even though that tier cannot fund the first commit period. This rule must be enforced on the server so invalid combinations cannot slip past UI checks. Promo-code support on standalone Stripe checkout also needed to remain intact after product clarified that spec requirement.

### How this is addressed
- Reject monthly tier `19` when paired with `hostingPlan='commit'` before creating a Stripe checkout session.
- Add regression coverage for the invalid monthly-19/commit combination and assert that no checkout session is created.
- Keep standalone Stripe checkout promo-code expectations aligned with product behavior while using intro pricing for the first-month discount.

## Verification
- `pnpm test -- apps/web/src/routers/kiloclaw-billing-router.test.ts`
- `pnpm typecheck`
- `pnpm format`

## Visual Changes
N/A

### Human Reviewer
- This PR is stacked on top of #2578. Review with that base in mind.
- Validate that the commit-hosting guard should stay tied to the current tier-to-credit relationship from the billing spec. If Kilo Pass pricing or included credits change, this validation rule will need to move with that policy.
- Validate that preserving promo-code support on standalone Stripe checkout is intended product behavior and matches the updated spec root PR.

### Code Reviewer Agent
<details>
  - Validation lives in `createKiloPassUpsellCheckout` and intentionally rejects before Stripe session creation.
  - Regression coverage was added in `apps/web/src/routers/kiloclaw-billing-router.test.ts` for the invalid monthly-19/commit path.
  - Followed `Kilo Pass Upsell Checkout` rule 4 from the billing spec and kept standalone checkout promo-code behavior aligned with updated product guidance.
</details>

### DIFF
--- a/apps/web/src/routers/kiloclaw-router.ts
+++ b/apps/web/src/routers/kiloclaw-router.ts
@@ -3334,7 +3334,7 @@ export const kiloclawRouter = createTRPCRouter({
         customer: stripeCustomerId,
         billing_address_collection: 'required',
         line_items: [{ price: priceId, quantity: 1 }],
-        allow_promotion_codes: false,
+        allow_promotion_codes: true,
         customer_update: { name: 'auto', address: 'auto' },
         tax_id_collection: { enabled: true, required: 'never' },
         success_url: successUrl,
--- a/apps/web/src/routers/kiloclaw-router.ts
+++ b/apps/web/src/routers/kiloclaw-router.ts
@@ -72,6 +72,7 @@ import {
 } from '@/lib/kiloclaw/stripe-price-ids.server';
 import { getStripePriceIdForKiloPass } from '@/lib/kilo-pass/stripe-price-ids.server';
 import { KiloPassTier, KiloPassCadence } from '@/lib/kilo-pass/enums';
+import { isKiloPassSelectionEligibleForKiloclawCommitUpsell } from '@/lib/kilo-pass/bonus';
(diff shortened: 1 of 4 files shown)


apps/web/src/routers/kiloclaw-router.ts
```diff
@@ -3411,6 +3411,12 @@ export const kiloclawRouter = createTRPCRouter({
       if (!stripeCustomerId) {
         throw new TRPCError({ code: 'BAD_REQUEST', message: 'Missing Stripe customer for user.' });
       }
+      if (input.hostingPlan === 'commit' && input.cadence === 'monthly' && input.tier === '19') {
```
RSO: Can't we base this on the actual number of credits granted under the plan? Otherwise, this will go stale if/when we ever update the tiers.

RSO (approved):
Approved to unblock, but please address my comment

apps/web/src/routers/kiloclaw-router.ts
jeanduplessis: I replaced the tier value guard with a shared helper, so the rule will now move with tier changes instead of going stale.

PR #2584, opened 2026-04-18
fix(kiloclaw): add activation-state billing contract

jeanduplessis:
## Summary
Adds explicit activation state to KiloClaw personal billing APIs and stops granting access before Stripe settlement completes.

### Why this change is needed
Stripe-created hosting rows exist before invoice settlement, but backend currently treats those rows as active access-granting subscriptions. That forces consumers to infer activation from status and funding fields, and it can unlock access before settlement converts the row into its activated credits-backed state. Backend contract needs to land first so UI follow-up can adopt one explicit field without mixing API and presentation changes.

### How this is addressed
- Added `activationState` to billing status subscription payloads and personal subscription list/detail payloads.
- Mapped `payment_source='stripe'` rows to `pending_settlement` and all other rows to `activated`.
- Updated shared access-state logic so pending-settlement rows do not grant subscription access.
- Preserved detached-row reprovision recovery by recognizing detached pre-settlement rows separately from access checks.
- Added router coverage for pending-settlement billing status, list/detail payloads, and access-gate denial.

## Verification
- `pnpm test -- apps/web/src/routers/kiloclaw-billing-router.test.ts apps/web/src/routers/admin-kiloclaw-user-router.test.ts`
- `pnpm typecheck`
- `pnpm format`
- No manual testing performed. Backend contract change only.

## Visual Changes
N/A

### Human Reviewer
- Validate stacked rollout boundary: this PR adds backend contract and access semantics only. Existing UI still ignores `activationState` until PR 7.
- Validate detached-row reprovision behavior: pending-settlement rows no longer grant access, but detached Stripe rows still bootstrap onto an instance when recovery logic needs them.

### Code Reviewer Agent
<details>

  - `getKiloClawSubscriptionActivationState()` treats `payment_source='stripe'` as `pending_settlement`, and `getKiloClawSubscriptionAccessReason()` now returns `null` for that state before status checks run.
  - `resolveDetachedAccessGrantingPersonalSubscription()` intentionally avoids the shared access helper now, because recovery still needs to detect detached pre-settlement rows even though they are not access-granting.
  - Payload additions land in `getBillingStatus`, `getActivePersonalBillingStatus`, `listPersonalSubscriptions`, and `getSubscriptionDetail`.
</details>


apps/web/src/lib/kiloclaw/access-state.ts
```diff
@@ -44,11 +45,22 @@ function subscriptionRecency(subscription: KiloClawSubscription): number {
...
+
+  return 'activated';
+}
+
 export function getKiloClawSubscriptionAccessReason(
   subscription: KiloClawSubscriptionAccessRecord | null | undefined,
   now = new Date()
 ): KiloClawAccessReason | null {
   if (!subscription) return null;
+  if (getKiloClawSubscriptionActivationState(subscription) === 'pending_settlement') return null;
```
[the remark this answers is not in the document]
jeanduplessis: Folded minimal UI compatibility into PR6 in 7775619b0. Pending-settlement rows now derive `bannerState = none` and `lockReason = null`, so this backend slice no longer shows subscribed banner or generic lock dialog before PR7 adds dedicated processing UI. Added unit coverage for both branches in `billing-types.test.ts`.

apps/web/src/app/(app)/claw/components/billing/billing-types.ts
```diff
@@ -175,6 +177,9 @@ export function deriveBannerState(billing: ClawBillingStatus): ClawBannerState {
 
 export function deriveLockReason(billing: ClawBillingStatus): ClawLockReason {
   if (!billing.hasAccess) {
+    if (billing.subscription?.activationState === 'pending_settlement') {
+      return null;
```
RSO, line 183: Why `null`, and not a specific string?

RSO (approved):

apps/web/src/app/(app)/claw/components/billing/billing-types.ts
jeanduplessis: In this instance we don't want the lock to show up in the UI because this is a temporary state in-between stripe webhook events.

So returning null means "there is no lock reason" and we shouldn't show anything in the UI.

I'll add a comment to the code to make this clear.

PR #2623, opened 2026-04-20
feat(claw): skip intro screen and auto-start personal onboarding

Drixled:
## Summary

Cuts a step out of the personal KiloClaw onboarding so new users land directly on `BotIdentityStep` while provisioning starts in the background.

- Removes the intro `CreateInstanceCard` from the personal create-first happy path and lands the user on `BotIdentityStep` as step 1 of 4 (or 5 with pairing).
- Adds a shared `startClawProvision` helper and a one-shot `autoProvisionOnMount` effect in `ClawOnboardingFlow` so provisioning starts automatically on fresh entries (legacy card is still rendered for org onboarding and as a failure fallback).
- Introduces a `skipCreateInstanceStep` flag in the onboarding state machine that only controls the render decision and step numbering, kept separate from the auto-start side effect so refreshes never trigger `provision` a second time.
- Adds a session-scoped `onboarding-in-progress` marker (via new `safeSessionStorage`) so a mid-onboarding refresh keeps the wizard mounted instead of jumping to `post-provisioning`/`complete`, without leaking across account switches or browser restarts.
- Re-seeds `createFlowStartedAt` when the marker is detected on mount so the `ClawNewLoader` status timestamp filter still lets fresh status data through after reload.
- Narrows `claw_page_viewed` back to personal create-first onboarding, adds a `trigger: 'auto' | 'manual'` field to `claw_create_instance_clicked`, and removes the now-duplicate page-view capture from `CreateInstanceCard`.
- Centralizes legacy and compact onboarding step numbers/totals in `ClawOnboardingFlow.state.ts` so leaf components no longer duplicate literal step defaults.
- Org onboarding is unchanged: no skip, no auto-start, same intro card.

## Verification

- [x] `pnpm --filter web typecheck`
- [x] `pnpm --filter web lint`
- [x] `pnpm --filter web test -- claw` (30 suites, 471 tests pass, including 5 new state-machine tests)
- [x] `pnpm format`
- [ ] Manual browser verification pending on a running dev instance.

## Visual Changes

| Before | After |
| ------ | ----- |
| Personal onboarding started on the `Your AI assistant, always on` intro card and required a `Get Started` click before provisioning began. | Personal onboarding starts directly on `BotIdentityStep` (step 1 of 4) with the provisioning banner visible, and provisioning is auto-triggered in the background. |

## Reviewer Notes

- Risk areas worth a careful look:
  - `ClawOnboardingFlow.tsx` auto-start effect: ref-guarded against strict-mode double-invocation and pending mutations; disabled for org flow; disabled on refresh via `autoProvisionOnMount` so `provision` is never called twice.
  - `ClawNewClient.tsx` refresh recovery: `createFlowStartedAt` is re-seeded when the `onboarding-in-progress` marker is read on mount, otherwise the provisioning step can't detect `instanceRunning` after reload.
  - `safeSessionStorage` in `apps/web/src/lib/localStorage.ts` is a new helper. Note: `window.localStorage`/`window.sessionStorage` are accessed at module init; on restricted browsers (sandboxed iframes, Firefox with `dom.storage.enabled=false`) this can throw `SecurityError`. Low-probability for authenticated `/claw/new` users but worth a follow-up to defer access.
- Rollout is intentionally scoped: legacy `CreateInstanceCard` remains in place as the auto-start-failure fallback and is still the org onboarding entry point, so the blast radius is personal flow only.

### DIFF
--- a/apps/web/src/app/(app)/claw/components/BotIdentityStep.tsx
+++ b/apps/web/src/app/(app)/claw/components/BotIdentityStep.tsx
@@ -6,6 +6,10 @@ import { AnimatePresence, motion, useReducedMotion } from 'motion/react';
 import { Button } from '@/components/ui/button';
 import { Input } from '@/components/ui/input';
 import { OnboardingStepView } from './OnboardingStepView';
+import {
+  LEGACY_ONBOARDING_IDENTITY_STEP,
+  LEGACY_ONBOARDING_TOTAL_STEPS,
+} from './ClawOnboardingFlow.state';
 import type { BotIdentity } from './claw.types';
 import { cn } from '@/lib/utils';
 
@@ -59,8 +63,8 @@ const NATURE_PRESETS: NaturePreset[] = [
 
 export function BotIdentityStep({
   instanceRunning,
-  currentStep = 2,
-  totalSteps = 5,
+  currentStep = LEGACY_ONBOARDING_IDENTITY_STEP,
+  totalSteps = LEGACY_ONBOARDING_TOTAL_STEPS,
   onContinue,
 }: {
   instanceRunning: boolean;
--- a/apps/web/src/app/(app)/claw/components/BotIdentityStep.tsx
+++ b/apps/web/src/app/(app)/claw/components/BotIdentityStep.tsx
@@ -6,10 +6,6 @@ import { AnimatePresence, motion, useReducedMotion } from 'motion/react';
 import { Button } from '@/components/ui/button';
 import { Input } from '@/components/ui/input';
 import { OnboardingStepView } from './OnboardingStepView';
-import {
-  LEGACY_ONBOARDING_IDENTITY_STEP,
-  LEGACY_ONBOARDING_TOTAL_STEPS,
-} from './ClawOnboardingFlow.state';
 import type { BotIdentity } from './claw.types';
 import { cn } from '@/lib/utils';
 
@@ -63,13 +59,13 @@ const NATURE_PRESETS: NaturePreset[] = [
 
 export function BotIdentityStep({
   instanceRunning,
-  currentStep = LEGACY_ONBOARDING_IDENTITY_STEP,
-  totalSteps = LEGACY_ONBOARDING_TOTAL_STEPS,
+  currentStep,
+  totalSteps,
   onContinue,
 }: {
   instanceRunning: boolean;
-  currentStep?: number;
-  totalSteps?: number;
+  currentStep: number;
+  totalSteps: number;
   onContinue: (identity: BotIdentity) => void;
 }) {
   const [botName, setBotName] = useState('');
--- a/apps/web/src/app/(app)/claw/components/BotIdentityStep.tsx
+++ b/apps/web/src/app/(app)/claw/components/BotIdentityStep.tsx
@@ -59,9 +59,13 @@ const NATURE_PRESETS: NaturePreset[] = [
 
 export function BotIdentityStep({
   instanceRunning,
+  currentStep = 2,
+  totalSteps = 5,
   onContinue,
 }: {
   instanceRunning: boolean;
+  currentStep?: number;
+  totalSteps?: number;
   onContinue: (identity: BotIdentity) => void;
 }) {
   const [botName, setBotName] = useState('');
@@ -113,8 +117,8 @@ export function BotIdentityStep({
 
   return (
     <OnboardingStepView
-      currentStep={2}
-      totalSteps={5}
+      currentStep={currentStep}
+      totalSteps={totalSteps}
       title="Give your bot an identity"
       description="Make it yours. You can always change this later."
       showProvisioningBanner={!instanceRunning}
--- a/apps/web/src/app/(app)/claw/new/ClawNewClient.tsx
+++ b/apps/web/src/app/(app)/claw/new/ClawNewClient.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 import { Loader2 } from 'lucide-react';
 import { useQuery } from '@tanstack/react-query';
 import { useKiloClawStatus } from '@/hooks/useKiloClaw';
@@ -22,6 +22,58 @@ import { WelcomePage } from '../components/billing/WelcomePage';
 
 const ClawOnboardingWithBoundary = withStatusQueryBoundary(ClawOnboardingFlow);
 
+export function getCreateFlowStartedAtFromMarker(
+  markerForCurrentUser: boolean,
+  now: () => number = Date.now
+): number | null {
+  return markerForCurrentUser ? now() : null;
+}
+
+export function getClawNewMode({
+  createFlowStartedAt,
+  hasActiveInstance,
+  onboardingInProgress,
+}: {
+  createFlowStartedAt: number | null;
+  hasActiveInstance: boolean;
+  onboardingInProgress: boolean;
+}): ClawOnboardingMode {
+  return createFlowStartedAt !== null || !hasActiveInstance || onboardingInProgress
+    ? 'create-first'
+    : 'post-provisioning';
+}
+
+export function getCreateFirstFlowFlags({
+  mode,
+  autoStartFailed,
+  onboardingInProgress,
+  hasActiveInstance,
+}: {
+  mode: ClawOnboardingMode;
+  autoStartFailed: boolean;
+  onboardingInProgress: boolean;
+  hasActiveInstance: boolean;
+}): {
+  skipCreateInstanceStep: boolean;
+  autoProvisionOnMount: boolean;
+} {
+  return {
+    skipCreateInstanceStep: mode === 'create-first' && !autoStartFailed,
+    autoProvisionOnMount:
+      mode === 'create-first' && !autoStartFailed && !onboardingInProgress && !hasActiveInstance,
+  };
+}
+
+export function shouldApplyCreateFlowCallback({
+  activeUserId,
+  callbackUserId,
+}: {
+  activeUserId: string | null;
+  callbackUserId: string | null;
+}): boolean {
+  return activeUserId === callbackUserId;
+}
+
 function LoadingState() {
   return (
     <div
@@ -39,18 +91,18 @@ function ClawNewLoader({
   billingUpdatedAt,
   skipCreateInstanceStep,
   autoProvisionOnMount,
-  personalUserId,
   onCreateFlowStarted,
   onCreateFlowFailed,
+  onCreateFlowComplete,
 }: {
   mode: ClawOnboardingMode;
   createFlowStartedAt: number | null;
   billingUpdatedAt: number;
   skipCreateInstanceStep: boolean;
   autoProvisionOnMount: boolean;
-  personalUserId: string | null;
   onCreateFlowStarted: () => void;
   onCreateFlowFailed: () => void;
+  onCreateFlowComplete: () => void;
 }) {
   const statusQuery = useKiloClawStatus();
 
@@ -67,9 +119,9 @@ function ClawNewLoader({
         createFlowStarted={createFlowStartedAt !== null}
         skipCreateInstanceStep={skipCreateInstanceStep}
         autoProvisionOnMount={autoProvisionOnMount}
-        personalUserId={personalUserId}
         onCreateFlowStarted={onCreateFlowStarted}
         onCreateFlowFailed={onCreateFlowFailed}
+        onCreateFlowComplete={onCreateFlowComplete}
       />
     );
   }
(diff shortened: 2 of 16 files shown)


apps/web/src/lib/kiloclaw/onboarding-progress.ts
```diff
@@ -0,0 +1,29 @@
...
+ * billing already reports an active instance. Without this marker, a reload
+ * during auto-started provisioning would jump the user to `post-provisioning`
+ * (or `complete`) and silently skip identity/permissions/channels input.
+ *
+ * This marker is intentionally session-scoped so stale progress does not leak
+ * across account switches or browser restarts.
+ *
+ * Scope: personal flow only. Organization onboarding is unchanged.
+ */
+const CLAW_PERSONAL_ONBOARDING_IN_PROGRESS_KEY = 'kiloclaw-personal-onboarding-in-progress';
```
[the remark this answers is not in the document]
Drixled: Fixed.

St0rmz1:
Nice work centralizing the step constants and pulling `startClawProvision` into a shared helper. The 5 new state-machine tests are a solid addition for the render decision logic.

One gap worth filling before this ships: there is no test coverage for the pieces most likely to regress, which are the effect wiring and the marker lifecycle rather than the pure state machine.

Specifically, these behaviors are currently only verified by manual testing:

1. The `hasAutoStartedRef` guard in `ClawOnboardingFlow.tsx` preventing double invocation of `startClawProvision` under React strict mode.
2. `ClawNewLiveClient` reading the `onboarding-in-progress` marker on mount with correct user scoping, setting it on create flow start, and clearing it on completion or error.
3. The auto-start failure path flipping `skipCreateInstanceStep` off and falling back to the legacy `CreateInstanceCard` with a manual retry button.

Why this matters: the effect has an `eslint-disable react-hooks/exhaustive-deps` comment, which is fine today because of the ref guard, but means a future edit to the dep list will not fail CI even if it silently breaks double-invocation protection. Same story for the marker: if someone later tweaks the `[currentUserId, userLoading]` deps, the wizard could stop persisting across refreshes without any test flagging it.

Suggested additions (all cheap, jsdom only, no network):

- A React Testing Library test that mounts `ClawOnboardingFlow` twice in strict mode with `autoProvisionOnMount=true` and asserts the provision mutation spy is called exactly once.
- A test that seeds the marker in `sessionStorage`, mounts `ClawNewLiveClient`, and asserts the wizard renders in create-first mode with the identity step visible.
- A test that drives `renderStep` to `complete` and asserts the marker is cleared.

Not a blocker, but these three tests would lock in the most fragile parts of the change.

apps/web/src/lib/localStorage.ts
```diff
@@ -15,28 +15,34 @@ const nullStorage = {
...
+        // Silently fail if storage is unavailable
+      }
+    },
+    removeItem: (key: string): void => {
+      try {
+        storage.removeItem(key);
+      } catch {
+        // Silently fail if storage is unavailable
+      }
+    },
```
St0rmz1, line 46: Small robustness regression worth tightening up here. The old implementation accessed `window.localStorage` inside each operation's try/catch, so if storage was disabled the wrapper degraded gracefully per call. After this refactor, `createSafeStorage(window.localStorage)` evaluates `window.localStorage` once at module import time.

In browsers that throw `SecurityError` on the getter itself (sandboxed iframes with storage disabled, Firefox with `dom.storage.enabled=false`, Safari with certain ITP configurations, cookie-blocked third-party contexts), this now throws synchronously at import, which takes down every module that imports from `@/lib/localStorage`. The PR description flags this as a follow-up, so you are already aware, but it is cheap to fix in this PR.

One option is to pass a getter instead of a reference:

```ts
function createSafeStorage(getStorage: () => Storage) {
  return {
    getItem: (key: string): string | null => {
      try { return getStorage().getItem(key); } catch { return null; }
    },
    setItem: (key: string, value: string): void => {
      try { getStorage().setItem(key, value); } catch {}
    },
    removeItem: (key: string): void => {
      try { getStorage().removeItem(key); } catch {}
    },
  };
}

export const safeLocalStorage = IS_SERVER ? nullStorage : createSafeStorage(() => window.localStorage);
export const safeSessionStorage = IS_SERVER ? nullStorage : createSafeStorage(() => window.sessionStorage);
```

Blast radius today is low (authenticated `/claw/new` users are unlikely to be in such a context), so this is not a blocker, but it keeps the degradation behavior the old code provided.

apps/web/src/app/(app)/claw/new/ClawNewClient.tsx
```diff
@@ -93,12 +111,47 @@ export function ClawNewClient({
...
+  const [onboardingInProgress, setOnboardingInProgress] = useState(false);
+  // Re-seed `createFlowStartedAt` when resuming onboarding after a refresh so
+  // `ClawNewLoader`'s status timestamp filter lets fresh status data through.
   const [createFlowStartedAt, setCreateFlowStartedAt] = useState<number | null>(null);
-  const onCreateFlowStarted = useCallback(() => setCreateFlowStartedAt(Date.now()), []);
-  const onCreateFlowFailed = useCallback(() => setCreateFlowStartedAt(null), []);
-
-  if (billingQuery.isLoading) {
+  // Falls back to showing the legacy intro card when auto-start fails so the
+  // user still has a manual retry button.
```
St0rmz1, line 175: The marker lifecycle is nicely thought through: read on mount, write on start, clear on terminal state. One subtle coupling worth calling out though: `clearPersonalOnboardingInProgress` is called from inside `ClawOnboardingFlowInner` when `renderStep` becomes `complete` or `error`, but the corresponding `onboardingInProgress` React state here is only (re)read by this `useEffect`, and its deps are `[currentUserId, userLoading]`. After the wizard finishes, `sessionStorage` is cleared but this local state stays `true` until the user navigates away and comes back.

Why it matters: today the user has already reached the terminal render step by the time this diverges, so `mode` no longer controls what they see and there is no user-visible bug. The concern is for future maintainers. Any new code that reads `onboardingInProgress` at this level (to gate analytics, conditionally render something, etc.) will see a stale `true` after completion, and the bug will only show up in whatever that new code does, not here.

Two straightforward options:

1. Lift the clear up to the parent via a new callback, e.g. `onCreateFlowCompleted`, and reset `onboardingInProgress` alongside the `sessionStorage` clear. Keeps the two stores in sync by construction.
2. Drop the local `onboardingInProgress` state entirely and derive it from `readPersonalOnboardingInProgress(currentUserId)` on each render, or from a single source of truth. More work but removes the class of bug.

Not a blocker for this PR, especially if you are planning to keep the state narrow.

apps/web/src/app/(app)/claw/new/ClawNewClient.tsx
```diff
@@ -93,12 +111,47 @@ export function ClawNewClient({
...
+  }, [currentUserId, userLoading]);
+
+  const onCreateFlowStarted = useCallback(() => {
+    markPersonalOnboardingInProgress(currentUserId);
+    setOnboardingInProgress(true);
+    setAutoStartFailed(false);
+    setCreateFlowStartedAt(Date.now());
+    setSyncedMarkerUserId(currentUserId);
+  }, [currentUserId]);
+
```
St0rmz1, line 206: Edge case worth noting. Both `onCreateFlowStarted` and `onCreateFlowFailed` are memoized with `[currentUserId]` and get captured in `ClawOnboardingFlowInner`'s auto-start effect closure. If a user switches accounts (or logs out) while a provision mutation is in flight, the mutation's `onError` handler will invoke the OLD `onCreateFlowFailed`, which calls `clearPersonalOnboardingInProgress(oldUserId)` (fine, correctly scoped to user A) and then `setAutoStartFailed(true)` on the `ClawNewLiveClient` state that now belongs to user B.

Why it matters: user B could briefly see the legacy intro card with no context, because user A's provision failed after the account switch. They click "Get Started" and recover, so this is a minor UX glitch in an uncommon flow rather than a correctness or security issue. Flagging because the prop drilling of `personalUserId` down through `ClawNewLoader`, `withStatusQueryBoundary`, and `OrgClawNewClient` makes this surface a bit wider than strictly necessary.

One cleaner alternative: derive `personalUserId` inside `ClawOnboardingFlowInner` directly via `useUser()` instead of passing it down as a prop. That both eliminates the prop-drilling chain (including the `personalUserId={null}` additions in `OrgClawNewClient.tsx` and `withStatusQueryBoundary.tsx`) and reads the current user at the moment the effect fires, which sidesteps the stale-closure case entirely. The failure callback could then just notify the parent that auto-start failed, without carrying a user ID.

Fine to skip for this PR if you want to keep the diff scoped; documenting here so we have a record.

apps/web/src/app/(app)/claw/components/BotIdentityStep.tsx
```diff
@@ -59,9 +63,13 @@ const NATURE_PRESETS: NaturePreset[] = [
 
 export function BotIdentityStep({
   instanceRunning,
+  currentStep = LEGACY_ONBOARDING_IDENTITY_STEP,
```
St0rmz1: Good move pulling these into shared constants in `ClawOnboardingFlow.state.ts`. One fragility to flag: every leaf step component now defaults its `currentStep` and `totalSteps` props to the `LEGACY_*` constants when the caller omits them. Today that works because `ClawOnboardingFlow.tsx` passes the props at every call site, but a future caller that forgets (or a test harness, or a Storybook story added later) will silently render "Step 2 of 5" on an identity screen that should be "Step 1 of 4" in the compact flow. The type system cannot catch this because the props are optional.

The leaf components also read a bit misleadingly now: `BotIdentityStep` looks like it "knows" it is step 2 of 5, when in fact the whole point of this refactor is that it does not know the flow at all.

Two cheap options, pick whichever fits your preference:

1. Make `currentStep` and `totalSteps` required props (drop the defaults). `ClawOnboardingFlow.tsx` already passes them everywhere in this PR, and `ClawOnboardingFakeWalkthrough.tsx` could pass them too. One-line type change per component, plus a few extra prop passes at the fake-walkthrough call sites.
2. Keep the defaults but drop a short comment noting they are legacy-flow only, so future maintainers do not mistake them for "the canonical values" when they add a new caller.

Either is fine; option 1 is cleaner but option 2 is basically free.

(Same pattern applies to `ChannelPairingStep.tsx`, `ChannelSelectionStep.tsx`, `PermissionStep.tsx`, and `ProvisioningStep.tsx`.)

St0rmz1 (approved):

apps/web/src/lib/localStorage.ts
Drixled: Fixed. I changed `createSafeStorage` to take a getter instead of a storage reference, so `window.localStorage` / `window.sessionStorage` are now accessed lazily inside the per-operation try/catch instead of at module import time. That restores the old degradation behavior in restricted browser contexts.

apps/web/src/app/(app)/claw/new/ClawNewClient.tsx
Drixled: Fixed. I added `onCreateFlowComplete` so the parent clears its own `onboardingInProgress` state when the flow reaches a terminal step, instead of only clearing `sessionStorage` from inside the child flow. That keeps the React state and the storage marker aligned.

apps/web/src/app/(app)/claw/new/ClawNewClient.tsx
Drixled: Tightened this up. I removed the `personalUserId` prop-drilling and now read the current user directly inside `ClawOnboardingFlowInner` via `useUser()`. On the parent side, I also gated the async create-flow callbacks so an old-user callback will clear the old user’s marker but won’t mutate the new user’s page state after an account switch.

apps/web/src/app/(app)/claw/components/BotIdentityStep.tsx
Drixled: Fixed. I made `currentStep` and `totalSteps` required across the onboarding leaf components and updated all call sites, including the fake walkthrough. That removes the silent fallback-to-legacy-numbering behavior for future callers.

Drixled:
Added coverage for the fragile onboarding logic. I added a focused ClawNewClient helper test file for marker/mode/callback behavior and extended the onboarding state tests to lock in the canonical legacy step constants used by the walkthrough and leaf steps. Full React strict-mode effect coverage would need a jsdom-style render harness that this test setup doesn’t currently include, but the callback/mode/marker behavior is now explicitly covered in the claw test suite. @St0rmz1

RSO (changes requested):
I'm not a fan of this `useEffect` to auto-trigger provisioning. I'm going to propose a different route tomorrow.

RSO:
FYI, I've merged #2679, which is a different take on making the identity step the first one. Let me know if anything is missing from this PR that you want me to port, but I suggest we close this PR.

PR #2630, opened 2026-04-20
Bot - Apply profile config to bot-spawned Cloud Agent sessions

eshurakov:
## Summary

KiloBot previously spawned Cloud Agent sessions with no profile-derived configuration, bypassing the repo-binding + default profile layering that the web tRPC routers apply. As a result, bot-initiated sessions never received env vars, encrypted secrets, or setup commands from the owner's default profile or any repo-bound profile.

- Call \`mergeProfileConfiguration\` inside \`spawnCloudAgentSession\` so bot paths match the profile layering of the web tRPC routers (layer 1: repo binding, layer 2: owner's effective default profile).
- Extract \`resolvePlatformIntegrationOwner\` to narrow the XOR \`owned_by_organization_id\` / \`owned_by_user_id\` into a discriminated \`OwnerRef\`, eliminating the two pre-existing \`as string\` casts.
- Extract \`resolveBotSessionProfile\` as a thin wrapper around \`mergeProfileConfiguration\` so the spawn tool stays focused on orchestration and profile resolution stays unit-testable.
- Surface profile-resolution failures as a user-facing error response instead of silently spawning a session without profile config.

For org-owned platform integrations, \`ticketUserId\` is threaded through so \`getEffectiveDefaultProfileId\` can fall back to the user's personal default when the org has none.

## Verification

- \`pnpm format\`
- \`pnpm typecheck\`

## Visual Changes

N/A

## Reviewer Notes

- Semantic change: existing bot-spawned sessions will now pick up the owner's default profile and any repo-bound profile. Users with a default profile who did not expect it to apply to bot-spawned sessions will see behavior change.
- Legacy \`spawnCloudAgentSession\` helpers in \`apps/web/src/lib/slack-bot.ts\` and \`apps/web/src/lib/discord-bot.ts\` still bypass profile resolution. They are expected to be consolidated onto the new \`bot/\` tool-loop agent; kept out of scope here.
- No new logging of \`prepareInput\` was added, so \`encryptedSecrets\` do not leak into logs.


apps/web/src/lib/bot/tools/resolve-platform-integration-owner.ts
```diff
@@ -0,0 +1,28 @@
...
+  platformIntegration: PlatformIntegrationOwnerColumns
+): OwnerRef {
+  if (platformIntegration.owned_by_organization_id) {
+    return { kind: 'org', id: platformIntegration.owned_by_organization_id };
+  }
+  if (platformIntegration.owned_by_user_id) {
+    return { kind: 'user', id: platformIntegration.owned_by_user_id };
+  }
+  throw new Error('Platform integration has no owner (both organization and user are null).');
+}
```
RSO: This method also exists as `ownerFromIntegration` in `agent-runner.ts`. Let's consolidate. Also, the throw  should be unnecessary because the database does not accept a row without both `owned_by_user_id` or `owned_by_organization_id`

RSO (approved):

RSO:
Please note that this will only affect the Pound bot. But I'm going to work on migrating all users over to the new bot this week, so that should be fine.

PR #2632, opened 2026-04-20
Mobile - Add Home Tab With KiloClaw And Agent Sessions Overview

eshurakov:
## Summary

Introduce a new **Home tab** as the mobile app's landing screen and restructure navigation around it.

- New **Home** tab (default) with:
  - Personalized greeting (time-based, fetches display name via new `getProfile` tRPC procedure)
  - KiloClaw instance cards with status badge and latest message preview
  - Recent agent sessions section with compact session rows
  - Promo cards for users without instances or sessions
  - New-task CTA
  - Pull-to-refresh
- **KiloClaw tab** now routes directly into the user's last-active chat when instances exist; the last-active instance is persisted in local storage so tab switches resume the right conversation. The raw tab screen is kept only as an empty state for users with zero instances.
- Per-instance KiloClaw settings/dashboard/billing/changelog routes moved out of the tab group to `/kiloclaw/[instance-id]/*` so they stack on top of tabs instead of nesting inside them, letting the settings UI take over the full screen.
- Added an **instance picker** form-sheet reachable by tapping the chat title; `ScreenHeader` now supports `onTitlePress` (with a chevron affordance) and `onBack`.
- Small polish: `repo-picker` shows an Unlock icon for public repos to balance the lock icon.

## Verification

- [x] `pnpm format`
- [x] `pnpm typecheck`


apps/mobile/src/lib/hooks/use-kiloclaw-latest-message.ts
```diff
@@ -0,0 +1,87 @@
...
+  messages?: {
+    text?: string;
+    created_at?: string;
+    user?: { id?: string; name?: string };
+  }[];
+};
+
+async function fetchLatestMessage(creds: StreamChatCredentials): Promise<LatestMessage | null> {
+  const res = await fetch(
+    `${STREAM_CHAT_API_BASE}/channels/messaging/${creds.channelId}/query?api_key=${creds.apiKey}`,
```
iscekic, line 30: Could we do this via their sdk somehow? 🤔

iscekic (approved):
LGTM

apps/mobile/src/lib/hooks/use-kiloclaw-latest-message.ts
eshurakov: I think we can, but this looks easier because it does only one http call. SDK may open websocket.

I'm leaning towards keeping it like this, for now

PR #2650, opened 2026-04-21
fix(web): redirect docs domains

RSO:
## Summary
- Redirect requests from `docs.kilo.ai` and `docs.kilocode.ai` to the canonical `kilo.ai/docs` path.
- Preserve the original request path under `/docs` so existing docs links continue resolving after the domain migration.

## Verification
- `pnpm typecheck`
- `pnpm format`

## Visual Changes
N/A

## Reviewer Notes
- Host-based redirects are ordered before existing path redirects in `apps/web/next.config.mjs` so docs subdomain traffic is handled first.

### DIFF
--- a/apps/web/next.config.mjs
+++ b/apps/web/next.config.mjs
@@ -23,6 +23,43 @@ function validateGitLfs() {
 validateGitLfs();
 
 const monorepoRoot = resolve(import.meta.dirname, '../..');
+const docsRedirectHosts = ['docs.kilo.ai', 'docs.kilocode.ai'];
+
+const docsDomainRedirects = docsRedirectHosts.flatMap(host => [
+  {
+    source: '/',
+    has: [
+      {
+        type: 'host',
+        value: host,
+      },
+    ],
+    destination: 'https://kilo.ai/docs',
+    permanent: true,
+  },
+  {
+    source: '/docs/:path*',
+    has: [
+      {
+        type: 'host',
+        value: host,
+      },
+    ],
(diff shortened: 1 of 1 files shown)


apps/web/next.config.mjs
```diff
@@ -92,6 +92,28 @@ const nextConfig = {
...
+      },
+      {
+        source: '/:path*',
+        has: [
+          {
+            type: 'host',
+            value: 'docs.kilocode.ai',
+          },
+        ],
+        destination: 'https://kilo.ai/docs/:path*',
```
markijbema: i think this will result in doc/docs, for instance, this is a working url:
https://docs.kilo.ai/docs/collaborate

I think you only need to rewrite the host

markijbema (changes requested):

apps/web/next.config.mjs
RSO: Oh, than I don't even need this!

RSO: Actually, we do 🤦 so `docs.kilo.ai` should go to `kilo.ai/docs`, but `docs.kilo.ai/docs/foo` should go to `kilo.ai/docs/foo`. Will push an update

RSO: Addressed in efc52d2f2: root docs domains redirect to `https://kilo.ai/docs`, existing `/docs/*` paths stay under `https://kilo.ai/docs/*`, and other paths continue to redirect under `/docs/*`.

PR #2670, opened 2026-04-22
fix(ai-gateway): strip NUL bytes before microdollar_usage insert

marius-kilocode:
## Summary

Postgres `text` columns reject NUL bytes with `22021 invalid byte sequence for encoding "UTF8": 0x00`, which crashes the `microdollar_usage` CTE insert in `toInsertableDbUsageRecord` and silently leaves the request un-billed. Observed rate: **~1.9k silent billing-write failures/day** across FIM, chat, and responses paths.

- **Fix:** add `stripNulBytesInPlace` helper, apply it to both `core` and `metadata` right before returning from `toInsertableDbUsageRecord`.
- **Observability:** 5%-sampled `captureMessage` with the sanitized field list so we can identify the upstream source and fix it at the origin.
- **Tests:** pure helper, realistic body-sourced NULs (prompt content, model IDs, upstream response fields), defensively-constructed header-sourced NULs, clean no-NUL no-op.

Sentry: [KILOCODE-WEB-1G3Z](https://kilo-code.sentry.io/issues/7352205702/) — 1,363 events over ~5 weeks, multiple LLM paths.

## Why sanitize at the DB boundary

The realistic NUL source is JSON-body-derived fields — prompt content (`system_prompt_prefix`, `user_prompt_prefix`), LLM response fields (`model`, `inference_provider`, `message_id`, `finish_reason`, `upstream_id`, `requested_model`). HTTP header-sourced fields (`http_user_agent`, `machine_id`, `session_id`, etc.) can't realistically carry NULs because Node's `Headers` constructor rejects them per RFC 7230 — documented in the test.

Sanitizing at the `toInsertableDbUsageRecord` boundary closes the bleed for every LLM path at once with a single chokepoint, and defensively covers header-sourced fields in case a future code path bypasses `Headers` validation.

Once the sampled `captureMessage` surfaces the actual source field(s), the plan is to sanitize upstream and remove the defensive sanitizer.

## Blast radius

- Event sample (last 50 from KILOCODE-WEB-1G3Z):
  - `/api/fim/completions` (Mistral Codestral) — 66%
  - `/api/openrouter/responses` (grok-code-fast-1:free) — 26%
  - `/api/openrouter/chat/completions` — 6%
  - `/api/gateway/chat/completions` — 2%
- No KiloClaw endpoints affected.
- Current behavior: request succeeds, user not billed, analytics under-count. Low \$/day leak, but the `microdollar_usage` table also feeds abuse detection and the 'first usage' PostHog event — downstream analytics are under-counting.

## Risk

Low. Sanitizer is a one-shot walk over two flat objects, only modifies strings that actually contain a NUL (early-exits via `indexOf`), and the behavior is exercised by unit tests. No DB or schema changes.

## Test plan

- `pnpm --filter @kilocode/web exec jest src/lib/ai-gateway/processUsage.test.ts` — all 27 tests pass (6 new).
- `pnpm --filter @kilocode/web typecheck` — clean.
- `pnpm -w exec oxlint --config .oxlintrc.json apps/web/src/lib/ai-gateway/processUsage.ts apps/web/src/lib/ai-gateway/processUsage.test.ts` — clean.

## Follow-up

- Watch the sampled `captureMessage` in Sentry for a few days to identify which field is the dominant NUL source.
- Once known, sanitize at the source (likely `extractPromptInfo` or an upstream body parser) and remove this defensive sanitizer.


apps/web/src/lib/ai-gateway/processUsage.ts
```diff
@@ -182,6 +211,27 @@ export function toInsertableDbUsageRecord(
...
+      level: 'warning',
+      tags: { source: 'toInsertableDbUsageRecord' },
+      extra: {
+        fields: dirtyFields,
+        kilo_user_id,
+        requested_model: usageContextInfo.requested_model,
+        provider,
+      },
+    });
+  }
```
chrarnoldus: can we just make it a normal `console.warn` if we don't want to flood sentry?

marius-kilocode: Good call — switched to `console.warn` in a7c5058. Also dropped the 5% sampling gate since Sentry quota is no longer a concern; that'll let the dominant source field surface faster in Axiom (~40 events/day × 100% beats × 5%).

chrarnoldus (approved):

PR #2679, opened 2026-04-22
feat(claw): make bot identity the first onboarding step

RSO:
## Summary

- Show `BotIdentityStep` as the initial screen on `/claw/new` instead of the static `CreateInstanceCard`. Submitting the identity step now triggers instance provisioning.

## Verification

- Manually verified that `/claw/new` opens directly on "Give your bot an identity" for new users.
- Checked that the provisioning banner does not appear before the user submits the identity form.

## Visual Changes

https://github.com/user-attachments/assets/f20191a8-8912-4fb1-b6c2-30385bc0d2c8


jeanduplessis (commented):
We are losing some posthog events with this change.
We should make sure they are not going to cause a knock-on effect for how we measure things.

apps/web/src/app/(app)/claw/components/ClawOnboardingFlow.tsx
```diff
@@ -156,13 +155,15 @@ function ClawOnboardingFlowInner({
...
   const { data: isServiceDegraded } = useClawServiceDegraded();
+  useFeatureFlagVariantKey('button-vs-card');
   const posthog = usePostHog();
 
   useEffect(() => {
-    if (!flowState.createSetupActive || hasCapturedIdentityView.current) return;
+    if (flowState.renderStep !== 'identity' || hasCapturedIdentityView.current) return;
     hasCapturedIdentityView.current = true;
+    posthog?.capture('claw_page_viewed');
     posthog?.capture('claw_setup_identity_viewed');
```
RSO, line 165: @pedroheyerdahl FYI, these two events are going to be identical from here on out, but I didn't want to break any reporting on your side, so I kept them both.

pedroheyerdahl: appreciate the heads up! I’ll change the dbt logic to rename old event with the new name before exposing it for analytics. I'm okay with deprecating it going forward to avoid confusion. thanks!

jeanduplessis (approved):

PR #2686, opened 2026-04-22
feat(admin): add blocked-at backfill

markijbema:
## Summary

- Add an admin backfill for historical blocked users missing `blocked_at`.
- Add `/admin/api/backfills/blocked-at` GET/POST endpoints and a button on `/admin/backfills`.
- Backfill `blocked_at` from each user's current `updated_at` as the best available historical approximation; leave `blocked_by_kilo_user_id` null because there is no reliable system user for attribution.

## Verification

N/A

## Visual Changes

N/A

## Reviewer Notes

- Soft-deleted users are excluded so GDPR cleanup stays intact.
- The blocker remains `Unknown` for backfilled historical rows; nullable attribution is intentional because the original blocker cannot be recovered.


RSO (approved):

RSO:
🔨

PR #2699, opened 2026-04-22
feat(gastown): instrument container cold-start and mayor availability timing

jrf0110:
## Summary

- **Part 1 (Town DO side):** Wraps `container.fetch('/agents/start')` in `container-dispatch.ts` with timestamps and emits a `container.agent_start_fetch` AE event with `containerFetchMs`, `statusCode`, and `wasSuccess`. Also wraps the health ping in `ensureContainerReady()` (Town.do.ts) with timing and emits `container.health_ping` (distinguishing `ok` from `timeout`) and `container.ready_observed` (with `timeSinceContainerStartMs` correlating Town DO clock to container boot time using `startedAt` from the health response).
- **Part 2 (Container side — `startTime` transmission):** Adds `getStartTime()` export to `process-manager.ts`, adds `startedAt` field to `HealthResponse` type and the `/health` endpoint, and fixes the hardcoded `uptime: 0` in `main.ts` to use `getUptime()`.
- **Part 3 (Mayor availability milestones):** Adds `agent.startup_phase` structured logs in `startAgent()` (process-manager.ts) for `db_hydrated`, `sdk_ready`, and `session_created` phases with per-phase and cumulative `elapsedMs`. Adds `git_done` phase log in `runAgent()` (agent-runner.ts). Adds `agent.startup_complete` log. Adds `agent.pty_connected` log in `POST /agents/:agentId/pty` for both reuse and creation paths.
- **Part 4 (agent.spawned AE event):** Threads `containerFetchMs` from `startAgentInContainer()` result (changed return type from `boolean` to `{ started, containerFetchMs }`) through `dispatchAgent()` in scheduling.ts into the `agent.spawned` AE event. Updates all three direct callers of `startAgentInContainer` in Town.do.ts to destructure the new return type.

## Verification

- TypeScript compilation not run (per environment constraints), but all type changes are consistent: `HealthResponse` gets `startedAt?: string`, `GastownEventData` gets new timing fields, `startAgentInContainer` return type updated from `Promise<boolean>` to `Promise<{ started: boolean; containerFetchMs: number }>`, and all callers updated accordingly.
- Code follows existing patterns: uses `writeEvent()` for AE events, uses `log.info()` for structured container logs, uses Zod `.safeParse()` for the health response body IO boundary.

## Visual Changes

N/A

## Reviewer Notes

- The outer `try/catch` in `ensureContainerReady()` now has a nested `try/catch` for the health fetch itself. The outer catch handles `buildContainerConfig`/`getTownContainerStub` failures (as before); the inner catch handles the fetch timeout/error and emits the `container.health_ping` timeout event before re-swallowing the error.
- `main.ts` uses `getUptime()` from `process-manager.ts` instead of a `const` before imports (TypeScript prohibits variable declarations before import statements in ESM).
- `analytics.util.ts` adds 2 new blobs (blob15=healthPingStatus, blob16=containerStartedAt) and 6 new doubles (double11–double16) for the timing fields. This stays within AE's 20-blob/20-double limit.

### DIFF
--- a/services/gastown/src/dos/Town.do.ts
+++ b/services/gastown/src/dos/Town.do.ts
@@ -2438,29 +2438,33 @@ export class TownDO extends DurableObject<Env> {
         }
       }
 
-      const started = await dispatch.startAgentInContainer(this.env, this.ctx.storage, {
-        townId,
-        rigId: `mayor-${townId}`,
-        userId: townConfig.owner_user_id ?? rigConfig?.userId ?? townId,
-        agentId: mayor.id,
-        agentName: 'mayor',
-        role: 'mayor',
-        identity: mayor.identity,
-        beadId: '',
-        beadTitle: combinedMessage,
-        beadBody: '',
-        checkpoint: agents.readCheckpoint(this.sql, mayor.id),
-        // conversationHistory is no longer needed — the mayor's kilo.db
-        // is persisted to KV and hydrated on boot, preserving the full
-        // session state across container evictions.
-        gitUrl: rigConfig?.gitUrl ?? '',
-        defaultBranch: rigConfig?.defaultBranch ?? 'main',
-        kilocodeToken,
-        townConfig,
-        rigs: await this.rigListForMayor(),
-      });
+      const { started: mayorStarted } = await dispatch.startAgentInContainer(
+        this.env,
+        this.ctx.storage,
+        {
+          townId,
+          rigId: `mayor-${townId}`,
+          userId: townConfig.owner_user_id ?? rigConfig?.userId ?? townId,
+          agentId: mayor.id,
+          agentName: 'mayor',
+          role: 'mayor',
+          identity: mayor.identity,
+          beadId: '',
+          beadTitle: combinedMessage,
+          beadBody: '',
+          checkpoint: agents.readCheckpoint(this.sql, mayor.id),
+          // conversationHistory is no longer needed — the mayor's kilo.db
+          // is persisted to KV and hydrated on boot, preserving the full
+          // session state across container evictions.
+          gitUrl: rigConfig?.gitUrl ?? '',
+          defaultBranch: rigConfig?.defaultBranch ?? 'main',
+          kilocodeToken,
+          townConfig,
+          rigs: await this.rigListForMayor(),
+        }
+      );
 
-      if (started) {
+      if (mayorStarted) {
         agents.updateAgentStatus(this.sql, mayor.id, 'working');
         this._mayorWorkingSince = Date.now();
         sessionStatus = 'starting';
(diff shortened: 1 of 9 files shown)


jrf0110:
Hold on don't merge yet. I'm reviewing this

services/gastown/src/util/analytics.util.ts
```diff
@@ -44,6 +44,15 @@ export type GastownEventData = {
...
   label?: string;
+  // Timing fields for container cold-start instrumentation
+  containerFetchMs?: number;
+  healthPingMs?: number;
+  timeSinceContainerStartMs?: number;
+  uptimeMs?: number;
+  statusCode?: number;
+  wasSuccess?: boolean;
+  containerStartedAt?: string;
+  healthPingStatus?: string;
```
jrf0110: Many of these fields do not make sense to me. Why have separate timing fields when we already have `durationMs` that could be combined with a new event type?

services/gastown/src/util/analytics.util.ts
```diff
@@ -44,6 +44,15 @@ export type GastownEventData = {
   durationMs?: number;
   value?: number;
   label?: string;
+  // Timing fields for container cold-start instrumentation
+  containerFetchMs?: number;
+  healthPingMs?: number;
+  timeSinceContainerStartMs?: number;
+  uptimeMs?: number;
+  statusCode?: number;
+  wasSuccess?: boolean;
```
jrf0110: wasSuccess should be inferred by the presence of error

services/gastown/src/util/analytics.util.ts
```diff
@@ -44,6 +44,15 @@ export type GastownEventData = {
   durationMs?: number;
   value?: number;
   label?: string;
+  // Timing fields for container cold-start instrumentation
+  containerFetchMs?: number;
+  healthPingMs?: number;
+  timeSinceContainerStartMs?: number;
+  uptimeMs?: number;
+  statusCode?: number;
```
jrf0110, line 50: I'm very surprised we didn't have status code previously. We should be populating statusCode for all applicable events (especially the http events)

services/gastown/src/util/analytics.util.ts
jrf0110: We should probably just use the value field instead of healthPingStatus

jrf0110:
Most of my complaints are about not re-using the perfectly fine and general fields that exist on analytics event objects already. If you can solve those issues, then you can merge

services/gastown/src/util/analytics.util.ts
jrf0110: Done — replaced all separate timing fields (containerFetchMs, healthPingMs, timeSinceContainerStartMs, uptimeMs) with the existing durationMs field. The event name already disambiguates which timing is being measured, so separate fields were redundant. Also removed statusCode/wasSuccess duplication from double slots.

services/gastown/src/util/analytics.util.ts
jrf0110: Done — removed wasSuccess. Failure is now inferred from the presence of the error field (absent = success), consistent with how other events in GastownEventData work.

services/gastown/src/util/analytics.util.ts
jrf0110: Acknowledged — statusCode is now populated in all container event emissions (container.agent_start_fetch and container.health_ping). It's stored as double11 in Analytics Engine.

services/gastown/src/dos/town/container-dispatch.ts
```diff
@@ -519,6 +521,17 @@ export async function startAgentInContainer(
...
 
+    const containerFetchMs = Date.now() - fetchStart;
+    writeEvent(env, {
+      event: 'container.agent_start_fetch',
+      townId: params.townId,
+      rigId: params.rigId,
+      agentId: params.agentId,
+      containerFetchMs,
+      statusCode: response.status,
+      wasSuccess: response.ok,
```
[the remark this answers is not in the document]
jrf0110: Fixed — the event emission is now split per code path. The 'already running' case emits the event without an error field (success path). The error case emits with error set. There's no longer a single write-before-special-case pattern.

services/gastown/src/dos/Town.do.ts
```diff
@@ -4370,12 +4382,48 @@ export class TownDO extends DurableObject<Env> {
...
+        const healthResp = await container.fetch('http://container/health', {
+          signal: AbortSignal.timeout(5_000),
+          headers,
+        });
+        const healthPingMs = Date.now() - t0;
+        writeEvent(this.env, {
+          event: 'container.health_ping',
+          townId,
+          healthPingMs,
+          healthPingStatus: 'ok',
```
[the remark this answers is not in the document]
jrf0110: Fixed — health ping now explicitly checks healthResp.ok. Non-2xx responses emit container.health_ping with error set to 'non-ok status ${healthResp.status}'. Only 2xx responses go down the success path and emit container.ready_observed.

services/gastown/src/util/analytics.util.ts
jrf0110: Done — removed healthPingStatus entirely. Failure is now expressed via the error field (absence = success), consistent with the pattern used for all other events.

PR #2707, opened 2026-04-22
feat(kiloclaw): add openclaw doctor controller API route

pandemicsyn:
## Summary

- Adds a controller-local doctor runner that starts `openclaw doctor [--fix] --non-interactive` inside the machine via the controller process, avoiding the Fly Machines exec API timeout while keeping the legacy Fly-exec doctor path available.
- Wires the async runner end-to-end through `KiloClawInstance` start/status/cancel RPC methods, `/api/platform/doctor-controller/{start,status,cancel}`, the internal client, admin tRPC, and the admin instance detail "Run Doctor (Controller)" dialog.
- Stores the current/last doctor run metadata and log on the instance volume under `/root/.openclaw/.kiloclaw/doctor-runs/`, so admins can close/reopen the dialog without cancelling or losing output.
- Refreshes the branch onto current `main`, preserving newer Kilo Chat/Morning Briefing controller routes, and changes controller build CalVer to UTC date+time granularity with optional fourth-segment comparison support.

## Verification

N/A (no manual verification performed after the rebase).

## Visual Changes

Admin instance detail now shows a second doctor action, "Run Doctor (Controller)", next to the existing Fly-exec "Run Doctor" button. The controller path opens a dialog with a `--fix` toggle, status/exit/duration badges, persisted output display, explicit cancel while running, and a re-run action. Screenshots not captured.

## Reviewer Notes

- The existing `/api/platform/doctor` endpoint and original "Run Doctor" button are intentionally unchanged so both paths can run side-by-side during migration.
- Controller doctor is async: `start` returns immediately, the admin UI polls `status`, and closing the dialog does not cancel the run. Cancellation is explicit via `/cancel`.
- The controller keeps only current/last run state, not multi-run history. Output is capped around 1 MiB and stored tail-preserving with a truncation marker.
- UI gating still uses controller CalVer for this PR, but controller builds now emit `YYYY.M.D.HHmm` so same-day controller changes can be distinguished. Explicit controller capability hints are intentionally deferred to a follow-up.
- Do not use `controllerCapabilitiesVersion` for this feature: that DO field tracks the worker→controller config contract for named routing/origin allowlist compatibility, not controller binary route availability.


RSO (approved):
Tried to figure out why you don't have a code review on this PR [here](https://kilo-code.slack.com/archives/C09S0PFRFMW/p1776927121350719)

services/kiloclaw/controller/src/routes/doctor.ts
```diff
@@ -0,0 +1,512 @@
...
+  });
+  if (run.timeoutTimer) {
+    clearTimeout(run.timeoutTimer);
+    run.timeoutTimer = null;
+  }
+  if (run.metadataFlushTimer) {
+    clearTimeout(run.metadataFlushTimer);
+    run.metadataFlushTimer = null;
+  }
+  activeRun = null;
```
jeanduplessis: **WARNING:** Cancelled or timed-out runs can overlap with a new doctor process

`finalizeRun` clears `activeRun` immediately for `timed_out`/`cancelled`, but those paths only sent `SIGTERM` and scheduled SIGKILL; the child can still be alive until the grace timer or `close`. Since the next `/start` gates only on `activeRun`, an immediate retry after timeout/cancel can spawn another `openclaw doctor` while the previous one is still running.

Trace:
1. `timeout`/`cancel` call `scheduleSigkill(run)` and then `finalizeRun(...)`.
2. `finalizeRun` sets `activeRun = null` before child `close`.
3. A subsequent start can pass the `if (activeRun)` gate, truncate `current.log`, and spawn a new doctor while the old process may still mutate config or append output.

Impact: Two `openclaw doctor --fix` processes can overlap during SIGTERM grace, risking concurrent state/config mutation and mixed persisted output. Keep the run exclusive until the child terminates, or gate starts on a still-terminating run.

services/kiloclaw/controller/src/routes/doctor.ts
```diff
@@ -0,0 +1,512 @@
...
+    }
+    await next();
+  });
+
+  app.post('/_kilo/doctor/start', async c => {
+    let body: unknown;
+    try {
+      body = await c.req.json();
+    } catch {
+      body = {};
```
jeanduplessis, line 380: **WARNING:** Malformed JSON defaults to a mutating doctor run

Any `c.req.json()` failure is treated as `{}`, then `fix` defaults to `true`, so malformed or truncated authenticated JSON starts `openclaw doctor --fix` instead of returning `400`.

Trace:
1. The catch path assigns `body = {}` for all JSON parse failures.
2. `{}` passes `DoctorRunBodySchema`.
3. `parsed.data.fix ?? true` selects the mutating `--fix` path.

Impact: A bad internal request body can accidentally trigger a repair run. Please distinguish an actually empty body from invalid JSON, or require valid JSON whenever a body is present.

apps/web/src/app/admin/components/KiloclawInstances/KiloclawInstanceDetail.tsx
```diff
@@ -1482,6 +1487,13 @@ export function KiloclawInstanceDetail({ instanceId }: { instanceId: string }) {
     cleanVersion(controllerVersion?.version),
     '2026.2.26'
   );
+  // /_kilo/doctor/start|status|cancel was added to the controller on 2026-05-08. Older
+  // controllers fall through to the catch-all proxy and return 404 —
+  // disable the button with a tooltip until they redeploy.
+  const supportsDoctorController = calverAtLeast(
+    cleanVersion(controllerVersion?.version),
+    '2026.5.8'
```
jeanduplessis: **WARNING:** Same-day stale controllers are incorrectly treated as supporting async doctor routes

The feature gate uses `calverAtLeast(..., '2026.5.8')`, but this PR also changes controller builds to `YYYY.M.D.HHmm` specifically to distinguish same-day controller changes. Because the parser treats a missing fourth segment as `0`, any `2026.5.8` controller satisfies this gate even if it was built before `/_kilo/doctor/start|status|cancel` existed.

Trace:
1. The UI gate is date-only: `2026.5.8`.
2. `parseCalver` now treats missing time as `0`.
3. Same-day controller builds before this route can still report `2026.5.8` and pass the gate.

Impact: Admins can see and click “Run Doctor (Controller)” on stale same-day images, then hit route-unavailable failures instead of the button staying disabled until redeploy. Use the exact build-time minimum that introduced the async routes, or gate on an explicit capability when available.

jeanduplessis (approved):
Approved. I found 3 warnings worth interrogating if you agree with them before/after merge: (1) controller doctor exclusivity is released before cancelled/timed-out children exit, (2) invalid JSON can default into `--fix`, and (3) the UI feature gate may be too broad for same-day controller builds. I left line notes with trace and impact.

PR #2715, opened 2026-04-22
feat(claw): skip permissions step in onboarding

Drixled:
## Summary
- Remove the live KiloClaw permissions screen from onboarding so users move directly from bot identity to channel selection.
- Default onboarding exec permissions to the YOLO preset (`never-ask`), while preserving the existing provisioning path that applies and persists the preset.
- Keep the permissions screen available in the fake walkthrough/Storybook path with its original progress count.

## Verification
- Smoke-tested a fresh fake user through `/claw/new`: identity opened as step 1 of 3, continuing went directly to channel selection as step 2 of 3, and the permissions screen was absent.
- Smoke-tested `/claw/new?fakeOnboardingStep=permissions`: fake walkthrough still shows the permissions screen as step 2 of 5.

## Visual Changes
N/A

## Reviewer Notes
- The Settings permission controls and shared permission cards are intentionally unchanged.
- Channel setup remains in onboarding for this PR; moving it to Settings-only needs separate guidance improvements.


apps/web/src/app/(app)/claw/components/ClawOnboardingFlow.tsx
```diff
@@ -251,24 +250,12 @@ function ClawOnboardingFlowInner({
             });
             provisionInstance(weatherLocation?.location);
           }
-          posthog?.capture('claw_setup_permissions_viewed');
```
RSO, line 254: Note, you’re removing a postbox event here, so make sure to notify @pedroheyerdahl when this gets merged.

apps/web/src/app/(app)/claw/components/ClawOnboardingFlow.tsx
```diff
@@ -385,7 +371,6 @@ function ClawOnboardingFlowInner({
       case 'identity':
         return renderIdentityStep();
       case 'permissions':
```
RSO: This feels strange, does this mean we render the channel step when the case is permissions?

RSO (changes requested):
> Keep the permissions screen available in the fake walkthrough/Storybook path with its original progress count.

Why? The fake onboarding is to resemble the real one, so why would we keep it there?

RSO (approved):

PR #2722, opened 2026-04-22
fix(web): align Exa default UI with Brave configuration

joshavant:
## Summary

Aligns the settings UI’s Exa/Brave state inference with controller behavior for unset Exa mode.

When `kiloExaSearchMode` is `null`, the UI now treats Exa as the default only if Brave is not configured. This prevents the settings page from presenting Exa as active for users with valid Brave configuration and unset Exa mode.

## Verification

- [ ] No manual verification performed.

## Visual Changes

N/A

## Reviewer Notes

This is a UI-only follow-up on top of the controller branch.
It adjusts derived state used by badges/toggles in Settings and does not change backend behavior.

### DIFF
--- a/apps/web/src/app/(app)/claw/components/SettingsTab.tsx
+++ b/apps/web/src/app/(app)/claw/components/SettingsTab.tsx
@@ -1084,10 +1084,6 @@ export function SettingsTab({
   const exaSearchDisplayMode =
     kiloExaSearchMode ?? (activeWebSearchProvider === 'kilo-exa' ? 'kilo-proxy' : 'disabled');
   const braveSearchEnabled = braveSearchConfigured && activeWebSearchProvider === 'brave';
-  const hasCustomActiveSearchProvider =
-    activeWebSearchProvider !== null &&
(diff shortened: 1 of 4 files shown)


joshavant:
Addressed the provider-source concern with a backend runtime signal instead of secret-based inference.

Changes in this PR now:
- Add `activeWebSearchProvider` to `GET /api/kiloclaw/config` (derived from live OpenClaw config via `/_kilo/config/read`; `null` if unavailable).
- Update Settings search state to use `activeWebSearchProvider` for Exa/Brave active status.
- Add a custom-provider notice when active provider is non-`brave`/`kilo-exa`.
- Keep Exa mode controls driven by `kiloExaSearchMode` for dashboard-managed intent.
- Add route helper tests for provider extraction (`deriveActiveWebSearchProvider`).

This avoids misclassifying active provider when users edit OpenClaw directly or when explicit provider is preserved across redeploys.

St0rmz1:
Closing. Going to solve this differently.

PR #2767, opened 2026-04-23
feat(observability): alert on failed credit-based instance enrollment

St0rmz1:
## Summary
Instruments the enrollWithCredits TRPC mutation so every attempt emits a structured JSON log line under billingFlow=credit_enrollment, with one "attempted" entry followed by exactly one "succeeded" or "failed" entry (tagged with a failureReason and durationMs). An enclosing try/catch around the whole mutation body guarantees the funnel shape holds even if upstream helpers (anchor resolution, prior-subscription lookup) throw. Captures all five failure modes including no_instance.

The emit helper makes level, billingFlow, and billingComponent authoritative so callers can't shadow them via a colliding key. The error field is truncated to 500 chars to limit PII blast radius on internal_error paths, and logCreditEnrollmentFailed destructures error out before spreading so the truncation is explicit.

Documents two Axiom monitors in services/kiloclaw-billing/docs/observability.md: credit-enrollment-internal-error (page, >=1 in 5m) and credit-enrollment-failure-spike (ticket, >=10 in 15m).

## Verification

Exercised the happy path against local dev:

1. Provisioned a KiloClaw instance as a user with $55 in credits.
2. Navigated to /claw/subscription, selected the Standard hosting plan, and clicked Pay $4.00 with Credits.
3. Confirmed both JSON log lines in the web dev stdout, with every field the Axiom monitors filter on present and correctly shaped (billingFlow, billingComponent, event, outcome, durationMs, userId, instanceId, plan):

```
{"userId":"b106fb5d...","plan":"standard","instanceId":"c676d001...","event":"credit_enrollment.attempted","outcome":"started","level":"info","billingFlow":"credit_enrollment","billingComponent":"web_trpc"}
{"userId":"b106fb5d...","plan":"standard","instanceId":"c676d001...","durationMs":138,"event":"credit_enrollment.succeeded","outcome":"completed","level":"info","billingFlow":"credit_enrollment","billingComponent":"web_trpc"}
```

Failure paths were not exercised live (each requires seeding a specific DB state). Instead, the catch block dispatch was reviewed against every known throw site inside enrollWithCreditsImpl and the router guard, and each branch is typed via the CreditEnrollmentFailureReason union so TypeScript catches drift.

pnpm run format:check, pnpm run lint (web), and pnpm run typecheck (web) all pass.

## Visual Changes

N/A. Observability only; no UI or customer facing changes.

| Before | After |
| ------ | ----- |
| N/A    | N/A   |

## Reviewer Notes

Focus areas:

* JSON log shape mirrors the existing billing side effects route, so Axiom queries on billingFlow work uniformly across flows.
* The enclosing try/catch enforces the funnel invariant: every credit_enrollment.attempted pairs with exactly one succeeded or failed entry, even if upstream helpers such as resolvePersonalBillingAnchor or hadPriorPaidSubscription throw. The no_instance branch is logged before throwing TRPCError, and the outer catch re throws TRPCError without logging to avoid double counting.
* Failure reason classification uses message substring matching because enrollWithCreditsImpl throws plain Errors today. This is brittle: if any thrown message is reworded both the HTTP code and the alert tag drift together. Worth a follow up story to switch to typed error classes and dispatch via instanceof.
* The error field is truncated to 500 chars via destructure then slice; known reason paths only emit fixed template strings, so the cap really only matters for the internal_error branch where raw ORM or driver text can surface.
* Two Axiom monitors still need to be created manually after merge using the filters documented in services/kiloclaw-billing/docs/observability.md. Thresholds (>=1 in 5m for internal_error, >=10 in 15m for any failure) are first guesses and should be retuned once a week of baseline data lands.

### DIFF
--- a/apps/web/src/routers/kiloclaw-router.ts
+++ b/apps/web/src/routers/kiloclaw-router.ts
@@ -3839,13 +3839,6 @@ export const kiloclawRouter = createTRPCRouter({
           instanceId: input.instanceId,
         });
         if (!anchorInstance) {
-          logCreditEnrollmentFailed({
-            userId: ctx.user.id,
-            plan: input.plan,
-            instanceId: input.instanceId,
-            failureReason: 'no_instance',
-            durationMs: Date.now() - startedAt,
-          });
           throw new TRPCError({
             code: 'BAD_REQUEST',
             message: 'Provision KiloClaw first before enrolling hosting with credits.',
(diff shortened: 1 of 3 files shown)


apps/web/src/routers/kiloclaw-router.ts
```diff
@@ -3841,20 +3859,62 @@ export const kiloclawRouter = createTRPCRouter({
...
-            : message.includes('Insufficient credit balance')
-              ? 'BAD_REQUEST'
-              : 'INTERNAL_SERVER_ERROR';
+        const {
+          code,
+          failureReason,
+        }: {
+          code: 'NOT_FOUND' | 'CONFLICT' | 'BAD_REQUEST' | 'INTERNAL_SERVER_ERROR';
+          failureReason: CreditEnrollmentFailureReason;
+        } = message.includes('not found')
```
catrielmuller: Create a simple function that fast return 
```
{
          code: 'NOT_FOUND' | 'CONFLICT' | 'BAD_REQUEST' | 'INTERNAL_SERVER_ERROR';
          failureReason: CreditEnrollmentFailureReason;
        }
```
Will be simplify this block of code

catrielmuller (approved):

PR #2768, opened 2026-04-23
feat(models): replace gpt-5.4 with gpt-5.5 in preferred models

lambertjosh:
Swap the recommended OpenAI entry to gpt-5.5 now that it is public. Keep both gpt-5.4 and gpt-5.5 in HEALTH_CHECK_EXCLUSIONS while traffic on the new model is too spiky to alert on reliably.


arimesser (approved):
Looks good to me and the models are now live and visible in Kilo -- @lambertjosh  do we want to test first? Or good to just merge?

chrarnoldus (approved):

PR #2769, opened 2026-04-23
feat(kilo-ops): Grafana operations dashboard on Cloudflare Containers

jrf0110:
## Summary

image

Adds a new `services/kilo-ops/` service that hosts Grafana OSS inside a Cloudflare Container and proxies authenticated traffic to it from a Hono Worker.

- **Auth**: Worker terminates Cloudflare Access via `jose`-based JWT verification and forwards the user identity to Grafana through `X-WEBAUTH-USER` (Grafana's auth-proxy mode). Dev environments bypass CF Access with a stub identity.
- **Analytics Engine proxying**: Cloudflare Containers cannot reach `api.cloudflare.com` directly (loopback protection), which breaks Grafana's ClickHouse datasource against the AE SQL endpoint. The Worker installs an `outboundByHost` handler on `GrafanaContainer` that intercepts requests to a fake `cf-ae` internal host and forwards them to the real AE SQL endpoint from the Worker runtime, injecting the bearer token server-side. The container never sees the AE token.
- **Dashboards**: Gastown Operations dashboard is provisioned from the container image via `/etc/grafana/provisioning/dashboards/` so it survives container cold starts.
- **Routing**: Request traffic is sharded across country-scoped Durable Object instances (`grafana-${country}`) based on `CF-IPCountry`.
- **Secrets**: `CF_ANALYTICS_API_KEY` and `GF_SECRET_KEY` come from Secrets Store bindings and are resolved in the Worker. The AE token lives only in the Worker runtime; the Grafana secret_key is passed into the container at start.

Deployed at `ops.kiloapps.io`.

## Verification

- Deployed to production and confirmed Gastown Operations dashboard loads with live Analytics Engine queries (GET `/analytics_engine/sql?query=SELECT…`) returning 200.
- Confirmed CF Access redirect flow + `X-WEBAUTH-USER` auto-signup inside Grafana.
- Confirmed container cold start after sleep (`sleepAfter = '1h'`) restores provisioned dashboards and datasource correctly.

## Visual Changes

N/A (infrastructure service; Grafana UI is unchanged from upstream OSS).

## Reviewer Notes

- **Handler scope tightness**: `GrafanaContainer.outboundByHost['cf-ae']` only accepts GET/POST and always pins the upstream URL to `env.CF_CLICKHOUSE_URL` (preserving only the query string). The container cannot influence which endpoint the worker hits.
- **`@cloudflare/containers` version**: pinned to `^0.3.2` because `outboundByHost`/`ContainerProxy` require the 0.3.x API. Other services in this repo are on mixed older versions — not consolidated here.
- **Ephemeral Grafana state**: only provisioned dashboards/datasources persist. User preferences (theme, stars, etc.) reset on cold start. An intercept-based persistence layer (`/api/user/preferences` → DO) was discussed but deferred to a follow-up.
- **Replaces PR #2166**: that branch was based on a stale gastown-staging branching point and diverged non-trivially from `main`. This is a clean single-commit branch off current `main`. #2166 should be closed.

### DIFF
--- a/services/kilo-ops/container/Dockerfile
+++ b/services/kilo-ops/container/Dockerfile
@@ -1,0 +1,12 @@
+FROM grafana/grafana-oss:latest
+
+# Install ClickHouse datasource plugin for Analytics Engine
+RUN grafana cli plugins install vertamedia-clickhouse-datasource
+
+# Copy Grafana config
+COPY grafana.ini /etc/grafana/grafana.ini
+
+# Copy provisioning configs and dashboards
+COPY provisioning /etc/grafana/provisioning
+
+EXPOSE 3000
--- a/services/kilo-ops/container/Dockerfile
+++ b/services/kilo-ops/container/Dockerfile
@@ -1,7 +1,9 @@
-FROM grafana/grafana-oss:latest
+FROM grafana/grafana-oss:13.0.1
 
-# Install ClickHouse datasource plugin for Analytics Engine
-RUN grafana cli plugins install vertamedia-clickhouse-datasource
+# Install ClickHouse datasource plugin for Analytics Engine.
+# Pinned so rebuilds at the same git SHA produce the same artifact and a
+# plugin takeover or bad release can't silently land on the next deploy.
+RUN grafana cli plugins install vertamedia-clickhouse-datasource 3.4.11
 
 # Copy Grafana config
 COPY grafana.ini /etc/grafana/grafana.ini
--- a/services/kilo-ops/container/grafana.ini
+++ b/services/kilo-ops/container/grafana.ini
@@ -19,7 +19,18 @@ auto_assign_org_role = Admin
 disable_login_form = true
 disable_signout_menu = true
 
+[auth.basic]
+# Kill the default admin:admin backdoor. The Worker strips the Authorization
+# header before forwarding, but this is defense-in-depth: a future change to
+# STRIPPED_REQUEST_HEADERS, a new ingress path, or a second binding that
+# talks to GRAFANA_CONTAINER directly would otherwise reopen admin:admin.
+enabled = false
+
 [security]
 admin_user = admin
 secret_key = ${GF_SECURITY_SECRET_KEY}
-allow_embedding = true
+# Leave embedding off unless we add a specific internal embedder and scope
+# it with Content-Security-Policy: frame-ancestors in the Worker. With the
+# SameSite=None CF_Authorization cookie, allow_embedding=true would let an
+# attacker page iframe ops.kiloapps.io and clickjack a logged-in admin.
+allow_embedding = false
--- a/services/kilo-ops/container/grafana.ini
+++ b/services/kilo-ops/container/grafana.ini
@@ -1,0 +1,25 @@
+[server]
+root_url = https://ops.kiloapps.io/
+serve_from_sub_path = false
+
+[auth.proxy]
+enabled = true
+header_name = X-WEBAUTH-USER
+header_property = username
+auto_sign_up = true
+sync_ttl = 60
+whitelist =
+
+[users]
+allow_sign_up = false
+auto_assign_org = true
+auto_assign_org_role = Admin
+
+[auth]
+disable_login_form = true
+disable_signout_menu = true
+
+[security]
+admin_user = admin
+secret_key = ${GF_SECURITY_SECRET_KEY}
+allow_embedding = true
--- a/services/kilo-ops/src/cf-access.middleware.ts
+++ b/services/kilo-ops/src/cf-access.middleware.ts
@@ -49,8 +49,16 @@ export function withCloudflareAccess({
         audience,
       });
 
-      // User tokens have 'email', service tokens have 'common_name'
-      const identity = (payload.email as string) ?? (payload.common_name as string) ?? 'unknown';
+      // User tokens carry 'email', service tokens carry 'common_name'. Reject
+      // any token that lacks both — falling back to a shared identity would
+      // grant that token Grafana's auto-signup Admin role.
+      const email = typeof payload.email === 'string' ? payload.email : null;
+      const commonName = typeof payload.common_name === 'string' ? payload.common_name : null;
+      const identity = email?.trim() || commonName?.trim();
+      if (!identity) {
+        console.warn('CF Access JWT validated but contains no email or common_name');
+        return c.json({ error: 'Unauthorized' }, 401);
+      }
       c.set('userIdentity', identity);
     } catch (e) {
       console.warn(
--- a/services/kilo-ops/src/cf-access.middleware.ts
+++ b/services/kilo-ops/src/cf-access.middleware.ts
@@ -1,0 +1,64 @@
+import type { MiddlewareHandler } from 'hono';
+import { createRemoteJWKSet, jwtVerify } from 'jose';
+import type { KiloOpsEnv } from './worker';
+
+/**
+ * Cloudflare Access JWT validation middleware using the jose library.
+ * https://developers.cloudflare.com/cloudflare-one/access-controls/applications/http-apps/authorization-cookie/validating-json/
+ *
+ * jose handles signature verification, key rotation (via JWKS endpoint),
+ * expiry, nbf, issuer, and audience checks.
+ */
+
+const jwksCache = new Map<string, ReturnType<typeof createRemoteJWKSet>>();
+
+function getJWKS(teamDomain: string) {
+  let jwks = jwksCache.get(teamDomain);
+  if (!jwks) {
+    jwks = createRemoteJWKSet(new URL(`${teamDomain}/cdn-cgi/access/certs`));
+    jwksCache.set(teamDomain, jwks);
+  }
+  return jwks;
+}
+
+export function withCloudflareAccess({
+  team,
+  audience,
+}: {
+  team: string;
+  audience: string;
+}): MiddlewareHandler<KiloOpsEnv> {
+  if (!/^[a-z0-9-]+$/.test(team)) {
+    throw new Error(`Invalid CF Access team name: ${team}`);
+  }
+  if (!/^[a-f0-9]{64}$/.test(audience)) {
+    throw new Error(`Invalid CF Access audience tag: ${audience}`);
+  }
+
+  const teamDomain = `https://${team}.cloudflareaccess.com`;
+
+  return async (c, next) => {
+    const token = c.req.header('Cf-Access-Jwt-Assertion');
+    if (!token) {
+      return c.json({ error: 'Unauthorized' }, 401);
+    }
+
+    try {
+      const { payload } = await jwtVerify(token, getJWKS(teamDomain), {
+        issuer: teamDomain,
+        audience,
+      });
+
+      // User tokens have 'email', service tokens have 'common_name'
+      const identity = (payload.email as string) ?? (payload.common_name as string) ?? 'unknown';
+      c.set('userIdentity', identity);
+    } catch (e) {
+      console.warn(
+        `CF Access JWT validation failed: ${e instanceof Error ? e.message : 'unknown'}`
+      );
+      return c.json({ error: 'Unauthorized' }, 401);
+    }
+
+    await next();
+  };
+}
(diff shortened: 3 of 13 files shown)


services/kilo-ops/src/cf-access.middleware.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    try {
+      const { payload } = await jwtVerify(token, getJWKS(teamDomain), {
+        issuer: teamDomain,
+        audience,
+      });
+
+      // User tokens have 'email', service tokens have 'common_name'
+      const identity = (payload.email as string) ?? (payload.common_name as string) ?? 'unknown';
```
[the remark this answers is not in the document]
jrf0110: Fixed in 890ecb542. Now explicitly rejects tokens that lack both `email` and `common_name`: both claims are pulled out with a `typeof === "string"` guard, trimmed, and if neither is non-empty the request is rejected with 401 (matching the other failure modes in this middleware).

services/kilo-ops/src/worker.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+  headers.set('X-WEBAUTH-USER', userIdentity);
+
+  const init: RequestInit = {
+    method: c.req.method,
+    headers,
+  };
+
+  if (c.req.method !== 'GET' && c.req.method !== 'HEAD') {
+    const body = await c.req.raw.arrayBuffer();
```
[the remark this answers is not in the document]
jrf0110: Fixed in 890ecb542. Swapped the `arrayBuffer()` buffering for `new Request(containerUrl, c.req.raw)`, which passes the body through as a stream. Header overrides (strip auth/cookie, set `X-WEBAUTH-USER`) are applied on the new Request directly. Dashboard imports and plugin uploads now flow through without sitting in Worker memory.

services/kilo-ops/container/grafana.ini
```diff
@@ -0,0 +1,25 @@
...
+auto_assign_org_role = Admin
+
+[auth]
+disable_login_form = true
+disable_signout_menu = true
+
+[security]
+admin_user = admin
+secret_key = ${GF_SECURITY_SECRET_KEY}
+allow_embedding = true
```
St0rmz1: Heads up on this line combined with the CF Access cookie defaults. Worth tightening before merge.

`allow_embedding = true` here strips `X-Frame-Options` from Grafana responses, and no `Content-Security-Policy: frame-ancestors` is being set in the Worker either. Cloudflare Access's `CF_Authorization` cookie defaults to `SameSite=None` (per the CF Access authorization cookie docs), so the browser WILL send it on cross-origin iframe requests.

The concrete attack path: an attacker hosts evil.com, a user already signed in to `ops.kiloapps.io` visits it, evil.com iframes `https://ops.kiloapps.io/`, CF Access validates the cookie, and Grafana renders inside the iframe under the victim's `X-WEBAUTH-USER`. A CSS overlay on top of the iframe can trick the user into clicking Grafana admin actions (delete dashboard, install plugin, create API key). Since everyone who passes CF Access lands as Grafana Admin (`auto_assign_org_role = Admin`), blast radius is whatever a Grafana admin can do.

Suggested fix: unless there is a specific embed use case, flip this off:
```ini
allow_embedding = false
```
If embedding IS needed for a specific internal page, keep this on but scope it in the Worker response instead:
```ts
response.headers.set(
  'Content-Security-Policy',
  "frame-ancestors 'self' https://internal.kiloapps.io"
);
```

Suggested test: once the Worker sets `frame-ancestors` (or Grafana sets `X-Frame-Options: DENY` via `allow_embedding = false`), a quick vitest asserting the header is present on a proxied response locks the invariant in so a future refactor of the response wrapper does not silently drop it.

services/kilo-ops/container/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM grafana/grafana-oss:latest
```
St0rmz1: Small reproducibility and supply chain concern on this Dockerfile.

Both the base image and the plugin are unpinned:
```dockerfile
FROM grafana/grafana-oss:latest
RUN grafana cli plugins install vertamedia-clickhouse-datasource
```

What this means in practice: two successive deploys of the same git SHA can pull different Grafana versions and different plugin versions, with no record in git of the change. If a Grafana CVE lands, or the plugin is taken over by a bad actor (community plugin, not first-party), the next rebuild silently bakes it in. Pinning gives you a known artifact and forces a deliberate PR when you want to move forward.

Suggested fix:
```dockerfile
FROM grafana/grafana-oss:11.3.0@sha256:<digest>
RUN grafana cli plugins install vertamedia-clickhouse-datasource 3.3.0
```

Separately worth considering: Grafana Labs maintains a first-party `grafana-clickhouse-datasource` plugin. Using the first-party one narrows the trust surface and generally gets faster updates against recent ClickHouse protocol changes. Not a blocker, just something to weigh.

services/kilo-ops/src/worker.ts
```diff
@@ -0,0 +1,150 @@
...
+    return null;
+  }
+}
+
+const app = new Hono<KiloOpsEnv>();
+
+app.get('/healthz', c => c.json({ ok: true }));
+
+app.use('/*', async (c, next) => {
+  if (c.env.ENVIRONMENT === 'development') {
```
St0rmz1: One defense-in-depth thought on this dev bypass.

```ts
if (c.env.ENVIRONMENT === 'development') {
  c.set('userIdentity', 'dev@kilo.dev');
  return next();
}
```

The only thing standing between the public internet and an unauthenticated `dev@kilo.dev` admin session is the `ENVIRONMENT` var in `wrangler.jsonc`. If a future `wrangler deploy --env dev` lands by mistake on the prod route, or a config merge sets `ENVIRONMENT: "development"` on the production environment block, the CF Access check silently goes away and anyone reaching `ops.kiloapps.io` becomes a Grafana admin. There is no second signal to catch that.

A cheap way to add a second signal: gate the bypass on something that can ONLY be true in local dev. Two reasonable options:

1. Require a specific host in the request (only local `wrangler dev` uses a localhost/`127.0.0.1` host):
```ts
const host = new URL(c.req.url).hostname;
const isLocalDev =
  c.env.ENVIRONMENT === 'development' &&
  (host === 'localhost' || host === '127.0.0.1');
if (isLocalDev) { ... }
```
2. Or invert the check and assert CF Access must be active whenever the request arrives at the prod hostname, crashing early if not:
```ts
if (new URL(c.req.url).hostname === 'ops.kiloapps.io' && c.env.ENVIRONMENT !== 'production') {
  return c.json({ error: 'misconfiguration' }, 500);
}
```

Suggested test: a vitest that calls the Worker with `ENVIRONMENT = 'development'` and a non-local hostname, asserting the request gets 401 and not the dev identity.

services/kilo-ops/container/grafana.ini
```diff
@@ -0,0 +1,25 @@
...
+[users]
+allow_sign_up = false
+auto_assign_org = true
+auto_assign_org_role = Admin
+
+[auth]
+disable_login_form = true
+disable_signout_menu = true
+
+[security]
```
St0rmz1, line 29: Defense-in-depth suggestion on this `[security]` block.

Two Grafana defaults combine in a way worth being explicit about:

1. `admin_password` defaults to `admin` (no `GF_SECURITY_ADMIN_PASSWORD` is set anywhere).
2. `[auth.basic] enabled` is `true` by default, and `disable_login_form = true` only hides the UI login form. Basic auth still answers on `/api/*` endpoints.

Today this is not exploitable because the Worker strips the `Authorization` header at `services/kilo-ops/src/worker.ts` before forwarding to the container, and the container has no public port. So `admin:admin` cannot reach Grafana from outside. Good.

The reason to close this anyway: the safety depends entirely on that strip list staying correct and on there being exactly one ingress forever. A future refactor of `STRIPPED_REQUEST_HEADERS`, a new debug route, or a second worker binding that talks to `GRAFANA_CONTAINER` directly would each silently open up `admin:admin`. One line of config removes that footgun.

Suggested fix (either is fine, first is simpler):
```ini
[auth.basic]
enabled = false
```
Or set a strong admin password via a Secrets Store binding and inject it through `this.envVars` alongside `GF_SECURITY_SECRET_KEY`:
```ts
GF_SECURITY_ADMIN_PASSWORD: await resolveSecret(this.env.GF_ADMIN_PASSWORD),
```

services/kilo-ops/src/worker.ts
```diff
@@ -0,0 +1,150 @@
...
+  const userIdentity = c.get('userIdentity');
+
+  const url = new URL(c.req.url);
+  const country = (c.req.header('cf-ipcountry') ?? DEFAULT_COUNTRY).toUpperCase();
+  const container = getGrafanaContainerStub(c.env, country);
+
+  // Forward the raw request so the body streams to the container instead of
+  // being buffered into Worker memory. Dashboard imports and plugin uploads
+  // can be multi-MB — buffering would risk the 128 MB isolate limit.
+  const containerUrl = `http://container${url.pathname}${url.search}`;
```
St0rmz1: Minor robustness suggestion on this line.

```ts
const containerUrl = `http://container${url.pathname}${url.search}`;
```

Works correctly today, but string-concatenating URL parts loses anything that is not `pathname` or `search` (fragment, future additions), and it makes the scheme swap implicit. Using `URL` to do the swap is slightly safer and reads more clearly:

```ts
const target = new URL(c.req.url);
target.protocol = 'http:';
target.host = 'container';
const forwarded = new Request(target, c.req.raw);
```

The `const url = new URL(c.req.url)` above becomes unnecessary in that case since `target` covers both the hostname lookup for country routing (you can still read `target` before mutating, or keep a separate `new URL` for country) and the forwarded request URL.

This is a nit, happy if you leave it. Flagging mainly because the pattern of building URLs by string concatenation tends to grow a bug tail over time.

services/kilo-ops/container/grafana.ini
jrf0110: Fixed in e294a041e. Went with the simpler option: set `[auth.basic] enabled = false`. This kills the default `admin:admin` backdoor entirely, independent of whether the Worker strip list stays correct. Auth proxy is the only path in now.

services/kilo-ops/container/grafana.ini
jrf0110: Fixed in e294a041e. Set `allow_embedding = false` (nothing currently needs to embed Grafana, and the CF_Authorization cookie being SameSite=None made the clickjacking path real). If we later need embedding from a specific internal origin, we can add `Content-Security-Policy: frame-ancestors ...` in the Worker response instead of opening this back up globally. Skipping the vitest for now — no worker-level unit tests exist yet for this service and standing up the test harness is out of scope for this PR.

services/kilo-ops/src/worker.ts
jrf0110: Fixed in e294a041e. Went with option 1 — the dev bypass now requires both `ENVIRONMENT === "development"` AND a localhost hostname (`localhost`, `127.0.0.1`, or `[::1]`). A misconfigured `wrangler deploy --env dev` onto the `ops.kiloapps.io` route would now fall through to CF Access validation because the hostname check fails, which is exactly the safety net you wanted. Skipping the vitest for the reason mentioned in the embedding thread.

services/kilo-ops/src/worker.ts
jrf0110: Done in e294a041e. Switched to the URL-API pattern you suggested — clone the URL, swap `protocol` and `host`, pass to `new Request(target, c.req.raw)`. Dropped the now-redundant `url` variable too.

services/kilo-ops/container/Dockerfile
jrf0110: Fixed in e294a041e. Pinned to `grafana/grafana-oss:13.0.1` and `vertamedia-clickhouse-datasource 3.4.11` (current latest as of today). Going with version tags rather than SHA digests to match the convention in the rest of the repo (e.g. `oven/bun:1-slim`, `docker.io/cloudflare/sandbox:0.8.9`). Re: the first-party `grafana-clickhouse-datasource` plugin — good callout. Leaving that for a follow-up; swapping datasources mid-dashboard is a bigger change because the query syntax between the two plugins differs and every panel in `gastown.json` would need audit.

jrf0110: Another note on the changing plugins, Cloudflare recommends using the vertamedia-clickhouse-datasource in their docs so that's what I went with. I'm assuming there's a good reason for this, but I haven't validated

St0rmz1 (approved):

PR #2814, opened 2026-04-25
feat(kiloclaw): add optional Pipelock proxy sidecar

luckyPipewrench:
## Summary

Adds an opt-in Pipelock integration for KiloClaw VMs. Set `KILOCLAW_PIPELOCK_ENABLED=1` on an instance and the controller generates a per-VM TLS interception CA, writes a managed Pipelock config, and starts Pipelock under the existing process supervisor before OpenClaw. The OpenClaw child gets proxy environment variables that route env-aware HTTP/HTTPS traffic through `127.0.0.1:8888` for DLP, prompt-injection, SSRF, and response scanning.

When the flag is unset, KiloClaw startup and generated OpenClaw config are byte-for-byte unchanged.

The integration is controller-only. A Worker-side toggle that lets users enable Pipelock per-instance from the dashboard would be a small follow-up if you want per-instance control; until then, the flag is set globally on the Worker for all instances or none.

A new Phase 7 in the controller startup runs CA generation, CA bundle build, config write, supervisor spawn, and a strict readiness gate before Phase 8 starts OpenClaw. Capability separation is enforced at the env-var boundary: the Pipelock sidecar gets an explicit allowlist environment so it cannot inherit agent secrets or proxy env vars (the latter would loop the proxy through itself). Failures in CA generation, config write, supervisor spawn, or readiness leave the controller in degraded mode without ever starting OpenClaw. Readiness requires Pipelock to report `status=healthy` with `forward_proxy_enabled`, `tls_interception_enabled`, and `response_scan_enabled` all true, and `kill_switch_active` not active.

V1 enforces traffic routing at the env-var layer. VM-level egress containment (iptables `OWNER`, network namespace, seccomp) would harden this further against a malicious agent and is documented as a recommended follow-up in `services/kiloclaw/docs/pipelock.md`.

No Pipelock license key is required. The features used here (forward proxy, TLS interception, DLP, prompt-injection scanning, SSRF, SSE response scanning) are all in Pipelock's free single-agent tier.

## Verification

- [x] Built the new pipelock install layer on `debian:bookworm-slim` (Kilo's runtime base) and verified pipelock v2.3.0 installs to `/usr/local/bin` with the SHA256 pin matching the upstream `checksums.txt` for the v2.3.0 tag
- [x] Ran `pipelock check --config` against the controller-generated YAML; v2.3.0 accepts every field including `response_scanning.sse_streaming.enabled`
- [x] Exercised the `ensurePipelockCa` + `ensurePipelockCaBundle` + `ensurePipelockConfig` chain end-to-end against the real pipelock v2.3.0 binary; verified file modes (`0o600` on the CA key and config), bundle contents (system roots plus per-VM CA), and idempotency on repeat invocation
- [x] Spawned `pipelock run --config` and confirmed the controller's `waitForPipelockReady` passes against the live `/health` response; confirmed the forward proxy responds to `CONNECT example.com:443` with `HTTP/1.1 200`
- [x] Confirmed the unset-flag path: no Pipelock invocation, no proxy env injection, OpenClaw supervisor receives unchanged env

The pipelock-specific image layer was built and verified. The rest of the KiloClaw image (chromium, ffmpeg, openclaw toolchain) is unchanged by this PR.

## Visual Changes

N/A

## Reviewer Notes

1. **Pipelock env allowlist** in `controller/src/pipelock.ts` (`PIPELOCK_ENV_ALLOWLIST`). Every entry is a deliberate concession. Push back on any that feels too permissive.
2. **Phase 7 ordering** in `controller/src/index.ts` and the three degraded-state labels (`pipelock-init`, `pipelock-start`, `pipelock-listen`). Each path has a test in `start-controller-degraded.test.ts`.
3. **Readiness gate** in `controller/src/pipelock.ts` (`waitForPipelockReady` + `isPipelockHealthReady`). Three feature flags plus kill-switch-not-active. Stricter than a TCP poll so a stray listener on `:8888` cannot unblock OpenClaw.
4. **Dockerfile pipelock layer** with per-arch SHA256 verification of the upstream tarball.
5. **`docs/pipelock.md` "Enforcement boundary" section** for the V1 application-level vs follow-up VM-level distinction.


jjmata:
💯 🚀 

Pipelock configurations/Helm chart-based support has been merged into [our Sure repo](https://github.com/we-promise/sure) for weeks now, and this would allow us to extend Kilo Claw offers to our users at sure.am without additional headaches.

Currently demo site users don't have the benefit of kicking the tires of "external AI" via OpenClaw (we support it, but currently show only the direct-to-OpenAI-compatible LLM calls in our site.  I would feel more comfortable deploying Sure + Kilo Claw if we could also demonstrate out commitment to data security by having this "on" by default in our bundle.

emilieschario:
@luckyPipewrench can you email emilie (at) kilocode (dot) ai for us to chat about this?  Thanks for kicking off with a PR.

emilieschario:
Closing per our discussion yesterday

jjmata:
Sad ... let's talk to the Render people @luckyPipewrench - they seem interested!

PR #2821, opened 2026-04-26
fix(ai-gateway): exclude DeepSeek models from Vercel routing

lambertjosh:
## Summary

DeepSeek in thinking mode requires assistant messages to round-trip \`reasoning_content\` back to the API. OpenRouter translates from \`providerOptions.openrouter.reasoning_details\` automatically; Vercel AI Gateway routes directly to DeepSeek without that translation, producing 400s on multi-turn reasoning conversations.

This adds a DeepSeek exclusion to \`shouldRouteToVercel\` so the default OpenRouter path is always used. Matches the existing exclusion pattern used for other incompatible models in earlier revisions of the gateway.

## Reported error

\`\`\`
[DeepSeek] The \`reasoning_content\` in the thinking mode must be passed back to the API.
providerMetadata.gateway.routing.resolvedProvider = "deepseek"
\`\`\`

## Related

Companion to [Kilo-Org/kilocode#9526](https://github.com/Kilo-Org/kilocode/pull/9526), which cherry-picks the upstream client-side fix for the OpenRouter path. That PR covers direct OpenRouter and \`@kilocode/kilo-gateway\` → OpenRouter. This PR covers the remaining \`@kilocode/kilo-gateway\` → Vercel → DeepSeek path.

### DIFF
--- a/apps/web/src/lib/ai-gateway/providers/vercel/index.test.ts
+++ b/apps/web/src/lib/ai-gateway/providers/vercel/index.test.ts
@@ -1,5 +1,8 @@
 import { describe, it, expect } from '@jest/globals';
-import { getAnthropicProviderOptionsForVercel } from '@/lib/ai-gateway/providers/vercel';
+import {
+  getAnthropicProviderOptionsForVercel,
+  shouldRouteToVercel,
+} from '@/lib/ai-gateway/providers/vercel';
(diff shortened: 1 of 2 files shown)


lambertjosh:
Not needed, needed to bump kilo gw version.

PR #2824, opened 2026-04-27
fix(code-review): ignore stale running slots

alex-alecu:
## Summary
- Code review dispatch now stops old stuck running reviews from using owner slots forever.
- The existing retry path for stuck queued reviews stays the same, and stuck running reviews are not retried or changed by this fix.
- Added coverage for the slot count rules so recent work still counts and old stuck work does not.

## Verification
1. In a local database, create an owner with recent running or queued reviews and confirm they still use slots.
2. Add a running review older than 90 minutes and confirm dispatch no longer counts it as using a slot.
3. Add a queued review older than 5 minutes and confirm dispatch still treats it as recoverable queued work, not active work.

## Visual Changes
N/A

## Reviewer Notes
Old stuck running reviews will still appear as running until a separate cleanup changes their saved state. This PR only restores dispatch capacity.


markijbema (approved):
Wont this introduce the risk of people having unlimited number of endlessly running review agents? How do we know those reviews running for 90 minutes aren't actually consuming tokens?

alex-alecu:
> Wont this introduce the risk of people having unlimited number of endlessly running review agents? How do we know those reviews running for 90 minutes aren't actually consuming tokens?

@markijbema I will address this in a followup PR. As said in Slack I wanted a PR that solves the imediate pain with minimal code.

PR #2848, opened 2026-04-28
Allow document/pdf upload for Claude, GPT and Grok 4 models

chrarnoldus:


marius-kilocode (approved):

apps/web/src/lib/ai-gateway/models.ts
```diff
@@ -63,6 +68,10 @@ export function isFreeModel(model: string): boolean {
   );
 }
 
+export function isPdfSupportingModel(model: string): boolean {
+  return isAnthropicModel(model) || isOpenAiModel(model) || isGrok4Model(model);
```
marius-kilocode, line 72: Does every OpenAI model support pdf?

chrarnoldus: probably, I guess it's an API feature

PR #2870, opened 2026-04-28
feat(mobile): show per-instance unread badges on dashboard

eshurakov:
## Summary

Adds per-instance unread badges to the KiloClaw cards on the mobile dashboard, so users can see at a glance which instances have unread chat messages.

Screenshot 2026-04-28 at 22 09 34


iscekic (approved):
Approving this one, with the note that I'll probably end up ripping most of it out anyway in my migration to kilo chat. 😄

PR #2879, opened 2026-04-29
feat(kiloclaw): route *.kiloclaw.ai and inject per-instance origins

pandemicsyn:
## Summary

Lays the groundwork for per-instance name-based virtual hosting on `<label>.kiloclaw.ai`, alongside the existing `claw.kilosessions.ai` custom domain.

This PR makes the worker **reachable** on the new host pattern and teaches it to hand each machine the right `Origin` allowlist, but does not change any user-facing URLs, cookie domains, or redirect targets. That cutover is a follow-up PR.

**What changed**

- **`wrangler.jsonc`** — added a wildcard Workers route `*.kiloclaw.ai/*` on a new `kiloclaw.ai` zone. Cloudflare Custom Domains do not support wildcard hostnames, so this uses a route pattern (non-`custom_domain`). The existing `claw.kilosessions.ai` custom domain entry is untouched. The bare apex `kiloclaw.ai` is **unaffected** by this change — `*.kiloclaw.ai/*` matches one-or-more-labels subdomains only, and no apex DNS record or route is added.
- **Hostname label helper** (`src/auth/hostname-label.ts`). Maps both sandboxId shapes to a DNS-safe hostname label (and back, for the host-based router in the next PR):

  | kind | sandboxId | label |
  | --- | --- | --- |
  | instance-keyed | `ki_{32hex}` | `i-{32hex}` |
  | legacy | `{base64url-body}` | `u-{body}` |

  Prefix (`i-` / `u-`) disambiguates the two cases without a database lookup. For realistic userIds (UUIDs, `oauth/provider:{sub}`, emails, Google numeric subs) the base64url encoder produces alnum-only output, so the label body is strictly RFC 1035 compliant (`[A-Za-z0-9]`). Pathological userIds containing `>`/`~`/`?`/DEL — none of which appear in production data — would yield `-` or `_` in the base64url body; the helper returns `null` for those and the caller skips injection rather than emitting a non-compliant label.

- **`OPENCLAW_ALLOWED_ORIGINS` — per-instance injection for all instances** (`src/gateway/env.ts`). OpenClaw's origin check (`src/gateway/origin-check.ts` upstream) does exact case-insensitive string matching; it does not interpret host wildcards. Rather than widening the allowlist with unsupported patterns or going allow-all, the worker now appends `https://<label>.kiloclaw.ai` to the comma-separated list at env-build time, using the label helper above. **Both legacy and instance-keyed sandboxes get a per-instance origin**, so `*.kiloclaw.ai` routing will work for every instance once the machine restarts.

- **`WORKER_CONTROLLER_CAPABILITIES_VERSION`** (`src/config.ts`) — new constant marking what the worker hands to a machine's controller at env-build time. Starts at 2; legacy pre-tracking instances are treated as 1.

- **`controllerCapabilitiesVersion` on the Instance DO** — new nullable field on `PersistedStateSchema`, hydrated in `loadState`, initialized in `resetMutableState`, and exposed via `getStatus()`. **Persisted only after the provider confirms the machine update was applied**, via a dedicated `markControllerCapabilitiesApplied` helper called at three success sites: after `startRuntime` + `persistProviderResult`, after `restartRuntime` + `persistProviderResult`, and after `createNewMachine` in the recovery path.

**Why capability tracking**

`OPENCLAW_ALLOWED_ORIGINS` is baked into a Fly machine at provision/start/restart time. Running machines don't pick up a new worker-side value until they restart. During the eventual cutover to `<label>.kiloclaw.ai`, the dashboard needs a per-instance signal to decide whether an instance is safe to route to the new host or must keep using `claw.kilosessions.ai` until it's restarted. `controllerCapabilitiesVersion` is that signal, written only once the new env is in place on a live machine.

**Out of scope (follow-up PRs)**

- Host-based instance routing in the catch-all proxy (`Host: <label>.kiloclaw.ai` → Instance DO, parsing via `sandboxIdFromHostnameLabel`).
- Access-gateway cookie semantics on the new host.
- Dashboard "Open instance" link generation.
- `KILOCLAW_CHECKIN_URL` / OAuth redirect URIs.
- Retiring `claw.kilosessions.ai`.

**Cloudflare-side prerequisites (not code)**

- `kiloclaw.ai` active as a zone on the production CF account.
- Proxied wildcard DNS record on `*.kiloclaw.ai` (originless `AAAA * → 100::`). No apex record added — bare `kiloclaw.ai` stays independent of this change.
- TLS cert SAN covers `*.kiloclaw.ai` (Universal handles this on CF nameservers).

## Verification

Planned smoke test after deploy:

- `curl https://claw.kilosessions.ai/health` still works (regression).
- `curl https://kiloclaw.ai/` does **not** hit this worker (apex is unaffected).
- TLS handshake on `https://foo.kiloclaw.ai` presents a cert whose SAN covers `*.kiloclaw.ai`.
- After an **instance-keyed** user restarts their machine, its `config.gateway.controlUi.allowedOrigins` contains `https://i-<32hex>.kiloclaw.ai` and its DO `getStatus()` reports `controllerCapabilitiesVersion: 2`.

## Visual Changes

N/A

## Reviewer Notes

- **No user-visible change in this PR.** The new wildcard route is dormant until `*.kiloclaw.ai` resolves, and no caller constructs URLs on the new host yet.
- **Apex `kiloclaw.ai` is intentionally untouched.** The route pattern `*.kiloclaw.ai/*` does not match the apex; no apex DNS record or route is added.
- **OpenClaw origin check** is exact-match only. Wildcard host patterns like `https://*.kiloclaw.ai` would be stored as literals and never match a real `Origin` header — intentionally avoided.
- **Legacy and instance-keyed are unified.** Both get a per-instance origin (`u-` / `i-` labels respectively) via the shared `hostnameLabelFromSandboxId` helper. There is no "legacy stays on the old host" fallback path.
- **Capability version is only written after provider success.** `buildUserEnvVars` no longer persists it. The three call sites of `markControllerCapabilitiesApplied` all fire after `persistProviderResult` returns; a provider failure aborts before the mark, keeping the DO state consistent with what's actually running.
- **Risk**: the capability version needs to be bumped every time we change what the controller is configured with (env var shape, config structure, etc.). The constant has a version-history docstring to make that expectation explicit.
- **Follow-up `hostnameLabelFromSandboxId` / `sandboxIdFromHostnameLabel`** are the two sides of the label boundary that PR2 (host-based routing) will consume — the inverse function is already implemented and tested here so the next PR stays small.

### DIFF
--- a/services/kiloclaw/src/config.ts
+++ b/services/kiloclaw/src/config.ts
@@ -18,17 +18,17 @@ export const OPENCLAW_BUILTIN_DEFAULT_MODEL = 'kilocode/anthropic/claude-opus-4.
  * Version marker for what the worker hands to a machine's controller at
  * provision/start/restart time. Bumped when we change the set of env vars,
  * controller config shape, or origin-check rules. Persisted per-instance as
- * `controllerCapabilitiesVersion` so the dashboard can tell whether a given
- * live machine was configured by a current worker (safe to route at a new
- * host, etc.) or is still running on a pre-cutover config and needs a restart.
+ * `controllerCapabilitiesVersion` so callers can tell which controller
+ * configuration contract a running machine was started with.
  *
  * Legacy instances (persisted as null) are treated as version 1.
  *
  * Version history:
  *   1 — implicit; pre-capability-tracking.
- *   2 — controller accepts wildcard entries in OPENCLAW_ALLOWED_ORIGINS
- *       (e.g. `https://*.kiloclaw.ai`). Machines at this version can serve
- *       the Control UI from any *.kiloclaw.ai subdomain.
+ *   2 — controller env contains a per-instance literal origin
+ *       `https://<label>.kiloclaw.ai` appended to OPENCLAW_ALLOWED_ORIGINS.
+ *       Origin checking remains exact string only; wildcard host patterns are
+ *       not supported.
  */
 export const WORKER_CONTROLLER_CAPABILITIES_VERSION = 2;
 
--- a/services/kiloclaw/src/config.ts
+++ b/services/kiloclaw/src/config.ts
@@ -14,6 +14,24 @@ export const OPENCLAW_INTERNAL_PORT = 3001;
  *  Used as fallback when the user clears their model selection. */
 export const OPENCLAW_BUILTIN_DEFAULT_MODEL = 'kilocode/anthropic/claude-opus-4.6';
 
+/**
+ * Version marker for what the worker hands to a machine's controller at
+ * provision/start/restart time. Bumped when we change the set of env vars,
+ * controller config shape, or origin-check rules. Persisted per-instance as
+ * `controllerCapabilitiesVersion` so the dashboard can tell whether a given
+ * live machine was configured by a current worker (safe to route at a new
+ * host, etc.) or is still running on a pre-cutover config and needs a restart.
+ *
+ * Legacy instances (persisted as null) are treated as version 1.
+ *
+ * Version history:
+ *   1 — implicit; pre-capability-tracking.
+ *   2 — controller accepts wildcard entries in OPENCLAW_ALLOWED_ORIGINS
+ *       (e.g. `https://*.kiloclaw.ai`). Machines at this version can serve
+ *       the Control UI from any *.kiloclaw.ai subdomain.
+ */
+export const WORKER_CONTROLLER_CAPABILITIES_VERSION = 2;
+
 /** Maximum time to wait for the machine to reach 'started' state.
  *  Fly's /wait endpoint caps at 60s (spec.json:1538). */
 export const STARTUP_TIMEOUT_SECONDS = 60;
--- a/services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
+++ b/services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
@@ -2397,6 +2397,7 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
     botNature: string | null;
     botVibe: string | null;
     botEmoji: string | null;
+    controllerCapabilitiesVersion: number | null;
   }> {
     await this.loadState();
 
@@ -2454,6 +2455,7 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
       botNature: this.s.botNature,
       botVibe: this.s.botVibe,
       botEmoji: this.s.botEmoji,
+      controllerCapabilitiesVersion: this.s.controllerCapabilitiesVersion,
     };
   }
 
--- a/services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
+++ b/services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
@@ -47,6 +47,7 @@ import {
   OPENCLAW_BUILTIN_DEFAULT_MODEL,
   RESTARTING_TIMEOUT_MS,
   STARTING_TIMEOUT_MS,
+  WORKER_CONTROLLER_CAPABILITIES_VERSION,
 } from '../../config';
 import {
   SECRET_CATALOG,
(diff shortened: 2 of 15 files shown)


services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
```diff
@@ -1938,6 +1943,7 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
       },
     });
     await this.persistProviderResult(startResult);
+    await markControllerCapabilitiesApplied(this.ctx, this.s);
```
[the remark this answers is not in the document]
pandemicsyn: fixing

services/kiloclaw/src/config.ts
```diff
@@ -14,6 +14,24 @@ export const OPENCLAW_INTERNAL_PORT = 3001;
...
+ * live machine was configured by a current worker (safe to route at a new
+ * host, etc.) or is still running on a pre-cutover config and needs a restart.
+ *
+ * Legacy instances (persisted as null) are treated as version 1.
+ *
+ * Version history:
+ *   1 — implicit; pre-capability-tracking.
+ *   2 — controller accepts wildcard entries in OPENCLAW_ALLOWED_ORIGINS
+ *       (e.g. `https://*.kiloclaw.ai`). Machines at this version can serve
+ *       the Control UI from any *.kiloclaw.ai subdomain.
```
St0rmz1: The v2 description here reads as the opposite of what this PR ships. The new test in `controller/src/config-writer.test.ts` ("passes allowed origins entries through as literal strings") and the env builder both confirm OpenClaw's origin check is exact string matching only. v2 specifically means the worker appends a single literal `https://<label>.kiloclaw.ai` per instance, not a wildcard.

Worth fixing now while v2 is fresh. When v3 lands later, whoever reads this docstring will read "controller accepts wildcard entries" and may design around an assumption that does not hold.

Suggested replacement for the v2 entry:

```
2: controller env contains a per-instance literal origin
   `https://<label>.kiloclaw.ai` (label format defined in
   src/auth/hostname-label.ts) appended to OPENCLAW_ALLOWED_ORIGINS.
   Origin check remains exact string only.
```

services/kiloclaw/src/schemas/instance-config.ts
```diff
@@ -361,6 +361,11 @@ export const PersistedStateSchema = z.object({
   // Each entry is a feature name (e.g. "npm-global-prefix") that gates runtime behavior.
   // New instances get the current feature set; legacy instances have an empty array.
   instanceFeatures: z.array(z.string()).default([]),
```
St0rmz1, line 363: This comment was accurate for the first commit, but the third commit moved the write out of `buildUserEnvVars` and into each ownership guarded `status = running` transition. The test added at `kiloclaw-instance.test.ts:1993` ("does NOT persist controllerCapabilitiesVersion during env build") is explicit about that.

The risk is that a future debugger greps for `buildUserEnvVars`, finds nothing, and either gives up or "fixes" the missing write by adding it back to env build, which would reintroduce exactly the bug the third commit fixed.

Suggested rewrite:

```
// Version of the controller config the running machine was last provisioned
// with. Written atomically with each `status = running` transition in
// _startInner, markRestartSuccessful, syncStatusWithFly, and
// completeUnexpectedStopRecovery, never in env build, so the DO only
// reports a version the running machine actually has. Null means legacy
// (treat as version 1). See WORKER_CONTROLLER_CAPABILITIES_VERSION in
// config.ts for semantics.
```

services/kiloclaw/src/durable-objects/kiloclaw-instance/reconcile.ts
```diff
@@ -877,12 +878,19 @@ export async function syncStatusWithFly(
     if (state.lastStartedAt === null) {
       state.lastStartedAt = Date.now();
     }
+    // Mark the capability version here too: this is the reconcile-driven
+    // success path that covers the case where _startInner / restart's
+    // waitForState timed out (408) but the machine did in fact start with the
+    // new env. Without this mark, a timed-out-but-successful restart would
+    // leave the DO reporting a stale pre-cutover capability version.
+    state.controllerCapabilitiesVersion = WORKER_CONTROLLER_CAPABILITIES_VERSION;
```
St0rmz1, line 884: The justification in the comment above ("the machine did in fact start with the new env") holds when the start was issued by the current worker version, but there is a narrow rollout window where it does not:

1. Previous worker started a machine. Env has no per-instance origin.
2. waitForState timed out (408), DO is stuck in `starting`.
3. This PR deploys.
4. Alarm fires, calls `syncStatusWithFly`, sees Fly is `started`, flips DO to `running` and stamps v2.

The dashboard then sees v2 and assumes the machine accepts `https://<label>.kiloclaw.ai` Origin headers, but the machine has the old env and would reject them. Window is bounded by `STARTING_TIMEOUT_MS`, blast radius is small, recoverable on next restart, but worth either:

(a) tightening this docstring to call out the assumption, or
(b) when the dashboard cutover PR lands, also gating "safe to route on new host" on `lastStartedAt > workerDeployTime`.

Not blocking, more of a heads up for the cutover PR.

St0rmz1 (approved):

PR #2886, opened 2026-04-29
feat(code-reviews): limit code reviews concurrency for individuals 

alex-alecu:
## Why

Personal code reviews could run too many at once, even when an account had little or no credit. That could use too much capacity and make queues harder to control.

## What changed

Organizations can still run up to 20 cloud code reviews at the same time. Personal accounts now get a smaller limit based on credit balance: 3 at once when they have more than $5, and 1 at once when they have $5 or less. The queue uses the new limit before starting more reviews, while old stuck review handling stays the same. Local debug request logs are also ignored so they do not get added by mistake.

## How to test

1. Run `pnpm --filter web typecheck`.
2. Start the test database with `pnpm test:db` if there is no local Postgres.
3. Run `pnpm test -- dispatch-pending-reviews`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ next-env.d.ts
 /dev/fixtures/
 /dev/*/fixtures/
 dev-debug-request-logs/*
+**/dev-debug-request-logs/
 /dev/logs/
 /dev/*/logs/
 
--- a/apps/web/src/lib/code-reviews/dispatch/dispatch-pending-reviews.test.ts
+++ b/apps/web/src/lib/code-reviews/dispatch/dispatch-pending-reviews.test.ts
@@ -1,9 +1,19 @@
 const mockDispatchReview = jest.fn();
+const mockGetAgentConfigForOwner = jest.fn();
+const mockPrepareReviewPayload = jest.fn();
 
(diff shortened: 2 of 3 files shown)


alex-alecu:
Manual test passed.

Tested:
- Direct local dispatch for a personal user with a $5.000001 balance and 5 pending reviews.
- Direct local dispatch for a personal user with exactly a $5 balance and 5 pending reviews.
- Direct local dispatch for an org with 18 running reviews and 5 pending reviews.
- Direct local stale-queued recovery for a default-tier personal user.

Verified:
- Funded personal owner dispatched 3 reviews, left 2 pending, and sent 3 worker `/review` requests with `skipBalanceCheck: true` and `agentVersion: v2`.
- Exactly-$5 personal owner dispatched 1 review, left 4 pending, and preserved the strict `> $5` threshold.
- Org owner dispatched 2 more reviews to reach 20 active reviews, leaving 3 pending.
- Stale queued review did not count as active, was re-queued with a fresh timestamp, and one pending review remained queued behind the default-tier limit.
- Local database cleanup left no manual test reviews, users, or organizations behind.

eshurakov (approved):

PR #2915, opened 2026-04-29
feat(kiloclaw): add lifecycle push notifications

eshurakov:
## Summary

- Add KiloClaw lifecycle push notifications for instance-ready and start-failed states.
- Route pushes through the notifications Worker and deep-link taps to instance chat.
- Keep production ready gating on low-load checkins; allow dev checkins to fire immediately for local testing.

## Verification

- Tested locally with KiloClaw and confirmed the notifications service received the ready lifecycle dispatch.

## Visual Changes

N/A

## Reviewer Notes

- Existing DO state defaults new lifecycle flags to already-sent to avoid retroactive pushes on deploy.


services/notifications/src/lib/notifications-service.ts
```diff
@@ -0,0 +1,79 @@
...
+        const accessToken = await this.env.EXPO_ACCESS_TOKEN.get();
+        return sendPushNotifications(messages, accessToken);
+      },
+      enqueueReceipts: async ticketTokenPairs => {
+        const receiptMsg: ReceiptCheckMessage = { ticketTokenPairs };
+        await this.env.RECEIPTS_QUEUE.send(receiptMsg, { delaySeconds: 900 });
+      },
+    });
+
+    console.log(
```
iscekic: Remove this?

iscekic (approved):

services/notifications/src/lib/notifications-service.ts
eshurakov: removed, thanks!

PR #2923, opened 2026-04-29
Kiloclaw pin consent and version change

St0rmz1:
## Summary

Two behavior changes to KiloClaw version pins.

1. Pins move from a hard lock to an advisory consent gate. Setting or removing a pin no longer requires special permissions. Either the user or an admin can write or clear the pin on an instance at any time.

2. Any version change action (Upgrade to Latest, the new admin Change Version button, the Kilo CLI recovery redeploy) now consults the pin. If a pin exists, the action requires explicit acknowledgement. After the caller confirms, the pin is cleared and the redeploy proceeds.

## What changed

Backend procedures (kiloclaw.restartMachine, admin.kiloclawInstances.restartMachine, organization.kiloclaw.restartMachine):

- New input flag (acknowledgePinRemoval on user and org, acknowledgeOverride on admin). When imageTag is set and a pin exists, the procedure throws PRECONDITION_FAILED with message PIN_EXISTS unless the flag is true.
- On acknowledgement, the pin row is deleted and the cleared state is pushed to the instance Durable Object via the existing pushPinToWorker helper before the redeploy fires.

Pin write mutations (setMyPin and removeMyPin on user and org routers):

- Removed the FORBIDDEN guards that previously blocked users from overwriting an admin pin. Pins are advisory now, so any caller can update them.

End user UI:

- UpgradeKiloClawDialog shows a warning when a pin is present and lists the pinned image tag.
- The redeploy or upgrade dialog inside InstanceControls shows the same warning when the upgrade radio is selected.
- The Kilo CLI recovery dialog mentions pin clearance on the existing amber warning so the user is not surprised when the pin disappears after recovery.
- All four user facing call paths now send acknowledgePinRemoval true so the dialog click acts as consent.

Admin UI on the instance detail page:

- New Change Version button next to Upgrade to Latest. Opens a dialog with a version selector populated from the available catalog. Surfaces a warning when a pin is set, plus a soft advisory when the chosen version is older than the running version.
- Upgrade to Latest is now pin aware. If a pin exists, clicking it routes to the Change Version dialog so the admin can see what is being overridden and confirm. A defensive PIN_EXISTS error fallback covers the rare race where a pin appears between the click time check and the backend gate.
- Replaced the isFlyProvider gate on the version buttons with a capability flag that also includes docker-local, so local dev exercises the same flow.

## Verification

Manual checks against local dev:

- [x] Change Version dialog opens, version selector lists available catalog tags, instance redeploys on the chosen tag
- [x] Override prompt shown and acknowledged when an admin pin is set
- [x] Override prompt shown and acknowledged when a user pin is set
- [x] Older version advisory shown when the chosen tag was published before the running tag
- [x] Backend test suite passes locally (68 tests across user and admin routers)
- [x] format check, typecheck, and lint all pass

## Visual Changes

N/A

### DIFF
--- a/apps/web/src/app/(app)/claw/components/ClawSettingsPage.tsx
+++ b/apps/web/src/app/(app)/claw/components/ClawSettingsPage.tsx
@@ -4,6 +4,7 @@ import { useCallback, useEffect, useState } from 'react';
 import { toast } from 'sonner';
 import { Settings } from 'lucide-react';
 import { useRouter } from 'next/navigation';
+import { TRPCClientError } from '@trpc/client';
 import type { KiloClawDashboardStatus } from '@/lib/kiloclaw/types';
 import { useKiloClawStatus, useKiloClawMutations, useKiloClawMyPin } from '@/hooks/useKiloClaw';
 import {
@@ -37,11 +38,13 @@ function ClawSettingsInner({
   const orgMutations = useOrgKiloClawMutations(organizationId ?? '');
   const mutations = organizationId ? orgMutations : personalMutations;
 
-  // Pin state for the upgrade-confirmation dialog. The dialog surfaces a
-  // warning when a pin exists; the click then acts as the consent that
-  // removes the pin and proceeds with the upgrade.
-  const personalPin = useKiloClawMyPin();
-  const orgPin = useOrgKiloClawMyPin(organizationId ?? '');
+  // Pin state for the upgrade-confirmation dialog. Only the active
+  // context queries (org vs personal); the inactive hook stays disabled
+  // so we don't fire an org getMyPin with an empty organizationId.
+  const personalPin = useKiloClawMyPin({ enabled: !organizationId });
+  const orgPin = useOrgKiloClawMyPin(organizationId ?? '', {
+    enabled: !!organizationId,
+  });
   const pin = organizationId ? orgPin.data : personalPin.data;
   const pinnedImageTag = pin?.image_tag ?? null;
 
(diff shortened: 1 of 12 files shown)


joshavant (approved):
1. **IMPORTANT** — Missing org-router regression tests for core behavior changes
     - **Area**: `apps/web/src/routers/organizations/organization-kiloclaw-router.ts` (restart gate + pin mutation
 semantics changed materially)
     - **Why it matters:** Core PR claim says behavior is changed across user/admin/org. Only user/admin got robust
 new tests.
     - **Concrete fix:** Add org-router tests mirroring personal coverage:
           - restart with `imageTag` + pin + no ack → `PIN_EXISTS`
           - restart with ack clears pin and proceeds
           - plain restart without `imageTag` ignores pin
           - `setMyPin`/`removeMyPin` can override/clear admin-set pin
 2. **IMPORTANT** — Concurrency window still exists in backend gate
     - **Area:** user/org/admin restart mutations (select pin, then later restart)
     - **Why it matters:** A pin created after SELECT but before restart can bypass consent for that specific request.
     - **Concrete fix:** Consider transaction/locking strategy or conditional mutation pattern that guarantees gate
 decision is based on latest pin state at action time. If intentionally accepted, document this as
 known/acceptable race.

Follow-ups (optional)

 - Add a small shared helper for PIN_EXISTS error handling in client components to reduce duplication.
 - Consider centralizing “ack only if warning rendered” logic into hook-level mutation wrappers for consistency.

PR #2954, opened 2026-04-30
Notifications - Add cloud agent push notifications

eshurakov:
## Summary

Terminal cloud-agent session events (completed / failed / interrupted) now send event to the notifications service.

## Verification

- [x] local testing till notifications service


iscekic (commented):
A couple from the bot

packages/notifications/src/rpc-schemas.ts
```diff
@@ -132,11 +132,35 @@ export const sendScheduledActionNoticeOutputSchema = z.object({
...
+export type SendCloudAgentSessionNotificationResult = z.infer<
+  typeof sendCloudAgentSessionNotificationOutputSchema
+>;
+
 // ── dispatchPush (internal DO RPC) ──────────────────────────────────
 
 export const dispatchPushInputSchema = z.object({
   userId: z.string().min(1),
-  presenceContext: z.string().min(1),
+  presenceContext: z.string().min(1).optional(),
```
iscekic: Making `presenceContext` optional turns an important behavior into an implicit default: any future `dispatchPush` caller that forgets the field silently bypasses presence suppression. For maintainability, can we make the choice explicit, e.g. `presence: { kind: "check"; context: string } | { kind: "skip" }`, or at least require `presenceContext: string | null` so every caller has to opt into check vs skip?

services/cloud-agent-next/wrangler.jsonc
```diff
@@ -277,6 +282,11 @@
           "service": "git-token-service-dev",
           "entrypoint": "GitTokenRPCEntrypoint",
         },
+        {
+          "binding": "NOTIFICATIONS",
+          "service": "notifications",
```
iscekic, line 284: This dev binding points at `notifications`, but this PR also defines a dev Worker named `notifications-dev` in the notifications config. That is easy to maintain incorrectly and can make `cloud-agent-next --env dev` talk to the wrong notifications Worker. Should this be `notifications-dev`, or should the config document why dev intentionally shares the top-level service?

services/cloud-agent-next/src/notifications-binding.ts
```diff
@@ -0,0 +1,30 @@
...
+ * `wrangler types` only sees `Fetcher` for service bindings; the actual RPC
+ * shape comes from the notifications worker's WorkerEntrypoint and is declared
+ * here so the generated file can be freely regenerated.
+ *
+ * Keep in sync with: services/notifications/src/index.ts (NotificationsService).
+ */
+
+import type { CloudAgentPushStatus } from './notifications/types.js';
+
+export type SendCloudAgentSessionNotificationParams = {
```
iscekic: These RPC params/result/status types are hand-copied even though `@kilocode/notifications` now owns the Zod schemas and schema-derived types. That gives us multiple drift points across the service boundary. Can we follow the existing `services/kiloclaw/src/notifications-binding.ts` pattern and import the shared `SendCloudAgentSessionNotificationParams`, `SendCloudAgentSessionNotificationResult`, and status type from `@kilocode/notifications` instead?

services/notifications/src/index.ts
```diff
@@ -216,6 +219,51 @@ export class NotificationsService extends WorkerEntrypoint<Env> {
     });
   }
 
+  async sendCloudAgentSessionNotification(
```
iscekic, line 222: This keeps the whole cloud-agent notification flow inline in the Worker entrypoint: validation, DB lookup, DO routing, push payload construction, and outcome mapping. The neighboring notification paths are easier to maintain because the core behavior is extracted behind testable helpers. Can we move this into something like `dispatchCloudAgentSessionPush` / `sendCloudAgentSessionNotificationCore` with injected deps, leaving this method as thin binding glue?

services/notifications/src/dos/NotificationChannelDO.ts
RSO, line 97: Why are we only marking the `idemKey` as pending when it's a badge and not a retry? The new notification in this PR has `badge:null`, and I'm wondering whether that means that there could be idempotency issues in case of a failure.

packages/notifications/src/rpc-schemas.ts
eshurakov: Addressed in 074b56d21: `presenceContext` is now required as `string | null`, and the Cloud Agent caller explicitly passes `null` to opt out of presence suppression.

services/cloud-agent-next/wrangler.jsonc
eshurakov: Addressed in 074b56d21: the `env.dev` binding now targets `notifications-dev`, and the notifications `dev` script launches the matching named environment while root `start` remains paired with root bindings.

services/cloud-agent-next/src/notifications-binding.ts
eshurakov: Addressed in 074b56d21: `notifications-binding.ts` now imports and re-exports the schema-derived contract types from `@kilocode/notifications`, with the workspace dependency declared directly.

services/notifications/src/dos/NotificationChannelDO.ts
eshurakov: Addressed in 074b56d21: badge-less Cloud Agent pushes now persist an `accepted` delivery marker after Expo acceptance and retry only receipt/stale-token bookkeeping, with regression coverage preventing a second visible push.

services/notifications/src/index.ts
eshurakov: Addressed in b86c0348c: Cloud Agent notification validation, authorization, push payload construction, and outcome mapping now live in `dispatchCloudAgentSessionPush` (`services/notifications/src/lib/cloud-agent-session-push.ts`). The Worker entrypoint only wires DB/DO dependencies, and the behavior/security tests now exercise the extracted core directly.

iscekic (approved):
Reviewed by Codex: no findings.

PR #2972, opened 2026-04-30
fix(kiloclaw): repair gateway-client scope baseline

joshavant:
## Summary

- Repair gateway-client scope-upgrade loops where OpenClaw's persistent device identity was originally paired as `cli`/`probe` with only `operator.read`.
- Merge scopes from gateway-client repair requests into the matching paired record by `deviceId`, including extra requested scopes such as `operator.talk.secrets`.
- Move paired-device scope remediation into its own pre-onboard/doctor bootstrap phase and document the phase in the controller spec.

## Verification

The fix was informed by readonly inspection of an affected Fly instance and OpenClaw v2026.4.15 source.

## Visual Changes

N/A

## Reviewer Notes

The repair-by-deviceId path is intentionally constrained to pending `gateway-client` operator repair requests and existing paired records that already have an operator token. This avoids widening arbitrary paired devices while covering OpenClaw's persisted-device-identity model, where `clientId` is metadata and `deviceId` is the authoritative pairing key.


services/kiloclaw/controller/src/bootstrap.ts
```diff
@@ -788,6 +812,48 @@ export type GatewayClientDeviceScopeRemediationResult = {
...
+  const tokens = record.tokens;
+  return isJsonRecord(tokens) && isJsonRecord(tokens[OPERATOR_TOKEN_ROLE]);
+}
+
+function readGatewayClientRepairRequests(
+  deps: Pick<BootstrapDeps, 'existsSync' | 'readFileSync'>
+): GatewayClientScopeRepairRequest[] {
+  if (!deps.existsSync(DEVICE_PENDING_PATH)) return [];
+
+  const pendingFile = JSON.parse(deps.readFileSync(DEVICE_PENDING_PATH, 'utf8')) as unknown;
```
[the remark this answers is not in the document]
joshavant: Fixed in 6ea365174: https://github.com/Kilo-Org/cloud/commit/6ea365174

services/kiloclaw/controller/src/bootstrap.ts
```diff
@@ -108,17 +114,46 @@ function isJsonRecord(value: unknown): value is JsonRecord {
   return typeof value === 'object' && value !== null && !Array.isArray(value);
 }
 
-function stringArrayEquals(left: unknown, right: readonly string[]): boolean {
+function stringArraySetEquals(left: unknown, right: readonly string[]): boolean {
   if (!Array.isArray(left) || left.length !== right.length) return false;
-  return left.every((value, index) => value === right[index]);
+  const rightSet = new Set(right);
+  return left.every(value => typeof value === 'string' && rightSet.has(value));
```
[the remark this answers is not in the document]
joshavant: Fixed in 7eaf8050c: https://github.com/Kilo-Org/cloud/commit/7eaf8050c

pandemicsyn (approved):

services/kiloclaw/controller/src/bootstrap.ts
```diff
@@ -806,21 +887,34 @@ export function remediateGatewayClientDeviceScopes(
...
+    const repairScopes = deviceId ? repairScopesByDeviceId.get(deviceId) : undefined;
+    const isGatewayClientPairing = value.clientId === GATEWAY_CLIENT_ID;
+    const shouldRepairByDeviceId =
+      Array.isArray(repairScopes) && repairScopes.length > 0 && hasOperatorToken(value);
+    if (!isGatewayClientPairing && !shouldRepairByDeviceId) continue;
     checked += 1;
 
+    const tokens = value.tokens;
+    const operatorToken = isJsonRecord(tokens) ? tokens[OPERATOR_TOKEN_ROLE] : undefined;
+    const mergedScopes = mergeStringLists(
```
St0rmz1, line 908: Behavior change worth confirming: this is now a monotonic merge across `scopes`, `approvedScopes`, the operator token's `scopes`, the baseline, and the repair request. The pre-PR version overwrote with the canonical baseline.

The practical effect is that once any scope lands in any of those three persisted fields, it can't be revoked by hand-editing `scopes` or `approvedScopes`, since the next boot re-merges from the others. If revocation should still be possible, intersecting with the baseline (plus repair scopes) would preserve that. If monotonic-grow is the intended contract here, a one-line comment naming that intent would help the next reader.

services/kiloclaw/controller/src/bootstrap.ts
```diff
@@ -788,6 +818,57 @@ export type GatewayClientDeviceScopeRemediationResult = {
...
+  if (!isJsonRecord(pendingFile)) {
+    console.warn('[controller] Device pending state is not an object, skipping repair lookup');
+    return [];
+  }
+
+  const repairs: GatewayClientScopeRepairRequest[] = [];
+  for (const value of Object.values(pendingFile)) {
+    if (!isJsonRecord(value)) continue;
+    if (value.clientId !== GATEWAY_CLIENT_ID) continue;
+    if (value.isRepair !== true) continue;
```
St0rmz1, line 847: The repair-by-deviceId path effectively treats `pending.json` as authoritative: any scope present on a `gateway-client` `isRepair: true` request gets merged into the paired record, with no allow-list intersection. The `hasOperatorRole` and `hasOperatorToken` guards prevent fresh grants but don't bound which scopes get added.

That's fine under the current trust model where `pending.json` is only written by local OpenClaw. The risk is future drift: if anything else ever stages requests into this file (for example an inbound pairing endpoint), this becomes an arbitrary-scope-grant against any deviceId that already has an operator token.

Two options to make that boundary harder to erode:
1. Intersect `repairScopes` with `GATEWAY_CLIENT_OPERATOR_SCOPES` so the repair path can never grant a scope outside the canonical set.
2. Add a short comment naming the trust assumption ("pending.json is trusted local state written by OpenClaw") so reviewers of future writers know to revisit this guard.

services/kiloclaw/controller/src/bootstrap.ts
joshavant: Documented the monotonic repair intent in 3f4d6c90d: https://github.com/Kilo-Org/cloud/commit/3f4d6c90d

services/kiloclaw/controller/src/bootstrap.ts
joshavant: Documented the pending.json trust boundary in 11eac752f: https://github.com/Kilo-Org/cloud/commit/11eac752f

PR #3000, opened 2026-05-01
feat(kiloclaw) Admin instance actions scheduler

St0rmz1:
## Summary

Adds a scheduling primitive for kiloclaw admin actions covering two action types: `scheduled_restart` (the worker DO redeploys on its current image at the chosen time) and `version_change` (the worker DO redeploys on a chosen image tag, with optional pin override). Three new tables back the feature: `kiloclaw_scheduled_actions` (parent), `kiloclaw_scheduled_action_stages`, and `kiloclaw_scheduled_action_targets`. The DO `alarm()` handler gains a wedge that queries Postgres for due targets and dispatches per `action_type`. Admin UI surfaces are: a new Scheduler tab on the kiloclaw dashboard, an upcoming action indicator on the per instance page, a Schedule for later toggle on the bulk Change Version dialog, and the same toggle on the per instance Change Version dialog. The Upgrade to Latest button keeps its immediate semantics (renamed "Upgrade to Latest Now") and now opens a confirmation dialog. Notifications are intentionally not in this PR; the admin tooling says so where it matters.

Architectural notes worth flagging:

- The DO apply path mirrors the `tracked_image_tag` sync wedge from PR #2942: `runScheduledActionApply` runs inside `alarm()` via `waitUntil`, so reconciliation is unaffected when Postgres is unreachable.
- No new DO storage field, no `scheduleAlarm()` math. The existing reconcile cadence (5 minutes for running instances, longer for hibernated) is the only timing source. The scheduled time is a "no earlier than" bound, not an exact fire time. Drift is up to one alarm interval and is surfaced to admins in the UI.
- Wake on schedule push: a new DO method `notifyScheduledActionPending()` calls `scheduleAlarm()` to refresh the alarm to a fresh near future timestamp. Exposed via worker route `POST /api/platform/scheduled-action/wake`. The web `scheduleAction` mutation fires this best effort after the inserts, batched at 20 concurrent so a 500 instance schedule does not open 500 outbound sockets at once.
- Concurrency hardening: the conflict check and inserts in `scheduleAction` run inside one `db.transaction(..., { isolationLevel: 'serializable' })` so two admins racing the same instance cannot both succeed. Outcome recording (`recordScheduledActionTargetOutcome`) wraps the target write and both counter writes in a single transaction so a partial failure rolls back cleanly. The DO apply path claims the target with a `pending` to `running` CAS before invoking the side effect so two concurrent waitUntil passes cannot both fire `restartMachine`.
- One pending scheduled action per instance: the API rejects with `CONFLICT` when any selected instance already has a pending action under a parent in `scheduled` or `running` status. Bulk schedules silently filter destroyed instances (mirrors the `bulkChangeVersion` Apply Now path) and reject only if every selected instance is destroyed.
- Per target cancel: a new endpoint `cancelScheduledActionTarget` drops one instance from a bulk schedule via atomic CAS on the target row and runs the same promotion sweep as the alarm path so an action whose targets are all individually cancelled does not stay `scheduled` forever.
- GDPR: `kiloclaw_scheduled_action_targets.user_id` references `kilocode_users.id`. The targets table is retained operational state; `softDeleteUser` is updated and a coverage test confirms it succeeds when the user owns scheduled action targets.

## Verification

- [x] Scheduled a restart on a dev instance from the Scheduler tab. Watched the wake call land, the alarm fire, the wedge run, the worker actually redeploy.
- [x] Scheduled a version change on a dev instance from the per instance Change Version dialog (Scheduled tab). Verified the source and target tag rendering, the override pins behavior, and that Apply Now still works alongside.
- [x] Bulk scheduled from the instances toolbar dialog in Schedule for later mode. Verified every selected instance appears as a target in the Scheduler tab and the parent shows the right total count.
- [x] Cancelled a single instance schedule from the upcoming action indicator. Confirmed the dialog shows just one option for single instance schedules.
- [x] Cancelled one instance from a bulk schedule (the dialog offered "Cancel only this instance" and "Cancel entire batch (N instances)"). Confirmed the parent stays running for the remaining targets.
- [x] Confirm dialog on Upgrade to Latest Now (the previous one click action was easy to fire by accident).
- [x] Local DB rebuild (`docker compose rm -s -f -v postgres`, `docker volume rm dev_postgres_data`, `pnpm drizzle migrate`) applied every migration cleanly with the regenerated 0109 migration after rebase.
- [x] 79 router tests pass. Format check, typecheck, and lint all clean.

## Visual Changes

| Before | After |
| ------ | ----- |
| (no Scheduler tab) | New Scheduler tab at `/admin/kiloclaw?tab=scheduler` with a sortable Recent actions table, separate Applied / Skipped / Failed columns, a Run at column, and a per row Detail dialog showing every target with source and target tags. |
| (no scheduled action awareness on the instance page) | Instance Information card always shows an "Upcoming scheduled action" row. Empty state is "None"; otherwise the action type renders in yellow, with run time, source and target tags labelled with OpenClaw versions, and an inline Cancel button. |
| (immediate only) | Change Version dialog gains a Now / Scheduled tab toggle. Bulk Change Version dialog gains Apply now / Schedule for later. Both warn that notifications are not implemented yet. |
| Upgrade to Latest fires on click | Renamed Upgrade to Latest Now and opens a confirmation dialog before firing. |

Screenshot 2026-05-01 at 2 10 03 PM

Screenshot 2026-05-01 at 2 10 32 PM

Screenshot 2026-05-01 at 2 11 01 PM

## Reviewer Notes

- The schedule path relies on the existing reconcile alarm cadence (no new alarm slot) and accepts up to one alarm interval of drift past the chosen time. The Scheduler tab copy says so. Tightening it would require pushing the actual scheduled timestamp into DO storage and modifying `scheduleAlarm` to pick `min(reconcileNextAt, scheduledActionNextAt)`. Deferred until drift becomes operational pain.
- The wake fan out to up to 500 DOs is best effort and concurrency capped at 20. A wake failure does not block the schedule; the next user activity will eventually rearm the alarm.
- Notifications are not in this PR. End user sessions are interrupted with no warning at the scheduled time. The existing immediate `bulkChangeVersion` path (PR #2975) has the same property; a follow up PR will add a notifications framework that makes both paths polite.


services/kiloclaw/src/durable-objects/kiloclaw-instance/scheduled-action-apply.ts
```diff
@@ -0,0 +1,263 @@
...
+        doWarn(ctx.state, 'scheduled-action-apply: record skipped failed', {
+          error: toLoggable(err),
+          targetId: target.target_id,
+        });
+      }
+      continue;
+    }
+
+    try {
+      const outcome = await dispatchByActionType(target, ctx, db);
```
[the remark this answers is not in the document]
St0rmz1: Fixed in db6880136. Apply path now claims the target with an atomic `pending` to `running` CAS via the new `claimScheduledActionTarget` helper before invoking the dispatcher. If two concurrent waitUntil passes find the same due row, only one wins the claim and proceeds to dispatch; the other gets 0 rows back and skips. Added `running` to the target status enum (TS only annotation, no migration since the column is `text`) and relaxed the CAS in `recordScheduledActionTargetOutcome` to accept either `pending` or `running` so both code paths still work.

apps/web/src/routers/admin-kiloclaw-instances-router.ts
```diff
@@ -1623,6 +1626,786 @@ export const adminKiloclawInstancesRouter = createTRPCRouter({
...
+              kiloclaw_scheduled_actions,
+              eq(
+                kiloclaw_scheduled_actions.id,
+                kiloclaw_scheduled_action_targets.scheduled_action_id
+              )
+            )
+            .where(
+              and(
+                inArray(kiloclaw_scheduled_action_targets.instance_id, liveInstanceIds),
+                eq(kiloclaw_scheduled_action_targets.status, 'pending'),
```
pandemicsyn: Theres a lot going on here but i think filter misses the transient running state. claimScheduledActionTarget flips the target to 'running' before dispatching restartMachine. 

A second scheduleAction call that lands during that window observes no matching target and inserts a new pending row under a fresh paren. we'd now have two live actions on the same instance.

pandemicsyn: Additionally, from the robot:


> services/kiloclaw/src/db/index.ts at the claimScheduledActionTarget body (look for the pending → running update; based on the diff that's around L392–405 of the current file). A cross-reference comment there ("callers filtering by status='pending' must also include 'running' to account for in-flight claims — see scheduleAction conflict check") will help whoever reads this function next.

apps/web/src/routers/admin-kiloclaw-instances-router.ts
```diff
@@ -1623,6 +1626,786 @@ export const adminKiloclawInstancesRouter = createTRPCRouter({
...
+          noticeSubject: z.string().max(120).default(''),
+          noticeBody: z.string().max(2000).default(''),
+        }),
+        z.object({
+          actionType: z.literal('version_change'),
+          instanceIds: z.array(z.string().uuid()).min(1).max(500),
+          // The image_tag the worker should redeploy on at the scheduled
+          // time. Must exist in kiloclaw_image_catalog with status='available'
+          // (validated in the procedure body, mirrors bulkChangeVersion).
+          imageTag: z
```
pandemicsyn, line 1668: Not sure this is really an issue. Going through this and gpt flagged:

> The version_change schema accepts the literal string 'latest', and there is no DB constraint preventing kiloclaw_image_catalog.image_tag = 'latest'. If such a catalog row exists, this schedule path stamps 'latest' into target_image_tag; at apply time applyVersionChangeForTarget passes that value to restartMachine, whose imageTag === 'latest' branch re-runs selectImageVersionForInstance at apply time instead of deploying the exact catalog row selected at schedule time. 
>That means a scheduled action may deploy "whatever is latest when it fires" rather than the concrete image the admin intended, and the audit log records metadata.imageTag: 'latest'. 
> Either reject 'latest' here explicitly or resolve it to a concrete catalog tag before inserting the scheduled action rows.bjects/kiloclaw-instance/index.ts:3353 — the if (options.imageTag === 'latest') branch). 

I think its recommendation might be pedantically correct, but I think supporting "latest" here would actually be fine too.

packages/db/src/schema.ts
```diff
@@ -4063,6 +4063,179 @@ export const kiloclaw_version_pins = pgTable('kiloclaw_version_pins', {
...
+  {
+    id: uuid()
+      .default(sql`gen_random_uuid()`)
+      .primaryKey()
+      .notNull(),
+
+    // Discriminator. PR 1 emits 'scheduled_restart'. PR 3 adds
+    // 'version_change'. Future types add new values; type-specific
+    // columns below stay nullable so unrelated types ignore them.
+    action_type: text().$type<'scheduled_restart' | 'version_change'>().notNull(),
```
pandemicsyn, line 4094: These show up in a bunch of spots (here, admin-kiloclaw-instances-router.ts,  services/kiloclaw/src/db/index.ts, ui spots, etc) could define them in packages/db/src/schema-types.ts and then import them from there?

apps/web/src/app/admin/components/KiloclawInstances/KiloclawInstanceDetail.tsx
```diff
@@ -1256,6 +1257,14 @@ export function KiloclawInstanceDetail({ instanceId }: { instanceId: string }) {
   const [resizeConfirmText, setResizeConfirmText] = useState('');
   const [changeVersionDialogOpen, setChangeVersionDialogOpen] = useState(false);
   const [changeVersionSelectedTag, setChangeVersionSelectedTag] = useState<string>('');
+  const [changeVersionMode, setChangeVersionMode] = useState<'now' | 'scheduled'>('now');
+  const [changeVersionScheduledAt, setChangeVersionScheduledAt] = useState<string>(() => {
+    // Default = now + 5 minutes (local time, datetime-local format).
+    const d = new Date(Date.now() + 5 * 60_000);
+    const pad = (n: number) => String(n).padStart(2, '0');
+    return `${d.getFullYear()}-${pad(d.getMonth() + 1)}-${pad(d.getDate())}T${pad(d.getHours())}:${pad(d.getMinutes())}`;
+  });
```
pandemicsyn: nit: only noticed because its at the start of the diff's i think, but caught my eye:

we've got this logic at least 3'ish times

pandemicsyn (approved):

apps/web/src/routers/admin-kiloclaw-instances-router.ts
St0rmz1: Good catch. Fixed in the next push: the conflict check now filters `status IN ('pending', 'running')` so a target in the brief claim window (after the apply path's `pending` to `running` CAS but before the final outcome write) is no longer missed. Error message updated to mention "pending or in-flight". Without this we could end up with two live actions on the same instance exactly as you described.

St0rmz1: Added a cross-reference comment in `claimScheduledActionTarget` calling out that callers checking "is there an in-flight action on this instance?" must include both 'pending' and 'running' in their status filter.

apps/web/src/routers/admin-kiloclaw-instances-router.ts
St0rmz1: Agreed. The 'latest' floating-tag behavior is a feature here: an admin scheduling "upgrade to latest at 3am tomorrow" usually wants whatever 'latest' resolves to at fire time, not what was 'latest' at schedule time. The audit log records what they picked ('latest') so the trail is honest about the intent. Leaving as is.

packages/db/src/schema.ts
St0rmz1: Done — moved to `packages/db/src/schema-types.ts` as `KiloClawScheduledActionStatus`, `KiloClawScheduledActionStageStatus`, and `KiloClawScheduledActionTargetStatus` (zod enum + inferred TS type for each). Schema columns now use the type alias via `.$type<...>()`, and the worker DB helpers import from there too. The router and UI haven't been migrated yet (they use string literals); easy follow-up if we want full coverage.

apps/web/src/app/admin/components/KiloclawInstances/KiloclawInstanceDetail.tsx
St0rmz1: Yep — the now+5min datetime-local seed appeared in 3 places (Scheduler tab, bulk Change Version dialog, per instance Change Version dialog). Extracted to `@/lib/kiloclaw/scheduled-action-form` as `defaultScheduledAt()`. Single source of truth if we ever tune the lead.

services/kiloclaw/src/db/index.ts
```diff
@@ -288,3 +293,333 @@ export async function updateGoogleOAuthConnectionTokenData(
...
+    SET status = CASE
+          WHEN s.applied_count > 0 OR s.skipped_count > 0 THEN 'completed'
+          ELSE 'failed'
+        END,
+        completed_at = COALESCE(s.completed_at, now())
+    WHERE s.scheduled_action_id IN (${idList})
+      AND s.status IN ('pending', 'running')
+      AND NOT EXISTS (
+        SELECT 1 FROM kiloclaw_scheduled_action_targets t
+        WHERE t.stage_id = s.id AND t.status = 'pending'
```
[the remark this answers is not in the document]
St0rmz1: Good catch — same bug pattern as the conflict-check fix but on the promotion sweep. Updated both NOT EXISTS clauses (parent + stage) to treat 'pending' and 'running' as unresolved in `maybePromoteScheduledActionsToCompleted` and the inline equivalent inside `cancelScheduledActionTarget`'s transaction. A parallel apply pass on a sibling target can no longer promote the parent while another target is still mid-dispatch.

PR #3044, opened 2026-05-05
feat(emails): add purchase transactional emails

evanjacobson:
## Summary

KiloClaw and credit purchases now send the transactional confirmations users expect after money or credits move, closing a gap where successful top-ups and subscription activations were only reflected in-app.

This adds Mailgun-backed transactional emails for credit top-ups and Stripe-based KiloClaw subscription starts, with durable send markers so webhook retries can recover missed sends without double-emailing users. The implementation follows the existing server-rendered email template conventions and extends the current KiloClaw email-log pattern for activation-period dedupe.

Credit-based KiloClaw subscription starts will follow in a separate PR.

<details>

- Adds reusable transactional email rendering/sending helpers for the new purchase emails.
- Adds a generic `transactional_email_log` table for idempotent non-KiloClaw purchase email markers.
- Extends `kiloclaw_email_log` with `period_start` so subscription-started emails dedupe per activation period (in practice, either trial-->paid or canceled-->paid) rather than forever per instance.
- Handles duplicate webhook/retry paths by using persisted purchase/change-log state to send once when the first attempt completed the business transaction but missed the email side effect.
- Leaves terminal address failures deduped, while clearing provider-not-configured markers so retries can attempt delivery after configuration is fixed.

</details>

## Verification

- [ ] Not manually verified; this change is primarily webhook/email side-effect behavior and was covered with automated tests rather than a browser flow.

## Visual Changes

N/A - no in-app UI changes. This adds new transactional email templates.

## Reviewer Notes

<details>

- Idempotency behavior around webhook retries and duplicate settlements.
- The `transactional_email_log` and `kiloclaw_email_log.period_start` schema changes.
- Provider failure handling: retryable `provider_not_configured` versus terminal invalid-recipient outcomes.
- Copy and links in the new transactional email templates.

</details>

### DIFF
--- a/apps/web/src/lib/user.ts
+++ b/apps/web/src/lib/user.ts
@@ -594,6 +594,7 @@ export class SoftDeletePreconditionError extends Error {
  * - referral_code_usages (financial, references anonymized user)
  * - kiloclaw_subscriptions, kiloclaw_earlybird_purchases, kiloclaw_email_log (retained records)
  * - kiloclaw_scheduled_action_targets (retained operational records;
+ * - top_up_email_log (retained outbox marker, financial record)
  *   user_id FK references the anonymized kilocode_users row — no PII
  *   stored directly on the target row)
  *
--- a/apps/web/src/lib/user.ts
+++ b/apps/web/src/lib/user.ts
@@ -594,7 +594,7 @@ export class SoftDeletePreconditionError extends Error {
  * - referral_code_usages (financial, references anonymized user)
  * - kiloclaw_subscriptions, kiloclaw_earlybird_purchases, kiloclaw_email_log (retained records)
  * - kiloclaw_scheduled_action_targets (retained operational records;
- * - top_up_email_log (retained outbox marker, financial record)
+ * - transactional_email_log (retained outbox marker, financial record)
  *   user_id FK references the anonymized kilocode_users row — no PII
  *   stored directly on the target row)
  *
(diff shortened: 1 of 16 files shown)


apps/web/src/lib/user.ts
```diff
@@ -594,6 +594,7 @@ export class SoftDeletePreconditionError extends Error {
  * - referral_code_usages (financial, references anonymized user)
  * - kiloclaw_subscriptions, kiloclaw_earlybird_purchases, kiloclaw_email_log (retained records)
  * - kiloclaw_scheduled_action_targets (retained operational records;
+ * - transactional_email_log (retained outbox marker, financial record)
```
evanjacobson, line 597: I landed on this decision considering that kiloclaw_email_log above is retained. Let me know if I should do the GDPR soft delete flow for this table instead.

packages/db/src/migrations/0110_giant_gabe_jones.sql
evanjacobson, line 1: My intention behind creating this table was to do the following:

1. Create a table that can be used for idempotency checks for non-KiloClaw emails
2. Allow the table to be used by future unknown, idempotent emails (or wherever we want to track emails sent)

packages/db/src/schema.ts
```diff
@@ -4386,14 +4394,17 @@ export const kiloclaw_email_log = pgTable(
...
+      .default(sql`'epoch'`),
     sent_at: timestamp({ withTimezone: true, mode: 'string' }).defaultNow().notNull(),
   },
   table => [
     uniqueIndex('UQ_kiloclaw_email_log_user_type_global')
       .on(table.user_id, table.email_type)
       .where(isNull(table.instance_id)),
-    uniqueIndex('UQ_kiloclaw_email_log_user_instance_type')
-      .on(table.user_id, table.instance_id, table.email_type)
+    uniqueIndex('UQ_kiloclaw_email_log_user_instance_type_period')
```
evanjacobson, line 4406: I was on the fence regarding keeping this in the business logic only. I'd be happy to remove this and do that instead.

evanjacobson:
#3049 #3050

PR #3045, opened 2026-05-05
fix(onboarding): gate dashboard escape hatch by auth state

Drixled:
## Summary

- Adds a `/get-started` dashboard escape hatch for signed-in users so they can skip the starting-point chooser and land on `/profile`.
- Hides that escape hatch for signed-out visitors, leaving only the existing `Already have an account? Sign in` path so the footer does not present two competing mini-actions.
- Adds focused end-to-end coverage for both auth states on `/get-started`. Architectural changes: none; this stays within the existing onboarding page component and its Playwright spec.

## Verification

- Not manually verified in a browser during this session; this was a small auth-state footer change and I relied on focused automated coverage while working.

## Visual Changes

| Before | After |
| ------ | ----- |
| _Signed-out `/get-started` footer showing both `Skip to dashboard` and `Sign in`._ | _Signed-out `/get-started` footer showing only `Sign in`._ |
| _Signed-in `/get-started` footer without a clear dashboard escape hatch._ | _Signed-in `/get-started` footer showing `Skip to dashboard`._ |

## Reviewer Notes

- Review both `/get-started` auth states: signed-out users should only see the sign-in prompt, while signed-in users should only see the dashboard escape hatch.
- `/profile` remains the escape hatch target to match the existing dashboard/profile landing route.


lambertjosh:
Hrm - I'm not seeing it when auth'd?

image

pandemicsyn (approved):

PR #3050, opened 2026-05-05
feat(emails): add top-up purchase confirmations

evanjacobson:
## Summary

Credit top-ups now send the transactional confirmation users expect after funds are added, closing a gap where successful top-ups were only reflected in-app.

This adds a Mailgun-backed credit top-up email with durable send markers so webhook retries can recover missed sends without double-emailing users. The implementation follows the existing server-rendered email template conventions and uses a generic transactional email log for idempotent purchase confirmations.

<details>

- Adds a new `creditsTopUp` transactional email template and sender.
- Adds `transactional_email_log` for idempotent non-KiloClaw purchase email markers keyed by email type and Stripe payment identity.
- Sends distinct manual top-up and auto top-up copy, including Stripe receipt links when available.
- Handles duplicate webhook/retry paths by using persisted credit transaction state to send once when the first attempt completed the top-up but missed the email side effect.
- Keeps top-up confirmation behavior disabled for `skipPostTopUpFreeStuff` flows, including retry recovery.

</details>

## Verification

- [x] automated tests
- [x] pnpm typecheck, drizzle check, format check

## Visual Changes

image

## Reviewer Notes

<details>

- Idempotency behavior around webhook retries and duplicate top-up settlements.
- The `transactional_email_log` schema/index change.
- Manual versus auto top-up copy and optional receipt-link rendering.
- `skipPostTopUpFreeStuff` handling on both first attempt and retry recovery paths.

</details>


apps/web/src/lib/user.ts
```diff
@@ -594,6 +594,7 @@ export class SoftDeletePreconditionError extends Error {
  * - referral_code_usages (financial, references anonymized user)
  * - kiloclaw_subscriptions, kiloclaw_earlybird_purchases, kiloclaw_email_log (retained records)
  * - kiloclaw_scheduled_action_targets (retained operational records;
+ * - transactional_email_log (retained outbox marker, financial record)
```
evanjacobson, line 597: The current state does not delete rows on GDPR delete request, per the standard set by kiloclaw_email_log. 

Please let me know if this should change.

packages/db/src/migrations/0113_black_george_stacy.sql
evanjacobson, line 1: My intention behind creating this table was to do the following:

Create a table that can be used for idempotency checks for non-KiloClaw emails
Allow the table to be used by future unknown, idempotent emails (or wherever we want to track emails sent)

apps/web/src/lib/credits.ts
```diff
@@ -111,10 +125,176 @@ export async function processTopUp(
       });
     }
 
+    await maybeSendTopUpConfirmationEmail({
```
jeanduplessis, line 153: This confirmation email is sequenced after `processFirstTopupBonus()` and the auto-top-up promo grant. If either unrelated bonus/promo step throws after the credit transaction has already committed, the `after()` callback stops before the confirmation marker/send runs, and Stripe may not retry because the webhook request itself already succeeded. That means a successfully credited top-up can still miss the confirmation email. Consider isolating the email send from bonus/promo failures (for example, catch/log those steps independently or run the confirmation in its own protected step).

apps/web/src/lib/credits.ts
```diff
@@ -111,10 +125,176 @@ export async function processTopUp(
...
+      return;
+    }
+
+    const receiptUrl = await resolveStripeReceiptUrl(stripeChargeOrInvoiceId);
+    const sendResult = await sendCreditsTopUpEmail({
+      to: user.google_user_email,
+      variant: isAutoTopUp ? 'auto' : 'manual',
+      amountCents: amountInCents,
+      creditsCents: amountInCents,
+      purchaseDate: new Date(),
```
jeanduplessis: On the retry-recovery path, this uses the retry/send time as the purchase date rather than the time the credit transaction/payment was originally created. For delayed webhook retries or manual Stripe replays, the email can show a date days after the user actually paid. Consider loading the committed `credit_transactions.created_at` (or using Stripe’s payment timestamp when available) for recovered sends.

jeanduplessis (approved):

apps/web/src/lib/credits.ts
evanjacobson: This finding was valid and addressed

apps/web/src/lib/credits.ts
evanjacobson: This finding was valid and addressed

PR #3051, opened 2026-05-05
feat(kiloclaw): named instance tiers + admin temporary size override

pandemicsyn:
### 1. Named instance-tier catalog

Replaces the admin UI's free-form `{cpus, memory_mb, cpu_kind}` resize with a named-tier catalog as the source of truth for provisioning and resizing. New shared package `@kilocode/kiloclaw-instance-tiers` consumed by both worker and web.

| Status | Key | Hardware |
|---|---|---|
| offered | `perf-1-3` (default) | 1× perf, 3 GB, 10 GB |
| offered | `perf-4-8` | 4× perf, 8 GB, 20 GB |
| offered | `perf-4-16` | 4× perf, 16 GB, 40 GB |
| legacy | `shared-2-3` | 2× shared, 3 GB, 10 GB |
| legacy | `shared-2-4` | 2× shared, 4 GB, 10 GB |

`resizeMachine(targetTierKey)` is upgrade-only and stopped-machine-only; downgrades, sidegrades, and legacy-as-target are rejected server-side. Upgrade policy lives in the catalog (`canUpgradeTo`) so future providers reuse it.

Per-provider:
- **Fly**: extends Fly Volume on storage growth, then persists tier. Volume-extend failure rolls back DO state. Alarm + live-check backfill legacy instances from `machine.config.guest`.
- **docker-local**: persists tier; container recreate via stop→resize→start picks up new `Memory`/`NanoCpus`. **No automatic backfill** for legacy docker-local instances (provider doesn't observe its own container; cure is re-provision or admin resize).
- **Northflank**: provisioning honors the tier via new `NF_DEPLOYMENT_PLAN_PERF_*` env vars (falling back to `NF_DEPLOYMENT_PLAN`). Resize rejected with "not yet supported."

DO state gains `instanceType` + `volumeSizeGb`; Postgres gains `kiloclaw_instances.instance_type` (nullable text, CHECK-constrained, partial index). Postgres syncs **only when DO state actually changes** — at provision insert, at resize, or at backfill that observes a previously-unknown shape. No alarm-tick churn.

Also fixes two stale bullets in `services/kiloclaw/AGENTS.md` that incorrectly claimed Next.js was the sole `kiloclaw_instances` writer (contradicts `.specs/kiloclaw-datamodel.md` rule 21 and the actual Worker insert path).

### 2. Admin temporary CPU/RAM override

Support occasionally needs to temporarily upgrade a customer's hardware to recover them (OOM, runaway process) and revert later. Tier resize is the wrong tool — it's upgrade-only by design and tier-coupled (rewrites billing-relevant fields). Adds a parallel admin-only override that wins over the tier-derived `machineSize` for runtime spec construction without touching `instanceType` or `volumeSizeGb`.

Key design points:

- **Parallel field, not mutation.** New persisted `adminMachineSizeOverride: MachineSize | null` + metadata. New `effectiveMachineSize(state) = override ?? machineSize` helper applied at every Fly-guest construction site. `state.machineSize` stays the billable tier hardware everywhere else (tier comparisons, customer dashboard, backfill).
- **Two presets, preset-agnostic DO.** `'perf-4-8'` / `'perf-4-16'` enforced at the tRPC + worker route boundary; the DO accepts any `MachineSize`. Adding a preset or relaxing to free-form is a tRPC-layer change.
- **Sticky until cleared.** Survives stop/start cycles. Tier resize auto-clears any active override (and reports `clearedOverride` in its response + audit log) so a tier change doesn't silently downgrade or free-upgrade the customer.
- **Customer-invisible.** `getStatus()` keeps returning the billable tier hardware. Override is admin-only by design — billing stays on the tier, support cost is the upgrade subsidy.
- **Backfill-aware.** `backfillMachineSizeFromFlyConfig` and `startExistingMachine`'s legacy branch both early-return when an override is active, so the live Fly guest (which reflects the override) doesn't get mistaken for tier hardware.
- **Apply-on-next-restart.** Set/clear can run on a stopped or running instance; the override applies on the next `startExistingMachine` (where `fly.updateMachine(guest=...)` enforces Fly's own stopped-state requirement). Northflank rejected for both.

### Out of scope

User-facing upgrade flow • per-tier billing/pricing • Northflank tier resize • volume downgrades • docker-local auto-backfill • override TTL/auto-expiry (additive) • outstanding-overrides admin dashboard view (the filter exists; a panel is a small follow-up) • alarm-driven volume-size reconciliation for `/extend-volume`.

## Verification

Manually verified end-to-end on docker-local and Fly: provision → resize `perf-1-3 → perf-4-8` → `docker inspect` confirms `Memory: 8 GiB`, `NanoCpus: 4` matching the catalog; provider-aware dialog Alerts render correctly. Set/clear override on a running Fly instance, restart, confirm Fly guest matches the override on the new machine.

## Visual Changes

Admin detail page: catalog-driven tier resize dialog, **Instance Tier** badge with optional `Override` chip, **Volume Size** row, new **Size Override** button + dialog (two radios + reason field, amber styling). When override is active: a banner above the Live Worker Status grid shows hardware + actor + reason.

Admin list page: per-row `Override` badge with tooltip; `Has size override` filter chip in the toolbar.

No user-facing surface changes.

## Reviewer Notes

- **Migration `0116_low_amazoness.sql`** adds both columns + indexes + CHECK; nullable, additive, no defaults.
- **AGENTS.md hard-invariant change** — two stale bullets corrected. Worth a separate eyeball.
- **`'custom'` is terminal.** Written by backfill when live Fly guest doesn't match any catalog entry; replaced by a real key on next resize. `resolveInstanceTypeFromState` self-heals the incoherent `custom + null machineSize` shape (read-side, doesn't write back).


apps/web/src/routers/admin-kiloclaw-instances-router.ts
```diff
@@ -2665,11 +2720,12 @@ export const adminKiloclawInstancesRouter = createTRPCRouter({
...
           .regex(/^vol_[a-zA-Z0-9]+$/, 'Invalid Fly volume ID'),
+        targetSizeGb: z.number().int().min(1).max(500),
       })
     )
     .mutation(async ({ input, ctx }) => {
       console.log(
-        `[admin-kiloclaw] extendVolume triggered by admin ${ctx.user.id} (${ctx.user.google_user_email}) app=${input.appName} volume=${input.volumeId} size=15GB`
+        `[admin-kiloclaw] extendVolume triggered by admin ${ctx.user.id} (${ctx.user.google_user_email}) app=${input.appName} volume=${input.volumeId} targetSizeGb=${input.targetSizeGb}`
       );
       const instance = await resolveInstance(input.userId, input.instanceId);
```
St0rmz1, line 2730: The other three admin hardware mutation routes added in this PR (`resizeMachine`, `setAdminMachineSizeOverride`, `clearAdminMachineSizeOverride`) all call `assertInstanceBelongsToUser(instance, input.userId)` immediately after `resolveInstance`. `extendVolume` was modified here (the new `targetSizeGb`) but did not pick up the same guard.

`resolveInstance(input.userId, input.instanceId)` calls `getInstanceById(instanceId)`, which queries only by id and does not filter by `user_id`. The platform route's `resolveInstanceDoKey` also ignores `userId` when `instanceId` is supplied. So an admin who passes `userId=A` together with `instanceId=B` where B belongs to user C ends up extending C's volume while the audit log row records `target_user_id=A`. The Fly app/volume consistency check on lines 2747 to 2752 passes as long as the supplied `appName` and `volumeId` match the resolved instance, which they do for C's instance. Fly volumes can grow but cannot be shrunk, so the storage change is permanent and attributed to the wrong user.

Suggestion: add `assertInstanceBelongsToUser(instance, input.userId);` immediately after the `resolveInstance` call. A regression test alongside the existing extendVolume tests would lock the behavior in so the guard does not silently regress in a future refactor.

services/kiloclaw/src/providers/northflank/index.ts
```diff
@@ -534,6 +575,11 @@ export const northflankProviderAdapter: InstanceProviderAdapter = {
     let providerState = getNorthflankProviderState(state);
     const projectId = providerState.projectId;
     const volumeName = providerState.volumeName ?? names.volumeName;
+    const deploymentPlan = resolveNorthflankDeploymentPlan(
```
St0rmz1, line 608: Storage sizing on Northflank did not pick up the same tier-from-state shape used here for the deployment plan. Both Northflank provisioning paths still hardcode `storageSizeMb: config.volumeSizeMb` (the global `NF_VOLUME_SIZE_MB`): `ensureProvisioningResources` at line 519 and `ensureStorage` at line 562. Meanwhile the DO's `provision()` now derives `state.volumeSizeGb` from the tier catalog (20 GB for `perf-4-8`, 40 GB for `perf-4-16`), and `getStatus()` returns that value to the customer dashboard.

Net effect for a freshly provisioned Northflank instance at any tier other than `perf-1-3`:
* Postgres `instance_type` and DO `state.volumeSizeGb` advertise the catalog tier (20 or 40 GB).
* Actual Northflank volume size is whatever `NF_VOLUME_SIZE_MB` resolves to (default 10240, so 10 GB).
* Customer fills past 10 GB and hits disk full errors with no signal anywhere that the persisted state diverged from reality.

The cleanest fix mirrors the deployment plan path: read `state.volumeSizeGb` (or look it up from the catalog when an `instanceType` is set) and pass that as `storageSizeMb` in both provisioning sites. If sizing per instance is not viable on Northflank because of storage class constraints, the alternative is to reject tiers other than the default on Northflank provisioning at the platform `/provision` route, matching how tier resize is already rejected on Northflank. Either way the persisted state should not promise more storage than the volume actually has.

A test in `providers/northflank/index.test.ts` parallel to the existing tier deployment plan coverage would seal the chosen behavior.

St0rmz1 (approved):

PR #3055, opened 2026-05-05
feat(gastown-container): add crash visibility + per-agent start mutex

jrf0110:
## Summary

Diagnostic changes for the investigation bead: town `4d82f099-ccb7-4eaf-8676-73562e0a27eb` is restarting its container every ~1.5–2 min for sustained periods. Root cause is not yet known — this PR adds the **observability and safety** we need to tell H1–H6 apart from production logs, plus fixes one concrete race that shows up in those logs.

**`main.ts` — crash visibility**
- New `unhandledRejection` listener logs `container.unhandled_rejection` with full message/stack, container uptime, and active-agent count. It does **not** call `process.exit`, so visibility is the only side effect. Bun/Node silently drop rejections without a handler, making fire-and-forget failures (`void saveDbSnapshot()`, `void subscribeToEvents()`, `setInterval(() => void sendHeartbeats())`) invisible today.
- Existing `uncaughtException` handler now also logs `name`, `uptimeMs`, and `activeAgents` alongside message/stack. Still fatal (`process.exit(1)`) — an exception escaping every try/catch is a genuine invariant break.
- `bootHydration()` is wrapped in try/catch at the call site so a rare synchronous throw before its first await doesn't crash the container.

**`main.ts` — OOM observability**
- 30s periodic `container.memory_usage` log (rss/heap/external/agents/uptime). Cadence matches the heartbeat. This is what catches H3 (memory leak + external SIGKILL) — those failures leave no stack behind.

**`process-manager.ts` — per-agentId startAgent mutex (fixes H6)**
- The `/agents/start` log line appears twice at the same millisecond for the same agentId in production logs. Both callers pass the re-entrancy check at the top of `startAgent` (because neither has committed a 'starting' record yet), then race on `startupAbortController`, `session.create()`, idle timers, `sdkInstance.sessionCount`, and the `agents` map — leaving leaked sessions and a confused lifecycle.
- Added `withStartAgentLock(agentId, fn)` (chained-promise mutex, same shape as the existing `sdkServerLock`) and wrapped the body of `startAgent` with it. The second concurrent caller now waits for the first to finish (or abort cleanly) before proceeding.
- Three unit tests cover: same-id serialisation, cross-id concurrency, and lock release on throw.

## Investigation plan (next step after this lands)

With this deployed to staging, pull 1–2 hours of logs for town `4d82f099-...` and classify:

| Hypothesis | Signal in logs after this PR |
| --- | --- |
| H1 (unhandled throw / rejection) | `container.unhandled_rejection` or `container.uncaught_exception` lines clustered right before each restart |
| H2 (corrupt kilo.db) | `${MANAGER_LOG} session.create failed … stale DB recovery` repeating |
| H3 (OOM / external kill) | `container.memory_usage` showing rssMB monotonically growing, followed by restart with NO preceding exception log |
| H4 (user-triggered stop/destroy) | Absence of any container-side crash log + tRPC `forceRestartContainer` / `destroyContainer` calls on the worker side |
| H5 (/refresh-token loop) | High cadence of `refresh_token.received` logs |
| H6 (concurrent /agents/start race) | Previously would manifest as leaked sessions / 'failed' agents. The mutex eliminates this class, so if restarts stop after deploy, H6 was the cause |

The follow-up bead/PR with the actual fix (once H1–H6 is narrowed down) is a separate deliverable — this PR is the instrumentation required to get there. The mutex is pre-emptive because the race it fixes is real and visible in the current logs regardless of whether it's the root cause of the restarts.

## Verification

- Container typecheck clean: `cd services/gastown/container && pnpm typecheck` ✅
- Container tests: `pnpm test` — 62 pass; 2 pre-existing JWT-mock failures in `plugin/client.test.ts` confirmed unrelated (reproduce on `main` without this patch).
- New tests in `process-manager.test.ts` directly exercise `withStartAgentLock` (same-agentId serialisation, different-agentId concurrency, lock release on throw).

## Visual Changes

N/A — backend-only instrumentation.

## Reviewer Notes

- The `unhandledRejection` handler intentionally does **not** exit. If production shows that a specific rejection leaves the process in a wedged state, we can escalate individually; visibility first.
- `withStartAgentLock` is `export`ed so the test file can exercise it without booting the full SDK harness. The only production caller is `startAgent` in the same module.
- `Promise.withResolvers` is Bun-native and already used in this codebase; no polyfill needed.
- No changes to `bootHydration`, `subscribeToEvents`, or `saveDbSnapshot` themselves — those audits can happen after we have the crash-class data from the new logs.

### DIFF
--- a/services/gastown/container/src/main.ts
+++ b/services/gastown/container/src/main.ts
@@ -2,7 +2,17 @@ import { startControlServer } from './control-server';
 import { log } from './logger';
 import { activeAgentCount, bootHydration, getUptime, listAgents } from './process-manager';
 
-log.info('container.cold_start', { uptime: getUptime(), ts: new Date().toISOString() });
+// Container-scoped identifiers for crash/diagnostic logs. The container is
+// pinned to a single town for its lifetime (see GASTOWN_TOWN_ID injection in
+// the deployer), so reading these once at module init is safe and lets us
+// emit them even when no agents are registered yet.
+const TOWN_ID = process.env.GASTOWN_TOWN_ID ?? null;
+
+log.info('container.cold_start', {
+  uptime: getUptime(),
+  ts: new Date().toISOString(),
+  townId: TOWN_ID,
+});
 
 // Bun (like Node) will ignore unhandled promise rejections unless a handler
 // is registered. Without this handler a rejection in a fire-and-forget path
@@ -20,6 +30,7 @@ process.on('unhandledRejection', (reason, promise) => {
       : { message: String(reason) };
   log.error('container.unhandled_rejection', {
     ...err,
+    townId: TOWN_ID,
     uptimeMs: getUptime(),
     activeAgents: activeAgentCount(),
     promise: String(promise),
@@ -31,6 +42,7 @@ process.on('uncaughtException', err => {
     message: err.message,
     stack: err.stack,
     name: err.name,
+    townId: TOWN_ID,
     uptimeMs: getUptime(),
     activeAgents: activeAgentCount(),
   });
@@ -57,6 +69,7 @@ setInterval(() => {
       heapUsedMB: Math.round(mem.heapUsed / 1024 / 1024),
       heapTotalMB: Math.round(mem.heapTotal / 1024 / 1024),
       externalMB: Math.round(mem.external / 1024 / 1024),
+      townId: TOWN_ID,
       uptimeMs: getUptime(),
       agents: listAgents().length,
       activeAgents: activeAgentCount(),
@@ -81,6 +94,7 @@ void (async () => {
     log.error('container.boot_hydration_failed', {
       message: err instanceof Error ? err.message : String(err),
       stack: err instanceof Error ? err.stack : undefined,
+      townId: TOWN_ID,
     });
   }
 })();
--- a/services/gastown/container/src/main.ts
+++ b/services/gastown/container/src/main.ts
@@ -1,11 +1,42 @@
 import { startControlServer } from './control-server';
 import { log } from './logger';
-import { bootHydration, getUptime } from './process-manager';
+import { activeAgentCount, bootHydration, getUptime, listAgents } from './process-manager';
 
 log.info('container.cold_start', { uptime: getUptime(), ts: new Date().toISOString() });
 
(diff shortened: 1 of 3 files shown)


services/gastown/container/src/process-manager.ts
```diff
@@ -71,6 +71,39 @@ export function isDraining(): boolean {
...
+// complete (or abort) before proceeding, and then observes a consistent
+// snapshot in `agents.get(agentId)`.
+const startAgentLocks = new Map<string, Promise<unknown>>();
+
+// Exported for tests that exercise the locking behaviour directly without
+// bringing up the whole SDK/process harness. Production callers should use
+// `startAgent` (which wraps `startAgentImpl` with this lock).
+export async function withStartAgentLock<T>(agentId: string, fn: () => Promise<T>): Promise<T> {
+  const previous = startAgentLocks.get(agentId) ?? Promise.resolve();
+  const deferred = Promise.withResolvers<void>();
```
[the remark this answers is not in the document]
jrf0110: Fixed in ff34e19c5. Replaced `Promise.withResolvers` with the explicit `new Promise` release pattern used by `sdkServerLock` above, so the mutex no longer depends on a newer Bun global during module import.

services/gastown/container/src/main.ts
```diff
@@ -1,20 +1,86 @@
...
+    ...err,
+    uptimeMs: getUptime(),
+    activeAgents: activeAgentCount(),
+    promise: String(promise),
+  });
+});
+
 process.on('uncaughtException', err => {
-  log.error('container.uncaught_exception', { error: err.message, stack: err.stack });
+  log.error('container.uncaught_exception', {
```
jrf0110, line 41: Is it possible to log town_id in these logs?

jrf0110: Done in cd2f2a99d. Added `townId` (read once from `process.env.GASTOWN_TOWN_ID` at module init since the container is pinned to a single town for its lifetime) to: container.cold_start, container.unhandled_rejection, container.uncaught_exception, container.memory_usage, and container.boot_hydration_failed. This means crash/diagnostic logs are correlatable with a town even when no agents are registered yet (e.g. pre-hydration crashes).

PR #3069, opened 2026-05-06
fix(bot): enforce linked account access checks

RSO:
## Summary
- Enforce platform integration ownership access before processing linked Kilo Bot requests.
- Unlink and re-prompt stale bot identities when the linked Kilo user no longer owns or belongs to the integration owner.
- Reuse the same ownership access helper during account-linking and add focused coverage for stale-link denial paths.

## Verification
N/A (backend-only bot authorization change; no manual UI flow exercised).

## Visual Changes
N/A

## Reviewer Notes
This intentionally treats Redis identity mappings as a cache. The PR no longer attempts GDPR or org-removal Redis scans; stale cache entries are invalidated opportunistically when encountered and cannot authorize requests without current integration access.


markijbema (commented):
LGTM 🤓 Absolutely chef's-kiss centralization of the ownership predicate into `canKiloUserAccessPlatformIntegration` — collapsing the `owned_by_organization_id` / `owned_by_user_id` / "neither" branching into a single authoritative call site is the kind of refactor that brings a tear to the eye of anyone who has ever debugged drift between two subtly-different copies of an auth check. The mention-handler change is the real star though: treating Redis identity mappings as a cache rather than a source of truth, and lazily invalidating on access via `unlinkKiloUser` → `promptLinkAccount` when the linked Kilo user no longer has integration access, is *exactly* the right invariant — capability checks gate on live ownership, the cache can never authorize beyond what the underlying ACL allows, and you sidestep the entire combinatorial nightmare of GDPR/org-removal Redis scans. The 403 + "Access Denied" assertion plus the stale-link branch coverage in `bot.test.ts` (with that delightfully cursed `globalThis.__botTestMentionHandler` capture trick to hoist the `onNewMention` callback out of the closure for direct invocation) means the denial paths are pinned down on both the linking and runtime-mention sides. Ship it. 🚀

markijbema (approved):

PR #3083, opened 2026-05-06
feat(kiloclaw) onboarding calendar connect step

St0rmz1:
## Summary

Add a Connect a calendar step to the KiloClaw onboarding wizard. New users see this between identity capture and channel selection. The step wraps the existing Google Calendar OAuth flow at /api/integrations/google/connect (read access only, no writes), with a Skip for now option that is always available.

To make the OAuth round trip land users back in the wizard rather than stranding them on Settings, this PR adds returnTo support to the Google OAuth state, connect, and callback routes. The signed state token validates returnTo against a strict relative path regex, blocks URI fragments, blocks path traversal segments (`.` / `..`), and caps length at 2048 characters. The callback honors returnTo on both success and error paths.

The Connect button gates on the kiloclaw instance row existing in the DB (a fast write that completes shortly after provisioning kicks off, distinct from the slower Fly machine readiness). While the row is missing, the button reads "Setting up your instance…" and is disabled. Skip for now stays enabled throughout, so the user is never blocked.

After the OAuth round trip, the wizard hydrates local bot identity state from the persisted instance row so the user resumes at the calendar step instead of restarting from identity. A watchdog effect cleans the URL and surfaces a soft warning if hydration cannot complete within 5 seconds.

PostHog tracking covers the full funnel:
- `claw_setup_calendar_viewed` (fires on render, matching the identity step pattern)
- `claw_setup_calendar_connect_clicked`
- `claw_setup_calendar_oauth_completed`
- `claw_setup_calendar_oauth_failed` (reason allowlisted; unknown codes bucketed as `unknown`)
- `claw_setup_calendar_resumed` (outcome: connected / error / unknown)
- `claw_setup_calendar_completed` (with `skipped` and `connected` flags)

## Verification

- [x] Provisioned a fresh kiloclaw instance through `/claw/new`. Identity step works as before. Calendar step renders with correct copy and the Recommended pill.
- [x] Clicked Connect Google Calendar within the first second of the calendar step rendering, before the kiloclaw row was visible. Button was disabled with "Setting up your instance…" copy. Once the row appeared, the button swapped to the active link.
- [x] Completed the Google OAuth consent flow with calendar readonly scope. Returned to `/claw/new?step=calendar&success=google_connected`. Wizard restored bot identity, set the step to calendar, showed a "Calendar connected" toast, and stripped the query params from the URL.
- [x] Cancelled the Google OAuth consent flow. Returned to `/claw/new?step=calendar&error=access_denied`. Wizard showed an error toast, fired `claw_setup_calendar_oauth_failed` with reason `'access_denied'`, and cleaned the URL.
- [x] Clicked Skip for now without connecting. Wizard advanced to channels.
- [x] Confirmed the existing Settings page Google connect flow still redirects to `/claw/settings` (returnTo unset, fallback path).
- [x] Confirmed crafted returnTo values are rejected before being baked into the signed state: `//evil.example.com`, `https://evil.example.com/foo`, `/claw/new#frag`, `/claw/../admin` all fall through to the no-returnTo fallback.

## Visual Changes

| Before | After |
| ------ | ----- |
|        |       |

## Reviewer Notes

- The Connect button gates on `flowState.instanceStatus !== null` (the DB row written by the provision API), not on Fly machine readiness. The OAuth state itself just needs to know which instance to associate the connection with; the gateway being up is only required later when calendar data is actually read.
- `instanceReady` is a required (non optional) prop on `CalendarConnectStepView` so any future caller that forgets to pass it fails typecheck. Same defense in depth pattern used for `assistantEmoji` previously.
- The watchdog effect runs alongside the resume effect with a 5 second deadline. If `botIdentity` hydration cannot complete from `flowState.instanceStatus`, the watchdog cleans the URL and shows a soft warning rather than leaving stale `?step=calendar` params lingering on the identity step.
- `buildGoogleRedirectPath` in the callback route requires a pre-encoded `key=value` fragment; `appendQueryParam` takes raw key plus value and encodes both. JSDoc on each function states the contract; the parameter name is `preEncodedQueryFragment` for emphasis.
- A separate race surfaced during testing where freshly provisioned instances default to always ask permissions even though the wizard sends never ask. Root cause is the controller boot reading env vars before the live config patch lands. Not in scope for this PR; tracked for a follow up.


apps/web/src/app/(app)/claw/components/ClawOnboardingFlow.tsx
```diff
@@ -295,9 +402,45 @@ function ClawOnboardingFlowInner({
             preset: DEFAULT_ONBOARDING_EXEC_PRESET,
             defaulted: true,
           });
-          posthog?.capture('claw_setup_channels_viewed');
           setBotIdentity(identity);
-          setOnboardingStep('channels');
+          setOnboardingStep('calendar');
```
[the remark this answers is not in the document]
St0rmz1: Implemented guards. We ensure the record exists first. Which should take 1 or 2 seconds tops.

apps/web/src/app/(app)/claw/components/ClawOnboardingFlow.state.ts
```diff
@@ -31,6 +39,7 @@ export const FAKE_ONBOARDING_STEP_PARAM = 'fakeOnboardingStep';
 
 export const CLAW_ONBOARDING_FAKE_STEPS = [
   'identity',
+  'calendar',
```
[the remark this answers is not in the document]
St0rmz1: Addressed with https://github.com/Kilo-Org/cloud/pull/3083/changes/2430f6bd9ddc10e6ab208547794d538d5095e2f2

apps/web/src/lib/integrations/google/oauth-state.ts
```diff
@@ -6,15 +6,44 @@ import { GoogleCapabilitySchema } from './capabilities';
...
 
+// Constrain returnTo to a relative path so it can never be hijacked into an
+// open-redirect to an external host. Must start with `/`, may contain a
+// non-protocol-style path, optionally followed by a query string. Fragments
+// are disallowed — buildGoogleRedirectPath / appendQueryParam append the
+// success/error param using a `?` or `&` separator, and a fragment in the
+// returnTo would push the appended param past the `#` where browsers ignore
+// it. Disallows `//` after the leading slash so we don't accidentally accept
+// protocol-relative URLs like `//evil.example.com`.
+const RETURN_TO_REGEX = /^\/(?!\/)[^?#]*(\?[^#]*)?$/;
```
jeanduplessis: **[CRITICAL]** `apps/web/src/lib/integrations/google/oauth-state.ts:17` - `returnTo` validation can still become an external redirect or normalized path traversal

**Warning:** The new `RETURN_TO_REGEX` and literal segment check still allow backslashes and percent-encoded dot segments. When the callback later passes the generated path to `new URL(..., APP_URL)`, values such as `/\evil.example.com/path` can normalize to an external host, and `/claw/%2e%2e/admin` normalizes outside the caller-written path.

**Trace:**
1. This line accepts any non-`?`/`#` path characters after a single leading slash; backslashes and encoded dot segments are not excluded.
2. I checked the callback construction in `apps/web/src/app/api/integrations/google/callback/route.ts`: success and failure paths are passed into `new URL(path, APP_URL)`. WHATWG URL normalization treats `/\evil.example.com/path` as `https://evil.example.com/path` and `/claw/%2e%2e/admin` as `/admin`.
3. Because `returnTo` is signed into OAuth state after this validation, a crafted connect URL can survive the round trip and redirect somewhere other than the relative path the caller wrote.

**Impact:** An authenticated user with an active instance can initiate Google OAuth with a crafted `returnTo` that becomes an external redirect after callback URL normalization. Encoded traversal variants also bypass the intended path anchoring.

apps/web/src/app/(app)/claw/components/ClawOnboardingFlow.state.ts
```diff
@@ -280,6 +289,13 @@ function getRenderStepDecision({
     };
   }
 
+  if (onboardingStep === 'calendar') {
```
jeanduplessis, line 323: **[WARNING]** `apps/web/src/app/(app)/claw/components/ClawOnboardingFlow.state.ts:292` - Calendar resume is ignored after the OAuth full-page reload when the flow remounts in `post-provisioning` mode

**Warning:** This new calendar branch sits after the `mode === 'post-provisioning'` early return, so `setOnboardingStep('calendar')` cannot make the wizard render the calendar step once the full-page OAuth round trip remounts the flow as post-provisioning.

**Trace:**
1. The PR adds a `calendar` onboarding step and a resume effect that calls `setOnboardingStep('calendar')` after returning to `/claw/new?step=calendar`.
2. I checked the surrounding state machine: `getRenderStepDecision` returns `complete` or `provisioning` for `mode === 'post-provisioning'` before reaching this branch. I also checked the `/claw/new` callers: after a full OAuth reload, local `createFlowStartedAt` is lost; if billing/status now sees the active instance, the flow mounts with `mode="post-provisioning"` (org flow has the same local-state reset).
3. In that real return-from-OAuth state, the new `calendar` branch is unreachable, so the resume query is consumed but the rendered step remains provisioning/complete.

**Impact:** Users who complete or cancel Google OAuth after the instance/subscription is visible can be sent to provisioning/complete, and running instances can auto-redirect to chat instead of resuming at the calendar step with the connected/error state.

apps/web/src/app/(app)/claw/components/ClawOnboardingFakeWalkthrough.tsx
```diff
@@ -208,7 +211,21 @@ function renderFakeStep({
...
+      return <BotIdentityStep {...stepProgress} onContinue={() => setStep('calendar')} />;
+    }
+    case 'calendar': {
+      return (
+        <CalendarConnectStepView
+          {...stepProgress}
+          connectUrl="#"
+          isConnected={false}
+          connectedAccountEmail={null}
+          instanceReady
```
jeanduplessis: **[WARNING]** `apps/web/src/app/(app)/claw/components/ClawOnboardingFakeWalkthrough.tsx:223` - Fake walkthrough passes an undefined `instanceReady` identifier

**Warning:** JSX shorthand `instanceReady` requires an in-scope variable, but `renderFakeStep` does not define or receive one.

**Trace:**
1. The PR makes `instanceReady` a required prop on `CalendarConnectStepView` and passes it here using shorthand.
2. I checked the surrounding `renderFakeStep` function and `RenderFakeStepInput`; neither includes an `instanceReady` binding.
3. This changed line therefore references an undefined identifier rather than passing a boolean literal or prop.

**Impact:** The dev-only fake walkthrough path fails to typecheck/compile when rendering the new calendar fake step.

jeanduplessis (changes requested):
Requesting changes based on the inline findings: returnTo normalization/open-redirect bypass, calendar resume being unreachable after OAuth remounts in post-provisioning mode, and the fake walkthrough undefined instanceReady reference.

apps/web/src/app/(app)/claw/components/ClawOnboardingFlow.state.ts
St0rmz1: Addressed in `7bf6ec92f` (post-prov resume) and `1b74a052c` (provisioning-step follow-up). Three layered changes:

1. **`getRenderStepDecision` post-prov branch now honors explicit wizard steps.** Inside the `mode === 'post-provisioning'` block, before falling through to the default `complete`/`provisioning` returns, we check `onboardingStep` for each in-flight wizard step (`calendar`, `channels`, `provisioning`, `pairing`) and return that render step. So once the resume effect sets `onboardingStep`, the state machine respects it even though mode is post-provisioning. Tests cover all four cases.

2. **Lazy-init `onboardingStep` from `?step=` on mount in `ClawOnboardingFlow.tsx`.** Without this, first render uses `'identity'`, the state machine returns `'complete'`, and PR-1's auto-redirect to `/chat` fires before the resume effect's `setOnboardingStep` can take effect. The lazy initializer reads `window.location.search` so first render already reflects the resume.

3. **Follow-up fix in `renderProvisioningStep`.** The original post-prov branch returned a static `ProvisioningStepView` with no `onComplete` callback. Fine for the original returning-user case, but routing wizard-resume traffic through it left the user stuck on the spinner with no path to pairing/done. Now the static view renders only when `mode === 'post-provisioning' && onboardingStep !== 'provisioning'`. Wizard-resume traffic flows through the full `ProvisioningStep` so its `onComplete` fires.

apps/web/src/app/(app)/claw/components/ClawOnboardingFlow.state.ts
```diff
@@ -245,6 +254,37 @@ function getRenderStepDecision({
...
+        reason: 'calendar resume requested; honor it even in post-provisioning mode',
+      };
+    }
+    if (onboardingStep === 'channels') {
+      return {
+        renderStep: 'channels',
+        reason: 'wizard resume on channels; honor it even in post-provisioning mode',
+      };
+    }
+    if (onboardingStep === 'provisioning') {
```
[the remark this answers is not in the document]
St0rmz1: Addressed in `1b74a052c`.

`renderProvisioningStep` previously short-circuited to the static `ProvisioningStepView` whenever `mode === 'post-provisioning'`. That was correct for the original case (returning user lands on `/claw/new` with a running instance — `onboardingStep` is still `'identity'` at mount, the static view renders briefly, the state machine flips to `'complete'`, PR-1's auto-redirect carries them to `/chat`).

But the recent post-prov resume fix routes wizard-resume traffic through the same branch when the user advances calendar → channels → provisioning after an OAuth round-trip. The static view has no `onComplete`, so the user got stuck on the spinner with no path to pairing/done.

The fix narrows the static-view condition:

```ts
if (mode === 'post-provisioning' && onboardingStep !== 'provisioning')
  return <ProvisioningStepView ... />;
return <ProvisioningStep ... onComplete={...} />;
```

When the resume reaches the provisioning step explicitly, it falls through to the full `ProvisioningStep`. Its `onComplete` fires, sets `onboardingStep` to `'pairing'` or `'done'`, and PR-1's auto-redirect takes over from there. The original returning-user case is unchanged because that path keeps `onboardingStep === 'identity'`.

jeanduplessis (approved):

PR #3084, opened 2026-05-07
feat(email): add org top-up confirmations

evanjacobson:
## Summary

Organization credit top-ups now send the transactional confirmation billing owners expect after team funds are added, closing the gap where successful org top-ups were only reflected in-app.

This extends the existing credit top-up email flow to organization purchases, including distinct manual and auto top-up copy, organization payment-detail links, Stripe receipt links when available, and durable send markers so webhook retries can recover missed sends without double-emailing recipients. The implementation follows the existing server-rendered email template conventions and reuses the transactional email log idempotency pattern already used for purchase confirmations.

<details>

- Adds organization-specific manual and auto variants to the existing credit top-up transactional email.
- Sends confirmations to active organization owners and billing managers only.
- Tracks organization top-up sends with `transactional_email_log.organization_id` so retry paths can remain idempotent.
- Recovers missed org top-up emails from duplicate webhook/payment retry paths when the credit transaction already exists.
- Preserves terminal NeverBounce handling while allowing retryable all-recipient failures to be retried later.
- Passes auto top-up context through both initial org auto top-up setup payments and recurring `invoice.paid` org auto top-ups.

</details>

## Verification

- [x]  test coverage

## Visual Changes

N/A

## Reviewer Notes

<details>

- Idempotency behavior around duplicate Stripe events and missed email recovery.
- Recipient selection for active org owners and billing managers.
- Manual versus auto organization top-up copy and organization payment-detail links.
- Failure handling when some or all email recipients are rejected or fail to send.
- The `transactional_email_log.organization_id` schema/index addition and GDPR retention note.

</details>

<details><summary>A few more details </summary>
<p>

- We are not falling into the unrelated post-commit failure pitfall from Jean’s first comment. In this PR, processTopupForOrganization runs the test-only promo grant before scheduling the email, but that branch is gated to NODE_ENV === 'test' at apps/web/src/lib/organizations/organization-billing.ts:345. In production there is no unrelated bonus/promo step ahead of scheduleOrganizationTopUpConfirmationEmail at apps/web/src/lib/organizations/organization-billing.ts:355.

- We are not falling into the retry-date pitfall from Jean’s second comment. Duplicate payment recovery loads the existing credit_transactions.created_at in getExistingOrganizationTopUpCreditTransaction and passes it as purchaseDate at apps/web/src/lib/organizations/organization-billing.ts:334-340. The email sender uses purchaseDate ?? new Date() at apps/web/src/lib/organizations/organization-billing.ts:165, so recovery uses the original transaction timestamp.

- We do share the known insert-before-send marker limitations described in PR #3050. This PR inserts a marker before sending at apps/web/src/lib/organizations/organization-billing.ts:141-149. That means a crash after marker insert but before provider acceptance can suppress a future retry. The code does mitigate multi-recipient failures by clearing the marker if all sends fail retryably at apps/web/src/lib/organizations/organization-billing.ts:220-222, but it is still not a durable outbox. That appears consistent with the explicit PR #3050 decision to keep this pattern uniform rather than solve it locally.

</p>
</details>


packages/db/src/migrations/0116_past_marauders.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE "transactional_email_log" ALTER COLUMN "user_id" DROP NOT NULL;--> statement-breakpoint
+ALTER TABLE "transactional_email_log" ADD COLUMN "organization_id" uuid;--> statement-breakpoint
+ALTER TABLE "transactional_email_log" ADD CONSTRAINT "transactional_email_log_organization_id_organizations_id_fk" FOREIGN KEY ("organization_id") REFERENCES "public"."organizations"("id") ON DELETE no action ON UPDATE no action;--> statement-breakpoint
+CREATE INDEX "IDX_transactional_email_log_organization_id" ON "transactional_email_log" USING btree ("organization_id");
```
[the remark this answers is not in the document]
evanjacobson: Thanks for checking this. We intentionally kept the generated index plain/non-partial here: this table currently has fewer than 300 rows, so the lock window is negligible, and keeping the Drizzle-generated migration as-is avoids hand-editing generated migration SQL. The partial predicate is not needed for the current lookup pattern or table size.

apps/web/src/lib/organizations/organization-billing.ts
```diff
@@ -73,13 +77,188 @@ export async function getOrCreateStripeCustomerIdForOrganization(
...
+}): Promise<void> {
+  const {
+    userId,
+    organization,
+    amountInCents,
+    stripeChargeOrInvoiceId,
+    isAutoTopUp,
+    purchaseDate,
+  } = params;
+  const recipients = await getTopUpConfirmationRecipientMembersForOrganization(organization.id);
```
jeanduplessis: **[WARNING] Recipient lookup failures escape the best-effort email path**

**Trace:**
1. The new org confirmation sender fetches recipients before entering its `try` block.
2. The surrounding error handling starts later, at the marker insert/send block, so failures from `getOrganizationMembers()`/DB reads in `getTopUpConfirmationRecipientMembersForOrganization()` are not caught or reported through the function’s `captureException` path.
3. `scheduleOrganizationTopUpConfirmationEmail()` awaits this function directly in automated-test mode and schedules it via `after()` otherwise, so this pre-try rejection is not handled consistently with the rest of the best-effort email flow.

**Impact:** A transient DB/read failure while loading org members can make the email task reject before the marker/idempotency logic runs. In awaited paths this can surface as a failed top-up call after the credit transaction has already committed; in production `after()` paths it can become an unhandled/background failure without the intended Sentry context.

apps/web/src/lib/organizations/organization-billing.ts
```diff
@@ -73,13 +77,188 @@ export async function getOrCreateStripeCustomerIdForOrganization(
...
+
+  if (recipients.length === 0) {
```
jeanduplessis: **[WARNING] No-recipient skips are not marked terminal, so retries can email future recipients for old purchases**

**Trace:**
1. When an org top-up has zero eligible owners/billing managers, the function returns before inserting a `transactional_email_log` marker.
2. The duplicate-payment path later treats an existing credit transaction with no marker as a missed email and calls `recoverOrganizationTopUpConfirmationEmailIfMissing()` / `scheduleOrganizationTopUpConfirmationEmail()`.
3. Because recipients are recomputed at retry time, adding an owner or billing manager after the original no-recipient skip lets a duplicate webhook/manual replay send the old top-up confirmation to someone who was not an eligible recipient at purchase time.

**Impact:** Organizations that temporarily have no eligible recipients can later leak historical top-up details/Stripe receipt links to newly added billing contacts on webhook replay or duplicate processing. If “no eligible recipients” is terminal for that purchase, it should still write a marker or otherwise distinguish “skipped intentionally” from “missed send.”

jeanduplessis (changes requested):

apps/web/src/lib/organizations/organization-billing.ts
evanjacobson: Recipient lookup now runs inside the same try block as marker creation and sending, so  failures are captured with org/user/payment context instead of escaping the best-effort email flow. 

If lookup fails (missing/unknown) after we insert the marker, we delete the marker to allow a future retry; if cleanup fails, we leave it in place to prefer missing one email over duplicate sends.

To avoid confusion:
- If lookup succeeds and returns zero recipients, then the marker stays (no future sends).
- If lookup fails before we know recipients, the marker is deleted and treated as missed/unknown, and a future retry may send to whatever eligible recipients exist then.

apps/web/src/lib/organizations/organization-billing.ts
evanjacobson: With the most recent changes, the function now inserts the transactional_email_log marker before checking recipients and keeps it when the recipient list is empty. Duplicate webhook/recovery paths now see the marker and won’t email users added later. Added regression coverage for “no recipients first, recipient added later.”

apps/web/src/lib/organizations/organization-billing.ts
evanjacobson: lookup failures escaping without the intended Sentry context is also addressed, and the error handling has been improved (more scopes)

jeanduplessis (approved):

PR #3087, opened 2026-05-07
fix(github): use integration app type for check run updates

alex-alecu:
## Summary
- Recreates the stale Seer fix from #983 against the current monorepo layout.

Before this PR:

- A code review integration could be installed with a specific GitHub App type, e.g. `standard` or another stored app type.
- Kilo would create/check/manage reviews using that integration.
- But later, when updating or cancelling the GitHub check run, some code paths accidentally defaulted back to the `standard` GitHub App.
- If the check run belonged to a different app type, GitHub would reject the update.
- That could leave the PR check stuck as queued/running, cancelled incorrectly, or not cleaned up.

> Kilo might finish or cancel a code review, but GitHub could still show the Kilo check as stuck because Kilo tried to update it with the wrong app credentials.

This PR fixes that by always passing the integration’s stored `github_app_type` into check-run updates and cleanup calls.

So the important outcome is:

- GitHub PR checks update correctly.
- Cancelled/retried reviews clean up their check runs correctly.
- Stale or orphaned check runs are less likely to block merges.
- Existing standard-app integrations still work as before.

### DIFF
--- a/apps/web/src/app/api/internal/code-review-status/[reviewId]/route.test.ts
+++ b/apps/web/src/app/api/internal/code-review-status/[reviewId]/route.test.ts
@@ -2,7 +2,7 @@ import { describe, expect, it, jest, beforeEach } from '@jest/globals';
 import type { NextRequest } from 'next/server';
 import type * as codeReviewsDbModule from '@/lib/code-reviews/db/code-reviews';
 import type * as platformIntegrationsModule from '@/lib/integrations/db/platform-integrations';
-import type { CloudAgentCodeReview } from '@kilocode/db/schema';
+import type { CloudAgentCodeReview, PlatformIntegration } from '@kilocode/db/schema';
 
 // --- Mock functions ---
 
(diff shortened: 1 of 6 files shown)


alex-alecu:
Manual test passed.

Tested:
- Retriggered an existing GitHub code review through the authenticated tRPC flow, creating a fresh `Kilo Code Review` check run with the stored standard GitHub app integration.
- Sent internal status callbacks to move the check run to `in_progress` and then `completed/cancelled`.
- Cancelled a pending GitHub review through `codeReviews.cancel` and verified check-run cleanup.
- Verified an app-type mismatch uses the stored app type by confirming the gate update fails before terminal DB persistence, then succeeds after restoring the stored app type.

Verified:
- GitHub check-run updates reached the expected queued, in-progress, and completed/cancelled states.
- The `cloud_agent_code_reviews` row persisted the matching check-run ID, session IDs, terminal status, and terminal reason.
- Invalid internal callback authentication returned 401 and did not change persisted state.

jeanduplessis (approved):

PR #3090, opened 2026-05-07
feat(linear): add Linear platform integration

RSO:
## Summary

Adds Linear as a new platform integration so workspaces can install Kilo as an app actor on a Linear workspace and have agent sessions triggered from issue mentions.

- New OAuth install/callback flow at `/api/integrations/linear/callback` using the `@chat-adapter/linear` adapter, with `actor=app` for agent-session installs.
- New tRPC router (`linearRouter`) with `getInstallation`, `getOAuthUrl`, `uninstallApp`, `updateModel`, plus a dev-only `devRemoveDbRowOnly`.
- New service layer (`linear-service.ts`) handling install upsert, ownership conflict detection (`LinearWorkspaceAlreadyConnectedError`), model selection (matches Slack defaults / org allow-list), and uninstall.
- Webhook handler for inbound Linear events including `OAuthApp.revoked`, which cleans up local state when a workspace admin uninstalls from Linear's side.
- New Linear integration page and details component (`LinearIntegrationDetails.tsx`) for both user-owned and org-owned installs, surfaced in the integrations hub.
- Schema: `PLATFORM.LINEAR` constant and a uniqueness constraint on `(platform=linear, platform_installation_id)` so the same workspace can't be connected to two different Kilo accounts.
- `fix(linear): revoke OAuth token on disconnect` — disconnect now POSTs to Linear's `/oauth/revoke` so reinstalling shows the consent screen instead of "already installed", matching how Slack disconnect already works.

## Verification

- [x] Connect Linear from a user account, then from an org — verify the workspace name + scopes show in the integration page.
- [x] Attempt to connect the same Linear workspace from a second Kilo account: shows the "already connected" error.
- [x] @-mention the Kilo app actor on a Linear issue and confirm an agent session starts and posts back.
- [x] Disconnect Linear from the integration page, then click reinstall: the Linear consent screen appears (not "Kilo Dev already installed").
- [x] Uninstall the OAuth app from Linear's settings UI: local `platform_integrations` row and chat-sdk installation are cleaned up via the webhook.
- [x] Switch the model in the Linear integration details and confirm a subsequent agent session uses it.

## Visual Changes

CleanShot 2026-05-07 at 18 22 39@2x

## Reviewer Notes

- Disconnect order matters: the access token is read from the chat-sdk adapter's Redis state *before* `deleteInstallation` wipes it, then `revokeLinearToken` is called, then the local cleanup proceeds. Revoke errors are swallowed so disconnect always succeeds locally even if Linear is unreachable.
- The webhook-driven `OAuthApp.revoked` path intentionally does not call revoke — Linear has already revoked at that point.
- `devRemoveDbRowOnly` is gated to `NODE_ENV === 'development'` and exists to re-test the OAuth flow without reinstalling on Linear's side.
- Migration `0116_classy_grey_gargoyle.sql` adds the workspace uniqueness constraint; safe to apply since no existing Linear rows.


apps/web/src/app/api/integrations/linear/callback/route.ts
```diff
@@ -0,0 +1,248 @@
...
+ *
+ * Falls back to the organizationId on failure so the install still succeeds
+ * with a human-readable label.
+ */
+async function fetchLinearWorkspaceName(
+  accessToken: string,
+  organizationId: string
+): Promise<string> {
+  try {
+    const response = await fetch(LINEAR_GRAPHQL_URL, {
```
RSO: Should we use the Linear SDK?

apps/web/src/lib/integrations/linear-service.ts
```diff
@@ -0,0 +1,461 @@
...
+ * unreachable or the token is already invalid.
+ */
+export async function revokeLinearToken(accessToken: string): Promise<boolean> {
+  try {
+    const body = new URLSearchParams({
+      token: accessToken,
+      token_type_hint: 'access_token',
+    });
+
+    const response = await fetch(LINEAR_REVOKE_URL, {
```
RSO, line 63: Can this be done through the Linear SDK as well?

apps/web/src/routers/linear-router.ts
```diff
@@ -0,0 +1,153 @@
...
+      ? `org_${input.organizationId}`
+      : `user_${ctx.user.id}`;
+    const state = createOAuthState(statePrefix, ctx.user.id);
+    return { url: linearService.getLinearOAuthUrl(state) };
+  }),
+
+  uninstallApp: baseProcedure.input(optionalOrgInput).mutation(async ({ ctx, input }) => {
+    if (input?.organizationId) {
+      await ensureOrganizationAccess(ctx, input.organizationId);
+      await requireActiveSubscriptionOrTrial(input.organizationId);
```
jrf0110: Perhaps we shouldn't require the active sub to uninstall?

apps/web/src/lib/bot/linear-link-state.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    const ageSeconds = Math.floor(Date.now() / 1000) - data.iat;
+    if (ageSeconds < 0 || ageSeconds > STATE_TTL_SECONDS) return null;
+
+    return {
+      userId: data.userId,
+      platformIntegrationId: data.platformIntegrationId,
+      organizationId: data.organizationId,
+      callbackPath: data.callbackPath,
+    };
```
jrf0110: Probably better to use zod for this parsing/validation

apps/web/src/lib/bot/platforms/linear-webhook.ts
```diff
@@ -0,0 +1,145 @@
...
+  );
+}
+
+async function handleLinearOAuthAppRevoked(
+  organizationId: string,
+  chat: Chat,
+  linearAdapter: LinearAdapter
+): Promise<void> {
+  try {
+    await deleteInstallationByOrganizationId(organizationId);
```
jrf0110: Perhaps move this delete to be after the linearAdapter.deleteInstallation call in case the upstream delete fails

jrf0110 (approved):

PR #3102, opened 2026-05-07
fix(code-review): retry sandbox prep

alex-alecu:
## Why

Some code reviews can fail when the sandbox has a short server problem while getting ready. A retry can make those reviews start instead of failing right away.

## What changed

Code reviews now try one more time when setup fails because the sandbox returned a server error. The retry uses a fresh session and remembers that it already retried, so it cannot loop forever. Billing problems, cancelled reviews, and clear setup mistakes still fail without a retry.

## How to test

1. Run `pnpm --filter @kilocode/worker-utils test -- src/cloud-agent-next-client.test.ts`.
2. Run `pnpm --filter kilo-code-review-worker test`.
3. Run `pnpm --filter cloud-agent-next test -- src/session-prepare.test.ts src/sandbox-recovery.test.ts`.
4. Run `pnpm --filter kilo-code-review-worker typecheck`.

### DIFF
--- a/packages/worker-utils/src/cloud-agent-next-client.test.ts
+++ b/packages/worker-utils/src/cloud-agent-next-client.test.ts
@@ -98,6 +98,7 @@ describe('CloudAgentNextFetchClient billing error detection', () => {
       expect(error).toBeInstanceOf(CloudAgentNextBillingError);
       const billingError = error as CloudAgentNextBillingError;
       expect(billingError.terminalReason).toBe('billing');
+      expect(billingError.procedure).toBe('prepareSession');
       expect(billingError.status).toBe(402);
     }
   });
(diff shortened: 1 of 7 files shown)


alex-alecu:
Manual test passed.

Tested:
- Started the code-review worker locally with stubbed internal API and cloud-agent-next services.
- Sent a v2 code review request where the first prepare call returned a retryable sandbox 500 marker and the second prepare call succeeded.
- Sent a v2 code review request where prepare returned a deterministic 400 setup error.

Verified:
- The sandbox 500 flow called prepare twice, initiated once with the retry session, and returned running status with the retry session IDs.
- The internal status callback received running updates and the retry session ID update.
- The deterministic 400 flow called prepare once, did not initiate, and ended failed without a retry log.

catrielmuller (approved):

PR #3106, opened 2026-05-07
fix(bot): make botRequestId required end-to-end so callbacks are never silently dropped

RSO:
## Summary

`createBotRequest` previously swallowed DB errors and returned `undefined`, which propagated through `processLinkedMessage` → `processMessage` → `runBotAgent` → `spawnCloudAgentSession`. A falsy `botRequestId` caused `callbackTarget` to become `undefined`, so Cloud Agent sessions started **without a callback** and the bot never posted results back. A missing `INTERNAL_API_SECRET` had the same silent-drop failure mode.

- `createBotRequest` now returns `Promise<string>` and throws on failure (no more swallow-and-return-`undefined`).
- `processLinkedMessage` captures the failure to Sentry with rich context (`platform`, `platformIntegrationId`, `userId`, `threadId`, `messageId`) and posts a user-visible error message before returning.
- `botRequestId` is tightened to `string` in `processMessage`, `RunBotAgentParams`, and `spawnCloudAgentSession`. All `if (params.botRequestId)` guards are removed; only the `resolvedCloudAgentSessionId` check remains where it is genuinely needed.
- `spawnCloudAgentSession` now explicitly throws (with Sentry capture) when `INTERNAL_API_SECRET` is missing, and builds `callbackTarget` unconditionally.

## Verification

- Triggered a local bot mention and confirmed the `bot_requests` row is created and `callbackTarget` is populated on the spawned session.
- Manually simulated a `createBotRequest` failure by pointing at an invalid user — the user-visible error message posts and Sentry receives the capture with the expected context.

## Visual Changes

N/A — no UI changes.

## Reviewer Notes

- The previous swallow-and-continue behaviour was hiding the production bug being investigated. After this change, previously-hidden DB failures will surface as user-visible error messages and Sentry events. That is the intended outcome.
- `link-account/route.ts` (`reprocessLinkedMessage`) already has its own try/catch + Sentry capture, so the new throw path is covered there. No user message is possible there (the linking tab has closed) — acceptable.
- The `bot-session-callback` route already passes `botRequestId: string`, so the type tightening is a no-op at that call site.
- Discord bot has its own separate `spawnCloudAgentSession` in `lib/discord-bot.ts` — out of scope.
- Added a test that the public `createBotRequest` throws on FK violation (`src/lib/bot/request-logging.test.ts`). The full `pnpm typecheck` was skipped per `AGENTS.md`; `pnpm --filter web typecheck` covers the changed surface.

### DIFF
--- a/apps/web/src/lib/bot/agent-runner.ts
+++ b/apps/web/src/lib/bot/agent-runner.ts
@@ -55,7 +55,7 @@ type RunBotAgentParams = {
   rawMessage?: Message;
   platformIntegration: PlatformIntegration;
   user: User;
-  botRequestId: string | undefined;
+  botRequestId: string;
   prompt: string;
   /** Pre-uploaded image attachments from the user's message (already in R2). */
   images?: Images;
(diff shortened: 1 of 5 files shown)


eshurakov (approved):
LGTM

PR #3117, opened 2026-05-07
docs(kiloclaw) tools systemd clarification

St0rmz1:
## Summary

Agents running inside KiloClaw containers were suggesting fixes that rely on systemd, such as `systemctl restart`, `journalctl`, and unit files. None of those work in this environment because:

1. systemd is never PID 1. In local Docker the controller daemon is PID 1. On Fly Machines `/fly/init` is PID 1. Other providers may inject their own init shim.
2. The `systemctl` and `systemd` binaries do exist on disk because apt pulls them in as transitive dependencies (chromium, dbus, and others), so `which systemctl` finds the binary and tricks agents into thinking systemd is available.
3. The systemd daemon is never running. `systemctl is-system-running` returns `offline`, and any call against a service fails with the standard "System has not been booted with systemd as init system" error.
4. `openclaw`, the gateway, and other background processes are managed by the controller's own supervisor module, not by init.

This change documents the above explicitly inside `services/kiloclaw/container/TOOLS.md` so the agent inside the container stops giving bad advice. Existing instances pick up the new section automatically on redeploy through the bounded section append already used for kilo cli, mitigations, and plugin install context. The KiloClaw changelog page gets a new bugfix entry dated May 7 2026 with `deployHint: upgrade_required`.

## Verification

- [x] Built the kiloclaw image locally and started a fresh container. Confirmed the workspace `/root/.openclaw/workspace/TOOLS.md` was seeded with the new `<!-- BEGIN:process-model -->` section.
- [x] Repeated against an already provisioned container with an older workspace TOOLS.md. Bootstrap log shows `TOOLS.md: added Process Model section` and the file size grew to include the new content. Idempotent on subsequent boots.
- [x] Inside the container: `cat /proc/1/comm` reports `MainThread` (Node), `ps -p 1 -o args=` reports `node /usr/local/bin/kiloclaw-controller.js`, `which systemctl` reports `/usr/bin/systemctl`, and `systemctl is-system-running` reports `offline`. All match the claims in the new section.
- [x] Repeated against a real Fly Machine. PID 1 is `/fly/init`, `systemctl is-system-running` is `offline`, no KiloClaw `.service` files exist. Matches the Fly entry in the new section.
- [x] `pnpm exec vitest run controller/src/bootstrap.test.ts` passes 98 tests, including the new smoke test that pins the load bearing directives in `PROCESS_MODEL_SECTION_CONFIG`.
- [x] `pnpm run format:check`, `pnpm --filter ./services/kiloclaw lint`, `pnpm --filter ./apps/web lint`, `pnpm --filter ./services/kiloclaw typecheck`, and `pnpm --filter ./apps/web typecheck` all clean.

## Visual Changes

The only change visible to users is a new bugfix row at the top of the KiloClaw changelog page (`/claw/changelog`).

| Before | After |
| ------ | ----- |
| Top entry was the May 5 2026 Auto models bugfix. | Top entry is the May 7 2026 systemd clarification bugfix with deploy hint `upgrade_required`. |

## Reviewer Notes

- The new section is implemented as a bounded TOOLS.md section so existing instances pick it up on redeploy. This mirrors the pattern used by `KILO_CLI_SECTION_CONFIG`, `KILOCLAW_MITIGATIONS_SECTION_CONFIG`, and `PLUGIN_INSTALL_SECTION_CONFIG`.
- The smoke test in `bootstrap.test.ts` pins the substantive directives (the lede about systemd not being the init system, `/fly/init`, `tini`, the ban on `systemctl` and `journalctl`, the reference to the controller's own supervisor). A drive by edit that strips the substance while keeping the markers will fail the test.
- The compiled openclaw `doctor` output continues to recommend "install/enable systemd or run the gateway under your supervisor" when the gateway is not running. The new TOOLS.md section gives the agent the right context to interpret that message as a known false signal rather than an action item.


services/kiloclaw/container/TOOLS.md
```diff
@@ -36,3 +35,19 @@ The Kilo CLI (`kilo`) is an agentic coding assistant for the terminal, pre-confi
...
+<!-- BEGIN:process-model -->
+
+## Process Model
+
+KiloClaw does NOT use systemd as its init system, in any deployment.
+
+- **PID 1 varies by deployment**, but is never systemd:
+  - Local dev (plain `docker run`): `node /usr/local/bin/kiloclaw-controller.js` — the controller daemon runs as PID 1 directly.
+  - Fly Machines (production today): `/fly/init` — Fly's own init shim, which then execs the controller as a child.
+  - Other providers: whatever init shim that platform injects (e.g. `tini`, `dumb-init`, or a provider-specific equivalent). The controller still runs as a child of that shim.
```
pandemicsyn: This seems like implementation details we don't need to shove into the context ? Should be enough to just tell openclaw "yo we don't use systemd, don't try to use it" like we do below.

services/kiloclaw/controller/src/bootstrap.ts
```diff
@@ -1178,6 +1178,33 @@ When running \`openclaw doctor\` or \`openclaw security audit\`, the following f
...
+<!-- BEGIN:process-model -->
+
+## Process Model
+
+KiloClaw does NOT use systemd as its init system, in any deployment.
+
+- **PID 1 varies by deployment**, but is never systemd:
+  - Local dev (plain \`docker run\`): \`node /usr/local/bin/kiloclaw-controller.js\` — the controller daemon runs as PID 1 directly.
+  - Fly Machines (production today): \`/fly/init\` — Fly's own init shim, which then execs the controller as a child.
+  - Other providers: whatever init shim that platform injects (e.g. \`tini\`, \`dumb-init\`, or a provider-specific equivalent). The controller still runs as a child of that shim.
```
pandemicsyn: Same comment as above.

pandemicsyn (approved):
lgtm, one comment inline but also fine as is i think

jeanduplessis (approved):

PR #3121, opened 2026-05-07
feat(contributors): add manual tier upgrade for enrolled champions

bturcotte520:
Allow existing champions to be upgraded

### DIFF
--- a/apps/web/src/lib/contributor-champions/service.ts
+++ b/apps/web/src/lib/contributor-champions/service.ts
@@ -803,28 +803,50 @@ export async function upgradeContributorChampionTier(input: {
   if (!row) throw new Error('Contributor not found');
   if (!row.enrolledTier) throw new Error('Contributor is not enrolled');
 
-  const currentCreditUsd = TIER_CREDIT_USD[row.enrolledTier];
+  // Coarse pre-check using the leaderboard snapshot. The authoritative check
+  // happens inside the transaction after the row lock is acquired.
   const newCreditUsd = TIER_CREDIT_USD[input.newTier];
-
-  if (newCreditUsd <= currentCreditUsd) {
+  if (newCreditUsd <= TIER_CREDIT_USD[row.enrolledTier]) {
     throw new Error(
       `New tier "${input.newTier}" must be higher than current tier "${row.enrolledTier}"`
     );
   }
 
-  const creditDifferentialUsd = newCreditUsd - currentCreditUsd;
   const newCreditAmountMicrodollars = toMicrodollars(newCreditUsd);
   // Prefer the explicit membership link over the email-derived match, same as enrollContributorChampion.
   const linkedKiloUserId = row.linkedKiloUserId ?? row.linkedUserId;
 
-  const creditGranted = await db.transaction(async tx => {
+  const { creditGranted, creditDifferentialUsd } = await db.transaction(async tx => {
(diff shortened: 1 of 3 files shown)


apps/web/src/lib/contributor-champions/service.ts
```diff
@@ -788,6 +788,115 @@ export async function enrollContributorChampion(input: {
...
+    const [membership] = await tx
+      .select({ enrolled_tier: contributor_champion_memberships.enrolled_tier })
+      .from(contributor_champion_memberships)
+      .where(eq(contributor_champion_memberships.contributor_id, input.contributorId))
+      .limit(1);
+
+    if (!membership?.enrolled_tier) throw new Error('Contributor membership not found');
+
+    const lockedCurrentCreditUsd =
+      TIER_CREDIT_USD[membership.enrolled_tier as ContributorTier] ?? 0;
```
johnnyeric: Suggest to use parseContributorTier for safety in case enrolled_tier has invalid data.
e.g.
``` 
const lockedCurrentTier = parseContributorTier(membership.enrolled_tier);
if (!lockedCurrentTier) throw new Error(`Invalid enrolled_tier "${membership.enrolled_tier}" in DB`);                                                                  
const lockedCurrentCreditUsd = TIER_CREDIT_USD[lockedCurrentTier];  
```

apps/web/src/lib/contributor-champions/service.ts
```diff
@@ -788,6 +788,115 @@ export async function enrollContributorChampion(input: {
...
+
+      if (linkedUser) {
+        const result = await grantCreditForCategory(linkedUser, {
+          credit_category: 'contributor-champion-credits',
+          amount_usd: lockedDifferentialUsd,
+          expiry_hours: CREDIT_EXPIRY_HOURS,
+          counts_as_selfservice: false,
+          dbOrTx: tx,
+        });
+        granted = result.success;
```
johnnyeric: The bot's concern has been addressed about the positive scenario, but if the grant fails somethings would still get committed in the transaction. I'd suggest to reject transaction in case of grant failure so that it can be retried cleanly.
suggestion (replacing current `granted = result.success` :
``` 
if (!result.success) {                                                                                                                     
  throw new Error('Failed to grant top-up credit; rolling back tier upgrade');                                                                                       
}
granted = true;

File: apps/web/src/lib/contributor-champions/service.ts [hunk 3]
```diff
@@ -788,6 +788,115 @@ export async function enrollContributorChampion(input: {
...
+    // leaderboard snapshot taken before the lock. Without this, two concurrent
+    // upgrade calls could both read the pre-upgrade tier, both compute the same
+    // differential, and both grant — double-granting the top-up.
+    const [membership] = await tx
+      .select({ enrolled_tier: contributor_champion_memberships.enrolled_tier })
+      .from(contributor_champion_memberships)
+      .where(eq(contributor_champion_memberships.contributor_id, input.contributorId))
+      .limit(1);
+
+    if (!membership?.enrolled_tier) throw new Error('Contributor membership not found');
```
johnnyeric: Would it make sense to separate the error logs for missing membership or not enrolled?
like:
``` 
if (!membership) throw new Error('Contributor membership not found');                                                                                                  
if (!membership.enrolled_tier) throw new Error('Contributor is not currently enrolled');  
```

johnnyeric (approved):

PR #3124, opened 2026-05-07
fix(gastown): make Ctrl+V paste work in mayor terminal on Windows/Linux

jrf0110:
## Summary

Fix Ctrl+V paste not working in the gastown mayor terminal on Windows/Linux Chrome. xterm.js intercepts Ctrl+V as a control byte (0x16) and suppresses the browser's native paste event, while macOS Cmd+V works fine since xterm.js lets it through. Two changes:

1. **Ctrl+V / Ctrl+Shift+V paste handling**: Merged into the existing `attachCustomKeys` handler (formerly `attachKittyEnterHandler`). Intercepts Ctrl+V on Windows/Linux, reads clipboard via `navigator.clipboard.readText()`, and calls `term.paste()` which correctly handles bracketed paste mode. macOS Cmd+V is explicitly excluded to avoid double-handling.

2. **Right-click paste**: Added `contextmenu` handler on the terminal container that pastes from the system clipboard, matching GNOME Terminal / VS Code terminal convention. Listener is properly cleaned up on unmount.

## Verification

- On Windows Chrome with clipboard read permission, Ctrl+V pastes clipboard text into the shell with bracketed paste markers preserved
- On macOS, Cmd+V continues to work via native browser paste (handler excludes metaKey)
- If clipboard permission is denied, Ctrl+V does nothing visible and does not send ^V to the shell or throw unhandled rejections
- Right-click in the terminal pastes from clipboard; listener is removed on unmount
- Kitty Enter (Shift+Enter, Ctrl+Enter, Alt+Enter) continues to work unchanged

## Visual Changes

N/A

## Reviewer Notes

The paste logic is intentionally minimal — no toast/error UI on permission denial, silent failure matches the user mental model. `attachCustomKeyEventHandler` only accepts one handler, so paste and Kitty Enter logic are combined in a single handler.

### DIFF
--- a/apps/web/src/app/(app)/gastown/onboarding/OnboardingStepRepo.tsx
+++ b/apps/web/src/app/(app)/gastown/onboarding/OnboardingStepRepo.tsx
@@ -1,7 +1,9 @@
 'use client';
 
-import { useState, useMemo, useCallback } from 'react';
+import { useState, useMemo, useCallback, useEffect } from 'react';
 import { useQuery } from '@tanstack/react-query';
+import { useSearchParams } from 'next/navigation';
+import { toast } from 'sonner';
 import { useTRPC } from '@/lib/trpc/utils';
 import { useUser } from '@/hooks/useUser';
 import { RepositoryCombobox, type RepositoryOption } from '@/components/shared/RepositoryCombobox';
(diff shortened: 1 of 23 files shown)


jrf0110:
I'm really surprised this stuff isn't handled by the clipboard plugin. Let's hold off on merging until I verify it works correctly. Don't merge until I comment back or resolve this thread.

PR #3136, opened 2026-05-08
feat(code-review): add code review alerts

alex-alecu:
## Summary

Move code-review pipeline alerting onto the standard on-call pattern: BetterStack polls a health endpoint and owns the incident lifecycle (dedupe, re-notification, escalation, optional Slack routing). The cloud app no longer posts directly to Slack.

- New endpoint `GET /api/code-reviews/up?key=kilo-code-reviews-health-check` runs all four code-review detectors against `cloud_agent_code_reviews` over the rolling 30-minute window:
  - `200 healthy: true` when no detectors trip
  - `503 healthy: false` with a JSON `alerts` payload (kind, label, severity, detector fields, `adminUrl`, `runbookUrl`) when one or more trip
  - On detector/DB exception, fail open with `200 healthy: true` and capture to Sentry, matching `/api/models/up`

Cross-repo PRs:
- Runbook: https://github.com/Kilo-Org/on-call/pull/39
- Handbook: https://github.com/Kilo-Org/handbook/pull/818

## Verification

- Hit the new route locally with the correct and incorrect key; confirmed `200`/`401` and the unhealthy `503` payload shape (kinds, fields, runbook URL).

## Reviewer Notes

- The BetterStack monitor must be created out of band: URL `https://api.kilo.ai/api/code-reviews/up?key=kilo-code-reviews-health-check`, alert on `503`, suggested cadence ~5 min, severity ticket, runbook link in monitor description. Until then, `503` responses don't fire any alert.
- Followups for the runbook + handbook PRs are linked above; the runbook URL referenced in health responses points at `main` of `Kilo-Org/on-call` and goes live when that PR merges.


apps/web/src/app/api/cron/code-review-alerts/route.ts
```diff
@@ -0,0 +1,175 @@
...
+): Promise<AlertResult> {
+  const alertKey = alertKeyForDetails(details);
+  const webhookUrl = SLACK_CODE_REVIEW_WEBHOOK_URL;
+
+  if (!webhookUrl) {
+    console.warn('SLACK_CODE_REVIEW_WEBHOOK_URL is not configured; skipping code review alert');
+    return { detector, alertKey, suppressed: false, slack: 'skipped' };
+  }
+
+  const { suppressed } = await checkAndRecordAlert(alertKey, severity);
```
alex-alecu: Should the cooldown be recorded only after Slack returns success? Right now a 429/5xx or timeout from the webhook can mark this alert as seen and suppress later cron runs even though no Slack message was delivered.

apps/web/src/lib/code-reviews/alerting/detectors.ts
```diff
@@ -0,0 +1,233 @@
...
+      COUNT(*) FILTER (
+        WHERE status = 'queued'
+          AND updated_at < NOW() - (${STUCK_QUEUED_MINUTES} * INTERVAL '1 minute')
+      ) AS stuck_queued_count,
+      COUNT(*) FILTER (
+        WHERE status = 'running'
+          AND COALESCE(started_at, updated_at, created_at) < NOW() - (${STUCK_RUNNING_MINUTES} * INTERVAL '1 minute')
+      ) AS stuck_running_count
+    FROM ${cloud_agent_code_reviews}
+    WHERE created_at >= NOW() - (6 * INTERVAL '1 hour')
```
alex-alecu: Should this include all currently queued/running reviews instead of filtering by created_at? A review that has been running for more than six hours still satisfies the >2h stuck condition but ages out of this detector.

apps/web/src/app/api/cron/code-review-alerts/route.ts
alex-alecu: checkAndRecordAlert records the cooldown KV entry before `postToSlack` on line 104. If the Slack POST then throws (timeout / 5xx / network), the dedup is already written for the full ticket cooldown (4h) and the next cron runs will be silently suppressed even though no alert was actually delivered. Should we record the alert only after a successful Slack post (or roll back the KV entry on failure)?

apps/web/src/lib/code-reviews/alerting/detectors.ts
alex-alecu: This `WHERE created_at >= NOW() - 6 hours` cap excludes any review whose `created_at` is older than 6h, even if it is still in `running` / `queued`. That means a review that has been running for >6h — the most stuck case this detector is meant to catch — is invisible to `stuck_running_count`. Was the intent to scope by `created_at` here, or should the bound use `updated_at`/`started_at` (or be dropped for non-terminal statuses) so long-stuck reviews still trip the alert?

apps/web/src/lib/code-reviews/alerting/detectors.ts
```diff
@@ -0,0 +1,233 @@
...
+};
+
+type ErrorReasonRow = {
+  reason: string;
+  count: CountValue;
+};
+
+const terminalStatusesSql = sql`('completed', 'failed', 'cancelled', 'interrupted')`;
+const failureStatusesSql = sql`('failed', 'interrupted')`;
+const benignTerminalReasonsSql = sql`('billing', 'user_cancelled', 'superseded')`;
```
alex-alecu: These three benign reasons live here in isolation while `CODE_REVIEW_TERMINAL_REASONS` in `packages/db/src/schema-types.ts` is the authoritative list. If a new non-system reason is later added there (e.g. another no-op cancellation cause), this hard-coded list will not be updated and the failure-rate / error-spike detectors will start counting it as a system failure. Should the canonical "benign" subset live alongside the enum so they stay in sync?

services/o11y/src/alerting/code-review-dedup-routes.ts
```diff
@@ -0,0 +1,39 @@
...
+  app.post(
+    '/alerting/code-review-dedup',
+    requireAdmin,
+    zodJsonValidator(CodeReviewDedupInputSchema),
+    async c => {
+      const { alertKey, severity } = c.req.valid('json');
+      const kvKey = codeReviewAlertKey(severity, alertKey);
+
+      const existing = await c.env.O11Y_ALERT_STATE.get(kvKey);
+      if (existing) return c.json({ suppressed: true });
```
alex-alecu: The existing `services/o11y/src/alerting/dedup.ts#shouldSuppress` checks the page-level key when severity is `'ticket'` so a higher-severity alert suppresses the lower one for the same dimension. This new dedup route accepts both severities in its zod schema but only checks the exact key — so once anyone emits a `'page'` code review alert, the same alertKey will re-fire as a ticket while a page is already active. Intentional, or should this mirror `shouldSuppress`?

apps/web/src/app/api/cron/code-review-alerts/route.ts
alex-alecu: Fixed in d7864b43. The route now records the dedup key only after Slack returns success, so a failed POST will not silence later cron runs.

alex-alecu: Fixed in d7864b43. The dedup KV write now happens after Slack succeeds, with a test that asserts no record on a 429.

apps/web/src/lib/code-reviews/alerting/detectors.ts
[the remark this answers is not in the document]
alex-alecu: Fixed in 3497065e. The query now scopes by `status IN (queued, running)` instead of the 6-hour `created_at` window, so old stuck reviews still trip the alert.

alex-alecu: Fixed in 3497065e. Outer filter is now `status IN (queued, running)` so reviews older than six hours still count.

alex-alecu: Fixed in 3497065e. Dropped the `created_at` cap and scoped by `status IN (queued, running)` instead, with a new test that covers reviews stuck for over six hours.

apps/web/src/lib/code-reviews/alerting/detectors.ts
alex-alecu: Fixed in 64863d46. Added a `CODE_REVIEW_BENIGN_TERMINAL_REASONS` export next to the enum and the detector now builds its SQL list from it.

services/o11y/src/alerting/code-review-dedup-routes.ts
alex-alecu: Fixed in 8b386db4. The check action now also looks up the page key when severity is ticket, mirroring `shouldSuppress`. Added two tests for both directions.

apps/web/src/app/api/cron/code-review-alerts/route.ts
```diff
@@ -0,0 +1,175 @@
...
+      } else if (result.slack === 'sent') {
+        sent += 1;
+        slack = 'sent';
+      } else if (result.slack === 'skipped' && slack !== 'sent') {
+        slack = 'skipped';
+      }
+    } catch (error) {
+      capturedErrors.push(new Error(`${detector.name} detector failed`, { cause: error }));
+      errorSummaries.push({ detector: detector.name, message: errorMessage(error) });
+      slack = errorSummaries.some(item => item.message.startsWith('Slack webhook failed'))
```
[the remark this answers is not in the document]
alex-alecu: Fixed in fb66b3d9. Replaced the `.some()` rescan with a direct `message.startsWith` check on the current error.

apps/web/src/app/api/cron/code-review-alerts/route.ts
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  const { suppressed } = await checkAlertSuppression(alertKey, severity);
+
+  if (suppressed) {
+    return { detector, alertKey, suppressed: true, slack: 'suppressed' };
+  }
+
+  const slack = await postToSlack(details, webhookUrl);
+  await recordAlertDelivery(alertKey, severity);
```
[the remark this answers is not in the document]
alex-alecu: Fixed in 8a044927. The route now wraps `recordAlertDelivery` in try/catch, captures to Sentry, and reports the alert as `sent`. Added a test that covers the dedup-write failure path.

alex-alecu:
Manual test passed.

Tested `GET /api/code-reviews/up` against the local Next.js dev server on `:3000` with seeded `cloud_agent_code_reviews` rows.

Tested:
- `GET /api/code-reviews/up` (no key) → expect 401
- `GET /api/code-reviews/up?key=wrong` → expect 401
- `GET /api/code-reviews/up?key=kilo-code-reviews-health-check` against an empty 30-min window → expect 200 healthy
- Same correct-key call with 8 `failed`/`timeout` rows + 5 stuck `queued` rows (`updated_at` 20m old) seeded → expect 503 with all four detectors tripping
- Same call after rewriting all 8 failures to `terminal_reason='user_cancelled'` (benign) → expect benign-reason exclusion silences `failure_rate` and `error_spike`
- Same call after marking one row `completed` → expect `no_completions` to silence
- Same call after deleting the seeded rows → expect 200 healthy again

Verified:
- 401 path returns `{"healthy":false}` for missing and wrong `key`.
- 200 healthy path returns `{"healthy":true,"alerts":[],"metadata":{"timestamp":"…","runbookUrl":"https://github.com/Kilo-Org/on-call/blob/main/runbooks/code-review-pipeline-health.md"}}`.
- 503 unhealthy payload contains all four alert kinds with the documented shape:
  - `failure_rate`: `rate=1`, `total=8`, `failures=8`, `topReason="timeout"`, `topReasonCount=8`
  - `stuck_reviews`: `queuedCount=6`, `runningCount=0`
  - `no_completions`: `createdCount=13`
  - `error_spike`: `reason="timeout"`, `count=8`, `total=8`, `share=1`
  Each alert carries `label`, `severity:"ticket"`, `adminUrl` (`…/admin/code-reviews`), and `runbookUrl` (on-call repo `main`).
- Benign `terminal_reason` (`user_cancelled`) correctly excluded from `failure_rate` and `error_spike`; `stuck_reviews` and `no_completions` continued to fire as expected.
- A single `completed` row in the rolling window correctly silences `no_completions` while `stuck_reviews` keeps firing.
- Database state (counts by `status`/`terminal_reason`) before and after each call matched what the detectors reported.
- After cleanup, `cloud_agent_code_reviews` returned to its pre-test state and the endpoint returned to 200 healthy.

alex-alecu:
The endpoint returns `503 healthy:false` if **any one** of the four detectors trips. All thresholds are currently hardcoded in:

`apps/web/src/lib/code-reviews/alerting/thresholds.ts`

```ts
CODE_REVIEW_ALERT_WINDOW_MINUTES = 30;

FAILURE_RATE_THRESHOLD = 0.25;
FAILURE_RATE_MIN_TERMINAL = 8;

STUCK_QUEUED_MINUTES = 15;
STUCK_RUNNING_MINUTES = 120;
STUCK_COUNT_THRESHOLD = 5;

NO_COMPLETIONS_MIN_CREATED = 5;

ERROR_SPIKE_FRACTION = 0.5;
ERROR_SPIKE_MIN_FAILURES = 6;
```

## 1. `failure_rate` alert

Looks at reviews **created in the last 30 minutes**.

### Counts as “terminal”

These statuses count in the denominator:

```ts
completed
failed
cancelled
interrupted
```

### Counts as “system failure”

Only these statuses count as failures:

```ts
failed
interrupted
```

…but only if `terminal_reason` is **not** one of the benign reasons:

```ts
billing
user_cancelled
superseded
```

So these are treated as system failures:

```ts
failed + timeout
failed + upstream_error
failed + sandbox_error
failed + unknown
failed + null/empty terminal_reason
interrupted + interrupted
interrupted + timeout
...
```

These are **not** counted as system failures:

```ts
cancelled
completed
failed + billing
failed + user_cancelled
failed + superseded
interrupted + billing
interrupted + user_cancelled
interrupted + superseded
```

### Threshold

Triggers when:

```text
terminal_count >= 8
AND
system_failure_count / terminal_count > 0.25
```

Important: it is **strictly greater than 25%**, not greater-or-equal.

Examples:

| Terminal reviews in last 30m | System failures needed to alert |
|---:|---:|
| 0–7 | No alert |
| 8 | 3+ |
| 9 | 3+ |
| 10 | 3+ |
| 11 | 3+ |
| 12 | 4+ |
| 16 | 5+ |
| 20 | 6+ |

So:

```text
8 terminal reviews, 2 system failures = 25% exactly → no alert
8 terminal reviews, 3 system failures = 37.5% → alert
```

## 2. `stuck_reviews` alert

This detector does **not** use the 30-minute created-at window. It scans all current `queued` and `running` reviews.

### Queued threshold

Triggers if:

```text
5 or more queued reviews have updated_at older than 15 minutes
```

SQL condition:

```sql
status = 'queued'
AND updated_at < NOW() - INTERVAL '15 minutes'
```

### Running threshold

Triggers if:

```text
5 or more running reviews have been running/stale for more than 120 minutes
```

SQL condition:

```sql
status = 'running'
AND COALESCE(started_at, updated_at, created_at) < NOW() - INTERVAL '120 minutes'
```

### Important detail

Queued and running counts are checked separately, not summed.

```text
4 stuck queued + 4 stuck running = no alert
5 stuck queued + 0 stuck running = alert
0 stuck queued + 5 stuck running = alert
5 stuck queued + 5 stuck running = alert
```

## 3. `no_completions` alert

Looks at reviews **created in the last 30 minutes**.

Triggers when:

```text
5 or more reviews were created in the last 30 minutes
AND
0 of those reviews have status = 'completed'
```

It counts all created reviews regardless of status:

```ts
pending
queued
running
failed
cancelled
interrupted
completed
```

But it only clears if at least one is:

```ts
completed
```

Examples:

```text
4 reviews created, 0 completed → no alert
5 reviews created, 0 completed → alert
20 reviews created, 0 completed → alert
20 reviews created, 1 completed → no alert
```

Cancelled/failed/interrupted reviews still count toward `createdCount`, but they do not satisfy the “completed” condition.

## 4. `error_spike` alert

Looks at system failures **created in the last 30 minutes**.

It considers only:

```ts
status IN ('failed', 'interrupted')
```

and excludes benign terminal reasons:

```ts
billing
user_cancelled
superseded
```

It groups failures by:

```ts
terminal_reason
```

with `null` or empty string treated as:

```ts
unknown
```

### Threshold

Triggers when:

```text
total non-benign failed/interrupted rows >= 6
AND
the top terminal_reason accounts for >= 50% of those failures
```

This one is **greater-or-equal to 50%**, because the code only returns healthy when `share < 0.5`.

Examples:

| System failures in last 30m | Same reason needed to alert |
|---:|---:|
| 0–5 | No alert |
| 6 | 3+ with same reason |
| 7 | 4+ with same reason |
| 8 | 4+ with same reason |
| 10 | 5+ with same reason |
| 20 | 10+ with same reason |

Example alerting cases:

```text
6 failures total:
  3 timeout
  3 upstream_error
→ alerts, top reason share = 50%

8 failures total:
  4 timeout
  2 upstream_error
  2 sandbox_error
→ alerts, timeout share = 50%

8 failures total:
  3 timeout
  3 upstream_error
  2 sandbox_error
→ no error_spike alert, top share = 37.5%
```

## Status / reason summary

| Status | Failure rate denominator? | Failure rate numerator? | Error spike? | No completions created count? |
|---|---:|---:|---:|---:|
| `completed` | Yes | No | No | Yes |
| `failed` | Yes | Yes, unless benign reason | Yes, unless benign reason | Yes |
| `cancelled` | Yes | No | No | Yes |
| `interrupted` | Yes | Yes, unless benign reason | Yes, unless benign reason | Yes |
| `queued` | No | No | No | Yes |
| `running` | No | No | No | Yes |
| `pending` | No | No | No | Yes |

Benign terminal reasons:

```ts
billing
user_cancelled
superseded
```

System/error terminal reasons include:

```ts
interrupted
timeout
upstream_error
sandbox_error
unknown
null / empty terminal_reason
```

## Overall endpoint behavior

Returns `503` if **any** detector trips:

```text
failure_rate OR stuck_reviews OR no_completions OR error_spike
```

Returns `200` when none trip:

```json
{
  "healthy": true,
  "alerts": []
}
```

Returns `503` with one or more alerts when tripped:

```json
{
  "healthy": false,
  "alerts": [
    {
      "kind": "failure_rate",
      "severity": "ticket",
      ...
    }
  ]
}
```

If a detector throws due to DB/query issues, the route captures the exception to Sentry and treats that detector as non-tripped. That means DB/query errors fail open rather than paging.

apps/web/src/app/api/code-reviews/up/route.ts
```diff
@@ -0,0 +1,87 @@
...
+    return NextResponse.json({ healthy: false }, { status: 401 });
+  }
+
+  // Run detectors independently so a single broken detector does not mask the
+  // alerts the others would have produced. Failed detectors are captured to
+  // Sentry; we still return whatever alerts the surviving detectors produced.
+  const evaluations = await Promise.all(
+    DETECTORS.map(async detector => {
+      try {
+        return await detector.evaluate(db);
```
jeanduplessis: **WARNING:** Detector queries can hang instead of failing open.\n\nTrace:\n1. The health route awaits each detector directly and only fails open after the detector promise rejects.\n2. Each detector calls `database.execute(...)` without a statement timeout or abort deadline. The referenced `/api/models/up` route sets a local `statement_timeout` before running its health query.\n3. If a detector query blocks on DB slowness/locks or runs past the BetterStack/platform timeout, no exception reaches this catch in time, so the request times out/returns a platform 5xx instead of the intended `200 healthy: true` fail-open response.\n\nImpact: A DB stall or slow query can create a false BetterStack incident and skip the Sentry capture path this route relies on for detector failures.

apps/web/src/lib/code-reviews/alerting/detectors.ts
```diff
@@ -0,0 +1,237 @@
...
+  created_count: CountValue;
+};
+
+type ErrorReasonRow = {
+  reason: string;
+  count: CountValue;
+};
+
+const terminalStatusesSql = sql`('completed', 'failed', 'cancelled', 'interrupted')`;
+const failureStatusesSql = sql`('failed', 'interrupted')`;
```
jeanduplessis: **WARNING:** Interrupted callbacks are normalized to `cancelled`, so these detectors miss them.\n\nTrace:\n1. The new failure-rate and error-spike detectors count system failures only when persisted `status` is `failed` or `interrupted`; `cancelled` is denominator-only.\n2. The real status ingestion path in `apps/web/src/app/api/internal/code-review-status/[reviewId]/route.ts` maps cloud-agent-next `status: 'interrupted'` to persisted `status: 'cancelled'`; the route tests explicitly assert this mapping.\n3. Therefore, an interrupted/error callback persisted through the application path becomes `cancelled`, not `interrupted`, and these predicates exclude it from both `system_failure_count` and `error_spike`.\n\nImpact: A real interruption/error spike can be undercounted or completely missed by the new health endpoint, especially when there are enough completions to keep `no_completions` silent.

jeanduplessis (approved):
Found two concrete alerting correctness issues:\n- detector DB queries do not have a timeout, so the intended fail-open behavior can become a request timeout/5xx;\n- persisted interrupted callbacks are normalized to cancelled, but the new failure detectors only count interrupted as a failure status.

apps/web/src/app/api/code-reviews/up/route.ts
alex-alecu: Fixed in 40e331b.

apps/web/src/lib/code-reviews/alerting/detectors.ts
alex-alecu: Fixed in dfc1ed0.

PR #3151, opened 2026-05-09
release: gastown-staging -> main

### LINKED ISSUE #3149: [Gastown] Misleading "GitHub API returned null" error when town has no GitHub token
## Summary

When a town's `git_auth` config has no GitHub token (none of `github_token`, `github_cli_pat`, or `platform_integration_id` are set), the PR status poller in the merge_request bead lifecycle fails with a misleading error message that suggests the GitHub API itself is the problem.

## What the user sees

After 10 consecutive null poll attempts, the merge_request bead is marked `failed` with this metadata:

```json
{
  "failureReason": "pr_poll_failed",
  "failureMessage": "Cannot poll PR status — GitHub API returned null 10 consecutive times. Check that a valid GitHub token is configured in town settings and that the GitHub API is reachable."
}
```

This is confusing because:
- The polecat that produced the PR successfully *created* the PR using a working credential, so users naturally assume "of course GitHub auth works, the agent just used it."
- The error implies the GitHub API was contacted and returned `null`, when in reality **no API call was made** — `resolveGitHubToken` returned `null` and `checkPRStatus` short-circuited before any fetch.

## Root cause

`services/gastown/src/dos/town/town-scm.ts:checkPRStatus` returns `null` for at least four distinct conditions, which the caller in `actions.ts` (the poll_pr action handler around line 1030) collapses into a single `poll_null_count` metric:

1. `resolveGitHubToken(ctx)` returned `null` (no token in town config and no platform integration available) — only logs a `console.warn` at town-scm.ts:70
2. `fetch` to `api.github.com/repos/.../pulls/N` returned a non-OK status (could be 404, 401, 403, 5xx, secondary rate limit) — only logs at town-scm.ts:86
3. `response.json()` failed
4. `GitHubPRStatusSchema.safeParse(json)` failed (Zod parse mismatch)

All four return `null` indistinguishably. The warn logs are present, but they don't surface to the bead's metadata or to the user.

The polecat side uses an entirely different code path (its own container-injected GitHub credential for `git push` / PR creation), so a working polecat does **not** imply the town worker can resolve a token. They are independent.

## Why this is hard to diagnose

The user has no signal that points to the actual cause. They see "GitHub API returned null" and a hint to check the token, but:
- They likely already verified the polecat created the PR successfully → "auth works"
- The token check in town settings UI may show a token field configured at the org/rig level but not at the **town** level, and the user can't easily tell which level `resolveGitHubToken` actually consults
- There's no way to tell from the bead whether the failure is "no token" vs. "bad token" vs. "GitHub 5xx" vs. "schema drift"

## Reproduction

1. Create a town without setting a `github_token` / `github_cli_pat` / `platform_integration_id` in its config (or have the platform integration fail). Let polecats use their own container creds.
2. Sling a bead that opens a PR.
3. The merge_request bead created by `gt_done` will poll `checkPRStatus`, get `null` 10x, and fail with the misleading message.

Real-world example: bead `c262038b-f24e-4e21-a89c-7fb3a5f9864f` in town `98172328-9bd1-4b59-ba3e-0ae627058e6b`, rig `b6cf4b32-4e1b-4558-a864-a2a8df7bb1de`, against PR https://github.com/Kilo-Org/cloud/pull/3148 — which is OPEN/MERGEABLE/CLEAN and trivially fetchable via a manually authenticated GraphQL query, ruling out actual GitHub API trouble.

### 1. Distinguish the null causes in `checkPRStatus` (high impact, low effort)

Return a discriminated union from `checkPRStatus` instead of `PRStatusResult | null`:

```ts
type PRStatusError =
  | { kind: 'no_token' }
  | { kind: 'http_error'; status: number; statusText: string }
  | { kind: 'invalid_response'; reason: 'json_parse' | 'schema_mismatch' }
  | { kind: 'unrecognized_url' };

type PRStatusOutcome =
  | { ok: true; result: PRStatusResult }
  | { ok: false; error: PRStatusError };
```

Then the poll_pr handler in `actions.ts` can write the specific error kind into the bead's `metadata.failureReason` / `failureMessage`. Examples:

- `no_token` → "Cannot poll PR status — no GitHub token configured for this town. Add a GitHub PAT or platform integration in town settings."
- `http_error 404` → "PR not found. Was the branch deleted before the PR could be polled?"
- `http_error 401` → "Town's GitHub token is invalid or expired."
- `http_error 403` → "Town's GitHub token lacks `pull-requests: read` permission for this repo, or hit a secondary rate limit."
- `http_error 5xx` → keep retrying with backoff rather than counting toward null threshold.
- `schema_mismatch` → "GitHub API response shape changed; please file a bug." (and include a few keys for the bug report).

### 2. Don't fail-fast on legitimate transient errors

Currently, *any* null counts toward the 10-strike threshold, including transient 5xx and rate limits. After the discriminated union split, only `no_token` / 4xx auth errors / repeated `schema_mismatch` should fail-fast; transient issues should reset or use a longer threshold.

### 3. Surface the actionable hint about which token level is consulted

In the failure message, name the resolution chain:

> "No GitHub token resolved. Tried (in order): town `git_auth.github_token`, town `github_cli_pat`, town platform integration, rig platform integration. Configure one of these in town or rig settings."

This would have saved the user in the repro above ~30 minutes of confusion since they were looking at polecat container auth, not town config auth.

### 4. UI: show a "Town GitHub token" health indicator in town settings

When the town has no resolvable GitHub token, show a yellow warning banner: "Polecats can still create PRs (they use their own credentials), but the town cannot poll PR status to land merged work. Configure a token below." This decouples the two concerns visually so users stop assuming the polecat's success implies a healthy town config.

## Area

Merge Queue / Refinery

## Context

- **Town ID:** 98172328-9bd1-4b59-ba3e-0ae627058e6b
- **Agent:** Mayor (0c952401-2aaa-4335-bee2-35036e90483c)
- **Rig ID:** b6cf4b32-4e1b-4558-a864-a2a8df7bb1de

*Filed automatically by the Mayor via `gt_report_bug`.*

jrf0110:
## Summary

Promotes 13 commits from `gastown-staging` to `main`. Six independent fix/feature groups plus follow-on hardening:

1. **Boot-hydration timeout fix** — unblocks `/agents/start` during container boot hydration and preserves mayor tools on prewarm.
2. **GitHub auth correctness** — fresh integration tokens instead of stale stored value, plus distinguished failure messages when no token resolves.
3. **Logging hygiene** — redundant request-logging middleware replaced with per-route Hono-param tagging.
4. **Developer tooling** — dev-only convoy debug endpoints and a deterministic review-then-land E2E test procedure.
5. **Hook-leak race + mayor `GH_TOKEN`** — fixes a cascade where phantom dispatch failures tore live SDK hooks out from under running agents (causing `gt_request_changes` 500s, silent merges, and triage role-mismatch escalations) and the mayor's prewarmed SDK shipped without `GH_TOKEN`.
6. **Follow-on hardening** — token-rotation ordering in `/refresh-token`, bootHydration resolve scoping, mayor-id RPC consolidation, request-tag middleware coverage for nested `:rigId` routes, sample-rate tuning, and formatting.

`TownContainerDO.max_instances` was lowered from 800 → 500 as part of commit 1.

### 1. Boot hydration + mayor prewarm fix (`2ffcef28f`, direct push)

Three independent fixes for the `startAgentInContainer` timeout regression observed after #2974, plus a tighter container-instance cap.

**Symptoms.** Production logs were filling with two error patterns since the last `gastown-staging`→`main` promotion:

```
[<DOMAIN>] startAgentInContainer: EXCEPTION for agent <UUID>: TimeoutError: The operation was aborted due to timeout
timeout after 6000ms: ensureSDKServer for <agentId>
```

**Root cause.** The control server starts accepting requests immediately at boot (`main.ts:83`), while `bootHydration()` runs concurrently and serialises every registry agent + the new mayor prewarm through the global `sdkServerLock` (createKilo reads `process.cwd()`/`process.env`). Fresh `/agents/start`, `/refresh-token`, and PATCH `/agents/:id/model` requests queued behind that work and the DO-side `AbortSignal.timeout(60s)` (resp. `REFRESH_AGENT_TIMEOUT_MS=6_000`) fired before they ever got the lock.

The mayor prewarm added in #3122 made things worse on two axes:

1. It built `KILO_CONFIG_CONTENT` from hardcoded model defaults, so the real `/agents/start` with the user's actual model triggered `ensureSDKServer`'s "config mismatch — evicting prewarmed server" path on every warm restart, doubling lock-holding time on the critical path the prewarm was supposed to speed up.
2. It was missing `GASTOWN_AGENT_ROLE`, `GASTOWN_AGENT_ID`, and `GASTOWN_TOWN_ID` from the prewarm env. `kilo serve` snapshots `process.env` at spawn, and `plugin/index.ts:66` keys mayor-tool registration off `GASTOWN_AGENT_ROLE === 'mayor'`. Without those, the prewarmed server booted with **no mayor tools**, and the cache hit on the next `/agents/start` handed that defective instance back to the user — manifesting as "mayor tools became unavailable."

##### 1. Hydration gate (`control-server.ts`, `process-manager.ts`)
New `awaitHydration()` exported from `process-manager.ts`: a promise that `bootHydration` replaces on entry and resolves in a `finally`. Awaited at the top of `/agents/start`, `/refresh-token`, and PATCH `/agents/:id/model` (before any `process.env` mutation in the model PATCH path so concurrent requests can't race on env writes before holding the SDK lock). Default-resolved at module init so test/dev contexts that never run hydration aren't blocked.

##### 2. Prewarm config matches `/agents/start` (`Town.do.ts`, `gastown.worker.ts`, `process-manager.ts`)
New `getMayorPrewarmContext()` on `TownDO` returns `{ agentId, model, smallModel, kilocodeToken, organizationId }` resolved the same way `_ensureMayor` resolves them (`config.resolveModel(townConfig, null, 'mayor')`). The `/api/towns/:townId/mayor-id` endpoint now returns that whole context so the container builds a `KILO_CONFIG_CONTENT` byte-identical to what the next `/agents/start` will send. Falls back to the bare `{ agentId }` shape for back-compat; the container skips prewarm when model/token aren't available rather than building a config that's guaranteed to mismatch.

##### 3. Mayor workdir + plugin env (`agent-runner.ts`, `process-manager.ts`)
- Exported `ensureMayorWorkspaceForTown(townId)` so `prewarmMayorSDK` materialises the workspace before `ensureSDKServer`'s `process.chdir` (was throwing `ENOENT` on cold containers).
- `buildPrewarmEnv` now mirrors the mayor-shaped subset of `buildAgentEnv`: `GASTOWN_AGENT_ID`, `GASTOWN_AGENT_ROLE='mayor'`, `GASTOWN_TOWN_ID`, `KILOCODE_FEATURE='gastown'`, `KILO_TEST_HOME`, `XDG_DATA_HOME`. New end-to-end test intercepts `createKilo` and asserts those keys are visible to the spawn.

##### 4. `wrangler.jsonc`
Lowered `TownContainerDO.max_instances` from 800 → 500 (manual change).

### 2. Remove manual request logging middleware (#3158, `a6cf1029b`)

Removes the redundant request-logging middleware in `gastown.worker.ts` that logged every request twice (`-->`/`<--` via `logger.info`) — already covered by the per-route `instrumented(c, route, handler)` AE event wrapper. Replaces the regex-based `logger.setTags` block with proper per-route tagging using Hono `c.req.param()` matching for `:orgId` / `:townId` / `:rigId` / `:agentId` prefixes. Net diff: ~30 deletions + ~25 additions.

Link: https://github.com/Kilo-Org/cloud/pull/3158

### 3. Convoy debug endpoints + E2E test procedure (`7f9121ffa`, direct push)

Adds three dev-only debug endpoints for autonomous convoy testing without going through the mayor LLM:

- `GET  /debug/towns/:townId/rigs` — list rigs in a town
- `POST /debug/towns/:townId/sling-convoy` — call `Town.slingConvoy()` directly
- `GET  /debug/towns/:townId/convoys` — list active convoys with progress

Documents the new endpoints and adds a Test C section to `services/gastown/docs/e2e-pr-feedback-testing.md` with a deterministic procedure for verifying review-then-land convoys end-to-end (sub-bead PRs into the convoy feature branch, then a landing PR into main). Captures known issues observed during verification: container MTU/TLS handshake failures with github.com, 'failed' blockers not gating dependents, and intermittent polecat skipping of sub-PR creation.

### 4. Fresh integration tokens for GitHub auth (`ce15a6fe7`, direct push)

`resolveGitHubToken` previously preferred `git_auth.github_token` over the platform integration. Since GitHub App installation tokens have a 1h TTL but `git_auth.github_token` is only written at rig creation (or rare manual refresh), every long-lived town with an integration was handing out an expired token to:

- Polecat/refinery `gh` CLI (via `GH_TOKEN` derived from `GIT_TOKEN` in the container), surfacing as "Failed to log in to github.com using token (GH_TOKEN). The token in GH_TOKEN is invalid."
- The worker-side PR poller (`checkPRStatus`, `checkPRFeedback`, `mergePR`, `areThreadsBlocking`) — 401 from api.github.com.
- The `/refresh-git-token` endpoint the container falls back to on auth failure — it returned the same expired token, so the retry just re-failed.

Fix flips priority to `github_cli_pat` → live integration → stored `github_token` (last-resort fallback for towns with no integration). Empty-string responses from the integration service now warn and fall back instead of silently failing. Resolves a fresh token at agent dispatch (`startAgentInContainer`), merge dispatch (`startMergeInContainer`), and rig setup (`setupRigRepoInContainer`) before stuffing `GIT_TOKEN` into envVars. `buildContainerConfig` now resolves a fresh token before serializing `git_auth.github_token` into the `X-Town-Config` header. Adds 6 unit tests covering the priority chain.

### 5. Distinguish null causes in PR status polling (#3160, `63873e425`)

Fixes #3149.

Replace `PRStatusResult | null` return type with discriminated `PRStatusOutcome` union in `checkPRStatus`. Each null cause (no token, HTTP error, invalid response, unrecognized URL, host mismatch) now surfaces a structured `PRStatusError` with actionable failure messages.

Key changes:
- `resolveGitHubToken` returns `GitHubTokenResolution` with resolution chain tracking which sources were tried (back-compat helper `resolveGitHubTokenString` exists for non-error-aware callers).
- `no_token` and non-transient HTTP errors (401/403/404) fail the bead immediately (1 strike).
- `invalid_response`/`unrecognized_url`/`host_mismatch` fail after 3 strikes.
- Transient HTTP errors (5xx/429) keep existing 10-strike behavior.
- `poll_transient_count` and `poll_non_transient_count` separate counters (replaces the cross-contaminated single `poll_null_count`); both reset on successful poll.
- `failureKind` persisted to bead metadata for analytics.
- AE event `pr.poll_failed` emitted on terminal failure.
- `resolveGitHubToken` tracks the configured integration source even when `GIT_TOKEN_SERVICE` binding is missing.

Link: https://github.com/Kilo-Org/cloud/pull/3160

### 6. Hook-leak race + mayor `GH_TOKEN` plumb-through (`cbbd120cf`, direct push)

Three independent gastown reliability fixes that came out of debugging a production town stuck in a dispatch loop with the mayor reporting `GH_TOKEN is NOT set` while the bead it was working on cycled `in_progress → open` every few minutes (`town_id=0532b9ef-04c8-488a-ac34-6eb1a3d47228`).

#### 6a. `dispatchAgent` catch path no longer transitions agent to `idle`

The catch in `scheduling.ts:dispatchAgent` was running an `UPDATE agent_metadata SET status='idle'` after any thrown error from `startAgentInContainer`. The exception path can fire after the container has already accepted `/agents/start` (e.g. `/refresh-token` raced a token rotation and threw late) — in which case the SDK session is alive and heartbeating. Marking the agent idle was tripping `reconcileAgents`' "idle agent hooked to live bead" rule, which tore the hook out from under the running session and made every subsequent `gt_request_changes` / `gt_triage_resolve` call fail with `is not hooked to a bead` until the session exited. Now we leave the agent as `working` and let the heartbeat-staleness check (90s) do the cleanup if the agent really did die. Also surfaces `getLastStartError()` in the `dispatch_failed` analytics event so future debugging doesn't have to guess at the root cause.

#### 6b. `reconcileAgents` skips fresh-heartbeat agents when reaping idle+hooked

`reconciler.ts:reconcileAgents` "idle agent hooked to live bead" rule now skips agents whose `last_activity_at` is fresh (<90s). The 90s window matches the same module's stale-heartbeat threshold, so a truly dead agent still gets reaped on a later tick — but a live SDK session caught mid-action by a phantom `dispatch_failed` keeps its hook.

#### 6c. `review-queue` unhooked-refinery `gt_done` no longer silently merges

`review-queue.ts:agentDone` previously fell back to marking the MR as `merged` when a refinery called `gt_done` with no `pr_url` after losing its hook. That same hook-leak race fires while the refinery is mid-review and trying to call `gt_request_changes` — so the silent-merge path was landing PRs the refinery was actively trying to reject (after `gt_request_changes` repeatedly hit "not hooked"). The fallback now fails the review (returning the source bead to `open`) and raises an escalation so a human can decide whether the refinery meant to approve or request changes.

#### 6d. Mayor prewarm propagates GitHub auth

`getMayorPrewarmContext` now resolves the GitHub token via the standard chain (`github_cli_pat` / platform integration / stored fallback) and returns `githubToken` + `githubCliPat` alongside the kilocode token. The container's `buildPrewarmEnv` (`process-manager.ts`) sets `GH_TOKEN`, `GIT_TOKEN`, `GITHUB_TOKEN`, and `GITHUB_CLI_PAT` on the prewarmed SDK's `process.env` so the mayor's bash tool sees credentials from boot. Previously these were only set by `buildAgentEnv` on `/agents/start`, which never runs while `ensureMayor` short-circuits on a warm session — the mayor's `gh auth status` reported "not logged in" until the SDK was torn down and rebuilt.

`PERSIST_ENV_KEYS` now includes `GH_TOKEN`/`GIT_TOKEN`/`GITHUB_TOKEN`/`GITHUB_CLI_PAT` so a token rotation arriving via `/agents/start` cache-hit reaches `process.env` (without these, only `KILO_CONFIG_CONTENT`/`OPENCODE_CONFIG_CONTENT`/`GASTOWN_ORGANIZATION_ID` were refreshed on cache-hit, leaving git auth pinned to whatever was set at prewarm time).

Also includes the prior staged-but-unrelated work that was already in the working tree: `hasActiveWork` thunk-based short-circuit refactor (avoids 4 SQL reads per check on hot towns) and `healthCheck` re-arms the alarm with the right cadence (idle vs active) so post-deploy health pings don't pin every idle town to the 5s fast loop.

### 7. Follow-on hardening (multiple commits)

A run of small fixes/refactors that landed on top of the changes above:

- **`ff992a383` `fix(gastown/container): move GASTOWN_CONTAINER_TOKEN assignment after awaitHydration`** — order-of-ops fix in `/refresh-token`: the local `newToken` is captured before the `await awaitHydration()`, then `process.env.GASTOWN_CONTAINER_TOKEN` is assigned only after the hydration gate clears. Without this, a mid-hydration token refresh could hand the prewarm a different token than what hydration captured, racing the SDK lock.
- **`acdec1028` `refactor(gastown/container): capture bootHydration resolve locally`** — moves the `resolveHydration` capture from a module global to a local in `bootHydration` so a future second call (periodic re-hydration, etc.) can't interleave with an in-flight one.
- **`c0c17ee67` `refactor(gastown): eliminate double RPC in /mayor-id`** — folds `getMayorAgentId` into `getMayorPrewarmContext` so the worker's `/api/towns/:townId/mayor-id` endpoint makes a single DO RPC instead of two.
- **`eb0525ffc` `fix(gastown): add rigId tag middleware for /api/users/:userId/rigs/:rigId routes`** — the per-route tagging block added in #3158 missed this prefix; rig-scoped logs from the user-API surface had no `rigId` tag.
- **`c65fbc226` `fix(gastown): correct misleading migration comment in non-transient poll counter`** — comment-only, drift from #3160 review.
- **`60224676d` `chore(gastown): apply oxfmt formatting`** — formatter pass.
- **`3f19681b9` `chore(gastown): Lower trace sample rate`** — Sentry trace sample rate tuning to reduce production sampling cost.

## Verification

- Unit-tested the hydration gate end-to-end with a fetch barrier (asserts awaiters block while `bootHydration` is in flight, release when it returns).
- Unit-tested the prewarm env shape end-to-end (drives `bootHydration` with a `/mayor-id` fetch mock, intercepts `createKilo`, asserts `GASTOWN_AGENT_ID`, `GASTOWN_AGENT_ROLE='mayor'`, `GASTOWN_TOWN_ID`, `GASTOWN_CONTAINER_TOKEN`, and a non-empty `KILO_CONFIG_CONTENT` are all visible at spawn time).
- Reviewed the `_ensureMayor` model-resolution path to confirm `resolveModel(townConfig, null, 'mayor')` is byte-identical to what `/agents/start` will send (mayor role ignores `rigOverride` entirely in `config.resolveModel`).
- For commit 6 (hook-leak / GH_TOKEN): behavior verified by code inspection against the production AE event timeline for town `0532b9ef-04c8-488a-ac34-6eb1a3d47228` — every `dispatch_failed` in that town hit the catch path, every cycling unhook hit the idle+hooked+live-bead rule, and every "GH_TOKEN is NOT set" mayor session was a prewarm cache-hit. Manual production verification of the actual fix is deferred to post-deploy AE/Sentry monitoring.
- Manual production verification deferred for commit 1 — those changes target a hot path that's hard to reproduce locally; will monitor Sentry / AE `mayor.ensure_decision: short_circuit_warm` and `agent.startup_phase` after merge.

## Visual Changes

N/A

## Reviewer Notes

- The `/api/towns/:townId/mayor-id` response shape is back-compat: the container's Zod schema (`MayorPrewarmResponse`) accepts both the new full-context shape (now including `githubToken` / `githubCliPat`) and the legacy `{ agentId }` shape with `.passthrough()`, and rolls back to "skip prewarm" on missing fields.
- The `organizationId` fallback chain in `buildPrewarmEnv` distinguishes `undefined` (older worker, fall back to `process.env`) from `null` (worker authoritatively says "no org") so a stale env-var value can't override an authoritative `null`.
- Commit 6's `dispatchAgent` catch-path change deliberately leaves the agent as `working` even when `startAgentInContainer` throws — the comment at the catch site explains the rationale and points at the symmetric stale-heartbeat reaper. If you read the diff and worry about leaked `working` rows, the heartbeat path in `reconcileAgents` (line ~723, 90s threshold) is the safety net.
- Commit 6's silent-merge fix in `review-queue.agentDone` is a behavior change: previously, an unhooked refinery calling `gt_done` without a `pr_url` would mark the MR `merged`; now it fails the review and raises an escalation. This is intentional — the previous behavior could land PRs the refinery wanted rejected — but it means towns currently relying on the silent-merge path will see a new escalation surface until the underlying hook-leak race is gone (which the same commit fixes upstream).
- Commit 6 also adds `GH_TOKEN`/`GIT_TOKEN`/`GITHUB_TOKEN`/`GITHUB_CLI_PAT` to `PERSIST_ENV_KEYS`. This is what enables hot token rotation to reach `process.env` even when `/agents/start` short-circuits to a cached SDK server. Previously the cache-hit path only refreshed `KILO_CONFIG_CONTENT`/`OPENCODE_CONFIG_CONTENT`/`GASTOWN_ORGANIZATION_ID`.
- Two SUGGESTION-level findings deferred from earlier code review still apply: (a) `prewarmMayorSDK` warns but doesn't bail on workdir-mismatch (cheap to harden later), (b) one negative-case timing assertion in the new test relies on a 10ms `setTimeout` (test still validates the positive case deterministically).
- The `refresh-git-token.handler.ts` change is a caller update for the new `GitHubTokenResolution` return type (was `string | null`).
- The `wrangler.jsonc` max_instances change (800→500) is from the boot hydration commit (`2ffcef28f`).

### DIFF
--- a/services/gastown/container/src/control-server.ts
+++ b/services/gastown/container/src/control-server.ts
@@ -263,13 +263,19 @@ app.post('/refresh-token', async c => {
   if (!body || typeof body !== 'object' || !('token' in body) || typeof body.token !== 'string') {
     return c.json({ error: 'Missing or invalid token field' }, 400);
   }
-  process.env.GASTOWN_CONTAINER_TOKEN = body.token;
+  // Capture the new token into a local so it survives the await below.
+  const newToken = body.token;
 
   // Wait for boot hydration to release the global sdkServerLock before
-  // we serialise N agent restarts through it. Without this, agent k
-  // queues behind every in-flight hydration ensureSDKServer and easily
-  // blows past REFRESH_AGENT_TIMEOUT_MS (6s) on warm-restart containers.
+  // we mutate process.env or serialise N agent restarts through it.
+  // Without this gate, a mid-hydration token refresh can cause
+  // buildPrewarmEnv to pick up a different token than the one hydration
+  // captured locally — matching the PATCH /agents/:id/model handler
+  // which also gates first.
   await awaitHydration();
+
+  // Now safe to assign: hydration is done, no concurrent env readers.
+  process.env.GASTOWN_CONTAINER_TOKEN = newToken;
 
   const activeAgents = listAgents().filter(a => a.status === 'running' || a.status === 'starting');
   log.info('refresh_token.received', {
--- a/services/gastown/container/src/control-server.ts
+++ b/services/gastown/container/src/control-server.ts
@@ -18,6 +18,7 @@ import {
   registerEventSink,
   refreshTokenForAllAgents,
   listAgents,
+  awaitHydration,
 } from './process-manager';
 import { log } from './logger';
 import { startHeartbeat, stopHeartbeat, notifyContainerReady } from './heartbeat';
@@ -264,6 +265,12 @@ app.post('/refresh-token', async c => {
   }
   process.env.GASTOWN_CONTAINER_TOKEN = body.token;
 
+  // Wait for boot hydration to release the global sdkServerLock before
+  // we serialise N agent restarts through it. Without this, agent k
+  // queues behind every in-flight hydration ensureSDKServer and easily
+  // blows past REFRESH_AGENT_TIMEOUT_MS (6s) on warm-restart containers.
+  await awaitHydration();
+
   const activeAgents = listAgents().filter(a => a.status === 'running' || a.status === 'starting');
   log.info('refresh_token.received', {
     agentCount: activeAgents.length,
@@ -311,6 +318,15 @@ app.post('/agents/start', async c => {
     console.error('[control-server] /agents/start: invalid request body', parsed.error.issues);
     return c.json({ error: 'Invalid request body', issues: parsed.error.issues }, 400);
   }
+
+  // Wait for boot hydration to release the global sdkServerLock. The
+  // control server starts accepting requests immediately at boot, before
+  // bootHydration finishes resuming registry agents and prewarming the
+  // mayor — without this gate, fresh dispatches queue behind every
+  // serialised SDK spawn and the DO-side AbortSignal.timeout(60s) fires
+  // before they ever get the lock, surfacing as the
+  // "startAgentInContainer EXCEPTION TimeoutError" pattern.
+  await awaitHydration();
 
   // Persist the organization ID as a standalone env var so it survives
   // config rebuilds (e.g. model hot-swap). The env var is the primary
@@ -385,6 +401,15 @@ app.patch('/agents/:agentId/model', async c => {
   if (!parsed.success) {
     return c.json({ error: 'Invalid request body', issues: parsed.error.issues }, 400);
   }
+
+  // Model hot-swap restarts the SDK server (see updateAgentModel) and
+  // contends for the same global sdkServerLock that boot hydration is
+  // holding. Wait for hydration to drain BEFORE the env mutations
+  // below: concurrent PATCH requests landing during hydration would
+  // otherwise race on process.env writes before any of them holds the
+  // SDK lock, and the env visible to the eventual `kilo serve` spawn
+  // would be non-deterministic.
+  await awaitHydration();
 
   // Update org billing context from the request body if provided.
   if (parsed.data.organizationId) {
--- a/services/gastown/container/src/process-manager.ts
+++ b/services/gastown/container/src/process-manager.ts
@@ -603,6 +603,15 @@ const PERSIST_ENV_KEYS = new Set([
   'KILO_CONFIG_CONTENT',
   'OPENCODE_CONFIG_CONTENT',
   'GASTOWN_ORGANIZATION_ID',
+  // Git auth: token rotations need to land in process.env so that
+  // bash subprocesses spawned by the SDK's bash tool see the fresh
+  // value. Without these here, a cache-hit on /agents/start returns
+  // the prewarmed SDK with whatever GH_TOKEN/GIT_TOKEN was set at
+  // prewarm time, even after the worker rotated them.
+  'GH_TOKEN',
+  'GIT_TOKEN',
+  'GITHUB_TOKEN',
+  'GITHUB_CLI_PAT',
 ]);
 
 async function ensureSDKServer(
@@ -2632,6 +2641,8 @@ type MayorPrewarmContext = {
   smallModel?: string;
   kilocodeToken?: string;
   organizationId?: string | null;
+  githubToken?: string;
+  githubCliPat?: string;
 };
 
 // Mirrors the response contract documented at
@@ -2645,6 +2656,8 @@ const MayorPrewarmResponse = z
     smallModel: z.string().optional(),
     kilocodeToken: z.string().optional(),
     organizationId: z.string().nullable().optional(),
+    githubToken: z.string().optional(),
+    githubCliPat: z.string().optional(),
   })
   .passthrough();
 
@@ -2665,9 +2678,18 @@ async function fetchMayorPrewarmContext(
     const json: unknown = await resp.json();
     const parsed = MayorPrewarmResponse.safeParse(json);
     if (!parsed.success) return null;
-    const { agentId, model, smallModel, kilocodeToken, organizationId } = parsed.data;
+    const { agentId, model, smallModel, kilocodeToken, organizationId, githubToken, githubCliPat } =
+      parsed.data;
     if (!agentId) return null;
-    return { agentId, model, smallModel, kilocodeToken, organizationId };
+    return {
+      agentId,
+      model,
+      smallModel,
+      kilocodeToken,
+      organizationId,
+      githubToken,
+      githubCliPat,
+    };
   } catch (err) {
     console.warn(`${MANAGER_LOG} fetchMayorPrewarmContext failed:`, err);
     return null;
@@ -2719,6 +2741,27 @@ function buildPrewarmEnv(ctx: MayorPrewarmContext, townId: string): Record<strin
       ? ctx.organizationId
       : (process.env.GASTOWN_ORGANIZATION_ID ?? null);
   if (organizationId) env.GASTOWN_ORGANIZATION_ID = organizationId;
+
+  // Plumb GitHub auth into the prewarmed SDK env so `gh` CLI and `git`
+  // subprocesses spawned from the mayor's bash tool see credentials.
+  // Mirror buildAgentEnv (agent-runner.ts:180-188): GITHUB_CLI_PAT wins
+  // for `gh` (PRs/issues appear under the user's identity), else fall
+  // back to the integration-resolved GIT_TOKEN.
+  //
+  // Without this, ensureMayor's short-circuit path returns a prewarmed
+  // SDK whose process.env is missing GH_TOKEN entirely — `gh auth status`
+  // reports "not logged in" until the SDK is torn down and rebuilt.
+  if (ctx.githubToken) {
+    env.GIT_TOKEN = ctx.githubToken;
+    env.GITHUB_TOKEN = ctx.githubToken;
+  }
+  if (ctx.githubCliPat) {
+    env.GITHUB_CLI_PAT = ctx.githubCliPat;
+  }
+  const ghToken = ctx.githubCliPat ?? ctx.githubToken;
+  if (ghToken) {
+    env.GH_TOKEN = ghToken;
+  }
 
   // Without the worker-resolved model, skip prewarm: any guess we make
   // here will almost certainly differ from /agents/start's resolved
--- a/services/gastown/container/src/process-manager.ts
+++ b/services/gastown/container/src/process-manager.ts
@@ -78,7 +78,6 @@ export function isDraining(): boolean {
 // the promise immediately so non-hydrating containers (tests, dev)
 // don't block; bootHydration replaces it on entry and resolves it on exit.
 let _hydrationComplete: Promise<void> = Promise.resolve();
-let _resolveHydration: () => void = () => {};
 
 export function awaitHydration(): Promise<void> {
   return _hydrationComplete;
@@ -2818,14 +2817,12 @@ async function prewarmMayorSDK(townId: string, apiUrl: string, token: string): P
  * contending for it themselves.
  */
 export async function bootHydration(): Promise<void> {
-  const LOG = '[boot-hydration]';
-  _hydrationComplete = new Promise<void>(resolve => {
-    _resolveHydration = resolve;
-  });
+  let resolve!: () => void;
+  _hydrationComplete = new Promise<void>(r => { resolve = r; });
   try {
-    await bootHydrationImpl(LOG);
+    await bootHydrationImpl('[boot-hydration]');
   } finally {
-    _resolveHydration();
+    resolve();
   }
 }
 
--- a/services/gastown/container/src/process-manager.ts
+++ b/services/gastown/container/src/process-manager.ts
@@ -2818,7 +2818,9 @@ async function prewarmMayorSDK(townId: string, apiUrl: string, token: string): P
  */
 export async function bootHydration(): Promise<void> {
   let resolve!: () => void;
-  _hydrationComplete = new Promise<void>(r => { resolve = r; });
+  _hydrationComplete = new Promise<void>(r => {
+    resolve = r;
+  });
   try {
     await bootHydrationImpl('[boot-hydration]');
   } finally {
--- a/services/gastown/container/src/process-manager.ts
+++ b/services/gastown/container/src/process-manager.ts
@@ -11,7 +11,11 @@ import { z } from 'zod';
 import * as fs from 'node:fs/promises';
 import type { ManagedAgent, StartAgentRequest } from './types';
 import { reportAgentCompleted, reportMayorWaiting } from './completion-reporter';
-import { buildKiloConfigContent, mayorWorkdirForTown } from './agent-runner';
+import {
+  buildKiloConfigContent,
+  ensureMayorWorkspaceForTown,
+  mayorWorkdirForTown,
+} from './agent-runner';
 import {
   getCurrentTownConfig,
   getLastAppliedEnvVarKeys,
@@ -64,6 +68,20 @@ let _draining = false;
 
 export function isDraining(): boolean {
   return _draining;
+}
+
+// Resolved when bootHydration() returns. /agents/start and /refresh-token
+// must await this before contending for the global sdkServerLock — without
+// this gate, fresh dispatches arriving during boot queue behind every
+// in-flight registry agent + the mayor prewarm and the DO-side 60s
+// AbortSignal.timeout fires before they ever get the lock. We resolve
+// the promise immediately so non-hydrating containers (tests, dev)
+// don't block; bootHydration replaces it on entry and resolves it on exit.
+let _hydrationComplete: Promise<void> = Promise.resolve();
+let _resolveHydration: () => void = () => {};
+
+export function awaitHydration(): Promise<void> {
+  return _hydrationComplete;
 }
 
 // Mutex for ensureSDKServer — createKilo() reads process.cwd() and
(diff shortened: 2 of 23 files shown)


services/gastown/src/gastown.worker.ts
```diff
@@ -672,6 +672,19 @@ app.use('/api/towns/:townId/mayor-id', async (c: Context<GastownEnv, string>, ne
...
+  //   will send (no eviction churn in ensureSDKServer).
+  // - When the mayor agent exists but no kilocode token is available,
+  //   return { agentId } only — the container will skip prewarm.
+  // - When there is no mayor at all, return { agentId: null } — the
+  //   container treats missing/null agentId as "no mayor, skip prewarm".
+  const ctx = await town.getMayorPrewarmContext();
+  if (ctx) {
+    return c.json({ success: true, ...ctx });
+  }
   const agentId = await town.getMayorAgentId();
```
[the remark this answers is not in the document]
jrf0110: Fixed in c0c17ee67 — consolidated to a single `getMayorPrewarmContext()` call. The method now returns `{ agentId }` in the partial case (mayor exists but no kilocode token), so the worker derives the fallback response without a second RPC hop. `null` only means "no mayor at all".

services/gastown/container/src/process-manager.ts
```diff
@@ -66,6 +70,20 @@ export function isDraining(): boolean {
...
 
+// Resolved when bootHydration() returns. /agents/start and /refresh-token
+// must await this before contending for the global sdkServerLock — without
+// this gate, fresh dispatches arriving during boot queue behind every
+// in-flight registry agent + the mayor prewarm and the DO-side 60s
+// AbortSignal.timeout fires before they ever get the lock. We resolve
+// the promise immediately so non-hydrating containers (tests, dev)
+// don't block; bootHydration replaces it on entry and resolves it on exit.
+let _hydrationComplete: Promise<void> = Promise.resolve();
+let _resolveHydration: () => void = () => {};
```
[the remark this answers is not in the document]
jrf0110: Fixed in acdec1028 — removed the `_resolveHydration` module global and captured `resolve` as a local inside `bootHydration()`, exactly as suggested. The existing test suite passes with the change.

services/gastown/src/dos/town/actions.ts
```diff
@@ -1179,57 +1238,190 @@ export function applyAction(ctx: ApplyActionContext, action: Action): (() => Pro
...
+                  rigId: failRigId,
+                  reason: error.kind,
+                  label: 'provider' in error ? error.provider : '',
+                  statusCode: error.kind === 'http_error' ? error.status : undefined,
+                });
+              }
+            } else {
+              // Non-transient, non-immediate errors (invalid_response only)
+              // count toward a lower 3-strike threshold.
+              // Migrate legacy poll_null_count into poll_non_transient_count on first read.
```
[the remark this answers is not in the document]
jrf0110: Good catch — fixed in c65fbc226. The comment now correctly states that no legacy `poll_null_count` migration is needed because `invalid_response` is a new error kind that couldn't have accumulated under the old classification.

jrf0110:
## Review observation dispositions

### Observation A — "Request/response logging removed without replacement"

Intentional — PR #3158 deletes those manual log lines because `instrumented()` already emits structured AE events with route, userId, townId, rigId, agentId, beadId, durationMs, and error per route. The new per-route Hono-param tagging middleware preserves the tagging side of the old block. No tracing observability is lost; structured tracing replaces it.

### Observation C — "Double `GIT_TOKEN_SERVICE.getToken` call per agent start"

Acknowledged. The second `GIT_TOKEN_SERVICE.getToken` is a KV cache hit, so the perf impact is negligible, but the duplication is real — will track as a separate cleanup since deduping requires changing `buildContainerConfig`'s signature beyond what's in scope for this release.

### Additional thread resolved

A 4th inline thread about a misleading migration comment in `actions.ts` was also addressed: the comment on the non-transient poll counter branch claimed `poll_null_count` migration, but the SQL doesn't include it (correctly, since `invalid_response` is a new error kind). Fixed in c65fbc226.

services/gastown/container/src/process-manager.ts
```diff
@@ -586,6 +603,15 @@ const PERSIST_ENV_KEYS = new Set([
   'KILO_CONFIG_CONTENT',
   'OPENCODE_CONFIG_CONTENT',
   'GASTOWN_ORGANIZATION_ID',
+  // Git auth: token rotations need to land in process.env so that
+  // bash subprocesses spawned by the SDK's bash tool see the fresh
+  // value. Without these here, a cache-hit on /agents/start returns
+  // the prewarmed SDK with whatever GH_TOKEN/GIT_TOKEN was set at
+  // prewarm time, even after the worker rotated them.
+  'GH_TOKEN',
```
jeanduplessis, line 610: **CRITICAL:** Persisting GitHub auth env keys in `process.env` can leak stale or wrong tokens into later SDK starts.

Trace:
1. These keys are now in `PERSIST_ENV_KEYS`, so `ensureSDKServer` skips restoring/deleting them after setting `process.env` for one SDK start.
2. On cache hits, `ensureSDKServer` only assigns truthy values for persisted keys and never clears absent ones.
3. `buildAgentEnv` falls back to `process.env` for `GITHUB_CLI_PAT`, `GIT_TOKEN`, and `GITHUB_TOKEN`, so a token from a previous rig/integration can be inherited by a later start whose request has no token or a different integration.

Impact: A later agent/mayor session can receive and use a stale or wrong `GH_TOKEN`/`GIT_TOKEN` from a previous start. This can leak GitHub credentials across rigs/integrations in the same town and can keep using a removed PAT until the container process is replaced.

services/gastown/container/src/control-server.ts
```diff
@@ -312,6 +325,15 @@ app.post('/agents/start', async c => {
...
   }
 
+  // Wait for boot hydration to release the global sdkServerLock. The
+  // control server starts accepting requests immediately at boot, before
+  // bootHydration finishes resuming registry agents and prewarming the
+  // mayor — without this gate, fresh dispatches queue behind every
+  // serialised SDK spawn and the DO-side AbortSignal.timeout(60s) fires
+  // before they ever get the lock, surfacing as the
+  // "startAgentInContainer EXCEPTION TimeoutError" pattern.
+  await awaitHydration();
```
jeanduplessis, line 335: **WARNING:** The hydration gate can block `/agents/start` indefinitely behind the unbounded boot registry fetch.

Trace:
1. `/agents/start`, `/refresh-token`, and model PATCH now await `awaitHydration()` before proceeding.
2. `bootHydrationImpl` resolves that gate only when the whole boot hydration path returns.
3. The `/container-registry` fetch in `bootHydrationImpl` has no `AbortSignal.timeout`, so a network/worker stall can keep the hydration promise unresolved forever.

Impact: A single stuck boot registry fetch can make a fresh container unable to start agents, refresh tokens, or apply model patches until the process is replaced. This is especially risky because the DO-side caller only sees timeouts while the container remains wedged behind the gate.

services/gastown/src/dos/town/actions.ts
```diff
@@ -330,9 +330,66 @@ const LOG = '[actions]';
...
+const PR_POLL_NON_TRANSIENT_THRESHOLD = 3;
+
 /** Minimum interval between PR polls per MR bead (ms) (#1632). */
 export const PR_POLL_INTERVAL_MS = 60_000; // 1 minute
 
+function failureMessageFor(error: PRStatusError): string {
+  switch (error.kind) {
+    case 'no_token':
+      return (
+        `No GitHub token resolved for this town. Tried (in order): ` +
```
jeanduplessis: **WARNING:** GitLab polling failures are surfaced with GitHub-specific failure messages.

Trace:
1. `PRStatusError` now carries `provider: 'github' | 'gitlab'`.
2. `checkPRStatus` creates GitLab `no_token` and `http_error` errors.
3. `failureMessageFor` ignores `error.provider` for `no_token`, 401, 403, and 404 and hard-codes GitHub token names/scopes.

Impact: GitLab MR polling can fail a bead immediately while telling the operator to refresh/configure GitHub credentials. That sends users to the wrong remediation path for missing/invalid GitLab tokens or GitLab 401/403/404 responses.

services/gastown/test/unit/pr-poll-thresholds.test.ts
```diff
@@ -0,0 +1,343 @@
...
+        expect(isNonTransient, `invalid_response should be in 3-strike bucket`).toBe(true);
+      } else {
+        expect(isNonTransient, `${error.kind} should NOT be in 3-strike bucket`).toBe(false);
+      }
+    }
+  });
+});
+
+describe('counter cross-contamination', () => {
+  it('9 transient errors then 1 non-transient should not fail (separate counters)', () => {
```
jeanduplessis, line 325: **SUGGESTION:** These counter cross-contamination tests do not execute the counter behavior they claim to protect.

Trace:
1. The test name covers the retry/counter state machine: “9 transient errors then 1 non-transient should not fail”.
2. The assertions only verify `_shouldFailImmediately` / `_shouldCountAsTransient` helper classification.
3. The actual `poll_transient_count` / `poll_non_transient_count` JSON update/reset SQL in `applyAction` is only described in comments, not exercised.

Impact: A regression in the SQL counter reset/increment logic could ship while this test remains green. Consider driving `applyAction` against a bead fixture or extracting the counter update into a testable helper.

jeanduplessis (approved):
Found 4 issues in the final combined diff:

- 1 critical auth/env leakage risk in container SDK env persistence.
- 2 warnings around hydration gating and GitLab-specific PR polling errors.
- 1 test-quality suggestion where counter behavior is documented but not executed.

I verified the current unresolved review-thread state (none unresolved) and checked the previously resolved/outdated threads against the final diff.

services/gastown/container/src/process-manager.ts
jrf0110: Fixed in e16b8b643. Git auth keys are no longer part of the persisted env set used during SDK creation restore/cleanup. Cache hits now apply Git token env separately and explicitly delete absent Git token keys, so a previous rig/integration token cannot leak into a later cached SDK start while still allowing token rotations to reach bash subprocesses.

services/gastown/container/src/control-server.ts
jrf0110: Fixed in e16b8b643. Added `AbortSignal.timeout(10_000)` to the boot hydration `/container-registry` fetch, so a stalled registry request returns through `bootHydrationImpl` and releases the hydration gate instead of wedging `/agents/start`, `/refresh-token`, and model patch requests indefinitely.

services/gastown/src/dos/town/actions.ts
jrf0110: Fixed in e16b8b643. `failureMessageFor` now labels missing/invalid tokens by provider, uses a GitLab-specific 403 remediation for merge request reads, and reports GitLab 404s as MR-not-found instead of GitHub PR guidance. Added unit coverage for GitLab-specific messages.

services/gastown/test/unit/pr-poll-thresholds.test.ts
jrf0110: Fixed in e16b8b643. Replaced the comment-only counter assertions with a testable `nextPollCounterState` transition helper and updated the tests to execute the transient/non-transient reset and threshold behavior directly.

PR #3156, opened 2026-05-09
fix(gastown): distinguish null causes in PR status polling (#3149)

### LINKED ISSUE #3149: [Gastown] Misleading "GitHub API returned null" error when town has no GitHub token

jrf0110:
## Summary

Replaces the `null` return from `checkPRStatus` with a discriminated `PRStatusOutcome` union (`{ ok: true, result }` | `{ ok: false, error }`), and similarly updates `resolveGitHubToken` to return `GitHubTokenResolution` that captures the resolution chain. This ensures each distinct failure mode (no token, HTTP errors, invalid response, unrecognized URL, host mismatch) surfaces a specific, actionable failure message instead of a generic "null" after 10 polls.

Key changes:
- **`checkPRStatus`** returns `PRStatusOutcome` with structured `PRStatusError` variants (no_token, http_error, invalid_response, unrecognized_url, host_mismatch)
- **`resolveGitHubToken`** returns `GitHubTokenResolution` tracking which sources were tried
- **Three-bucket failure strategy**: no_token/non-transient HTTP fail immediately (1 strike), invalid_response/unrecognized_url/host_mismatch fail after 3 strikes, transient 5xx/429 keep existing 10-strike behavior
- **`poll_null_count` resets to 0** on successful poll
- **AE telemetry**: `pr.poll_failed` event emitted on terminal failure
- Updated all callers including `refresh-git-token.handler.ts` and `Town.do.ts`

Fixes #3149.

## Verification

- Unit tests cover `checkPRStatus` error discrimination (no_token, 401/403/404, 5xx/429 transient, invalid_response, unrecognized_url, host_mismatch)
- Unit tests cover `failureMessageFor` producing actionable messages for each error kind
- Unit tests cover `shouldFailImmediately` and `shouldCountAsTransient` threshold logic
- Integration test covers end-to-end no_token failure through DO alarm

## Visual Changes

N/A

## Reviewer Notes

The `platformIntegrationId` is not passed in the `Town.do.ts` `checkPRStatus` context — this matches the pre-existing behavior. The `refresh-git-token.handler.ts` does pass it, which is the intended pattern for rig-level token resolution.

### DIFF
--- a/services/gastown/src/dos/town/actions.ts
+++ b/services/gastown/src/dos/town/actions.ts
@@ -22,7 +22,7 @@ import * as reviewQueue from './review-queue';
 import * as patrol from './patrol';
 import { getRig } from './rigs';
 import { parseGitUrl } from '../../util/platform-pr.util';
-import type { PRStatusResult } from './town-scm';
+import type { PRStatusOutcome, PRStatusError } from './town-scm';
 
 // ── Bead mutations ──────────────────────────────────────────────────
 
@@ -300,8 +300,8 @@ export type ApplyActionContext = {
   dispatchAgent: (agentId: string, beadId: string, rigId: string) => Promise<boolean>;
   /** Stop an agent's container process. */
   stopAgent: (agentId: string) => Promise<void>;
-  /** Check a PR's status via GitHub/GitLab API. Returns PRStatusResult or null. */
-  checkPRStatus: (prUrl: string) => Promise<PRStatusResult | null>;
+  /** Check a PR's status via GitHub/GitLab API. Returns PRStatusOutcome. */
+  checkPRStatus: (prUrl: string) => Promise<PRStatusOutcome>;
   /** Check PR for unresolved review comments and failing CI checks. */
   checkPRFeedback: (prUrl: string) => Promise<PRFeedbackCheckResult | null>;
   /** Merge a PR via GitHub/GitLab API. */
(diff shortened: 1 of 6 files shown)


jrf0110:
Superseded by #3160. Closing as duplicate from convoy retry loop.

PR #3158, opened 2026-05-09
chore(gastown): remove manual request logging middleware

jrf0110:
## Summary

Removes the redundant request-logging middleware from `gastown.worker.ts` that logged every HTTP request twice (incoming + outgoing) via `logger.info()`. This is already covered by the per-route `instrumented(c, route, handler)` wrapper from `analytics.middleware.ts` which emits structured AE events with `route`, `userId`, `townId`, `rigId`, `agentId`, `beadId`, `durationMs`, and `error` for every wrapped route.

Two pieces removed:
1. The `import { logger } from './util/log.util'` line (now unused at the worker level)
2. The entire request-logging middleware block (regex constants + `app.use('*', ...)` with `logger.setTags`, `logger.info` calls)

Preserved: `timingMiddleware` and `useWorkersLogger` middleware — these are independent infrastructure that other code depends on.

## Verification

Manual review of the diff confirms 30 lines deleted, 0 added, single file modified.

## Visual Changes

N/A

## Reviewer Notes

The `logger.setTags` call from URL regexes is safe to remove — those tags were only consumed by the two `logger.info` lines being removed. Auth middleware already sets `orgId`/`userId` tags via JWT-derived values. Town.do.ts runs in a separate isolate and does its own `setTags`.

### DIFF
--- a/services/gastown/src/gastown.worker.ts
+++ b/services/gastown/src/gastown.worker.ts
@@ -128,7 +128,6 @@ import { townAuthMiddleware } from './middleware/town-auth.middleware';
 import { orgAuthMiddleware } from './middleware/org-auth.middleware';
 import { adminAuditMiddleware } from './middleware/admin-audit.middleware';
 import { timingMiddleware, instrumented } from './middleware/analytics.middleware';
-import { logger } from './util/log.util';
 import { useWorkersLogger } from 'workers-tagged-logger';
 import type { MiddlewareHandler } from 'hono';
 import { handleGetTownConfig, handleUpdateTownConfig } from './handlers/town-config.handler';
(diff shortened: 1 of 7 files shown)


services/gastown/src/gastown.worker.ts
```diff
@@ -173,35 +172,6 @@ app.use('*', timingMiddleware);
...
-  const path = c.req.path;
-  // Tag with route params immediately so all downstream logs (auth,
-  // handlers, DO calls) inherit them. Auth-derived tags (userId, orgId)
-  // are set by kiloAuthMiddleware and orgAuthMiddleware when they run.
-  logger.setTags({
-    orgId: RE_ORG.exec(path)?.groups?.orgId,
-    townId: RE_TOWN.exec(path)?.groups?.townId,
-    rigId: RE_RIG.exec(path)?.groups?.rigId,
-    agentId: RE_AGENT.exec(path)?.groups?.agentId,
-  });
```
jrf0110, line 198: It'd still be nice to set tags on requests, but maybe we can do it in a way more structured way like:


```tsx
app.use('/api/towns/:townId', async (c, next) => {
  logger.setTags({ townId: c.req.param('townId' })
  await next()
})
```

PR #3183, opened 2026-05-11
feat(code-reviews): support REVIEW.md

alex-alecu:
## Summary
- Add support for repository-level `REVIEW.md` guidance in code review prompts and record when that guidance is used.
- Keep `REVIEW.md` disabled by default through the DB-backed `disable_review_md` setting so the feature can be enabled selectively during testing.
- Preserve the settings UI/API toggle so explicit `disable_review_md: false` enables repository guidance.

## Verification
N/A - no manual verification was performed for this backend default change.

## Visual Changes
N/A

## Reviewer Notes
The default-off behavior is intended as a short-lived rollout flag. Existing configs with `disable_review_md` missing now skip repository `REVIEW.md`; explicitly setting it to `false` enables the feature.


apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
```diff
@@ -139,50 +145,56 @@ export async function prepareReviewPayload(
...
-        const tokenData = await generateGitHubInstallationToken(
-          integration.platform_installation_id,
-          appType
-        );
+        const tokenData = await generateGitHubInstallationToken(installationId, appType);
         githubToken = tokenData.token;
+        const [repoOwner, repoName] = review.repo_full_name.split('/');
+
+        if (repoOwner && repoName) {
+          repositoryReviewInstructions = await fetchRepositoryReviewInstructions({
```
alex-alecu: Could the `REVIEW.md` fetch run in parallel with the existing GitHub state lookups instead of blocking them here?

apps/web/src/lib/integrations/platforms/github/adapter.ts
```diff
@@ -635,6 +635,56 @@ export async function getPRHeadCommit(
...
+export async function fetchGitHubRootTextFileAtRef(params: {
+  installationId: string;
+  owner: string;
+  repo: string;
+  path: string;
+  ref: string;
+  appType?: GitHubAppType;
+}): Promise<string | null> {
+  const { installationId, owner, repo, path, ref, appType = 'standard' } = params;
+  const tokenData = await generateGitHubInstallationToken(installationId, appType);
```
alex-alecu: Could this helper accept the existing installation token so each review does not perform a second token-generation request just to fetch `REVIEW.md`?

apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
alex-alecu: Could `fetchRepositoryReviewInstructions` be added to the `Promise.all([...])` on line 194 instead of being awaited sequentially before it? The four GitHub requests look independent, so this currently adds one full round-trip of latency to every review. Same pattern on the GitLab branch at line 262 vs the `Promise.all` on line 283.

apps/web/src/lib/integrations/platforms/github/adapter.ts
alex-alecu: Done in 9f6d245d3.

apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
alex-alecu: Done in 8baa8a5e7.

alex-alecu: Done in 8baa8a5e7 and 8d90d5334.

apps/web/src/lib/code-reviews/prompts/repository-review-instructions.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  const content = rawContent
+    .replace(/\r\n/g, '\n')
+    .replace(/\r/g, '\n')
+    .split('')
+    .filter(character => !isUnsafeControlCharacter(character))
+    .join('')
+    .trim();
+
+  if (content.trim().length === 0) return null;
```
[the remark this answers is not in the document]
alex-alecu: Done in 7c29553.

alex-alecu:
Manual test passed.

Tested:
- Dev-only code review payload preparation flow for a seeded GitHub review row with REVIEW.md disabled/default-off.
- Prompt generation path reached prepareReviewPayload and generated a GitHub review payload without repository REVIEW.md guidance.

Verified:
- REVIEW.md metadata columns exist on cloud_agent_code_reviews.
- Persisted metadata remained repository_review_instructions_used=false, repository_review_instructions_ref=null, repository_review_instructions_truncated=false.
- Generated prompt used default review guidance and did not include the REVIEW.md instructions section.

packages/db/src/schema-types.ts
```diff
@@ -756,6 +756,7 @@ export const CodeReviewAgentConfigSchema = z.object({
   selected_repository_ids: z.array(z.number()).optional(),
   // Manually added repositories (for GitLab where pagination limits results)
   manually_added_repositories: z.array(ManuallyAddedRepositorySchema).optional(),
+  disable_review_md: z.boolean().optional(),
```
jrf0110, line 759: nit; It's just a little bit wonky that the field name is for controlling this feature is phrased as the negative (i.e. disable review.md?) rather than the positive `review_md_enabled`. Might consider adding a default value to the zod schema as well so undefined's don't need to be considered

jrf0110 (approved):

packages/db/src/schema-types.ts
alex-alecu: I'll address it in a followup PR prob next in which I enable by default REVIEW.md for all orgs

Orgs still need to add the REVIEW.md file

PR #3184, opened 2026-05-11
fix(kiloclaw): surface pending destroy cleanup

pandemicsyn:
## Summary

- Surface whether KiloClaw instance destruction has fully finalized provider cleanup or still has pending machine/volume resources.
- Add durable destroy timing state plus `reconcile.destroy_pending` / throttled `reconcile.destroy_stuck` telemetry so non-finalized cleanup remains visible while DO alarms keep retrying.
- Propagate the DO destroy result through `/api/platform/destroy`, log confirmed/pending/request-failed outcomes in the billing lifecycle sweep, and clarify the billing spec’s accepted-destroy vs provider-finalized semantics.

## Verification

- No manual verification was performed; this change is backend lifecycle telemetry/contract behavior and was verified with automated service tests and type checks instead.

## Visual Changes

N/A

## Reviewer Notes

- Billing still proceeds with local destroyed-state transitions per the existing spec; this PR makes pending provider cleanup explicit rather than blocking billing on immediate Fly deletion.
- The DO remains the retry owner: pending IDs stay in durable state, provider 404 counts as confirmed deletion, and stuck telemetry is throttled after 15 minutes and then hourly.

### DIFF
--- a/services/kiloclaw/src/durable-objects/kiloclaw-instance/reconcile.ts
+++ b/services/kiloclaw/src/durable-objects/kiloclaw-instance/reconcile.ts
@@ -110,7 +110,7 @@ export function emitDestroyPendingTelemetry(
   doWarn(state, 'Destroy incomplete, alarm will retry', details);
 }
 
-async function maybeEmitDestroyStuckTelemetry(
+export async function maybeEmitDestroyStuckTelemetry(
   ctx: DurableObjectState,
   state: InstanceMutableState,
   rctx: ReconcileContext
--- a/services/kiloclaw/src/durable-objects/kiloclaw-instance/reconcile.ts
+++ b/services/kiloclaw/src/durable-objects/kiloclaw-instance/reconcile.ts
@@ -11,6 +11,8 @@ import {
   RESTARTING_MAX_TIMEOUT_MS,
   RECOVERING_TIMEOUT_MS,
   WORKER_CONTROLLER_CAPABILITIES_VERSION,
+  DESTROY_STUCK_THRESHOLD_MS,
+  DESTROY_STUCK_TELEMETRY_INTERVAL_MS,
   getProactiveRefreshThresholdMs,
 } from '../../config';
 import { resolveInstanceTypeLabel } from '@kilocode/kiloclaw-instance-tiers';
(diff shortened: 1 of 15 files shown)


services/kiloclaw/src/durable-objects/kiloclaw-instance/reconcile.ts
```diff
@@ -1318,7 +1397,10 @@ async function retryPendingDestroy(
   await recoverBoundMachineForDestroy(flyConfig, ctx, state, rctx);
   await tryDeleteMachine(flyConfig, ctx, state, rctx);
   await tryDeleteVolume(flyConfig, ctx, state, rctx);
-  await finalizeDestroyIfComplete(ctx, state, rctx, markDestroyedInPostgres);
+  const result = await finalizeDestroyIfComplete(ctx, state, rctx, markDestroyedInPostgres);
+  if (!result.finalized) {
```
jeanduplessis, line 1401: **WARNING:** This stuck-destroy telemetry is only wired into the Fly retry path. `retryPendingDestroy()` is reached from `reconcileWithFly()` when `state.status === 'destroying'`, but the non-Fly alarm branch in `index.ts` calls `retryNonFlyDestroy()` + `finalizeDestroyIfComplete(...)` and ignores the result before rescheduling. That means Northflank/Docker destroy retries can remain pending with `lastDestroyError*` persisted but never emit the new 15-minute/hourly `reconcile.destroy_stuck` telemetry promised by the PR.

Trace:
1. This changed line emits stuck telemetry only after Fly `retryPendingDestroy`.
2. The non-Fly alarm `destroying` path was checked in `index.ts` and does not call this helper or `emitDestroyPendingTelemetry`.
3. Non-Fly providers have real pending destroy states (`northflank`, `docker-local`), so repeated provider cleanup failures retry silently.

Impact: Pending non-Fly provider cleanup can retry indefinitely without the new stuck telemetry, making failed Northflank/Docker resource deletion much harder to detect operationally.

jeanduplessis (approved):
Found one warning: the new stuck-destroy telemetry is wired for Fly alarm retries but not for the non-Fly `destroying` alarm path. No other verified issues found in the changed lines.

PR #3196, opened 2026-05-12
fix(gateway): strip embedding dimensions before upstream requests

marius-kilocode:
The Kilo indexing client uses the catalog's `dimension` metadata to size local vector stores. That value is client/control-plane metadata, not a safe upstream request parameter for every embedding provider.

After restoring the catalog response in https://github.com/Kilo-Org/cloud/pull/3193, clients can again hydrate `modelDimension` from Cloud and send `dimensions` to `/api/openrouter/embeddings`. The gateway then forwarded that field unchanged to OpenRouter/provider APIs. Models like Mistral/Codestral reject it with `extra_forbidden` because their embeddings endpoint does not accept a `dimensions` override.

This strips `dimensions` from the upstream request body in the same place we already strip deprecated/client-only fields like `user` and Mistral-specific `output_dimension`/`output_dtype`. The gateway can still accept the field from clients for compatibility, but upstreams only receive supported payload fields.

### DIFF
--- a/apps/web/src/lib/ai-gateway/embeddings/embedding-request.test.ts
+++ b/apps/web/src/lib/ai-gateway/embeddings/embedding-request.test.ts
@@ -2,7 +2,7 @@ import { describe, it, expect } from '@jest/globals';
 import { buildUpstreamBody } from './embedding-request';
 
 describe('buildUpstreamBody', () => {
-  it('should forward all standard fields and strip Mistral-specific and deprecated fields', () => {
+  it('should forward supported fields and strip client-only, Mistral-specific, and deprecated fields', () => {
     const result = buildUpstreamBody({
       model: 'google/text-embedding-004',
       input: ['text1', 'text2'],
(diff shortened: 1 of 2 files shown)


imanolmzd-svg (approved):

johnnyeric (approved):

marius-kilocode:
Updated this PR after the follow-up `Bad Request` report.

The client sends `encoding_format: "base64"` because the OpenAI SDK used by the extension has historically parsed numeric embeddings incorrectly. That field is also OpenAI SDK/API-specific and is not safe to forward to all upstream embedding providers through OpenRouter. I now strip it from the upstream request together with `dimensions`.

To preserve the client contract, the proxy now checks the original request. If the client requested `encoding_format: "base64"` and upstream returns numeric embeddings, Cloud converts those numeric float arrays back into OpenAI-style base64 embeddings before returning the response to the client. So upstreams no longer see unsupported request fields, while the Kilo client still receives the base64 response shape it expects.

Validation:

- `pnpm --filter web test -- apps/web/src/lib/ai-gateway/embeddings/embedding-request.test.ts`
- `pnpm run format:check`

PR #3197, opened 2026-05-12
fix(gateway): correct Codestral embedding dimension

marius-kilocode:
Kilo indexing uses `/api/gateway/embedding-models` as the source of truth for local vector-store dimensions. After #3196 stopped forwarding `dimensions` to upstream providers, `mistralai/codestral-embed-2505` returns its native 1536-dimensional vectors.

The catalog still advertised Codestral as 256 dimensions, so the client created a 256-dimensional Qdrant collection and then upserted 1536-dimensional vectors. Qdrant rejects that with `Wrong input: Vector dimension error: expected dim: 256, got 1536`, which surfaced in the extension as a generic `Bad Request`.

This updates the Codestral catalog entry to 1536 so the vector store is created with the dimension the gateway actually returns.


marius-kilocode:
Was fixed elsewhere

PR #3210, opened 2026-05-12
feat(kiloclaw) morning briefing config table

St0rmz1:
## Summary

Adds a morning briefing interests feature. Users can pick news topics during onboarding or from settings, and the briefing scopes its daily web search to those topics.

A new Postgres table holds desired state for the morning briefing: an enabled flag, cron, timezone, and the list of interest topics. The plugin running inside the instance container is still authoritative for runtime behavior. The kiloclaw worker is the only writer to this table, matching the ownership pattern used for the kiloclaw_instances table.

The onboarding wizard gains a step where admins pick topics from a preset list or type their own. Picking at least one topic also turns the morning briefing on, using the browser timezone. Settings has a matching editor so admins can change their topics later. Both surfaces use the same tRPC mutation.

The kiloclaw plugin receives the topics through a new HTTP endpoint and persists them into its local config file. On each briefing run the plugin builds the web search query from those topics. When no topics are present the plugin falls back to the original default query.

The morning briefing section in settings is now a collapsible card so the page stays compact when no action is required.

Users without admin access go through onboarding without seeing the interests step.

## Verification

- [x] Onboarding flow as an admin. Pick a few topics. Confirm the briefing turns on and the next run uses the chosen topics.
- [x] Settings page. Open the morning briefing card, change interests, confirm save succeeds.
- [ ] Standard account through onboarding. Confirm the interests step is hidden and the wizard completes without it.
- [x] Older kiloclaw image. Confirm settings shows the upgrade prompt instead of the editor.

## Visual Changes

| Before | After |
| ------ | ----- |
|        |       |

## Reviewer Notes

The plugin runs inside the kiloclaw container image, so the new plugin route ships with the next image build. Adding any new gateway path under \`/_kilo/morning-briefing/\` requires three updates: a plugin handler, a controller proxy, and a worker helper. Each layer allowlists the path explicitly. Skipping the controller proxy returns a 404 that surfaces as \`controller_route_unavailable\` from the worker.

The Postgres write happens through \`waitUntil\` and tolerates failure. The plugin config file remains authoritative for runtime behavior. If the Postgres write fails the briefing still works, and the row is repaired lazily on the next status read.

Topic caps (20 entries, with each topic limited to 64 characters) are enforced at the tRPC layer, the worker route, and the plugin handler. The Zod trim runs at the worker boundary, with a matching trim inside the plugin as defense in depth.

Concurrent writes to \`config.json\` on the plugin side are serialized through a write queue keyed by the config file path, so an interests save and a reconcile pass cannot clobber each other.

The morning briefing card in settings matches the google account section style: collapsed by default with a status line, expand to manage.


services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
```diff
@@ -3640,24 +3642,132 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
...
+          pg.instanceId
+        )
+      );
+    }
+
+    // Augment the plugin response with interest_topics from Postgres.
+    // Other fields (enabled/cron/timezone) come from the plugin so a
+    // running-state mismatch (e.g. operator-edited config.json) surfaces.
+    if (pg?.row) {
+      return { ...pluginStatus, interestTopics: pg.row.interest_topics };
```
jeanduplessis: ⚠️ WARNING: This can permanently surface stale interest topics after a best-effort Postgres mirror failure.

Trace:
1. `updateBriefingInterests` writes the plugin first, then mirrors `interestTopics` to Postgres in `waitUntil`; `syncMorningBriefingConfigToPostgresHelper` catches and swallows DB failures.
2. `getMorningBriefingStatus` reads both plugin and Postgres, but when `pg.row` exists it overwrites `pluginStatus.interestTopics` with `pg.row.interest_topics`.
3. The lazy repair path only runs when `pg.row === null`, so a transient Hyperdrive/write failure after the plugin accepted a change leaves an existing row stale; later status reads keep returning the stale row instead of the plugin’s authoritative config.

Impact: Settings and any Postgres-backed analytics/admin reads can show old topics while the plugin actually runs with new topics. Clearing or changing topics can appear to fail or silently revert in the UI whenever the best-effort mirror write fails after a successful plugin write.

apps/web/src/routers/kiloclaw-router.ts
```diff
@@ -2838,6 +2842,23 @@ export const kiloclawRouter = createTRPCRouter({
     return client.runMorningBriefing(ctx.user.id, workerInstanceId(instance));
   }),
 
+  updateBriefingInterests: clawAccessProcedure
```
jeanduplessis, line 2845: ⚠️ WARNING: The admin-only interests gate is only enforced in the UI, not on this mutation.

Trace:
1. The changed onboarding/settings code gates the interests UI on `currentUser?.is_admin === true` / `canSeeMorningBriefing = !!user?.is_admin`.
2. This new personal mutation uses `clawAccessProcedure`, which checks KiloClaw access/subscription but not `ctx.user.is_admin`; the org variant similarly uses `organizationMemberMutationProcedure`, which allows any organization member with an active subscription/trial.
3. A non-admin can bypass the hidden UI by calling the new tRPC mutation directly and update the plugin/Postgres interest topics.

Impact: The admin-only rollout contract is bypassed. Non-admin personal users, or non-admin org members, can change briefing topics for an instance even though both onboarding and settings intentionally hide the feature from them.

services/kiloclaw/plugins/kiloclaw-morning-briefing/src/index.lifecycle.test.ts
```diff
@@ -942,4 +953,234 @@ describe('morning briefing lifecycle', () => {
...
+        response
+      );
+
+      expect(response.statusCode).toBe(200);
+      const configPath = path.join(harness.stateDir, 'morning-briefing', 'config.json');
+      const stored = (await readJson(configPath)) as { interestTopics: string[] };
+      expect(stored.interestTopics).toEqual(['Tech']);
+    });
+
+    it('serialises concurrent interest writes so neither is silently clobbered', async () => {
```
jeanduplessis: 💡 SUGGESTION: This test does not verify the serialization race it claims to cover.

Trace:
1. The test claims to verify serialized concurrent writes, but it accepts either complete input as the final on-disk value.
2. An unserialized last-writer-wins implementation would also typically leave one of the two complete inputs in `config.json`, so this assertion would still pass without proving the queue ran.
3. The production race described in the PR is a slow reconcile holding a stale `StoredConfig` and clobbering an interests write; this test does not run reconcile or assert preservation of unrelated config fields across that path.

Impact: A regression in the config write queue, especially the reconcile-vs-interests clobber path this PR is trying to protect, can pass the new test while still losing user-selected topics.

jeanduplessis (approved):
Found two concrete correctness issues and one test coverage gap:

- Status can keep showing stale interest topics when the plugin write succeeds but the best-effort Postgres mirror fails for an existing row.
- The new interests mutation does not enforce the admin-only gate server-side, so hidden UI can be bypassed.
- The new concurrent-write test does not actually prove serialization or cover the reconcile clobber path it describes.

services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
St0rmz1: Fixed. `getMorningBriefingStatus` now treats the plugin as authoritative for `interestTopics`. When the plugin reports topics and the Postgres row disagrees, the plugin value is returned on the current read and a `waitUntil` write fires so the row catches up. Lazy backfill on `pg.row === null` still covers legacy instances. The new path covers the case where the row exists but is stale, which was the gap.

apps/web/src/routers/kiloclaw-router.ts
St0rmz1: Fixed. Added an explicit `if (!ctx.user.is_admin) throw FORBIDDEN` at the top of `updateBriefingInterests` in both the personal router and the org variant. Matches the UI gate. A direct tRPC call from a user without admin access now fails in the mutation handler.

services/kiloclaw/plugins/kiloclaw-morning-briefing/src/index.lifecycle.test.ts
St0rmz1: Fixed. Replaced with a test that uses a barrier to deterministically reproduce the reconcile clobber path. The harness now accepts an optional `cronAddBarrier` promise. The test preloads a config with `enabled: false`, fires enable so reconcile starts, holds reconcile inside `cron add`, fires an interests update while reconcile waits, releases the barrier, and asserts that the final config has reconcile's enable spread (`enabled: true`, `cronJobId` set) and the fresh `interestTopics: ['Tech', 'AI']`, plus `cron` and `timezone` preserved from the preload. Without `queueConfigWrite`, the spread would clobber `interestTopics` back to an empty array.

PR #3219, opened 2026-05-12
fix(kiloclaw) admin polling wake guard

St0rmz1:
## Summary

Polling against a stopped KiloClaw instance was causing Fly's HTTPS proxy to wake the machine within seconds of any stop request, even with `services[0].autostart: false` set on the machine config. Fly treats that flag as a hint rather than a guarantee for single machine apps, so an inflight request from the admin UI's polling loop would trigger a fresh start. Workflows that rely on a stable stopped state (resize, tier change, volume work) were effectively unusable from the admin UI.

This change adds a guard at the worker layer for the three polling routes that proxy through Fly's edge to port 18789:

- `/api/platform/gateway/status`
- `/api/platform/controller-version`
- `/api/platform/morning-briefing/status`

Each route now reads the Durable Object status first via `stub.getStatus()` and short circuits when the status is anything other than `running`. The response in that case is HTTP 200 with the payload `{ ok: false, reason: 'instance_not_running', status }`. Mirrors the existing pattern at `/api/platform/gateway/ready` so polling does not flood logs and Sentry with 5xx noise. The downstream Durable Object method is never called, which means no traffic reaches Fly's proxy and the machine stays asleep.

`/api/platform/debug-status` is intentionally not guarded. `getDebugState` reads Durable Object storage directly and never proxies through Fly's edge, so the wake bug does not apply there.

Frontend changes:

- Response types in `apps/web/src/lib/kiloclaw/types.ts` become a discriminated union of the existing ok shape and the new `InstanceNotRunningSentinel` variant.
- New narrowing helpers (`gatewayStatusOk`, `controllerVersionOk`, `morningBriefingStatusOk`) drop the sentinel for consumers that only care about the ok payload.
- The admin instance detail page now renders an inline "Instance is X. Start the machine to see gateway status." hint during the brief window between a user stop and the page picking up the new instance state.
- The sentinel type and predicate live in `apps/web/src/lib/kiloclaw/instance-not-running-sentinel.ts`. The mobile app imports the same module through a thin re-export at `apps/mobile/src/lib/kiloclaw/instance-not-running-sentinel.ts` so Metro can resolve it at bundle time (Metro does not honor `compilerOptions.paths` and the existing custom resolver in `apps/mobile/metro.config.js` only remaps for `cloud-agent-sdk` callers).

## Verification

- [x] Reproduced the original bug on a Fly instance by stopping via the admin UI with the page open. Fly event log showed `start source=proxy` within 2 seconds of the user stop, and `fly logs` showed the proxy layer logging `Starting machine` from inside the stopping window.
- [x] Confirmed the same repro after the fix did not wake the machine.
- [x] `pnpm run typecheck` passes for the entire workspace.
- [x] `pnpm run lint` passes for the entire workspace.
- [x] `pnpm run format:check` passes.
- [x] The `services/kiloclaw` test suite runs 1852 tests, including 10 new tests in `services/kiloclaw/src/routes/platform-poll-short-circuit.test.ts` that cover stopped, starting, and running states across the three guarded routes plus an assertion that `/debug-status` is intentionally not guarded.

## Visual Changes

N/A. The only visible change in the admin UI is a brief stopped state hint during the window between a user stop and the page picking up the new instance state. That branch was previously unreachable because the polling either succeeded or threw an error toast.

## Reviewer Notes

- The guard adds one Durable Object RPC per poll. Status reads are inexpensive storage reads. KV caching was considered but introduces a correctness problem: a cached `running` value within a TTL window would let one polling request through and wake the machine, regressing the bug this PR fixes. Accepting the extra RPC is the safer tradeoff.
- Mobile shares the sentinel module with web rather than duplicating it. The web file is canonical, mobile imports through a thin re-export so any future shape changes only need to land in one place.
- If Durable Object status drifts from the real machine state, the guard could return a sentinel for a machine that is actually running. That is harmless in practice because the next poll succeeds once status reconciles, and the UI handles the transient sentinel by rendering the stopped state hint.


services/kiloclaw/src/routes/platform.ts
```diff
@@ -709,6 +709,45 @@ function jsonError(message: string, status: number, code?: string): Response {
...
+  instanceId: string | undefined
+): Promise<Response | null> {
+  const status = await withResolvedDORetry(
+    env,
+    userId,
+    instanceId,
+    stub => stub.getStatus(),
+    'getStatus'
+  );
+  if (status.status === 'running') return null;
```
jeanduplessis: ⚠️ The guard still falls through when `getStatus()` returns stale `running`.

Trace:
1. `shortCircuitIfNotRunning()` returns `null` whenever DO status is `running`, so the route continues to the proxied controller call.
2. `getStatus()` returns cached DO state and only dispatches the live Fly check via `waitUntil`, so provider drift or an in-flight stop can still leave this value as `running`.
3. In that stale/racing state, the next call to `getGatewayProcessStatus()`, `getControllerVersion()`, or `getMorningBriefingStatus()` still reaches Fly’s HTTPS proxy and can wake the machine.

Impact: polling can still resurrect a machine during stop/resize/tier/volume workflows, which is the core bug this PR is trying to prevent.

jeanduplessis (changes requested):
Found one remaining edge case in the wake guard: it trusts cached DO `running` status, so stale/racing `running` can still fall through to the Fly-proxied controller call and wake the machine. The rest of the sentinel plumbing and UI narrowing looked consistent.

services/kiloclaw/src/routes/platform.ts
St0rmz1: fixed in https://github.com/Kilo-Org/cloud/pull/3219/changes/4d6ed582c049856b734d77c242e1af2f34512a86

jeanduplessis (approved):

PR #3222, opened 2026-05-12
feat(integrations): add DoltHub OAuth integration (dev-only)

jrf0110:
l## Summary
Add a complete DoltHub OAuth integration gated to development environments only. The DoltHub OAuth app is pending admin approval, so all UI surfaces, API routes, and backend logic are disabled in production.

Changes include:
- **Platform definition**: Added `dolthub` to integration constants and platform definitions, enabled only when `IS_DEVELOPMENT` is true.
- **OAuth service module** (`lib/integrations/dolthub-service.ts`): Authorization URL generation, token exchange, token refresh, user info fetch, installation CRUD, and token validation with automatic refresh.
- **tRPC router** (`routers/dolthub-router.ts`): `getInstallation`, `disconnect`, and `verifyToken` endpoints — all return 404/not-found in production.
- **API routes**: `/api/integrations/dolthub/connect` (initiates OAuth) and `/api/integrations/dolthub/callback` (handles callback, with CSRF-protected state verification).
- **UI pages**: `/integrations/dolthub` and `/organizations/[id]/integrations/dolthub` with `notFound()` in production.
- **Integration card**: Wired into `IntegrationsHub` with loading and installation status.
- **Environment variables**: `DOLTHUB_APP_DEV_CLIENT_ID` and `DOLTHUB_APP_DEV_CLIENT_SECRET` added to `.env.local.example`, `.env.test`, and `config.server.ts`.
- **Tests**: Comprehensive unit tests for `dolthub-service` (token flows, installation CRUD, token refresh) and `dolthub-router` (production guards, verify token smoke test).

## Verification
- Verified lint (`oxlint`) and typecheck (`tsgo`) pass.
- Tests were reviewed for coverage of production guards, success paths, and error paths. (Database-dependent tests could not be executed in this environment due to missing Postgres/Docker.)

## Visual Changes

image

image

image

image

## Reviewer Notes
- All entry points (pages, API routes, tRPC endpoints, service functions) have independent production guards.
- No database schema changes; reuses the existing `platform_integrations` table.
- Client ID stored in `.env.local.example` is a public OAuth client ID (not a secret).

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -53,7 +53,7 @@ DISCORD_CLIENT_ID=your-discord-client-id
 DISCORD_CLIENT_SECRET=your-discord-client-secret
 
 # DoltHub OAuth (dev only — app pending admin approval)
-DOLTHUB_APP_DEV_CLIENT_ID=dhoci.v1.tdg4bfv15v24adbpc2fqeqddugotg64nd9puisim322d0p452i50
+DOLTHUB_APP_DEV_CLIENT_ID=your-dolthub-client-id
 DOLTHUB_APP_DEV_CLIENT_SECRET=your-dolthub-client-secret
 # ============================================================================
 # Billing & Stripe (Required for billing features)
--- a/.env.local.example
+++ b/.env.local.example
@@ -51,6 +51,10 @@ LINKEDIN_CLIENT_SECRET=your-linkedin-client-secret
 # Discord OAuth (optional)
 DISCORD_CLIENT_ID=your-discord-client-id
 DISCORD_CLIENT_SECRET=your-discord-client-secret
+
+# DoltHub OAuth (dev only — app pending admin approval)
+DOLTHUB_APP_DEV_CLIENT_ID=dhoci.v1.tdg4bfv15v24adbpc2fqeqddugotg64nd9puisim322d0p452i50
+DOLTHUB_APP_DEV_CLIENT_SECRET=your-dolthub-client-secret
 # ============================================================================
 # Billing & Stripe (Required for billing features)
 # ============================================================================
--- a/apps/web/src/app/api/integrations/dolthub/callback/route.ts
+++ b/apps/web/src/app/api/integrations/dolthub/callback/route.ts
@@ -1,0 +1,163 @@
+import type { NextRequest } from 'next/server';
+import { NextResponse } from 'next/server';
+import { getUserFromAuth } from '@/lib/user.server';
+import { ensureOrganizationAccess } from '@/routers/organizations/utils';
+import type { Owner } from '@/lib/integrations/core/types';
+import { captureException, captureMessage } from '@sentry/nextjs';
+import { IS_DEVELOPMENT } from '@/lib/constants';
+import {
+  exchangeDoltHubOAuthCode,
+  fetchDoltHubUser,
+  upsertDoltHubInstallation,
+} from '@/lib/integrations/dolthub-service';
+import { verifyOAuthState } from '@/lib/integrations/oauth-state';
+import { APP_URL } from '@/lib/constants';
+
+function buildDoltHubRedirectPath(state: string | null, queryParam: string): string {
+  const verified = state ? verifyOAuthState(state) : null;
+  const owner = verified?.owner;
+
+  if (owner?.startsWith('org_')) {
+    return `/organizations/${owner.replace('org_', '')}/integrations/dolthub?${queryParam}`;
+  }
+  if (owner?.startsWith('user_')) {
+    return `/integrations/dolthub?${queryParam}`;
+  }
+  return `/integrations?${queryParam}`;
+}
+
+/**
+ * DoltHub OAuth Callback
+ *
+ * Called when user completes the DoltHub OAuth flow.
+ * Exchanges the authorization code for tokens and stores the integration.
+ */
+export async function GET(request: NextRequest) {
+  if (!IS_DEVELOPMENT) {
+    return NextResponse.json({ error: 'not_found' }, { status: 404 });
+  }
+
+  try {
+    const { user, authFailedResponse } = await getUserFromAuth({ adminOnly: false });
+    if (authFailedResponse) {
+      return NextResponse.redirect(new URL('/users/sign_in', APP_URL));
+    }
+
+    const searchParams = request.nextUrl.searchParams;
+    const code = searchParams.get('code');
+    const state = searchParams.get('state');
+    const error = searchParams.get('error');
+
+    if (error) {
+      captureMessage('DoltHub OAuth error', {
+        level: 'warning',
+        tags: { endpoint: 'dolthub/callback', source: 'dolthub_oauth' },
+        extra: { error, state },
+      });
+
+      return NextResponse.redirect(
+        new URL(buildDoltHubRedirectPath(state, `error=${encodeURIComponent(error)}`), APP_URL)
+      );
+    }
+
+    if (!code) {
+      captureMessage('DoltHub callback missing code', {
+        level: 'warning',
+        tags: { endpoint: 'dolthub/callback', source: 'dolthub_oauth' },
+        extra: { state, allParams: Object.fromEntries(searchParams.entries()) },
+      });
+
+      return NextResponse.redirect(
+        new URL(buildDoltHubRedirectPath(state, 'error=missing_code'), APP_URL)
+      );
+    }
+
+    const verified = verifyOAuthState(state);
+    if (!verified) {
+      captureMessage('DoltHub callback invalid or tampered state signature', {
+        level: 'warning',
+        tags: { endpoint: 'dolthub/callback', source: 'dolthub_oauth' },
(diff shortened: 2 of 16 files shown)


.env.local.example
```diff
@@ -51,6 +51,10 @@ LINKEDIN_CLIENT_SECRET=your-linkedin-client-secret
 # Discord OAuth (optional)
 DISCORD_CLIENT_ID=your-discord-client-id
 DISCORD_CLIENT_SECRET=your-discord-client-secret
+
+# DoltHub OAuth (dev only — app pending admin approval)
+DOLTHUB_APP_DEV_CLIENT_ID=dhoci.v1.tdg4bfv15v24adbpc2fqeqddugotg64nd9puisim322d0p452i50
+DOLTHUB_APP_DEV_CLIENT_SECRET=your-dolthub-client-secret
```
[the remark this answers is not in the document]
jrf0110: Fixed — replaced the real client ID with `your-dolthub-client-id` placeholder in `.env.local.example`.

apps/web/src/app/api/integrations/dolthub/callback/route.ts
```diff
@@ -0,0 +1,163 @@
...
+    if (owner.type === 'org') {
+      await ensureOrganizationAccess({ user }, owner.id);
+    } else if (user.id !== owner.id) {
+      return NextResponse.redirect(new URL('/integrations?error=unauthorized', APP_URL));
+    }
+
+    const tokens = await exchangeDoltHubOAuthCode(code);
+
+    const userInfo = await fetchDoltHubUser(tokens.accessToken);
+    const username = userInfo?.username ?? 'unknown';
```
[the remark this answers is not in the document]
jrf0110: Fixed — when `fetchDoltHubUser` returns no username, we now abort the installation and redirect with `error=user_info_failed` instead of falling back to `unknown` and proceeding.

apps/web/src/app/api/integrations/dolthub/callback/route.ts
```diff
@@ -0,0 +1,151 @@
...
+        new URL(buildDoltHubRedirectPath(state, 'error=missing_code'), APP_URL)
+      );
+    }
+
+    const verified = verifyOAuthState(state);
+    if (!verified) {
+      captureMessage('DoltHub callback invalid or tampered state signature', {
+        level: 'warning',
+        tags: { endpoint: 'dolthub/callback', source: 'dolthub_oauth' },
+        extra: { code: '***', state, allParams: Object.fromEntries(searchParams.entries()) },
```
jeanduplessis: **CRITICAL:** Invalid-state logging leaks the OAuth authorization code via `allParams`

**Warning:** This branch redacts `code` separately, but `allParams: Object.fromEntries(searchParams.entries())` includes the original callback query string, including the raw `code` whenever DoltHub redirects back with both `code` and an invalid/tampered/expired `state`.

**Trace:**
1. The callback reads `const code = searchParams.get('code')`.
2. On invalid state, the Sentry payload sets `code: '***'` but also logs `allParams`.
3. `allParams` reintroduces the sensitive authorization code into Sentry/logging.

**Impact:** A DoltHub OAuth authorization code can be stored in Sentry when state verification fails. Authorization codes are short-lived credentials exchangeable for tokens, so this is a credential exposure path.

apps/web/src/lib/integrations/dolthub-service.ts
```diff
@@ -0,0 +1,275 @@
...
+    const newTokens = await refreshDoltHubAccessToken(metadata.refresh_token);
+    const newExpiresAt = newTokens.expiresIn ? Date.now() + newTokens.expiresIn * 1000 : null;
+
+    await db
+      .update(platform_integrations)
+      .set({
+        metadata: {
+          ...metadata,
+          access_token: newTokens.accessToken,
+          refresh_token: newTokens.refreshToken,
```
jeanduplessis: **WARNING:** Refresh can erase a still-valid refresh token when DoltHub omits a replacement

**Warning:** The parser allows refresh responses without `refresh_token`, but the refresh persistence path always writes `refresh_token: newTokens.refreshToken`. If DoltHub returns only a new access token, this overwrites the existing refresh token with `null`.

**Trace:**
1. `refresh_token` is optional in `DoltHubTokenPayloadSchema`.
2. Missing `refresh_token` becomes `newTokens.refreshToken === null`.
3. The update persists that `null` over `metadata.refresh_token`.

**Impact:** The integration continues until the refreshed access token expires, then future refresh attempts fail because the previously valid refresh token was erased. Users must reconnect unnecessarily.

apps/web/src/lib/integrations/dolthub-service.ts
```diff
@@ -0,0 +1,275 @@
...
+      .returning();
+
+    if (!updated) {
+      throw new Error('DoltHub installation update returned no rows');
+    }
+
+    return updated;
+  }
+
+  const [created] = await db
```
jeanduplessis: **WARNING:** Concurrent OAuth completions can create duplicate active installations for the same owner

**Warning:** The DoltHub “upsert” is a select-then-insert without a transaction/conflict target, and inserted rows do not set `platform_installation_id`. The existing unique indexes include `platform_installation_id`, and Postgres allows multiple `NULL` values, so duplicate rows for the same owner/platform are not prevented.

**Trace:**
1. Two OAuth callbacks for the same owner can both run `getInstallation(owner)` before either insert commits.
2. Both then insert active DoltHub rows with no `platform_installation_id`.
3. Later `getInstallation(...).limit(1)` and `uninstall` operate on only one arbitrary row.

**Impact:** Duplicate/retry OAuth completion can leave multiple active DoltHub token rows. Disconnect may delete one while another active token remains in the database, causing stale installation state and credential cleanup gaps.

jeanduplessis (changes requested):
Found three issues that should be addressed before merge: one credential-logging issue in the invalid OAuth state path, one token refresh persistence bug that can drop a valid refresh token, and one duplicate-row race in the DoltHub installation upsert path. I also confirmed the previous review threads are resolved and did not find PostHog tracking changes in this PR.

apps/web/src/app/api/integrations/dolthub/callback/route.ts
jrf0110: Fixed — redacted the authorization code in allParams before logging by mapping search params and replacing code values with '***'.

apps/web/src/lib/integrations/dolthub-service.ts
jrf0110: Fixed — token refresh now preserves the existing refresh_token when the DoltHub response omits it, using newTokens.refreshToken ?? metadata?.refresh_token.

apps/web/src/lib/integrations/dolthub-service.ts
jrf0110: Fixed — upsertDoltHubInstallation now sets platform_installation_id to a synthetic value (dolthub-{type}-{id}), which makes the existing partial unique indexes enforce one installation per owner. uninstall was also updated to delete all matching rows to prevent cleanup gaps.

jeanduplessis (approved):

apps/web/src/app/api/integrations/dolthub/callback/route.ts
jrf0110: Follow-up in 5ad46eccc: extended the same redaction to the missing-code branch (it was still passing raw `searchParams.entries()` into `allParams`). Hoisted `redactedParams` so any future logging path in this handler is safe by default — no individual call site has to remember to filter.

apps/web/src/lib/integrations/dolthub-service.ts
jrf0110: Follow-up in 5ad46eccc: also fall back to `metadata.scope` when DoltHub omits `scope` in the refresh response — RFC 6749 says an omitted `scope` retains the prior value, same rule as `refresh_token`. The previous fix only handled `refresh_token`, so a refresh response without `scope` was still wiping the stored scope to `null`.

Added a test (`preserves existing refresh_token and scope when refresh response omits them`) that fakes a refresh response with neither `refresh_token` nor `scope` and asserts both are kept.

apps/web/src/lib/integrations/dolthub-service.ts
jrf0110: Follow-up in 5ad46eccc: the synthetic `platform_installation_id` made duplicate rows constraint-illegal at the DB level, but the surrounding code was still doing select-then-insert. That left a narrower race: two concurrent OAuth completions where neither sees an existing row both attempt the insert; one wins and the loser surfaces a Postgres unique-violation as an unhandled error to the user, even though their token was successfully committed by the winner.

Converted `upsertDoltHubInstallation` to `onConflictDoUpdate` against the existing partial unique index (`UQ_platform_integrations_owned_by_(user|org)_platform_inst`), which collapses both paths into a single atomic write — both callers now get a successful row back and the last writer wins on metadata.

Added a test `concurrent upserts for the same owner produce a single row` that fires three parallel `upsertDoltHubInstallation` calls and asserts exactly one row remains.

PR #3224, opened 2026-05-13
fix(code-reviews): retry failed attempts

alex-alecu:
## Summary
- Keep per-attempt code review history while preserving a stable parent `codeReviewId` for the review page and external status links.
- Route each new retry attempt to a fresh Durable Object keyed by `reviewId:attemptId`, and thread `attemptId` through dispatch, status/events/cancel reads, retries, and callback processing so stale callbacks only update the attempt they belong to.
- Update the web dispatch and retry flows to use the current attempt row, treat terminal dispatch responses as failures, and start infra retries in a fresh retry-attempt DO instead of resetting the failed attempt in place.

## Verification
- Confirmed a retried code review now starts from a fresh attempt-specific worker identity instead of reusing stale terminal state from a prior attempt.
- Confirmed attempt-aware callbacks update only the matching attempt and stale callbacks do not overwrite the latest review state.
- Confirmed manual retrigger and infra retry flows both target the new attempt identity end to end.

## Visual Changes
N/A

## Reviewer Notes
- `retry-fresh` now requires the failed attempt identity plus the new retry attempt identity to spin up a fresh DO; legacy fallback routes remain in place for older callers that still use plain `reviewId`.
- Worker integration tests were updated to match the current internal sandbox-retry behavior, which keeps the original DO queued with retry intent while fresh retry attempts use a separate DO.


alex-alecu:
Manual test passed.

Tested:
- Automatic code-review retry flow: a running v2 review received a retryable `container shutdown: SIGTERM` failure, started a fresh retry, and completed on the retry session.
- Non-retryable failure flow: a maximum-runtime failure ended the review without creating an infra retry.
- Review detail Session Log attempt selector and admin Retry-aware metrics/all-attempts CSV flows.

Verified:
- The parent review stayed active during the retry and finished completed on the fresh session.
- Attempt history recorded failed attempt 1 and completed `infra_failure` attempt 2 with the retry link and distinct session IDs.
- The non-retryable failure created zero `infra_failure` attempts.
- The detail page defaulted to the completed retry attempt, switched to the failed attempt with `attemptId`, and showed the selected attempt session.
- Admin metrics changed between final-outcome and all-attempts modes, and all-attempts CSV included attempt columns.

services/code-review-infra/src/code-review-orchestrator.ts
```diff
@@ -696,6 +696,88 @@ export class CodeReviewOrchestrator extends DurableObject<Env> {
...
+        reviewId: this.state.reviewId,
+        reason: params.reason,
+        previousCloudAgentSessionId,
+        previousSessionId,
+        previousCliSessionId,
+        previousSandboxId,
+      }
+    );
+
+    await this.runWithCloudAgentNext();
```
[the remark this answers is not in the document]
alex-alecu: Fixed in 25b4af3.

apps/web/src/lib/code-reviews/db/code-reviews.ts
```diff
@@ -77,6 +122,352 @@ export async function getCodeReviewById(reviewId: string): Promise<CloudAgentCod
...
+        cliSessionId: params.cliSessionId,
+        executionId: params.executionId,
+      });
+    }
+
+    const shouldUpdateLatestPending =
+      latestAttempt.status === 'pending' ||
+      (params.sessionId !== undefined && latestAttempt.session_id === params.sessionId) ||
+      (params.cliSessionId !== undefined && latestAttempt.cli_session_id === params.cliSessionId) ||
+      (!latestAttempt.session_id && !latestAttempt.cli_session_id);
```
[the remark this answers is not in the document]
alex-alecu: Fixed in 3da3983.

apps/web/src/app/api/internal/code-review-status/[reviewId]/route.ts
```diff
@@ -529,6 +602,57 @@ export async function POST(
       });
     }
 
+    if (
+      isRetryableInfraFailure(status, terminalReason, errorMessage) &&
```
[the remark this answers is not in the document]
Comment (alex-alecu) [reply apps/web/src/app/api/internal/code-review-status/[reviewId]/route.ts]: Fixed in c45a98d.

services/code-review-infra/src/code-review-orchestrator.ts
```diff
@@ -696,6 +708,70 @@ export class CodeReviewOrchestrator extends DurableObject<Env> {
...
+      {
+        reviewId: this.state.reviewId,
+        failedAttemptId: this.state.attemptId,
+        retryAttemptId: params.retryAttemptId,
+        reason: params.reason,
+        status: started.status,
+      }
+    );
+
+    return started.status === 'queued' || started.status === 'running';
```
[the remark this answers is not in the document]
alex-alecu: Fixed in 8ceed22ba.

apps/web/src/lib/code-reviews/dispatch/dispatch-pending-reviews.ts
```diff
@@ -356,19 +391,12 @@ async function handleAmbiguousDispatchFailure(
...
-
-    logExceptInTest('[dispatchReview] Mirrored terminal Worker status after dispatch failure', {
+    logExceptInTest('[dispatchReview] Worker returned terminal status for fresh dispatch', {
       reviewId: review.id,
+      attemptId,
       status: workerStatus.status,
-      mirrored,
     });
-    return true;
+    return false;
```
alex-alecu: Could this leave the DB row queued when the probe finds a terminal DO state? Since `tryDispatchPendingReviews` treats `false` as a non-dispatch, the review can sit queued and retry the same terminal attempt.

services/code-review-infra/src/code-review-orchestrator.ts
```diff
@@ -210,6 +218,7 @@ export class CodeReviewOrchestrator extends DurableObject<Env> {
     this.state.sessionId = undefined;
     this.state.cliSessionId = undefined;
     this.state.sandboxId = undefined;
+    this.state.status = 'queued';
```
alex-alecu, line 222: How does this queued state get picked up after this run returns? The only alarm was set by `start()`, so if it already fired for the current `runReview`, this sandbox retry path appears to save `queued` without scheduling another run or creating the fresh retry attempt.

apps/web/src/lib/code-reviews/dispatch/dispatch-pending-reviews.ts
alex-alecu: Pre-PR this branch mirrored the worker's terminal status onto the parent review (via `updateCodeReviewStatusIfNonTerminal`); now it just returns false. Since `ensureCurrentCodeReviewAttemptFromReview` reuses the same `attempt.id` on every redispatch, if the original callback never landed, won't the stale-claim sweep keep redispatching this review every `STALE_CLAIM_MINUTES` against the same `reviewId:attemptId` DO that always returns terminal? Should we still mark the parent review/attempt failed here so the row can exit the recovery loop?

apps/web/src/lib/code-reviews/db/code-reviews.ts
```diff
@@ -77,6 +163,487 @@ export async function getCodeReviewById(reviewId: string): Promise<CloudAgentCod
...
+        codeReviewId: params.codeReviewId,
+        status: 'running',
+        sessionId: params.sessionId,
+        cliSessionId: params.cliSessionId,
+        executionId: params.executionId,
+      });
+    }
+
+    const shouldUpdateLatestPending =
+      latestAttempt.status === 'pending' ||
```
alex-alecu: When the latest attempt is a freshly created infra-retry (status `pending`, no session yet), this `shouldUpdateLatestPending` branch fires on any delayed `running` callback that lacks `attemptId` and overwrites the new retry's `session_id`/`cli_session_id` with the OLD session ids. After that, the superseded-session guard in route.ts will reject the real new-session callbacks as stale and the retry attempt's state never recovers. Should this legacy fallback refuse to overwrite an attempt that is clearly a retry (e.g. has `retry_of_attempt_id` set) when the caller didn't supply an `attemptId`?

apps/web/src/lib/integrations/platforms/gitlab/webhook-handlers/merge-request-handler.ts
```diff
@@ -198,25 +199,98 @@ export async function handleMergeRequestCodeReview(
...
+        cancelledReviews
+          .filter(
+            review =>
+              review.platform === 'gitlab' &&
+              review.platformProjectId != null &&
+              review.headSha.length > 0
+          )
+          .map(async review => {
+            try {
+              const pratToken = await getOrCreateProjectAccessToken(
```
alex-alecu: All `cancelledReviews` here belong to the same MR / same `platformProjectId`, but `getOrCreateProjectAccessToken` is called once per review inside `Promise.allSettled`, which involves a DB lookup and potentially a GitLab API call. Could the PAT be resolved once outside the map (or memoized per `platformProjectId`) and reused across the `setCommitStatus` calls?

apps/web/src/routers/admin-code-reviews-router.ts
```diff
@@ -615,18 +826,51 @@ export const adminCodeReviewsRouter = createTRPCRouter({
...
+        .innerJoin(
+          cloud_agent_code_review_attempts,
+          eq(cloud_agent_code_review_attempts.code_review_id, cloud_agent_code_reviews.id)
+        )
+        .where(and(...conditions))
+        .orderBy(
+          desc(cloud_agent_code_reviews.created_at),
+          desc(cloud_agent_code_review_attempts.attempt_number)
+        )
+        .limit(10000);
```
alex-alecu: `.limit(10000)` here is applied after the `innerJoin` with `cloud_agent_code_review_attempts`, so a review with N attempts consumes N rows of the cap; the non-`all_attempts` branch caps on review rows directly. Doesn't this mean the export silently returns far fewer reviews when retries exist, and can split a single review's attempts across the cutoff? Should the cap be on distinct review ids first (e.g. subselect top N reviews by `created_at`, then join attempts)?

services/code-review-infra/src/code-review-orchestrator.ts
```diff
@@ -696,6 +708,73 @@ export class CodeReviewOrchestrator extends DurableObject<Env> {
...
+
+    if (!params.retryAttemptId) {
+      return false;
+    }
+
+    this.state.sandboxRetryAttempted = true;
+    await this.saveState();
+
+    const retryId = this.env.CODE_REVIEW_ORCHESTRATOR.idFromName(
+      `${this.state.reviewId}:${params.retryAttemptId}`
```
alex-alecu: `services/code-review-infra/src/index.ts` adds a `doNameForAttempt(reviewId, attemptId)` helper that every external HTTP route (`POST /review`, `GET /reviews/:id/{status,events}`, `/cancel`, `/retry-fresh`) routes through, but here the orchestrator builds the retry DO name inline as `${this.state.reviewId}:${params.retryAttemptId}`. If the helper's join format ever changes, this site will silently drift and the retry would run in a different DO than `/status`/`/events`/`/cancel` resolve to. Should this go through the same shared helper?

apps/web/src/routers/code-reviews/code-reviews-router.ts
```diff
@@ -620,10 +642,20 @@ export const codeReviewRouter = createTRPCRouter({
...
+        if (input.attemptId && !attempt) {
+          throw new TRPCError({
+            code: 'NOT_FOUND',
+            message: 'Code review attempt not found',
+          });
+        }
+
         return successResult({
-          cloudAgentSessionId: review.session_id ?? null,
+          cloudAgentSessionId: attempt?.session_id ?? review.session_id ?? null,
```
jeanduplessis: **WARNING:** Explicit attempt selection falls back to the parent review session when that attempt has no session.

Trace:
1. This endpoint validates `attemptId`, but then returns `attempt?.session_id ?? review.session_id ?? null`.
2. `CodeReviewStreamView` passes the selected `attemptId` when switching attempts, and attempt rows allow `session_id`/`cli_session_id` to be null for pending attempts or failures before session assignment.
3. For an explicit attempt, null means “this attempt has no session”; falling back to the parent review can show the latest attempt’s session for a different selected attempt.

Impact: Selecting a failed/pending attempt with no session can display the parent/latest attempt’s live stream or history instead of showing no session for that attempt.

services/code-review-infra/src/code-review-orchestrator.ts
jeanduplessis, line 222: **WARNING:** Internal sandbox retry can leave the review queued locally with no runner scheduled.

Trace:
1. `tryRetryFreshSessionAfterSandboxError` now sets `this.state.status = queued` and returns `true`, but no longer calls `runWithCloudAgentNext()`.
2. The only automatic queued runner is the fallback alarm scheduled in `start()`. If that alarm has already fired while the review was running, `alarm()` no-ops for `running` and no new alarm is scheduled here.
3. This path also only saves local DO state; it does not update the parent DB to queued or enqueue a new `waitUntil` runner.

Impact: A sandbox/internal-server failure after the initial fallback alarm can strand the DO in queued state while the DB remains running, with no fresh execution started.

apps/web/src/routers/admin-code-reviews-router.ts
```diff
@@ -87,58 +124,87 @@ function buildOwnershipFilter(
...
-
 const ownershipTypeExpr = sql<string>`CASE WHEN ${cloud_agent_code_reviews.owned_by_organization_id} IS NOT NULL THEN 'organization' ELSE 'personal' END`;
 const waitSecondsExpr = sql<number>`EXTRACT(EPOCH FROM (${cloud_agent_code_reviews.started_at} - ${cloud_agent_code_reviews.created_at}))`;
 const validStartedWaitCondition = sql`${cloud_agent_code_reviews.started_at} IS NOT NULL AND ${cloud_agent_code_reviews.started_at} >= ${cloud_agent_code_reviews.created_at}`;
+const attemptWaitSecondsExpr = sql<number>`EXTRACT(EPOCH FROM (${cloud_agent_code_review_attempts.started_at} - ${cloud_agent_code_review_attempts.created_at}))`;
+const validAttemptStartedWaitCondition = sql`${cloud_agent_code_review_attempts.started_at} IS NOT NULL AND ${cloud_agent_code_review_attempts.started_at} >= ${cloud_agent_code_review_attempts.created_at}`;
+
+function buildBaseConditions(input: FilterInput): SQL[] {
+  return [
+    gte(cloud_agent_code_reviews.created_at, input.startDate),
```
jeanduplessis: **WARNING:** All-attempts reports still filter and group by parent review timestamps.

Trace:
1. `buildBaseConditions()` always filters on `cloud_agent_code_reviews.created_at`.
2. The all-attempts branches join `cloud_agent_code_review_attempts` and switch status/error/duration columns, but keep date filters and day buckets based on the parent review.
3. Retry attempts have their own `created_at`/`started_at`/`completed_at`, so all-attempts mode should not attribute them to the parent review’s creation date.

Impact: All-attempts metrics, charts, error drilldowns, and CSV exports are wrong around date boundaries: attempts can be omitted from the day they ran or included in ranges where only the parent review was created.

apps/web/src/lib/code-reviews/dispatch/dispatch-pending-reviews.ts
jeanduplessis: **WARNING:** Terminal dispatch responses leave the queued claim uncleared.

Trace:
1. The ambiguous-dispatch terminal branch no longer mirrors the terminal worker status.
2. The review was already claimed as `queued` before dispatch, and returning `false` from this fulfilled path only affects the result count; it does not release the claim or mark the review/attempt terminal.
3. A terminal worker status therefore leaves a recent queued row that will block capacity until stale, then likely retry the same terminal DO state.

Impact: If dispatch times out but the worker reports `completed`, `failed`, or `cancelled`, the review can remain stuck as queued instead of being released or finalized.

jeanduplessis (approved):
Found four concrete retry/attempt edge-case issues:
- selected attempts with null sessions can fall back to the parent/latest session,
- internal sandbox retry can strand a DO in queued state without scheduling a runner,
- all-attempts admin reporting uses parent review timestamps instead of attempt timestamps,
- terminal worker status after ambiguous dispatch leaves the DB claim queued.

apps/web/src/lib/code-reviews/dispatch/dispatch-pending-reviews.ts
alex-alecu: Fixed in 0bd59df.

services/code-review-infra/src/code-review-orchestrator.ts
alex-alecu: Fixed in 1a06bb4.

apps/web/src/lib/code-reviews/dispatch/dispatch-pending-reviews.ts
alex-alecu: Fixed in 0bd59df.

apps/web/src/lib/code-reviews/db/code-reviews.ts
alex-alecu: Fixed in e4109c1.

apps/web/src/lib/integrations/platforms/gitlab/webhook-handlers/merge-request-handler.ts
alex-alecu: Fixed in 1a04f9c.

apps/web/src/routers/admin-code-reviews-router.ts
alex-alecu: Fixed in 68cef49.

services/code-review-infra/src/code-review-orchestrator.ts
alex-alecu: Fixed in 3ee9b9c.

apps/web/src/routers/code-reviews/code-reviews-router.ts
alex-alecu: Fixed in 6d95467.

services/code-review-infra/src/code-review-orchestrator.ts
alex-alecu: Fixed in 1a06bb4.

apps/web/src/routers/admin-code-reviews-router.ts
alex-alecu: Fixed in 68cef49.

apps/web/src/lib/code-reviews/dispatch/dispatch-pending-reviews.ts
alex-alecu: Fixed in 0bd59df.

alex-alecu:
Manual test passed.

Tested:
- Started the dev review stack and exercised an attempt-specific code-review worker dispatch for a local seeded review.
- Triggered the internal status callback flow from running to retryable infra failure and verified a fresh retry attempt was started.
- Queried worker status for the fresh retry attempt and verified legacy status lookup without attemptId stays isolated.
- Sent a stale callback for the first attempt after the retry attempt existed.

Verified:
- Worker returned 202 with the original attemptId and routed status to the attempt-specific Durable Object.
- Postgres recorded attempt 1 as failed and created attempt 2 with retry_reason=infra_failure and retry_of_attempt_id pointing to attempt 1.
- Fresh retry worker status returned 200 for attempt 2.
- Stale callback returned "Stale callback from superseded attempt" and did not change the parent review state.
- Temporary local review data was cleaned up after verification.

PR #3227, opened 2026-05-13
fix(code-reviews): update health alerts

alex-alecu:
## Why

https://github.com/Kilo-Org/cloud/pull/3136 followup

Old review rows could make the health check look broken long after they stopped mattering. This made alerts noisy and harder to trust.

## What changed

The health check now looks at reviews that started recently instead of old queued or running rows. It marks the system unhealthy when too many recent reviews take more than an hour, or when too many recent started reviews end with system errors. Queued reviews no longer count until they start, and very old running reviews stop affecting health once they are outside the recent window. Billing issues, user cancels, and replaced reviews still do not count as system errors.

## How to test

1. Run `pnpm test -- apps/web/src/lib/code-reviews/alerting/detectors.test.ts apps/web/src/app/api/code-reviews/up/route.test.ts`.
2. Run `scripts/typecheck-all.sh --changes-only`.
3. Run `pnpm format`.

### DIFF
--- a/apps/web/src/app/api/code-reviews/up/route.test.ts
+++ b/apps/web/src/app/api/code-reviews/up/route.test.ts
@@ -51,7 +51,8 @@ describe('GET /api/code-reviews/up', () => {
 
   function reviewValues(overrides: Partial<CodeReviewInsert> = {}) {
     const sequence = reviewSequence++;
-    const timestamp = minutesAgo(5);
+    const startedAt = minutesAgo(10);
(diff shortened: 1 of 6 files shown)


jeanduplessis (approved):
No issues found.

Reviewed the final combined diff and PR context, including current PR-head contents and related caller/callee/state-transition context for the code review health detectors and endpoint. No verified issues were found on changed lines.

PR #3232, opened 2026-05-13
feat(mobile): add data-sharing consent gate

iscekic:
## Summary

- Add `user.getMe` plus a mobile current-user hook so the app can key consent by account.
- Add versioned device-local consent storage and boot gating that re-prompts when `CURRENT_CONSENT_VERSION` is bumped.
- Add the consent/details screens and a Profile "Privacy choices" revoke action that signs the user out.

## Verification

- [ ] Not run: manual mobile flows require a dev build/device session that is not available from this environment.

## Visual Changes

Not captured; the new consent and details screens need device/simulator screenshots during manual verification.

## Reviewer Notes

- Consent versioning lives in `apps/mobile/src/lib/consent.ts`; increment `CURRENT_CONSENT_VERSION` to re-prompt accepted users on their next boot.
- Please focus manual review on cold start while already signed in, per-account isolation, and the revoke/sign-out path.

### DIFF
--- a/apps/mobile/src/app/_layout.tsx
+++ b/apps/mobile/src/app/_layout.tsx
@@ -31,6 +31,7 @@ import { Toaster } from 'sonner-native';
 import { Button } from '@/components/ui/button';
 import { Text } from '@/components/ui/text';
 import { AuthProvider, useAuth } from '@/lib/auth/auth-context';
+import { shouldStartAppsFlyer } from '@/lib/appsflyer-consent';
 import { initAppsFlyer } from '@/lib/appsflyer';
 import { consentModeForSearchParam } from '@/components/consent/consent-mode';
 import { hasAcceptedConsent, subscribeToConsentChanges } from '@/lib/consent';
@@ -161,6 +162,27 @@ function RootLayoutNav() {
 
     return unsubscribe;
   }, [token, userId]);
+
+  useEffect(() => {
+    if (
+      !shouldStartAppsFlyer({
+        hasToken: token != null,
+        consentChecked,
+        needsConsent,
+      })
+    ) {
+      return;
+    }
+
+    async function startAppsFlyer() {
+      if (Platform.OS === 'ios') {
+        await requestTrackingPermissionsAsync();
+      }
+      initAppsFlyer();
+    }
+
+    void startAppsFlyer();
+  }, [token, consentChecked, needsConsent]);
 
   useEffect(() => {
     if (isLoading) {
@@ -302,16 +324,6 @@ function RootLayout() {
       navigationIntegration.registerNavigationContainer(ref);
     }
   }, [ref]);
-
-  useEffect(() => {
-    async function startAppsFlyer() {
-      if (Platform.OS === 'ios') {
-        await requestTrackingPermissionsAsync();
-      }
-      initAppsFlyer();
-    }
-    void startAppsFlyer();
-  }, []);
 
   useEffect(() => {
     const subscription = setupNotificationResponseHandler();
--- a/apps/mobile/src/app/_layout.tsx
+++ b/apps/mobile/src/app/_layout.tsx
@@ -28,6 +28,8 @@ import { Platform, View } from 'react-native';
 import { GestureHandlerRootView } from 'react-native-gesture-handler';
 import { Toaster } from 'sonner-native';
 
+import { Button } from '@/components/ui/button';
+import { Text } from '@/components/ui/text';
 import { AuthProvider, useAuth } from '@/lib/auth/auth-context';
 import { initAppsFlyer } from '@/lib/appsflyer';
 import { consentModeForSearchParam } from '@/components/consent/consent-mode';
@@ -82,7 +84,7 @@ setupNotificationHandler();
 checkInitialNotification();
 
 function RootLayoutNav() {
-  const { token, isLoading: authLoading } = useAuth();
+  const { token, isLoading: authLoading, signOut } = useAuth();
   const { updateRequired, isChecking: updateChecking } = useForceUpdate();
   const [fontsLoaded, fontsError] = useFonts({
     JetBrainsMono_500Medium,
@@ -92,7 +94,12 @@ function RootLayoutNav() {
   const pathname = usePathname();
   const { mode } = useGlobalSearchParams<{ mode?: string }>();
   const router = useRouter();
-  const { userId, isLoading: userIdLoading } = useCurrentUserId({ enabled: token != null });
+  const {
+    userId,
+    isLoading: userIdLoading,
+    isError: userIdError,
+    refetch: refetchUserId,
+  } = useCurrentUserId({ enabled: token != null });
   const [consentChecked, setConsentChecked] = useState(false);
   const [needsConsent, setNeedsConsent] = useState(false);
 
@@ -182,6 +189,11 @@ function RootLayoutNav() {
         router.replace('/(auth)/login');
       }
     } else {
+      if (userIdError) {
+        void SplashScreen.hideAsync();
+        return;
+      }
+
       if (userIdLoading || !consentChecked) {
         return;
       }
@@ -215,6 +227,7 @@ function RootLayoutNav() {
     inForceUpdate,
     router,
     userIdLoading,
+    userIdError,
     consentChecked,
     needsConsent,
     onConsentRoute,
@@ -225,6 +238,7 @@ function RootLayoutNav() {
   const showingForceUpdate = updateRequired && inForceUpdate;
   const needsAuth = !token && !inAuthGroup;
   const needsAppRedirect = token != null && inAuthGroup;
+  const hasUserBootstrapError = token != null && userIdError;
   const consentLoading =
     token != null && !consentChecked && !inAuthGroup && !inForceUpdate && !onConsentRoute;
   const needsConsentRedirect = consentChecked && needsConsent && !onConsentRoute;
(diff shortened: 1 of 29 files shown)


apps/mobile/src/components/consent/consent-card.tsx
```diff
@@ -0,0 +1,136 @@
...
+  };
+
+  const handleAccept = async () => {
+    if (!userId) {
+      toast.error('Could not load your account. Please try again.');
+      return;
+    }
+
+    await acceptConsent(userId);
+    router.replace('/(app)/(tabs)' as Href);
```
jeanduplessis, line 48: **CRITICAL:** Accepting consent redirects back to the consent screen

Trace:
1. The root layout sets `needsConsent` from `hasAcceptedConsent(userId)` only when `token` or `userId` changes.
2. On the no-consent/version-bumped path, `needsConsent` becomes `true` and the root layout routes to `/(app)/consent`.
3. `ConsentCard.handleAccept` writes the accepted version and routes to tabs, but `token` and `userId` do not change, so the root layout never re-runs the SecureStore check. When the pathname leaves `/consent`, the still-true `needsConsent` branch redirects back to `/(app)/consent`.

Impact: First-time users, or users re-prompted after `CURRENT_CONSENT_VERSION` is bumped, cannot get past the consent screen after tapping “Accept and continue.”

apps/mobile/src/app/_layout.tsx
```diff
@@ -121,31 +160,63 @@ function RootLayoutNav() {
       } else {
         router.replace('/(auth)/login');
       }
-    } else if (inAuthGroup) {
-      router.replace('/(app)');
     } else {
+      if (userIdLoading || !consentChecked) {
```
jeanduplessis, line 196: **WARNING:** `getMe` failure leaves signed-in users stuck behind the hidden boot gate

Trace:
1. `useCurrentUserId` returns `isError`, but the root layout only reads `userId` and `isLoading`.
2. If `user.getMe` fails after retries, React Query can produce `userId === undefined` and `userIdLoading === false`; the consent effect sets `consentChecked` to `false`.
3. The navigation effect returns on `!consentChecked`, while `hidden` remains true via `consentLoading`/redirect state. There is no error route, toast, sign-out, or visible retry path.

Impact: A signed-in user with a transient/backend `user.getMe` failure can remain on a blank/splash-hidden app indefinitely instead of seeing a recoverable error state.

jeanduplessis (changes requested):
Found two issues in the consent gate flow: accepting consent does not refresh the root gate state, and `getMe` errors can leave signed-in users hidden behind the boot gate with no recovery path.

apps/mobile/src/components/consent/consent-card.tsx
iscekic: Fixed in 68583fb. Consent writes now publish a local consent-change event, and the root gate subscribes to refresh needsConsent immediately after accept/revoke instead of waiting for token or userId to change.

apps/mobile/src/components/consent/consent-details.tsx
```diff
@@ -0,0 +1,81 @@
...
+          title="Kilo Gateway (our backend)"
+          what="Account ID, request metadata, token usage."
+          why="Authentication, routing requests, billing reconciliation."
+          who="Kilo (kilo.ai)."
+        />
+        <Section
+          title="Analytics & attribution"
+          what="App events (opens, screens viewed, feature use), device type, app version, install source."
+          why="Measure app performance and understand which channels bring new users."
+          who="[MMP vendor - TBD]."
```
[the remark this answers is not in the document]
iscekic: Resolved. The placeholder was replaced with privacy-policy-backed wording so the consent details screen no longer renders TBD vendor text.

apps/mobile/src/components/profile-screen.tsx
```diff
@@ -75,6 +78,28 @@ export function ProfileScreen() {
...
+      'Revoke data sharing consent?',
+      'Kilo Code needs this consent to function. Revoking will sign you out. You can accept again on next sign-in.',
+      [
+        { text: 'Cancel', style: 'cancel' },
+        {
+          text: 'Revoke and sign out',
+          style: 'destructive',
+          onPress: () => {
+            void (async () => {
+              if (userId) {
```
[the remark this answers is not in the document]
iscekic: Resolved by changing Profile > Privacy choices to open the consent screen in review mode. Revocation now happens from the consent screen with an explicit user-id guard before sign-out.

apps/mobile/src/lib/consent.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    expect(await hasAcceptedConsent('user-1')).toBe(false);
+  });
+
+  it('returns true after acceptConsent for the same user', async () => {
+    const { CURRENT_CONSENT_VERSION, acceptConsent, hasAcceptedConsent } =
+      await import('./consent');
+
+    await acceptConsent('user-1');
+    expect(store.get('consent-accepted:user-1')).toBe(String(CURRENT_CONSENT_VERSION));
```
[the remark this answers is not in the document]
iscekic: Fixed in 68583fb. The consent storage tests now use the dash-separated key and cover the old-version and unversioned paths against the key that hasAcceptedConsent actually reads.

apps/mobile/src/app/_layout.tsx
iscekic: Fixed in a726b41. The root gate now reads the getMe error state and renders a visible account-load error with Retry and Sign out instead of leaving the app hidden behind the boot gate.

apps/mobile/src/lib/consent.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    await acceptConsent('user-1');
+    await revokeConsent('user-1');
+    expect(await hasAcceptedConsent('user-1')).toBe(false);
+  });
+
+  it('notifies listeners when consent changes for a user', async () => {
+    const { acceptConsent, revokeConsent, subscribeToConsentChanges } = await import('./consent');
+    const changes: string[] = [];
+
+    subscribeToConsentChanges(change => {
```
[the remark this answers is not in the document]
iscekic: Fixed in cd3b8f2. The listener test now captures the unsubscribe function and calls it after the accept/revoke assertions so the module-level listener set does not leak into later tests.

apps/mobile/src/components/consent/consent-details.tsx
```diff
@@ -0,0 +1,81 @@
...
+          who="Anthropic, OpenAI, Google, and other providers you select per request."
+        />
+        <Section
+          title="Kilo Gateway (our backend)"
+          what="Account ID, request metadata, token usage."
+          why="Authentication, routing requests, billing reconciliation."
+          who="Kilo (kilo.ai)."
+        />
+        <Section
+          title="Analytics & attribution"
```
jeanduplessis, line 46: **WARNING:** Analytics consent is collected after AppsFlyer can already initialize

Trace:
1. This screen adds an explicit analytics/attribution consent disclosure.
2. `apps/mobile/src/app/_layout.tsx` still runs the startup effect unconditionally: iOS ATT is requested and `initAppsFlyer()` is called before `RootLayoutNav` finishes `hasAcceptedConsent(userId)`.
3. `apps/mobile/src/lib/appsflyer.ts` initializes the AppsFlyer SDK with install conversion data enabled, so first launch / cold start can contact the attribution provider before this new consent gate is accepted.

Impact: Signed-in users who have not accepted the new consent version can still initialize third-party analytics/attribution before consenting.

apps/mobile/src/app/_layout.tsx
```diff
@@ -94,6 +117,50 @@ function RootLayoutNav() {
...
+    let cancelled = false;
+
+    async function checkConsent() {
+      if (!token || !userId) {
+        setConsentChecked(false);
+        setNeedsConsent(false);
+        return;
+      }
+
+      const accepted = await hasAcceptedConsent(userId);
```
jeanduplessis: **WARNING:** SecureStore consent read failures leave the boot gate hidden

Trace:
1. The boot gate now awaits `hasAcceptedConsent(userId)` before setting `consentChecked`.
2. `hasAcceptedConsent` calls `SecureStore.getItemAsync(...)` without catching errors, and `checkConsent()` is launched with `void checkConsent()` with no catch/error state.
3. If SecureStore rejects, `consentChecked` remains `false`; `consentLoading` keeps the app at opacity 0 with no retry/sign-out recovery path.

Impact: A signed-in user can get stuck behind the hidden boot gate if the local consent read fails.

dev/local/mobile-env.ts
```diff
@@ -0,0 +1,205 @@
...
+      host = arg.slice('--host='.length);
+    } else if (arg === '--help' || arg === '-h') {
+      console.log('Usage: pnpm dev:env:mobile [--host <lan-ip>]');
+      process.exit(0);
+    }
+  }
+  return { host };
+}
+
+function isUsableIpv4(value: string | undefined): value is string {
```
jeanduplessis, line 40: **SUGGESTION:** IPv4 validation accepts impossible octets

Trace:
1. The new mobile env generator uses `isUsableIpv4` to validate `--host` / detected host values.
2. The regex only checks shape, so values like `999.999.999.999` pass.
3. `main()` then writes URLs using that invalid host to both mobile and root env files.

Impact: A typo in `--host` can silently generate unusable local dev URLs instead of failing fast.

jeanduplessis (approved):
Found two concrete consent-gate issues and one dev-tool validation nit. The most important issue is that analytics/attribution can initialize before the new consent gate is accepted.

apps/mobile/src/components/consent/consent-details.tsx
iscekic: Fixed in 19d0c16. AppsFlyer startup now runs from the authenticated layout only after consent has been checked and accepted, so ATT and SDK initialization no longer happen before the consent gate clears.

apps/mobile/src/app/_layout.tsx
iscekic: Fixed in 2b6555e. Consent storage reads now go through a checked gate result; SecureStore read failures are captured and surface a visible privacy-choices load error with Retry and Sign out instead of leaving the app hidden behind the boot gate.

dev/local/mobile-env.ts
iscekic: Fixed in a631004. The mobile env host validator now parses IPv4 octets and rejects values outside 0-255, with a regression assertion for 999.999.999.999.

PR #3234, opened 2026-05-13
fix(kiloclaw): stage bundled plugin deps in image

pandemicsyn:
## Summary

- Stage OpenClaw bundled plugin runtime dependencies in an image-baked `OPENCLAW_PLUGIN_STAGE_DIR` so doctor and gateway startup do not repair default bundled plugins one-by-one at boot.
- Mirror the staged dependency prewarm in the local KiloClaw Dockerfile and default the controller env var while preserving explicit overrides.
- Document the controller env var in the KiloClaw controller spec and add a new KiloClaw changelog entry for the bundled-plugin startup optimization.

## Verification

- Built a fresh local KiloClaw image tagged `kiloclaw:plugin-stage-dir`.
- Inspected the built image and confirmed `OPENCLAW_PLUGIN_STAGE_DIR=/usr/local/share/openclaw-plugin-runtime-deps` plus a populated OpenClaw 2026.4.23 staged dependency root containing representative bundled plugin dependencies.
- Ran `services/kiloclaw/scripts/controller-entrypoint-smoke-test.sh` against the fresh image: 7 passed, 0 failed.
- Ran `services/kiloclaw/scripts/controller-proxy-auth-smoke-test.sh` against the fresh image.
- Ran `services/kiloclaw/scripts/controller-smoke-test.sh` against the fresh image: 15 passed, 0 failed.

## Visual Changes

N/A

## Reviewer Notes

This only changes bundled OpenClaw plugin runtime dependency staging. User-installed npm, ClawHub, marketplace, or local plugins are not bundled plugin candidates and keep their normal install behavior.

The staged root is intentionally outside `/root` because Fly mounts the persistent instance volume at `/root`; placing image-baked deps there would be shadowed at runtime and could reintroduce persistent package growth. The controller sets the same default env var only when unset, so explicit image/operator overrides still win.

### DIFF
--- a/.specs/kiloclaw-controller.md
+++ b/.specs/kiloclaw-controller.md
@@ -275,6 +275,10 @@ tag (16 bytes)`.
      supervisor.
    - `GOG_KEYRING_PASSWORD=kiloclaw` — required for the gog keyring
      backend (not a secret; see gog-credentials.ts).
+   - `OPENCLAW_PLUGIN_STAGE_DIR=/usr/local/share/openclaw-plugin-runtime-deps`
+     when unset — keep bundled OpenClaw plugin runtime dependencies in
+     the image-baked stage directory instead of mutating `/root` or each
+     bundled plugin directory during doctor/gateway startup.
 
 ### Feature Flags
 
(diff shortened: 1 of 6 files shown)


pandemicsyn:
https://github.com/Kilo-Org/cloud/actions/runs/25824092827


looks like it worked fine in dev , _noticable_ less cpu usage on obvs, but also actually memory?

SCR-20260513-ovkc

the reboot at 1300 was the one that led to poke at this.

jeanduplessis (approved):

PR #3237, opened 2026-05-13
fix(cloud-agent): pre-import kilo session in sync prepare path to fix ingest race

eshurakov:
## Summary

- Slack bot sessions (`autoInitiate=false`) were producing empty sessions in the dashboard because no data ever reached the session-ingest worker.
- Root cause: race condition between `Session.Event.Created` (~7ms) and `KiloSessions.init()` completing its `Bus.subscribe` registration (~35ms). The wrapper started without `--session-id`, so the CLI created a fresh session live — firing the event into the race window. With no handler registered, `bootstrap()` was never called, `session_share` was never written, and all `IngestQueue.flush()` calls failed silently.
- Fix: in the sync prepare path, pre-generate a `kiloSessionId`, write a minimal session JSON, run `kilo-restore-session.js --file` to import it into the CLI's SQLite, then pass `--session-id` to `ensureWrapper`. The wrapper calls `Session.get()` instead of `Session.create()`, so `Session.Event.Created` never fires — identical to how the `autoInitiate=true` (web UI) path already works.

## Verification

1. Trigger a Slack bot session via `pnpm --filter web script:run bot replay-slack-spawn <session-id>`
2. Check container logs — `ingest flush: no session_share found` should not appear
3. Confirm session data appears in the Kilo dashboard

## Visual Changes

N/A

## Reviewer Notes

The fix reuses `generateKiloSessionId()` (already imported) and the exact `kilo-restore-session.js --file` pattern from `async-preparation.ts` steps 8–9, which has been running in production for the `autoInitiate=true` web UI path. No new code paths introduced.

### DIFF
--- a/services/cloud-agent-next/src/router/handlers/session-prepare.ts
+++ b/services/cloud-agent-next/src/router/handlers/session-prepare.ts
@@ -562,17 +562,56 @@ const prepareSessionHandler = internalApiProtectedProcedure
         await writeAuthFile(sandbox, sessionHome, ctx.authToken);
         await writeGlobalRules(sandbox, sessionHome, cloudAgentSessionId);
 
-        // 11. Start wrapper (which starts kilo server in-process and creates session)
-        logger.info('Starting wrapper');
-        const { client: _wrapperClient, sessionId: kiloSessionId } =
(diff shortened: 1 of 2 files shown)


iscekic (approved):
Approving to unblock

PR #3250, opened 2026-05-14
fix(kiloclaw) volume cleanup enumerate and cap retries

St0rmz1:
## Problem

Production analytics surfaced eight kiloclaw sandboxes accumulating more than 12,000 `destroy_volume_failed` events each over a two week window, with no alerting. Investigation traced the cause to two structural gaps in `tryDeleteVolume`:

1. The function only ever targeted `state.pendingDestroyVolumeId`. Recovery clones and any volumes the DO had lost track of were never touched and lived on in Fly indefinitely.
2. On any failure other than 404, the function logged, persisted the error, and returned. The next alarm fired the same call. There was no attempt limit, no escalation, and no exit from the retry loop. A single stuck volume could retry on every alarm forever.

Combined, the bug pattern explains a meaningful slice of the unexplained portion of the monthly Fly volume bill.

## Approach

Two changes, one PR.

### 1. Cap retries on the existing destroy path

`tryDeleteVolume` now tracks a `destroyVolumeAttempts` counter on the DO's persisted state. After 50 consecutive failures (roughly an hour at the current alarm cadence) the function:

* Emits `reconcile.destroy_volume_abandoned_after_max_retries` with the volume id, attempt count, last error, and last status. Alerting can hook this event.
* Clears `pendingDestroyVolumeId` so the destroy loop can finalize and the DO can wipe its state.
* Resets the counter to 0 on success or 404 so a single transient failure does not erode the cap on the next cycle.
* Resets the counter to 0 inside `destroy()` itself, guaranteeing a fresh attempt window for every new destroy invocation.

The cap is intentionally generous. The goal is to stop the runaway loop, not to be aggressive about giving up on transient failures.

### 2. Add an orphan volume sweep

After `tryDeleteVolume` clears its pointer, the new `tryDeleteOrphanVolumes` runs on the same alarm. It calls `fly.listVolumes` for the app and destroys any volume whose name matches `volumeNameFromSandboxId(state.sandboxId)`.

This catches the cases the old logic missed: original volumes whose pointer got overwritten by a recovery clone, abandoned clones from prior recovery cycles, anything that never made it into the DO's `pendingDestroyVolumeId`.

Safety guards:

* The name match is exact. The filter is critical on shared `acct-*` apps so we never touch a sibling sandbox's volume.
* Skips volumes in `pending_destroy`, `destroying`, or `destroyed` states. Fly is already tearing those down.
* Skips volumes still attached to a machine. The normal pending destroy flow handles those.
* Best effort. Failures on individual volumes are logged but do not fail the alarm.

## Files changed

| File | What |
|------|------|
| `services/kiloclaw/src/schemas/instance-config.ts` | Added `destroyVolumeAttempts` to the persisted state Zod schema |
| `services/kiloclaw/src/durable-objects/kiloclaw-instance/types.ts` | Added `destroyVolumeAttempts: number` to `InstanceMutableState` |
| `services/kiloclaw/src/durable-objects/kiloclaw-instance/state.ts` | Wired the new field through load, reset, and create defaults |
| `services/kiloclaw/src/durable-objects/kiloclaw-instance/reconcile.ts` | Main fix: cap and escalation in `tryDeleteVolume`, new `tryDeleteOrphanVolumes` |
| `services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts` | `destroy()` resets the counter on every fresh invocation |
| `services/kiloclaw/src/fly/types.ts` | `FlyVolume.state` union now includes `pending_destroy` (Fly returns it; the type was missing it) |
| `services/kiloclaw/src/durable-objects/kiloclaw-instance.test.ts` | Nine new tests |

## Test coverage

Nine new tests in `kiloclaw-instance.test.ts`:

* Counter increments on each failure and keeps the pending pointer.
* At the cap, the abandon event fires with sandbox attribution and the destroy finalizes cleanly.
* Successful delete resets the counter.
* 404 resets the counter (existing already gone behavior preserved).
* `destroy()` invocation resets the counter so a previous cycle's value does not bleed in.
* Orphan sweep destroys volumes that match the sandbox name and skips volumes belonging to other sandboxes.
* Orphan sweep skips `pending_destroy` and attached volumes.
* Orphan sweep does not run while the primary pending pointer is set.
* Abandon and sweep coexist on the same alarm: the sweep can complete cleanup the moment the abandon fires.

Full kiloclaw suite passes (1922 tests). Format, lint, and typecheck all clean.

## Behavior after deploy

For the eight currently stuck sandboxes:

* Their next alarm fires.
* `tryDeleteVolume` tries the same call as before. If the underlying Fly state has recovered (404 now), the existing already gone handler clears state and the loop ends.
* If the 412 or other failure persists, the counter increments. Within a window of about an hour the cap is reached, the abandon event fires, and the loop ends.
* The orphan sweep on the same alarm gets one more best effort attempt. If the underlying transient condition has resolved, that final attempt may complete the cleanup.
* Either way, the loop ends.

To watch progress, the standing `log-check.sh` script counts `reconcile.destroy_volume_failed` events per day. The chronic rate of about 6,500 per day should drop to near zero within 24 to 48 hours after deploy.

## Open caveats

* The orphan sweep relies on volume name matching. Today's architecture guarantees one sandbox per Fly app (instance keyed sandboxes route to per instance `inst-*` apps; legacy sandboxes route to per user `acct-*` apps with at most one legacy sandbox per user) so the name filter is safe. If a future migration ever places multiple instances inside a shared Fly app, the truncation in `volumeNameFromSandboxId` becomes a collision hazard. A TODO marker is in place at the top of `tryDeleteOrphanVolumes`.
* Alerting on `reconcile.destroy_volume_abandoned_after_max_retries` is not wired in this PR (separate follow up). Once it ships, watch for false positives where the sweep silently cleans up the volume after the abandon event fires.

## Related follow ups

* Wire alerting on the new abandon event.
* Org wide volume janitor as a backstop for orphans this code does not reach (for example, sandboxes whose DO state has been wiped entirely).
* Force destroy admin endpoint to cover the case where `destroyed_at` is set and the admin destroy button is hidden.


services/kiloclaw/src/durable-objects/kiloclaw-instance/reconcile.ts
```diff
@@ -1456,20 +1490,129 @@ export async function tryDeleteVolume(
...
+    if (vol.name !== expectedName) continue;
+    // Fly is already tearing the volume down — let it finish.
+    if (
+      vol.state === 'pending_destroy' ||
+      vol.state === 'destroying' ||
+      vol.state === 'destroyed'
+    ) {
+      continue;
+    }
+    if (vol.attached_machine_id) {
```
jeanduplessis, line 1695: **WARNING:** Attached orphan volumes are skipped, then the DO finalizes and forgets them.

Trace:
1. The new sweep skips a matching orphan volume whenever `attached_machine_id` is set.
2. `retryPendingDestroy()` immediately calls `finalizeDestroyIfComplete()` after the sweep; finalization only checks `pendingDestroyMachineId` / `pendingDestroyVolumeId`, then deletes the alarm and wipes DO state.
3. For an orphan volume that was lost from `pendingDestroyVolumeId` while still attached to an old machine, `recoverBoundMachineForDestroy()` cannot see it, so the sweep logs/skips it and finalization stops future retries.

Impact: Matching attached orphan volumes can survive destroy permanently until an external janitor/manual cleanup handles them.

jeanduplessis (approved):
Found one warning: attached orphan volumes discovered by the new sweep are skipped but do not prevent destroy finalization, so the DO can wipe state and stop retrying while leaving the attached volume/machine on Fly.

PR #3252, opened 2026-05-14
fix(code-reviews): block interactive agents

alex-alecu:
## Why

Code reviews could get stuck when the agent asked a question or waited for input. No one can answer those prompts during an automatic review, so the review waited until it timed out.

## What changed

Code review agents now get clear rules that they must never ask questions or start anything interactive. The runtime also enforces this by failing fast if a code review enters a question or permission state. Interactive shells, editors, pagers, login prompts, and similar commands are blocked for code reviews. This makes stuck reviews fail clearly instead of waiting for a long timeout.

## How to test

1. Start a cloud code review and confirm it still reads the diff and posts a summary.
2. Trigger or simulate a question request during a code review and confirm the review fails instead of waiting.
3. Trigger or simulate a permission request during a code review and confirm it is rejected instead of waiting.
4. Run `pnpm --filter cloud-agent-next test -- test/unit/wrapper/snapshot.test.ts src/session-service.test.ts`.
5. Run `pnpm --filter cloud-agent-next typecheck`.


jeanduplessis (approved):

services/cloud-agent-next/wrapper/src/connection.ts
```diff
@@ -194,6 +227,20 @@ export function createConnectionManager(
...
+          callbacks.onTerminalError('Code review attempted to ask an interactive question');
+          return;
+        }
+
+        if (sessionStatus.type === 'permission' || pendingPermission) {
+          rejectCodeReviewPermission(pendingPermission?.id, config.kiloClient);
+          callbacks.onTerminalError('Code review attempted to request interactive permission');
+          return;
+        }
+      }
```
eshurakov: do we need this code if we remove question tool from the agent?

services/cloud-agent-next/wrapper/src/connection.ts
```diff
@@ -194,6 +227,20 @@ export function createConnectionManager(
...
+      if (isCodeReviewJob(state)) {
+        if (sessionStatus.type === 'question' || pendingQuestion) {
+          rejectCodeReviewQuestion(pendingQuestion?.id, config.kiloClient);
+          callbacks.onTerminalError('Code review attempted to ask an interactive question');
+          return;
+        }
+
+        if (sessionStatus.type === 'permission' || pendingPermission) {
+          rejectCodeReviewPermission(pendingPermission?.id, config.kiloClient);
+          callbacks.onTerminalError('Code review attempted to request interactive permission');
```
eshurakov: we call a terminal callback, but the session may continue and we're not terminating a session

alex-alecu:
@eshurakov I will followup with another PR to remove the question tool from the runtime in the first place to avoid allowing the LLM to question and then rejecting in the wrapper

PR #3253, opened 2026-05-14
docs(agents): document Workers client lifetime rule

jeanduplessis:
## Summary
Document repository guidance that prevents unsafe module-scope caching of transport-owning clients in Workers and Durable Objects.

### Why this change is needed
Cloudflare isolates and Durable Object module memory can outlive individual execution contexts. Reusing database clients or similar I/O-backed SDK objects from module scope risks stale bindings and cross-context runtime failures, while Hyperdrive already provides the connection-pooling layer needed for external databases.

### How this is addressed
- Add explicit `AGENTS.md` guidance against module-scope caching of DB clients, pools, and request/context-bound SDK objects.
- Direct future changes toward approved per-use helpers such as `getWorkerDb(...)`.
- Preserve valid Durable Object instance-local storage wrappers derived from constructor state.
- Require a documented safety case before any module-scope client caching optimization is introduced.

## Human Verification
- Reviewed rule wording against current repository patterns for `getWorkerDb(...)` and Durable Object instance-local SQLite/Drizzle state.

### Human Reviewer Flags
- Guidance intentionally distinguishes unsafe module-scope transport state from valid Durable Object instance fields created from constructor inputs.
- Rule codifies runtime-safety expectations rather than changing application behavior.


AGENTS.md
```diff
@@ -52,6 +52,13 @@ Before making or reviewing UI changes under `apps/web` — components, routes/pa
 - When the linter flags an unused variable, investigate the root cause — do not blindly prefix with `_`.
 - Use existing dependencies before implementing custom solutions. Check `package.json` for what's available.
 
+## Workers & Durable Objects
+
+- Do not cache database clients, pools, or other transport-owning/request-context-bound SDK objects in module scope for Cloudflare Workers or Durable Objects. Workers reuse isolates across requests, Durable Object classes in the same Worker can share module memory across object instances, and stale module-scope I/O state can cause cross-context runtime failures.
+- Create external database clients through approved per-use helpers such as `getWorkerDb(...)`; let Hyperdrive own pooling. Only cache pure data or context-independent values in module scope.
+- Durable Object instance fields are valid for object-local state created from constructor inputs, such as SQLite/Drizzle wrappers over `state.storage`.
+- If optimization appears to require module-scope client caching, stop and document why lifetime, transport ownership, binding freshness, and Cloudflare runtime behavior make it safe before implementing it.
+
```
jrf0110, line 61: I think we can simplify this with:

> Global or module-scoped singletons are generally an anti-pattern. Do not use them unless they can't be avoided.

I might be alone in this, but I think the sort of pattern we're trying to discourage here applies to almost all software development (not just Workers and DOs)

jrf0110 (approved):

AGENTS.md
jeanduplessis: I agree module-scoped singletons deserve skepticism in general. I kept this narrower because the failure mode here is unusually concrete in Workers/DOs: transport-owning or context-bound clients can survive request/object contexts and fail at runtime or retain stale bindings.

A generic singleton warning would also catch valid module-scope constants, pure caches, schemas, and similar context-independent values, so I think this rule stays more actionable if it names the unsafe class and the approved exception/pattern directly. We can refine the rule going forward if we pick up further patterns of unsafe usage.

PR #3255, opened 2026-05-14
Feat/kiloclaw briefing GitHub scope and diagnostics

St0rmz1:
## Summary

The morning briefing's GitHub source had three bugs that combined to make the brief report "No open issues found in accessible repositories" for users who actually had issues to surface.

1. The `gh search issues` query had no scope qualifier (`author:`, `assignee:`, `involves:`, etc.), so it ran a global search across all of GitHub instead of filtering to the authenticated user. With no scope, the search returned essentially nothing useful.
2. The query passed `sort:updated-desc` as part of the search string, which is not a valid GitHub search qualifier. `gh search issues` expects sort ordering via the `--sort` and `--order` flags. The literal `sort:updated-desc` was either ignored or treated as plain text.
3. When the query returned zero items, the brief emitted a single misleading sentence that gave the user no way to tell whether the token was the wrong type, had missing scopes, or simply had nothing to surface.

This PR fixes all three:

* Replaces the search invocation with `gh search issues --state open --involves @me --sort updated --order desc --limit 12 --json title,url,updatedAt`. `--involves @me` is the right default scope for a "what is on my plate" briefing (issues the user created, was assigned, was mentioned in, or commented on).
* Adds a new helper module (`github-utils.ts`) that classifies the token by prefix (`ghp_` classic, `github_pat_` fine grained, `ghs_` app, `gho_` OAuth), parses the `X-OAuth-Scopes` response header from `gh api -i user`, and (for fine grained tokens) counts accessible repositories via `gh api user/repos --paginate --jq '. | length'`.
* Rewrites the empty result message per token type. Classic tokens with the wrong scopes get the granted scopes, the missing scopes, and the `gh auth refresh` command to fix it. Fine grained tokens get the accessible repo count and an explanation of the resource owner model. App, OAuth, and unrecognized tokens get a generic fallback. The source status footer is also rewritten to drop the misleading "in accessible repositories" wording.
* All failures in the diagnostic path are swallowed into a safe fallback so a broken diagnostic cannot break the brief itself.

This is a plugin only change. No database migration, no new endpoints, no auth flow changes.

## Verification

Built the plugin bundle locally, swapped it onto a running container, restarted the gateway, and triggered a fresh briefing against a real GitHub token.

* Before: the original query returned an empty array, and the brief read "No open issues found in accessible repositories" even though the authenticated account had a known open issue.
* After: the new scoped query returned the expected issue, and the brief rendered it under `## GitHub` with the correct title, URL, and updated timestamp.
* 104 vitest tests passing in `services/kiloclaw/plugins/kiloclaw-morning-briefing`: 31 new helper unit tests, 5 new lifecycle diagnostic tests, 68 existing tests unchanged.

## Visual Changes

N/A

## Reviewer Notes

* The diagnostic gathering function (`gatherGithubEmptyResultContext` in `index.ts`) runs only on the empty result path, not on every brief. It adds at most two extra `gh api` calls in that branch, with 10 and 15 second timeouts respectively.
* Token classification is a pure prefix string match against GitHub's documented token formats. No network call.
* For fine grained tokens, `gh api user/repos --paginate --jq '. | length'` paginates through accessible repos and sums the count. For users with hundreds of repos this is a few sequential requests; acceptable since it only runs when the search returned zero items.
* The new lifecycle tests use `vi.stubEnv` to scope `GITHUB_TOKEN` per test and call `vi.unstubAllEnvs()` in `afterEach` so test ordering does not matter and host environment cannot leak in.


services/kiloclaw/plugins/kiloclaw-morning-briefing/src/index.ts
```diff
@@ -611,7 +620,108 @@ async function collectGithub(api: {
...
+          .filter(value => Number.isFinite(value))
+          .reduce((sum, value) => sum + value, 0);
+      }
+    } catch {
+      // Leave accessibleRepoCount=0 — the message still reads sensibly.
+    }
+    return {
+      tokenType: 'fine-grained',
+      login,
+      accessibleRepoCount,
```
jeanduplessis, line 712: **WARNING:** Fine-grained PAT diagnostic renders "0 repositories" when the `user/repos` call fails or times out — conflates "fetch failed" with "truly zero accessible repos".

Trace:
1. `accessibleRepoCount` is mutated only under `if (code === 0)`. On non-zero exit (rate limit, 401, 5xx) and on timeout (`runCommandWithTimeout` returns `code: number | null`; null is the timeout signal per `command-utils.ts:29–45`'s `isTimeoutExecutionError`), the count stays at its initial `0`.
2. The PR's own reviewer note acknowledges `--paginate` "is a few sequential requests" for users with hundreds of repos — exactly the population most likely to hit the 15s timeout.
3. Downstream renderers assert the count as fact:
   - `github-utils.ts` empty-section: `` `Token can see ${ctx.accessibleRepoCount} repositories it was explicitly granted access to.` ``
   - `github-utils.ts` empty-summary: `` `0 issues — fine-grained PAT for ${login} sees ${count} repos` ``

**Impact:** A fine-grained PAT user whose `user/repos --paginate` times out or 4xx's reads "Token can see 0 repositories … switch to a classic PAT with `repo` scope, or have the resource owner generate a fine-grained PAT scoped to those repos." That actively steers them toward replacing a token whose real repo set is unknown.

Suggest tracking success separately (e.g., `accessibleRepoCount: number | null` or a sibling `repoCountFetched: boolean`) and rendering "could not enumerate accessible repositories" when `code !== 0`. The diagnostic try/catch already promises a safe fallback — extend it to "we don't know" rather than asserting a wrong number.

services/kiloclaw/plugins/kiloclaw-morning-briefing/src/index.ts
```diff
@@ -611,7 +620,108 @@ async function collectGithub(api: {
...
+ * Best-effort diagnostics for the empty-result path. Runs lightweight
+ * `gh api` calls to figure out what token the user is on and what it can
+ * see, so the brief can render an actionable "no issues" message instead
+ * of a generic one. All failures here are absorbed into an 'unknown'
+ * fallback so a broken diagnostic doesn't break the brief.
+ */
+async function gatherGithubEmptyResultContext(
+  api: GithubApiRunner
+): Promise<GithubEmptyResultContext> {
+  const tokenType = classifyGithubToken(readGithubTokenFromEnv());
```
jeanduplessis, line 640: **SUGGESTION:** Users authenticated via `gh auth login` (no `GH_TOKEN`/`GITHUB_TOKEN` exported) will always classify as `'unknown'`, even when `gh api user` returns a real login and `gh search issues` works fine.

Trace:
1. `resolveGithubReady` treats `gh auth status` exit 0 as configured — includes the default interactive `gh auth login` flow that stores creds in `~/.config/gh/hosts.yml`.
2. `readGithubTokenFromEnv()` returns `undefined` in that case → `classifyGithubToken` returns `'unknown'` → classic/fine-grained branches are unreachable regardless of what `gh` is actually using.
3. Fallback render then prints "(unknown token). Could not determine accessible repositories or scopes from this token type." for a meaningful chunk of real users.

**Impact:** Non-blocking diagnostic degradation. Consider `gh auth token` as a fallback source for the classifier so the `gh auth login` case still gets typed copy. Brief itself is unaffected.

jeanduplessis (approved):
Reviewed end-to-end: PR body matches the diff, both prior bot threads are resolved, and the new query shape + diagnostic context type/render split are sound. Two non-blocking observations on the empty-result diagnostic path — both behavioral concerns about the fine-grained / unknown-token branches, neither blocking the core fix.

- WARNING: fine-grained PAT branch renders "0 repositories" when `gh api user/repos --paginate` returns non-zero / times out (`runCommandWithTimeout` returns `code: number | null`; the existing `if (code === 0)` skip leaves `accessibleRepoCount = 0`, which downstream copy asserts as fact).
- SUGGESTION: users authenticated via `gh auth login` (no `GH_TOKEN`/`GITHUB_TOKEN` set) always classify as `unknown` since the classifier reads only from env. Consider falling back to `gh auth token` so the classifier has a value in the common interactive-auth case.

No PostHog surface; tests are behavioral and well-isolated (`vi.stubEnv` + `vi.unstubAllEnvs`).

PR #3258, opened 2026-05-15
feat(model-eval-ingest): sync promoted bench evals

lambertjosh:
## Summary
- Add a dedicated Cloudflare `model-eval-ingest` puller that reads promoted evals from kilo-bench over a Service Binding, stores append-only audit rows, and recomputes public `modelStats.benchmarks.kiloBench` caches.
- Extend cloud persistence, migrations, GDPR anonymization, and admin-only operations for sync-now / ingest history while keeping bench URLs and promoter email out of public model stats JSON.
- Preserve fractional benchmark scores and weighted aggregate semantics, including latest-promotion-wins recompute behavior per model/task tuple.

## Verification
- Ran the bench dashboard Worker and cloud model-eval-ingest Worker locally, seeded a promoted `swift-falcon-local-e2e` bench eval, triggered the scheduled sync endpoint, and verified Postgres stored the ingest row and public `kiloBench` cache.
- Re-triggered the local scheduled sync to verify the ingest remained idempotent, then inspected the cache payload to confirm it excluded bench URLs and promoter email while retaining ISO `lastPromotedAt`.

## Visual Changes
N/A

## Reviewer Notes
- The new worker expects a `BENCH_DASHBOARD` Service Binding plus the shared internal API secret; the web admin trigger uses `MODEL_EVAL_INGEST_URL` to call `/internal/sync`.
- `model_eval_ingest.promoted_by_email` is retained for audit but anonymized by the account soft-delete flow.
- The generated Drizzle migration introduces the append-only ingest table and indexes needed for idempotency and cache recompute queries.


lambertjosh:
image
image
image

lambertjosh:
Note that if/when this is merged, we will then need to:

1. Wait for CI to deploy model-eval-ingest
2. Capture its deployed URL - i.e. workers.dev
3. Set MODEL_EVAL_INGEST_URL in Vercel production to value.
4. Redeploy Vercel / let the next web deploy pick it up

packages/db/src/schema.ts
```diff
@@ -2975,6 +2997,55 @@ export const modelStats = pgTable(
 export type ModelStats = typeof modelStats.$inferSelect;
 export type NewModelStats = typeof modelStats.$inferInsert;
 
+export const model_eval_ingest = pgTable(
+  'model_eval_ingest',
```
jrf0110: nit; Most other table names follow a pluralized convention (i.e. the table name tells us what objects are in this table). I'd recommend renaming the table to something like `model_evals` or `model_eval_ingestions`

packages/db/src/schema.ts
```diff
@@ -2975,6 +2997,55 @@ export const modelStats = pgTable(
...
+    avg_cache_read_tokens: decimal('avg_cache_read_tokens', {
+      precision: 16,
+      scale: 6,
+      mode: 'number',
+    }),
+    avg_execution_ms: decimal('avg_execution_ms', { precision: 16, scale: 6, mode: 'number' }),
+    promoted_at: timestamp('promoted_at', { withTimezone: true, mode: 'string' }).notNull(),
+    promoted_by_email: text('promoted_by_email').notNull(),
+    promotion_note: text('promotion_note'),
+    ingested_at: timestamp('ingested_at', { withTimezone: true, mode: 'string' })
```
jrf0110: nit; We generally use the name `created_at`. `model_eval_investions.created_at` semi-semantically tells us when the thing was ingested

packages/db/src/schema.ts
```diff
@@ -2975,6 +2997,55 @@ export const modelStats = pgTable(
...
+    n_errored: integer('n_errored').notNull(),
+    avg_cost_usd: decimal('avg_cost_usd', { precision: 14, scale: 8, mode: 'number' }),
+    avg_input_tokens: decimal('avg_input_tokens', { precision: 16, scale: 6, mode: 'number' }),
+    avg_output_tokens: decimal('avg_output_tokens', { precision: 16, scale: 6, mode: 'number' }),
+    avg_cache_read_tokens: decimal('avg_cache_read_tokens', {
+      precision: 16,
+      scale: 6,
+      mode: 'number',
+    }),
+    avg_execution_ms: decimal('avg_execution_ms', { precision: 16, scale: 6, mode: 'number' }),
```
jrf0110: Do we need floating point precision for a value measured in `ms`?

packages/db/src/schema.ts
```diff
@@ -2975,6 +2997,55 @@ export const modelStats = pgTable(
...
+    overall_score: decimal('overall_score', { precision: 12, scale: 8, mode: 'number' }).notNull(),
+    n_errored: integer('n_errored').notNull(),
+    avg_cost_usd: decimal('avg_cost_usd', { precision: 14, scale: 8, mode: 'number' }),
+    avg_input_tokens: decimal('avg_input_tokens', { precision: 16, scale: 6, mode: 'number' }),
+    avg_output_tokens: decimal('avg_output_tokens', { precision: 16, scale: 6, mode: 'number' }),
+    avg_cache_read_tokens: decimal('avg_cache_read_tokens', {
+      precision: 16,
+      scale: 6,
+      mode: 'number',
+    }),
```
jrf0110: I'd argue all of these should be integers

packages/db/src/schema.ts
```diff
@@ -2975,6 +2997,55 @@ export const modelStats = pgTable(
...
+    provider: text('provider').notNull(),
+    model: text('model').notNull(),
+    model_stats_id: uuid('model_stats_id').references(() => modelStats.id),
+    variant: text('variant'),
+    task_source: text('task_source').notNull(),
+    n_total_trials: integer('n_total_trials').notNull(),
+    total_score: decimal('total_score', { precision: 14, scale: 6, mode: 'number' }).notNull(),
+    overall_score: decimal('overall_score', { precision: 12, scale: 8, mode: 'number' }).notNull(),
+    n_errored: integer('n_errored').notNull(),
+    avg_cost_usd: decimal('avg_cost_usd', { precision: 14, scale: 8, mode: 'number' }),
```
jrf0110: We generally use microdollars for reporting cost

lambertjosh: Thanks. This is coming from the bench data which reports/shows in dollars today, so would you be OK to convert this into microdollars as part of the ingest? (read in dollars from bench, convert to microdollars, store in cloud db)

packages/db/src/schema.ts
lambertjosh: No! Ugh. Fixing.

packages/db/src/schema.ts
lambertjosh: Yea, they are averages but we should just round. We don't really need this level of precision.

packages/db/src/schema.ts
lambertjosh: Fixed.

packages/db/src/schema.ts
lambertjosh: Fixed

packages/db/src/schema.ts
lambertjosh: All integers now with the exception of microdollar which is bigint, as otherwise the limit would be an avg cost of ~2000usd. Since that is only a factor of 100 off from results we have seen, I bumped it to be safe.

packages/db/src/schema.ts
lambertjosh: Adjusted to microdollars.

lambertjosh:
@jrf0110 - ready for re-review I think. Thanks for the review so far.

apps/web/src/lib/model-eval-ingest-client.ts
```diff
@@ -0,0 +1,45 @@
...
+): Promise<ModelEvalSyncResult> {
+  if (!MODEL_EVAL_INGEST_URL) {
+    throw new Error('MODEL_EVAL_INGEST_URL is not configured');
+  }
+
+  const response = await fetch(`${MODEL_EVAL_INGEST_URL}/internal/sync`, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      'x-internal-api-key': INTERNAL_API_SECRET,
```
jrf0110, line 31: I thought you had moved to using cloudflare access service credentials? If so, you'd need to plumb through your client_id/secret and add the headers:

```
curl -H "CF-Access-Client-Id: <CLIENT_ID>" -H "CF-Access-Client-Secret: <CLIENT_SECRET>" https://app.example.com
```

jrf0110 (approved):

apps/web/src/lib/model-eval-ingest-client.ts
lambertjosh: @jrf0110 - I did for the rest of it. This is authentication between the Vercel app and the Worker, to manually trigger a re-scan for promoted models. (Otherwise it's 15min)

This was an existing secret/pattern used by other app->worker requests, so just re-used.

I can swap it out and implement a new Client Access based approach if we want.

PR #3260, opened 2026-05-15
perf(kiloclaw): index-backed first-instance lookup for setup promo

RSO:
## Summary

The `userIsWithinFirstKiloClawInstanceWindow` query in `apps/web/src/lib/kiloclaw/setup-promo.ts` was responsible for ~21% of read-replica time-consumed.

The original query used `min(created_at) >= now() - interval` and filtered only on `user_id`. Every existing index on `kiloclaw_instances.user_id` is partial on `destroyed_at IS NULL`, so the planner couldn't use them — destroyed rows must still count for "first instance" semantics, otherwise a returning user could destroy their first instance and re-qualify for the setup-promo window indefinitely.

Two changes:

1. **Schema** — add a non-partial `(user_id, created_at)` btree index (`IDX_kiloclaw_instances_user_id_created_at`). This is the only index that can serve a `user_id`-only lookup over the full instance history.
2. **Query rewrite** — replace the `min()` aggregate with `ORDER BY created_at LIMIT 1` plus a JS-side window check. `EXPLAIN` confirms this becomes an index-only scan returning a single row.

The function's behavior is unchanged. New tests in `setup-promo.test.ts` (added in the first commit on this branch) lock in the semantics — including the destroyed-rows-still-count case — and pass against both the old and new implementations.

The migration uses `CREATE INDEX CONCURRENTLY` (precedent: `0115_real_energizer.sql`) since `kiloclaw_instances` is a hot table and a blocking index build would lock writes.

## Verification

- Confirmed with `EXPLAIN` against the local test database that the rewritten query is an index-only scan using `IDX_kiloclaw_instances_user_id_created_at`, no heap fetches, no sequential scan.
- Spot-checked the same `EXPLAIN` for the original `min()` query with the new index in place — planner correctly rewrites it to a Limit + Index Only Scan, confirming the index alone would have resolved the perf issue.

## Visual Changes

N/A

## Reviewer Notes

- The migration starts with `COMMIT;` and ends with `BEGIN;` to allow `CREATE INDEX CONCURRENTLY` to run outside the transaction Drizzle wraps migrations in. Same pattern as `packages/db/src/migrations/0115_real_energizer.sql`. Worth a sanity check from whoever owns the migration apply path.
- Behavior is unchanged but worth confirming the product intent: destroyed instances still count toward "first instance," meaning a user who created their first instance two months ago, destroyed it, and creates a new one today is **not** within the window. This matches the previous `min()` semantics.
- The rewrite intentionally does the time comparison in JS instead of SQL. Two reasons: it keeps the SQL purely an index-lookup with no expression evaluation, and it makes the query plan independent of the planner's `min()`-via-index optimization (which can occasionally be defeated by parallel workers / unusual stats).

### DIFF
--- a/apps/web/src/lib/kiloclaw/setup-promo.test.ts
+++ b/apps/web/src/lib/kiloclaw/setup-promo.test.ts
@@ -1,0 +1,175 @@
+import { beforeEach, describe, expect, it } from '@jest/globals';
+import { kiloclaw_instances } from '@kilocode/db/schema';
+
+import { cleanupDbForTest, db } from '@/lib/drizzle';
+import { insertTestUser } from '@/tests/helpers/user.helper';
+
+import { userIsWithinFirstKiloClawInstanceWindow } from './setup-promo';
+
+const HOUR_MS = 60 * 60 * 1000;
+
+function hoursAgo(hours: number): string {
+  return new Date(Date.now() - hours * HOUR_MS).toISOString();
+}
+
+async function insertInstance(params: {
+  userId: string;
+  createdAt: string;
+  destroyedAt?: string | null;
+  sandboxSuffix?: string;
+}) {
+  const id = crypto.randomUUID();
+  await db.insert(kiloclaw_instances).values({
+    id,
+    user_id: params.userId,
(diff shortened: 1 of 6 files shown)


jeanduplessis (approved):
Reviewed the final combined diff and PR context. Approving per request without line comments.

PR #3269, opened 2026-05-15
fix(billing): normalize side-effect timestamp payloads

jeanduplessis:
## Summary
Normalize KiloClaw billing side-effect timestamps before strict API validation, and harden Vercel workflow setup used by CI/deploy jobs.

### Why this change is needed
Credit-renewal side-effect requests can carry Drizzle/Postgres timestamp text such as `2026-04-29 01:16:12.945+00`. The internal side-effects route intentionally requires strict ISO datetime strings, so those payloads were rejected with HTTP 400 `Invalid body` during paid conversion and auto top-up flows.

The branch also carries workflow hardening for `pnpm dlx vercel` resolution behavior and production deploy ordering already present in local deploy workflow edits.

### How this is addressed
- Normalize DB-backed renewal and user timestamps to UTC ISO before billing side-effects POST bodies are built.
- Keep strict route-side datetime validators unchanged and add production-shape regression coverage for the exact rejected fields.
- Record agent guidance for serializing DB-backed timestamp strings at strict JSON boundaries.
- Apply highest pnpm resolution mode to Vercel CI usage and keep production deploy workflow prerequisites explicit.

## Human Verification
- `pnpm --filter web test -- --runInBand "src/app/api/internal/kiloclaw/billing-side-effects/route.test.ts"`
- `pnpm --filter kiloclaw-billing test -- "src/lifecycle.test.ts"`

### Human Reviewer Flags
- Side-effects contract remains strict; normalization happens at worker egress rather than widening accepted API input.
- PR includes CI/deploy workflow hardening alongside billing regression fix because both were present in current branch scope.

### Code Reviewer Agent
<details>

- Route regression tests prove Postgres-shaped timestamps fail strict validation before worker normalization.
- Lifecycle regression tests prove paid-conversion and auto-top-up payloads emit UTC ISO strings after normalization.
- Existing queue timestamp normalization was generalized instead of creating separate one-off serializers.

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,8 @@ permissions:
 
 env:
   VERCEL_VERSION: '53.3.1'
+  # pnpm dlx vercel can hit ERR_PNPM_MISSING_TIME with default resolution mode.
+  PNPM_CONFIG_RESOLUTION_MODE: 'highest'
 
 # Optional repo/org variables for runner migration:
 # - RUNNER_DEFAULT_LABEL=blacksmith-2vcpu-ubuntu-2404
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -13,6 +13,8 @@ concurrency:
 
 env:
   VERCEL_VERSION: '53.3.1'
+  # pnpm dlx vercel can hit ERR_PNPM_MISSING_TIME with default resolution mode.
+  PNPM_CONFIG_RESOLUTION_MODE: 'highest'
 
 jobs:
   check-production-db-startup:
@@ -48,7 +50,6 @@ jobs:
   run-migrations:
     runs-on: ${{ vars.RUNNER_DEFAULT_LABEL || 'ubuntu-latest' }}
     timeout-minutes: 15
-    needs: check-production-db-startup
     environment: production
 
     steps:
@@ -79,7 +80,7 @@ jobs:
   deploy-app:
     runs-on: ${{ vars.RUNNER_LARGE_LABEL || 'ubuntu-24.04-8core' }}
     timeout-minutes: 30
-    needs: run-migrations
+    needs: [check-production-db-startup, run-migrations]
     environment: production
 
     env:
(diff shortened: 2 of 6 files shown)


.github/workflows/deploy-production.yml
```diff
@@ -48,7 +50,6 @@ jobs:
   run-migrations:
     runs-on: ${{ vars.RUNNER_DEFAULT_LABEL || 'ubuntu-latest' }}
     timeout-minutes: 15
-    needs: check-production-db-startup
     environment: production
```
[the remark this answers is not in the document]
jeanduplessis: This was discussed internally and decided to accept the risk for now and save the time from serializing the jobs.

.github/workflows/ci.yml
```diff
@@ -17,6 +17,8 @@ permissions:
 
 env:
   VERCEL_VERSION: '53.3.1'
+  # pnpm dlx vercel can hit ERR_PNPM_MISSING_TIME with default resolution mode.
+  PNPM_CONFIG_RESOLUTION_MODE: 'highest'
```
iscekic, line 21: This seems wrong... Why don't we pin the version of the vercel cli instead?

.github/workflows/deploy-production.yml
```diff
@@ -13,6 +13,8 @@ concurrency:
 
 env:
   VERCEL_VERSION: '53.3.1'
+  # pnpm dlx vercel can hit ERR_PNPM_MISSING_TIME with default resolution mode.
+  PNPM_CONFIG_RESOLUTION_MODE: 'highest'
```
iscekic, line 17: Ditto

iscekic (changes requested):

.github/workflows/ci.yml
jeanduplessis: We do pin the version, however, a transitive dependency of the Vercel CLI, undici, doesn't specify time in its metadata causing failures. It's sporadic, but its happened a few times in CI runs today.

image

iscekic (approved):
Per discussion on Slack - approving to unblock the bugfix, @jeanduplessis to follow up with a revert of the pnpm setting when the registry bug (missing time) is fixed.

PR #3272, opened 2026-05-15
fix(code-reviews): disable interactive tools

alex-alecu:
## Why

Code reviews could get stuck when the agent asked a question that no one could answer. A review should keep moving without waiting for chat input.

## What changed

Code-review runs now start without the tools that let the agent ask questions or enter planning mode. If a question or permission request still shows up, the wrapper rejects it and does not treat it like healthy activity. Normal cloud-agent sessions still keep their current interactive behavior. Tests cover new prompts, live events, and reconnect snapshots.

## How to test

1. Run `pnpm --filter cloud-agent-next test src/execution/orchestrator.test.ts test/unit/wrapper/snapshot.test.ts test/unit/wrapper/reconnection.test.ts`.
2. Run `pnpm --filter cloud-agent-next typecheck`.
3. Start a code-review run and confirm it does not wait on question or plan tools.

### DIFF
--- a/services/cloud-agent-next/src/execution/orchestrator.ts
+++ b/services/cloud-agent-next/src/execution/orchestrator.ts
@@ -348,7 +348,7 @@ export class ExecutionOrchestrator {
             originalToken: initContext.kilocodeToken,
             kilocodeModel: initContext.kilocodeModel ?? 'default',
             originalOrgId: orgId,
-            createdOnPlatform: initContext.createdOnPlatform,
+            createdOnPlatform: this.getCreatedOnPlatform(plan),
             appendSystemPrompt: existingMetadata.appendSystemPrompt,
             profile: mergeFastPathProfile(initContext.profile, existingMetadata.profile),
           }),
--- a/services/cloud-agent-next/src/execution/orchestrator.ts
+++ b/services/cloud-agent-next/src/execution/orchestrator.ts
@@ -31,6 +31,12 @@ import { withSandboxInternalServerErrorRecovery } from '../sandbox-recovery.js';
 /** Maximum time allowed for workspace preparation (resume, init, fast path). */
 const PREPARE_WORKSPACE_TIMEOUT_MS = 10 * 60 * 1000;
 
+const CODE_REVIEW_DISABLED_TOOLS = {
+  question: false,
+  plan_enter: false,
+  plan_exit: false,
+} satisfies Record<string, boolean>;
+
 function withWorkspacePreparationTimeout<T>(operation: Promise<T>, step: string): Promise<T> {
   return withTimeout(
     operation,
@@ -229,6 +235,7 @@ export class ExecutionOrchestrator {
       autoCommit: wrapper.autoCommit,
       condenseOnComplete: wrapper.condenseOnComplete,
       execution,
+      tools: this.getToolOverrides(plan),
     };
 
     if (fileParts.length > 0) {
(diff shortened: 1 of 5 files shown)


services/cloud-agent-next/src/execution/orchestrator.ts
```diff
@@ -487,6 +494,12 @@ export class ExecutionOrchestrator {
     );
   }
 
+  private getToolOverrides(plan: ExecutionPlan): Record<string, boolean> | undefined {
+    return this.getCreatedOnPlatform(plan) === 'code-review'
```
alex-alecu, line 498: Could this metadata fallback drift from the fast-path session env? In the prepared-session fast path, `getToolOverrides` can use `existingMetadata.createdOnPlatform` from metadata, but the wrapper still receives `KILO_PLATFORM` from `initContext.createdOnPlatform` only, so question/permission rejection would not run on that same path.

services/cloud-agent-next/wrapper/src/connection.ts
```diff
@@ -226,23 +226,33 @@ export function createConnectionManager(
 
       const pendingQuestion = questions.find(q => q.sessionID === kiloSessionId);
       const pendingPermission = permissions.find(p => p.sessionID === kiloSessionId);
+      const codeReviewJob = isCodeReviewJob(state);
+      const skipStatusForCodeReview =
+        codeReviewJob && (sessionStatus.type === 'question' || sessionStatus.type === 'permission');
```
alex-alecu: The `'question' | 'permission'` interactive-status check is hardcoded both here and again in the real-time loop around lines 525-537. Could we extract a shared predicate so the snapshot and real-time paths cannot drift if a third interactive status type is ever added?

services/cloud-agent-next/wrapper/src/connection.ts
```diff
@@ -510,23 +519,20 @@ export function createConnectionManager(
             if (eventType === 'question.asked') {
               const questionId = typeof properties.id === 'string' ? properties.id : undefined;
               rejectCodeReviewQuestion(questionId, config.kiloClient);
-              callbacks.onSseEvent?.();
               continue;
```
alex-alecu, line 522: With the `callbacks.onSseEvent?.()` calls removed from these branches (also at 503, 529, and 536), only `server.heartbeat` resets the 15s SSE transport watchdog. If a code-review session receives a burst of filtered question/permission/status events with no heartbeat for >15s, will that fire a spurious `reconnectEventSubscription()` plus a fresh snapshot round-trip, even though transport is clearly healthy when these events arrive? Was excluding the arrivals from `onSseEvent` intentional given that the watchdog measures transport liveness rather than session liveness?

services/cloud-agent-next/src/execution/orchestrator.ts
alex-alecu: Fixed in c6213fc1c.

services/cloud-agent-next/wrapper/src/connection.ts
alex-alecu: Done in c9901610b.

services/cloud-agent-next/wrapper/src/connection.ts
alex-alecu: Fixed in 675a23bfc.

jrf0110 (approved):

services/cloud-agent-next/wrapper/src/connection.ts
```diff
@@ -226,23 +230,32 @@ export function createConnectionManager(
 
       const pendingQuestion = questions.find(q => q.sessionID === kiloSessionId);
       const pendingPermission = permissions.find(p => p.sessionID === kiloSessionId);
+      const codeReviewJob = isCodeReviewJob(state);
+      const skipStatusForCodeReview = codeReviewJob && isInteractiveStatusType(sessionStatus.type);
+
+      if (codeReviewJob) {
+        rejectCodeReviewQuestion(pendingQuestion?.id, config.kiloClient);
+        rejectCodeReviewPermission(pendingPermission?.id, config.kiloClient);
+      }
```
eshurakov: I don't think we should be doing that in the snapshot as it is a non mutable "getter"

alex-alecu: Done in 6489a8d.

PR #3278, opened 2026-05-15
feat(kiloclaw) briefing local news

St0rmz1:
## Summary

Adds a Local News section to the Kiloclaw morning briefing and a Settings card for editing the location the briefing uses.

The Local News source is opt in via the user's interest topics. When the user selects `Local News`, the briefing runs a tiered web search keyed off the user's explicit location. Each tier widens the radius and time window so the brief still surfaces something useful when a tighter query returns nothing. The first tier that returns at least 3 unique items stops the escalation; results are capped at 10. If no explicit location is set, the section renders a nudge that points to the Settings card. The user's IANA timezone is captured for context only and never used as a query source, since `America/Los_Angeles` covers the entire US Pacific coast and would produce misleading results.

Settings gets a new Morning Briefing card with a Location editor. Edits write to the plugin's `config.json` via a new POST route on the plugin and a matching gateway route on the controller. The plugin reads the stored location at the start of each brief, so changes take effect on the next run with no container restart. Clearing falls back to the env var captured during onboarding, then to the no location nudge.

## Verification

- [x] `pnpm --filter kiloclaw-morning-briefing test` passes; new tests cover the tier loop, the no provider and no location branches, the throw on every tier swallow path, and dedupe edge cases.
- [x] Manual: built the kiloclaw image with `services/kiloclaw/scripts/build-local-image.sh`, ran the briefing locally, set and cleared the location through Settings, and confirmed the next brief switched between the tiered Local News section and the no location nudge with no restart.

## Visual Changes

| Before | After |
| ------ | ----- |
|        |       |

## Reviewer Notes

Plugin (`services/kiloclaw/plugins/kiloclaw-morning-briefing/`)
- `local-news-utils.ts`: location resolution, tier query builder, dedupe, section title and footer summary helpers. Includes a stored override path used by the Settings editor.
- `index.ts`: `collectLocalNews` runs the tier loop, swallows per tier search errors, and reports a status entry. New POST `/api/plugins/kiloclaw-morning-briefing/user-location` route accepts `{ userLocation: string | null }` and persists to `config.json`.
- Test files cover unit and lifecycle paths for the new source.

Controller and web
- New `/_kilo/morning-briefing/user-location` gateway route.
- New Morning Briefing Settings card with a Location editor. The status payload now exposes `userLocation`; tRPC and the `KiloClawDashboardStatus` type were updated to match.

Smaller related changes
- Onboarding walkthrough copy and state tests pick up the new interest topic context.
- Gateway adds a passthrough for the user location route on the instance side.

Notes
- Tier escalation stops at the first tier that returns 3 or more unique items; results capped at 10. Both constants live at the top of `local-news-utils.ts`.
- Stored location takes priority over the env var captured at provision time. Whitespace only counts as unset and falls through to the env var, then to the nudge.
- Local News never reads the timezone as a query source, only as decoration on the no location nudge.


services/kiloclaw/plugins/kiloclaw-morning-briefing/src/index.lifecycle.test.ts
```diff
@@ -105,6 +106,34 @@ async function createHarness(options?: {
...
+   */
+  webSearch?: {
+    providers?: Array<{ id?: string }>;
+    resultsPerQuery?: Record<
+      string,
+      { provider?: string; results?: Array<{ title: string; url: string; summary?: string }> }
+    >;
+    defaultResult?: { provider?: string; results?: Array<{ title: string; url: string }> };
+    /** When set, every call to `search()` rejects with this error. */
+    searchThrows?: Error;
```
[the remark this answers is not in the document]
St0rmz1: This is exercised at line 1692 ("swallows webSearch.search() throws across all tiers and reports 0 results"). The test wires `searchThrows: new Error('rate limit exceeded')` and asserts the brief still ships with a "No local news found" message and the source summary records 0 results. The harness option is intentional and covered.

apps/web/src/app/(app)/claw/components/SettingsTab.tsx
```diff
@@ -502,6 +502,164 @@ function GoogleAccountCard({
...
+      }
+    );
+  }
+
+  function handleClear() {
+    mutations.updateConfig.mutate(
+      { userLocation: null },
+      {
+        onSuccess: () => {
+          toast.success('Location cleared. Local News will be disabled until you set one.', {
```
jeanduplessis: WARNING: Clearing here does not necessarily disable Local News. The plugin treats `userLocation: null` as "remove stored override, then fall back to `KILOCLAW_USER_LOCATION`", so instances provisioned with an onboarding location can keep generating Local News from that env-backed value after this toast says it is disabled.

Trace:
1. Clear mutation sends `{ userLocation: null }`.
2. Local News resolution falls back from null stored override to env-backed onboarding location.
3. UI confirmation says Local News is disabled, which is false for that state.

Impact: User believes clear disabled Local News, but next briefing can still query old onboarding location.

services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
```diff
@@ -979,6 +979,29 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
...
+      if (userLocation !== previousUserLocation) {
+        try {
+          await gateway.updateMorningBriefingUserLocation(this.s, this.env, {
+            userLocation,
+          });
+        } catch (err) {
+          // Mirror the rest of this method's pattern of letting writes
+          // through gateway helpers raise. Worth a warn log so the
+          // failure is visible without blocking the DO save.
+          doWarn(this.s, 'updateMorningBriefingUserLocation failed', {
```
jeanduplessis: WARNING: This best-effort sync can fail while the config mutation still succeeds, leaving the running plugin on stale location state even though Settings says the next briefing will use the new value.

Trace:
1. DO/profile state is persisted first.
2. Plugin `config.json` sync is attempted afterward and both thrown errors and unsupported-route `null` outcomes are effectively ignored by the caller.
3. The plugin reads its own stored config on the next run, so stale plugin state can survive the accepted save.

Impact: Location saves can look successful while the next briefing still uses the previous location until a later successful sync or restart/redeploy.

services/kiloclaw/plugins/kiloclaw-morning-briefing/src/index.ts
```diff
@@ -1771,6 +2071,61 @@ export default definePluginEntry({
...
+        try {
+          const body = asObject(await readRequestBody(req));
+          // Three accepted shapes:
+          //   - { userLocation: "Novato, CA" }  → set
+          //   - { userLocation: null }          → clear
+          //   - { userLocation: "" }            → clear (trim-empty)
+          // Any other type (number, array, object) is rejected.
+          if (
+            body.userLocation !== null &&
+            body.userLocation !== undefined &&
```
jeanduplessis: WARNING: Missing or malformed `userLocation` payloads clear the stored override instead of being rejected. `undefined` passes this guard, `next` defaults to `null`, and the route persists that null.

Trace:
1. New route contract is documented as `{ userLocation: string | null }`.
2. Controller JSON parse fallback can yield `{}`.
3. `{}` reaches this handler as `body.userLocation === undefined`, then writes `null`.

Impact: Invalid JSON, `{}`, or a buggy caller that omits the field can erase the saved location instead of returning 400.

services/kiloclaw/plugins/kiloclaw-morning-briefing/src/index.ts
```diff
@@ -837,38 +899,215 @@ async function collectWebSearch(
...
+      api.logger.warn?.(`Morning briefing web search topic '${topic}' failed: ${errorText}`);
+    }
+  }
+
+  if (accumulated.length === 0) {
+    return {
+      source: 'web',
+      configured: true,
+      ok: true,
+      summary: 'Web search returned no results',
```
jeanduplessis, line 962: WARNING: If every per-topic web-search request throws, this path reports "Web search returned no results" with `ok: true` instead of surfacing a failed source.

Trace:
1. Each topic failure is swallowed and only logged.
2. When all searches fail, `accumulated` remains empty.
3. Empty accumulation falls through to the success/no-results response, so the briefing failure list never records the outage.

Impact: Rate limits, auth breaks, or provider outages are misclassified as valid empty search results, hiding real failures from status and delivery diagnostics.

services/kiloclaw/plugins/kiloclaw-morning-briefing/src/index.ts
```diff
@@ -794,38 +831,63 @@ async function collectGithub(api: GithubApiRunner): Promise<SourceCollectionResu
...
+  // Caller already filtered out "Local News" — that has its own source.
+  // If nothing remains, the user hasn't opted into general web news.
+  // Surface a nudge instead of forcing a hardcoded fallback query.
+  const cleanedTopics = interestTopics.map(topic => topic.trim()).filter(topic => topic.length > 0);
+  if (cleanedTopics.length === 0) {
+    return {
+      source: 'web',
+      configured: false,
+      ok: true,
+      summary: 'No general-interest topics selected',
```
jeanduplessis, line 884: WARNING: Empty-interest configs no longer use the legacy fallback engineering-news query. Older instances with no stored topics now render a nudge instead of the general web headlines they previously received.

Trace:
1. Deleted code explicitly preserved a hardcoded fallback query for empty topic sets.
2. Stored config still defaults missing or malformed `interestTopics` to `[]`.
3. This new branch treats `[]` as "no web source configured" and skips the old behavior.

Impact: Existing briefings can silently lose general web-search content after deploy, especially legacy instances that predate saved briefing interests.

jeanduplessis (approved):
Reviewed final combined diff plus PR body, unresolved review threads, controller spec context, PR-head source, and changed tests.

Findings focus on concrete changed-line regressions:
- clear-location UX contradicts env fallback behavior,
- running-instance location sync can succeed in UI but miss plugin state,
- malformed/missing user-location requests can clear config,
- all web-topic search failures are misreported as no results,
- empty-interest legacy briefs lose prior fallback web coverage.

Submitting as Approve per request, with warnings left inline for follow-up.

services/kiloclaw/plugins/kiloclaw-morning-briefing/src/index.ts
St0rmz1: This was an intended change as the results, when no interests selected, ran the risk of being very inappropriate. 
If none are selected, then don't run Web Search. Instead let the user know that they can select some interests for general web search/news if they'd like. but don't run web search if no topics selected.

PR #3281, opened 2026-05-15
fix(kiloclaw): run destruction sweep every 15 minutes

jeanduplessis:
## Summary

- Reduce deadline-to-destroy lag by scheduling `instance_destruction` every 15 minutes instead of relying only on hourly lifecycle kickoff.
- Keep hourly lifecycle and daily trial inactivity cron behavior unchanged.
- Use standalone queue work that reuses the existing sweep execution path while preventing downstream lifecycle chaining.

## Verification

- N/A - no manual runtime exercise for cron trigger or queue delivery in local environment.

## Visual Changes

N/A

## Reviewer Notes

- `standalone_instance_destruction` intentionally completes after `runSweep(...)`; unlike hourly lifecycle messages, it never calls `nextSweep(...)`.
- Normal hourly `instance_destruction` still advances to `past_due_cleanup`, preserving existing sweep order.

### DIFF
--- a/services/kiloclaw-billing/src/index.ts
+++ b/services/kiloclaw-billing/src/index.ts
@@ -6,6 +6,7 @@ import {
   BILLING_HOURLY_CRON,
   BILLING_QUEUE_MAX_RETRIES,
   BILLING_SWEEP_ORDER,
+  INSTANCE_DESTRUCTION_QUARTER_HOURLY_CRON,
   TRIAL_INACTIVITY_DAILY_CRON,
   TRIAL_INACTIVITY_STOP_CANDIDATE_SWEEP,
   TRIAL_INACTIVITY_SWEEP,
@@ -40,6 +41,12 @@ const LifecycleQueueMessageSchema = z.object({
   kind: z.literal('lifecycle'),
   runId: z.string().uuid(),
   sweep: z.enum(BILLING_SWEEP_ORDER),
+});
+
+const StandaloneInstanceDestructionQueueMessageSchema = z.object({
+  kind: z.literal('standalone_instance_destruction'),
+  runId: z.string().uuid(),
+  sweep: z.literal('instance_destruction'),
 });
 
 const TrialInactivityQueueMessageSchema = z.object({
@@ -109,6 +116,7 @@ const CreditRenewalTerminalFailureQueueMessageSchema = z.object({
 
 const BillingQueueMessageSchema = z.discriminatedUnion('kind', [
   LifecycleQueueMessageSchema,
+  StandaloneInstanceDestructionQueueMessageSchema,
   TrialInactivityQueueMessageSchema,
   TrialInactivityStopCandidateQueueMessageSchema,
   CreditRenewalDiscoveryQueueMessageSchema,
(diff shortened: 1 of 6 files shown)


services/kiloclaw-billing/wrangler.jsonc
```diff
@@ -20,7 +20,7 @@
     "cpu_ms": 120000,
   },
   "triggers": {
-    "crons": ["0 * * * *", "0 8 * * *"],
+    "crons": ["0 * * * *", "*/15 * * * *", "0 8 * * *"],
```
St0rmz1: At minute 0 both crons fire. The hourly chain eventually reaches `instance_destruction` while the new standalone path may already be running it. `runInstanceDestructionSweep` selects all rows where `destruction_deadline < now` and only nulls that field after destroying, so two concurrent runs can fetch the same rows before either updates them.

Practical impact looks bounded but real:
- `destroyInstance` treats a 409 from claw as expected, so the external destroy is safe.
- `trySendEmail` dedupes via the `kiloclaw_email_log` unique row plus `onConflictDoNothing`, so duplicate emails are dropped.
- `markInstanceDestroyedWithPersonalSubscriptionCollapse` reads then writes inside a transaction, so a concurrent run can operate on stale state.
- `insertLifecycleChangeLogBestEffort` is not deduped, so duplicate change log rows can land per destroyed instance.

Two options:
1. Move the standalone cron off the top of the hour, for example `5,20,35,50 * * * *`, so it never collides with the hourly trigger.
2. Claim each row atomically by including `destruction_deadline IS NOT NULL` in the candidate filter and updating it to NULL with a `RETURNING` clause before the destroy call.

Option 1 is the smaller change.

services/kiloclaw-billing/src/index.ts
```diff
@@ -441,6 +479,12 @@ export const handler: ExportedHandler<BillingWorkerEnv, BillingQueueMessage> = {
                 subscriptionId: parsed.data.subscriptionId,
                 renewalBoundary: parsed.data.renewalBoundary,
               });
+            } else if (parsed.data.kind === 'standalone_instance_destruction') {
+              await runSweep(env, parsed.data, message.attempts);
```
St0rmz1, line 483: `runInstanceDestructionSweep` is unpaginated. It selects all eligible rows in one query and processes them sequentially with external API calls and DB transactions, with no `pageBudget`, no wallclock guard, and no continuation. With this PR it now runs five times per hour (4 standalone plus 1 from the hourly chain). If the destruction backlog grows, each invocation can stretch toward the 120s `cpu_ms` ceiling, which increases overlap risk and queue timeouts.

`credit_renewal_discovery` and (in PR 3280) `trial_expiry` were both paginated for the same reason. Two options:
1. Treat as a follow up and paginate this sweep separately, similar to the trial expiry change.
2. As a minimal defensive change in this PR, add a `LIMIT N` to the candidate query so a single sweep cannot process more than a bounded batch in one message.

St0rmz1 (approved):

services/kiloclaw-billing/wrangler.jsonc
jeanduplessis: Adjusted the standalone schedule to `5,20,35,50 * * * *`, keeping the 15-minute cadence off the top of the hour instead of starting alongside the hourly kickoff. I also bounded destruction candidates in the follow-up fix, which keeps standalone runs shorter.

PR #3282, opened 2026-05-15
fix(cloud-agent-next): translate local backend URLs

eshurakov:
## Summary

- Keep local Cloud Agent backend URL examples as worker-reachable `localhost` values.
- Translate localhost backend and provider URLs to `host.docker.internal` only when injecting sandbox env/config.
- Preserve those worker-local defaults in dev env sync and cover the behavior with tests.

## Verification

N/A - no manual verification performed; this is local dev URL/env shaping with no UI flow.

## Visual Changes

N/A

## Reviewer Notes

The key boundary is that worker-side local fetches still use `localhost`, while sandbox-injected values are rewritten for Docker container reachability.


jeanduplessis (approved):
Reviewed final combined diff, PR body, and unresolved review-thread state. No issues found.

PR #3285, opened 2026-05-16
fix(kiloclaw): skip destroyed instances in expiry sweep

jeanduplessis:
## Summary
- Exclude already-destroyed KiloClaw instances from subscription-expiry sweep candidates.
- Prevent repeated no-op expiry skip warnings for rows that no longer need enforcement.
- Add SQL-shape coverage so destroyed instances stay filtered at query time.

## Verification
- N/A

## Visual Changes
- N/A

## Reviewer Notes
- Existing row-level destroyed-instance guard stays in place as defensive fallback.
- Targeted verification run in session:
  - `pnpm --filter kiloclaw-billing test -- src/lifecycle.test.ts`
  - `pnpm --filter kiloclaw-billing typecheck`

### DIFF
--- a/services/kiloclaw-billing/src/lifecycle.test.ts
+++ b/services/kiloclaw-billing/src/lifecycle.test.ts
@@ -5224,6 +5224,51 @@ describe('complementary inference ended sweep', () => {
   });
 });
 
+describe('subscription expiry sweep', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    mockGetWorkerDb.mockReset();
+    vi.spyOn(console, 'log').mockImplementation(() => undefined);
+    vi.spyOn(console, 'error').mockImplementation(() => undefined);
+  });
+
+  it('filters destroyed instances in SQL', async () => {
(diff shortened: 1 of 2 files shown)


St0rmz1:
One thing worth flagging for parity: two sibling sweeps in this same file have an identical structure (left join `kiloclaw_instances`, defensive `if (row.instance_destroyed_at) continue` inside the loop, but no SQL-level `isNull(destroyed_at)` filter on the candidate query):

- `runPastDueCleanupSweep` at `services/kiloclaw-billing/src/lifecycle.ts:3007-3028` (in-loop guard at line 3051)
- `runTrialWarningSweep` at `services/kiloclaw-billing/src/lifecycle.ts:3258-3281` (in-loop guard at line 3304)

Functionally those sweeps are still correct because the in-loop guard catches destroyed rows. The motivation for this PR (cutting repeated no-op skip warnings on every run) will probably show up there too over time, since they hit the same defensive log path. If you want to keep this PR laser-focused, a follow-up note in the description or a tracking ticket would be enough so the parity work does not get lost.

St0rmz1 (approved):

PR #3286, opened 2026-05-17
fix(ai-gateway): log upstream transport failures safely

jeanduplessis:
## Summary
- Add safe upstream fetch failure instrumentation so transport incidents identify provider, target host, path, and failure family.
- Preserve existing request behavior by rethrowing original fetch errors unchanged.
- Add regression coverage for timeout-shaped failures and confirm secrets/body data stay out of emitted metadata.

## Verification
- Not manually verified.

## Visual Changes
N/A

## Reviewer Notes
- Failure-family mapping intentionally stays narrow and deterministic; unknown runtime shapes fall back to `unknown`.
- Automated check run in workspace: `pnpm test -- apps/web/src/tests/openrouterApi.timeout.test.ts --runInBand`.

### DIFF
--- a/apps/web/src/lib/ai-gateway/providers/upstream-request.ts
+++ b/apps/web/src/lib/ai-gateway/providers/upstream-request.ts
@@ -178,36 +178,40 @@ export async function upstreamRequest({
       signal: combinedSignal,
     });
   } catch (error) {
-    const cause = error instanceof Error ? error.cause : undefined;
-    const errorName = getErrorName(error);
-    const errorMessage = getErrorMessage(error);
-    const causeCode = getCauseCode(cause);
-    const causeName = getCauseName(cause);
-    const causeMessage = getCauseMessage(cause);
-    const failureFamily = classifyUpstreamFetchFailure({ errorName, causeCode, causeName });
-    const failureMetadata = {
-      providerId: provider.id,
-      targetHost: getProviderTargetHost(provider.apiUrl),
-      path,
-      failureFamily,
-      errorName,
-      errorMessage,
-      ...(causeCode && { causeCode }),
-      ...(causeName && { causeName }),
-      ...(causeMessage && { causeMessage }),
-    };
+    try {
+      const cause = error instanceof Error ? error.cause : undefined;
+      const errorName = getErrorName(error);
+      const errorMessage = getErrorMessage(error);
+      const causeCode = getCauseCode(cause);
+      const causeName = getCauseName(cause);
+      const causeMessage = getCauseMessage(cause);
+      const failureFamily = classifyUpstreamFetchFailure({ errorName, causeCode, causeName });
+      const failureMetadata = {
+        providerId: provider.id,
+        targetHost: getProviderTargetHost(provider.apiUrl),
+        path,
+        failureFamily,
+        errorName,
+        errorMessage,
+        ...(causeCode && { causeCode }),
+        ...(causeName && { causeName }),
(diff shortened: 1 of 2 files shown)


apps/web/src/lib/ai-gateway/providers/upstream-request.ts
```diff
@@ -44,14 +152,48 @@ export async function upstreamRequest({
...
+    const causeName = getCauseName(cause);
+    const causeMessage = getCauseMessage(cause);
+    const failureFamily = classifyUpstreamFetchFailure({ errorName, causeCode, causeName });
+    const failureMetadata = {
+      providerId: provider.id,
+      targetHost: getProviderTargetHost(provider.apiUrl),
+      path,
+      failureFamily,
+      errorName,
+      errorMessage,
```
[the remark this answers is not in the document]
jeanduplessis: Fixed. Upstream failure messages now redact embedded HTTP(S) URLs before adding them to structured metadata, and Sentry receives a sanitized logged error instead of the raw thrown fetch error so URL-bearing exception text does not bypass that redaction. Added regression coverage with an ECONNRESET-shaped error whose top-level and cause messages contain secret-bearing URLs.

apps/web/src/lib/ai-gateway/providers/upstream-request.ts
```diff
@@ -44,14 +168,50 @@ export async function upstreamRequest({
...
+        tags: {
+          source: 'ai-gateway-upstream-fetch',
+          provider: provider.id,
+          failure_family: failureFamily,
+        },
+        extra: failureMetadata,
+      });
+    }
+
+    throw error;
```
jrf0110: While it _probably won't happen_, I think it's worth wrapping this new block in a catch so that we preserve the original error if any of these error enrichment functions end up throwing an error.

```suggestion
    try {
    const cause = error instanceof Error ? error.cause : undefined;
    const errorName = getErrorName(error);
    const errorMessage = getErrorMessage(error);
    const causeCode = getCauseCode(cause);
    const causeName = getCauseName(cause);
    const causeMessage = getCauseMessage(cause);
    const failureFamily = classifyUpstreamFetchFailure({ errorName, causeCode, causeName });
    const failureMetadata = {
      providerId: provider.id,
      targetHost: getProviderTargetHost(provider.apiUrl),
      path,
      failureFamily,
      errorName,
      errorMessage,
      ...(causeCode && { causeCode }),
      ...(causeName && { causeName }),
      ...(causeMessage && { causeMessage }),
    };

    if (!(failureFamily === 'abort' && signal?.aborted)) {
      errorExceptInTest('AI gateway upstream fetch failed', failureMetadata);
      captureException(createLoggedFetchFailure(errorName, errorMessage), {
        level: 'error',
        tags: {
          source: 'ai-gateway-upstream-fetch',
          provider: provider.id,
          failure_family: failureFamily,
        },
        extra: failureMetadata,
      });
    }
    }

    throw error;
```

jrf0110 (approved):

apps/web/src/lib/ai-gateway/providers/upstream-request.ts
jeanduplessis: Fixed. The upstream fetch catch path now guards diagnostic enrichment/reporting separately, so any secondary failure there cannot replace the original fetch rejection. Added regression coverage with a throwing cause getter to verify callers still receive the original transport error.

PR #3289, opened 2026-05-17
feat(abuse): send new events and identity metadata to abuse service (#3288)

### LINKED ISSUE #3288: Send new events and identity metadata to the abuse service
## Overview

The abuse service (`abuse.kiloapps.io`) recently added a generic event-ingest endpoint and expanded the existing auth-event payload. Cloud already calls `/api/classify`, `/api/auth-event`, and `/api/usage/cost` via `apps/web/src/lib/ai-gateway/abuse-service.ts`. This issue tracks wiring up the rest of what the abuse service expects.

There are two pieces:

1. A new `POST /api/events` envelope endpoint that takes a batch of typed cloud events.
2. A larger `POST /api/auth-event` payload with additional optional identity-metadata fields.

Everything is fire-and-forget from cloud's perspective. Failures are logged but never block the user-facing flow, same convention as the existing abuse-service helpers.

---

## 1. New endpoint: `POST /api/events`

Add a new helper alongside `classifyRequest` / `reportCost` / `reportAuthEvent` in `apps/web/src/lib/ai-gateway/abuse-service.ts`.

**Envelope:**

```ts
type EventsBatchPayload = {
  events: CloudEvent[];
};

type CloudEvent =
  | { type: 'user.blocked';                                 occurred_at?: number; data: UserEventData }
  | { type: 'user.unblocked';                               occurred_at?: number; data: UserEventData }
  | { type: 'user.deleted';                                 occurred_at?: number; data: { kilo_user_id: string } }
  | { type: 'user.email_changed';                           occurred_at?: number; data: UserEmailChangedData }
  | { type: 'org.member_added';                             occurred_at?: number; data: OrgEventData }
  | { type: 'org.member_removed';                           occurred_at?: number; data: OrgEventData }
  | { type: 'org.created';                                  occurred_at?: number; data: OrgEventData }
  | { type: 'org.deleted';                                  occurred_at?: number; data: OrgEventData }
  | { type: 'billing.credit_purchased';                     occurred_at?: number; data: BillingCreditPurchasedData }
  | { type: 'billing.kilo_pass_changed';                    occurred_at?: number; data: BillingKiloPassChangedData }
  | { type: 'stripe.payment_method.attached';               occurred_at?: number; data: StripeEventData }
  | { type: 'stripe.payment_method.detached';               occurred_at?: number; data: StripeEventData }
  | { type: 'stripe.charge.dispute.created';                occurred_at?: number; data: StripeEventData }
  | { type: 'stripe.charge.dispute.funds_withdrawn';        occurred_at?: number; data: StripeEventData }
  | { type: 'stripe.radar.early_fraud_warning.created';     occurred_at?: number; data: StripeEventData }
  | { type: 'stripe.charge.failed';                         occurred_at?: number; data: StripeEventData }
  | { type: 'stripe.payment_intent.succeeded';              occurred_at?: number; data: StripeEventData };

type UserEventData = {
  kilo_user_id: string;
  reason?: string | null;
  actor_email?: string | null;  // who performed the action, if applicable
};

type UserEmailChangedData = {
  kilo_user_id: string;
  previous_email?: string | null;
  email: string;  // the new email
};

type OrgEventData = {
  kilo_user_id: string;
  organization_id: string;
  role?: string | null;
  plan?: string | null;
  has_sso?: boolean;
  in_free_trial?: boolean;
};

type BillingCreditPurchasedData = {
  kilo_user_id: string;
  microdollars_acquired: number;
  total_microdollars_acquired?: number;  // running lifetime total after this purchase
};

type BillingKiloPassChangedData = {
  kilo_user_id: string;
  tier?: string | null;     // e.g. 'free', 'pro'
  status?: string | null;   // e.g. 'active', 'cancelled'
  streak_months?: number;
};

// Mirror the relevant subset of the Stripe webhook event object
type StripeEventData = {
  id?: string;
  type?: string;
  customer?: string | { id: string } | null;
  data?: { object?: { amount?: number; customer?: ...; decline_code?: string; [k: string]: unknown } };
  decline_code?: string;
  amount?: number;
  // Passthrough — additional fields are accepted
};
```

`occurred_at` is an optional epoch-ms timestamp; if omitted the abuse service stamps it on receipt.

The endpoint accepts a batch (`{ events: [...] }`). Cloud can either send one event per call or batch up to a few hundred per call — both work.

### Auth & headers

Same as the existing endpoints: send the `CF-Access-Client-Id` and `CF-Access-Client-Secret` headers from `ABUSE_SERVICE_CF_ACCESS_CLIENT_ID` / `ABUSE_SERVICE_CF_ACCESS_CLIENT_SECRET`. Reuse `fetchAbuseService(...)`.

### Suggested API

```ts
export async function reportEvents(payload: EventsBatchPayload): Promise<void> {
  await fetchAbuseService('/api/events', payload, 'events');
}
```

### Where to emit each event

Best-effort mapping; verify each call site matches the cloud codebase's actual flow:

| Event | Suggested call site |
|---|---|
| `user.blocked` / `user.unblocked` | Wherever an admin / automated flow sets the user's blocked state |
| `user.deleted` | User deletion flow |
| `user.email_changed` | After an email change is committed |
| `org.member_added` / `org.member_removed` | Membership mutations |
| `org.created` / `org.deleted` | Org lifecycle mutations |
| `billing.credit_purchased` | Successful credit purchase (post-Stripe charge or post-grant) |
| `billing.kilo_pass_changed` | Tier / status / streak transitions |
| `stripe.*` | The existing Stripe webhook handlers — emit one event per relevant Stripe webhook type, reusing the original event JSON for `data` |

Cloud is the source of truth for which user is which. Each event payload includes `kilo_user_id` so the abuse service can route it to the right per-identity record.

### Forward-compatibility

The abuse service accepts unknown `type` values too — it logs them once and moves on. Cloud can ship new event types ahead of the abuse service if needed; nothing breaks.

---

## 2. Expanded `POST /api/auth-event` payload

The existing `AuthEventPayload` type in `apps/web/src/lib/ai-gateway/abuse-service.ts` is a strict subset of what the abuse service now accepts. Adding the new fields lets the abuse service do its job without us having to backfill identity metadata via separate events.

All new fields are **optional**; existing payloads continue to work. Send whatever you have at signup/signin time.

```ts
export type AuthEventPayload = {
  // --- existing fields (unchanged) ---
  kilo_user_id: string;
  event_type: 'signup' | 'signin';
  email: string;
  account_created_at?: string;          // ISO 8601, was required, now optional in schema
  ip_address?: string | null;
  geo_city?: string | null;
  geo_country?: string | null;
  ja4_digest?: string | null;
  user_agent?: string | null;
  auth_method?: AuthProviderId | null;
  stytch_session_id?: string | null;

  // --- NEW: user.* metadata ---
  hosted_domain?: string | null;          // Google-Workspace-style hosted domain
  signup_ip?: string | null;              // Stable snapshot, not the current request IP
  signup_ja4_digest?: string | null;
  signup_geo_country?: string | null;
  customer_source?: string | null;       // Product/campaign that brought the user in
  is_bot?: boolean | null;
  is_admin?: boolean | null;
  is_blocked?: boolean | null;
  completed_welcome_form?: boolean | null;
  has_linkedin_url?: boolean | null;
  has_github_url?: boolean | null;
  has_discord_verified?: boolean | null;
  cohorts?: string[] | null;             // Cohort membership names

  // --- NEW: auth.* metadata ---
  has_validation_stytch?: boolean | null;
  has_validation_novel_card_with_hold?: boolean | null;
  stytch_verdict_action?: string | null;
  stytch_is_authentic_device?: boolean | null;
  stytch_device_type?: string | null;
  stytch_hardware_fingerprint?: string | null;
  auth_providers?: string[] | null;      // Distinct providers this user has linked

  // --- NEW: org.* metadata ---
  org_memberships?: Array<{
    organization_id: string;
    role?: string | null;
    plan?: string | null;
    has_sso?: boolean | null;
    in_free_trial?: boolean | null;
  }> | null;
};
```

Call sites: `reportAuthEvent` is already invoked from the signup / signin flows (currently in `apps/web/src/lib/user.ts` per-grep). The change is purely additive — populate the new fields with whatever's available in scope at call time and pass them through.

For `signin` events specifically: it's worth re-sending the metadata fields that may have changed since signup (`is_blocked`, `is_admin`, `cohorts`, `org_memberships`, etc.) so the abuse service stays current without needing a separate sync path.

---

## Implementation notes

- **All calls remain fire-and-forget.** The `fetchAbuseService(...)` helper already swallows errors and returns `null`. New helpers should follow the same pattern.
- **`ABUSE_SERVICE_URL` may be unset in non-production.** The existing helpers no-op when it isn't configured; new helpers should too.
- **Batching is optional.** A single-event batch is fine; the abuse service handles either shape. Don't build a queue/buffer unless it falls naturally out of an existing pattern.
- **`occurred_at`** should be the epoch-ms timestamp of the underlying business event when known (e.g. `Stripe webhook.created * 1000`), otherwise omit and let the abuse service stamp it.
- **Stripe events** can pass the relevant slice of the Stripe webhook event object straight through as `data` — the abuse-service schema is intentionally permissive (`passthrough`) on Stripe-shaped data.

---

## Acceptance criteria

- [ ] New `reportEvents` (or equivalent) helper added next to the existing `classifyRequest` / `reportCost` / `reportAuthEvent` helpers, hitting `POST /api/events`.
- [ ] Each of the 17 event types is emitted from at least the most obvious call site in the cloud codebase. Test coverage matches the existing abuse-service helpers' coverage style.
- [ ] `AuthEventPayload` type extended with the new optional fields; the existing `signup` and `signin` emit paths populate every field they have in-scope.
- [ ] No regressions to the three existing endpoints (`/api/classify`, `/api/usage/cost`, `/api/auth-event`).
- [ ] All new code paths swallow errors and never block user-facing flows, matching the existing pattern.

---

## References

- Existing client: `apps/web/src/lib/ai-gateway/abuse-service.ts`
- Existing call sites: `apps/web/src/lib/user.ts` (auth events), `apps/web/src/lib/ai-gateway/processUsage.ts` (cost), `apps/web/src/app/api/openrouter/[...path]/route.ts` (classify)

jrf0110:
## Summary

- Adds `reportEvents` calls for three previously missing production call sites: Stytch autoban (`user.blocked`), admin cancel-and-refund (`user.blocked`), and org admin credit grant (`billing.credit_purchased`).
- Expands `fireAuthEvent` in `user.ts` to populate the new `AuthEventPayload` fields (`auth_providers`, `org_memberships`) via parallel DB queries at auth time, keeping the call fire-and-forget.
- The `reportEvents` helper, all CloudEvent types, the expanded `AuthEventPayload` type, and the majority of the 17 event emit sites were already implemented on this branch prior to this commit. This completes coverage for the remaining gaps identified during codebase audit.

Closes #3288

## Verification

- Existing stytch.test.ts, cancel-and-refund.test.ts, and user.test.ts exercise the modified code paths; `ABUSE_SERVICE_URL` is unset in CI so all new `reportEvents`/`reportAuthEvent` calls no-op safely.
- The Stytch autoban path now uses `.returning()` to confirm the row was actually updated before firing the event, matching the conditional guard that was already present.

## Visual Changes

N/A

## Reviewer Notes

- `fireAuthEvent` is now async; all three call sites already used `void` so the change is non-blocking.
- The org admin credit grant emits `billing.credit_purchased` only for positive grant amounts and only when `created_by_kilo_user_id` is set (orgs created before that column existed have it null).
- `user.email_changed` has no call site because the email field is immutable in normal flows (tied to OAuth provider); no change needed.
- One-off backfill scripts that also set `blocked_at` were intentionally skipped per the bead guidance (low priority for one-off migration scripts).

### DIFF
--- a/apps/web/src/lib/abuse/bulkBlock.ts
+++ b/apps/web/src/lib/abuse/bulkBlock.ts
@@ -64,7 +64,7 @@ export async function bulkBlockUsers(
     void reportEvents({
       events: updated.map(u => ({
         type: 'user.blocked' as const,
-        data: { kilo_user_id: u.id, reason, actor_email: blockedByKiloUserId ?? null },
+        data: { kilo_user_id: u.id, reason, actor_email: null },
       })),
     });
   }
--- a/apps/web/src/lib/abuse/bulkBlock.ts
+++ b/apps/web/src/lib/abuse/bulkBlock.ts
@@ -2,6 +2,7 @@ import { db } from '@/lib/drizzle';
 import { kilocode_users } from '@kilocode/db/schema';
 import { and, eq, inArray, isNull, or, sql } from 'drizzle-orm';
 import { successResult, type CustomResult } from '@/lib/maybe-result';
+import { reportEvents } from '@/lib/ai-gateway/abuse-service';
 
 export type BulkBlockResponse = CustomResult<
   { updatedCount: number },
@@ -59,6 +60,15 @@ export async function bulkBlockUsers(
     .where(inArray(kilocode_users.id, valid))
     .returning({ id: kilocode_users.id });
 
+  if (updated.length > 0) {
+    void reportEvents({
+      events: updated.map(u => ({
+        type: 'user.blocked' as const,
+        data: { kilo_user_id: u.id, reason, actor_email: blockedByKiloUserId ?? null },
+      })),
+    });
+  }
+
   return successResult({ updatedCount: updated.length });
 }
 
@@ -86,6 +96,15 @@ export async function unblockBulkBlockedUsers(
       )
     )
     .returning({ id: kilocode_users.id });
+
+  if (rows.length > 0) {
+    void reportEvents({
+      events: rows.map(u => ({
+        type: 'user.unblocked' as const,
+        data: { kilo_user_id: u.id },
+      })),
+    });
+  }
 
   return { updatedCount: rows.length };
 }
--- a/apps/web/src/lib/credits.ts
+++ b/apps/web/src/lib/credits.ts
@@ -139,7 +139,11 @@ export async function processTopUp(
   };
 
   if (dbOrTx) {
-    after(emitCreditPurchasedEvent);
+    if (IS_IN_AUTOMATED_TEST) {
+      emitCreditPurchasedEvent();
+    } else {
+      after(emitCreditPurchasedEvent);
+    }
   } else {
     emitCreditPurchasedEvent();
   }
--- a/apps/web/src/lib/credits.ts
+++ b/apps/web/src/lib/credits.ts
@@ -124,7 +124,7 @@ export async function processTopUp(
     return false;
   }
 
-  if (!dbOrTx) {
+  const emitCreditPurchasedEvent = () => {
     void reportEvents({
       events: [
         {
@@ -136,6 +136,12 @@ export async function processTopUp(
         },
       ],
     });
+  };
+
+  if (dbOrTx) {
+    after(emitCreditPurchasedEvent);
+  } else {
+    emitCreditPurchasedEvent();
   }
 
   if (skipPostTopUpFreeStuff) return true;
--- a/apps/web/src/lib/credits.ts
+++ b/apps/web/src/lib/credits.ts
@@ -124,17 +124,19 @@ export async function processTopUp(
     return false;
   }
 
-  void reportEvents({
-    events: [
-      {
-        type: 'billing.credit_purchased',
-        data: {
-          kilo_user_id: user.id,
-          microdollars_acquired: creditAmountInMicrodollars,
+  if (!dbOrTx) {
+    void reportEvents({
+      events: [
+        {
+          type: 'billing.credit_purchased',
+          data: {
+            kilo_user_id: user.id,
+            microdollars_acquired: creditAmountInMicrodollars,
+          },
         },
-      },
-    ],
-  });
+      ],
+    });
+  }
 
   if (skipPostTopUpFreeStuff) return true;
 
--- a/apps/web/src/lib/credits.ts
+++ b/apps/web/src/lib/credits.ts
@@ -12,6 +12,7 @@ import { grantCreditForCategory } from '@/lib/promotionalCredits';
 import { IS_IN_AUTOMATED_TEST } from '@/lib/config.server';
 import { sendCreditsTopUpEmail } from '@/lib/email';
 import { client as stripeClient } from '@/lib/stripe-client';
+import { reportEvents } from '@/lib/ai-gateway/abuse-service';
 
 export type StripeConfig = { type: 'stripe'; stripe_payment_id: string };
 
@@ -122,6 +123,18 @@ export async function processTopUp(
     }
     return false;
   }
+
+  void reportEvents({
+    events: [
+      {
+        type: 'billing.credit_purchased',
+        data: {
+          kilo_user_id: user.id,
+          microdollars_acquired: creditAmountInMicrodollars,
+        },
+      },
+    ],
+  });
 
   if (skipPostTopUpFreeStuff) return true;
 
--- a/apps/web/src/lib/kilo-pass/cancel-and-refund.ts
+++ b/apps/web/src/lib/kilo-pass/cancel-and-refund.ts
@@ -14,6 +14,7 @@ import { getKiloPassStateForUser } from '@/lib/kilo-pass/state';
 import { releaseScheduledChangeForSubscription } from '@/lib/kilo-pass/scheduled-change-release';
 import { fromMicrodollars } from '@/lib/utils';
 import { KiloPassPaymentProvider } from '@/lib/kilo-pass/enums';
+import { reportEvents } from '@/lib/ai-gateway/abuse-service';
 
 type Db = typeof defaultDb;
 
@@ -233,6 +234,21 @@ export async function cancelAndRefundKiloPassForUser({
 
     return balanceReset;
   });
+
+  if (!user.blocked_reason) {
+    void reportEvents({
+      events: [
+        {
+          type: 'user.blocked',
+          data: {
+            kilo_user_id: userId,
+            reason,
+            actor_email: adminKiloUserId,
+          },
+        },
+      ],
+    });
+  }
 
   return {
     status: 'cancelled_and_refunded',
--- a/apps/web/src/lib/kilo-pass/cancel-and-refund.ts
+++ b/apps/web/src/lib/kilo-pass/cancel-and-refund.ts
@@ -243,7 +243,7 @@ export async function cancelAndRefundKiloPassForUser({
           data: {
             kilo_user_id: userId,
             reason,
-            actor_email: adminKiloUserId,
+            actor_email: null,
           },
         },
       ],
--- a/apps/web/src/lib/kilo-pass/stripe-handlers-subscription-events.ts
+++ b/apps/web/src/lib/kilo-pass/stripe-handlers-subscription-events.ts
@@ -4,6 +4,7 @@ import { kilo_pass_subscriptions } from '@kilocode/db/schema';
 
 import { db } from '@/lib/drizzle';
 import { eq } from 'drizzle-orm';
+import { reportEvents } from '@/lib/ai-gateway/abuse-service';
 
 import { KiloPassError } from '@/lib/kilo-pass/errors';
 import { appendKiloPassAuditLog } from '@/lib/kilo-pass/issuance';
@@ -38,6 +39,9 @@ export async function handleKiloPassSubscriptionEvent(params: {
   }
 
   const { kiloUserId, tier, cadence } = metadata;
+
+  let finalStatus: string | undefined;
+  let finalStreakMonths: number | undefined;
 
   await db.transaction(async tx => {
     await appendKiloPassAuditLog(tx, {
@@ -95,7 +99,13 @@ export async function handleKiloPassSubscriptionEvent(params: {
           provider_subscription_id: subscription.id,
         },
       })
-      .returning({ id: kilo_pass_subscriptions.id });
+      .returning({
+        id: kilo_pass_subscriptions.id,
+        current_streak_months: kilo_pass_subscriptions.current_streak_months,
+      });
+
+    finalStatus = stripeStatus;
+    finalStreakMonths = upserted[0]?.current_streak_months ?? 0;
 
     const kiloPassSubscriptionId = upserted[0]?.id;
 
@@ -122,5 +132,20 @@ export async function handleKiloPassSubscriptionEvent(params: {
         });
       }
     }
+  });
+
+  void reportEvents({
+    events: [
+      {
+        type: 'billing.kilo_pass_changed',
+        occurred_at: subscription.created * 1000,
+        data: {
+          kilo_user_id: kiloUserId,
+          tier,
+          status: finalStatus ?? null,
+          streak_months: finalStreakMonths,
+        },
+      },
+    ],
   });
 }
--- a/apps/web/src/lib/kilo-pass/stripe-handlers-subscription-events.ts
+++ b/apps/web/src/lib/kilo-pass/stripe-handlers-subscription-events.ts
@@ -138,7 +138,6 @@ export async function handleKiloPassSubscriptionEvent(params: {
     events: [
       {
         type: 'billing.kilo_pass_changed',
-        occurred_at: subscription.created * 1000,
         data: {
           kilo_user_id: kiloUserId,
           tier,
--- a/apps/web/src/lib/organizations/organizations.ts
+++ b/apps/web/src/lib/organizations/organizations.ts
@@ -26,6 +26,7 @@ import { logExceptInTest } from '@/lib/utils.server';
 import { APP_URL } from '@/lib/constants';
 import { createAuditLog } from '@/lib/organizations/organization-audit-logs';
 import { failureResult, successResult } from '@/lib/maybe-result';
+import { reportEvents } from '@/lib/ai-gateway/abuse-service';
 
 export async function getOrganizationById(
   id: Organization['id'],
@@ -185,11 +186,11 @@ export async function createOrganization(
   company_domain?: string,
   plan?: 'teams' | 'enterprise'
 ): Promise<Organization> {
-  return await db.transaction(async tx => {
+  const organization = await db.transaction(async tx => {
     const now = new Date();
     const trialEndDate = new Date(now.getTime() + TRIAL_DURATION_DAYS * 24 * 60 * 60 * 1000);
 
-    const [organization] = await tx
+    const [org] = await tx
       .insert(organizations)
       .values({
         name,
@@ -209,15 +210,34 @@ export async function createOrganization(
 
     if (!userId || !addUserAsOwner) {
       // If no user ID is provided or addUserAsOwner is false, return the organization without adding a member
-      return organization;
+      return org;
     }
     await tx.insert(organization_memberships).values({
-      organization_id: organization.id,
+      organization_id: org.id,
       kilo_user_id: userId,
       role: 'owner',
     });
-    return organization;
+    return org;
   });
+
+  if (userId) {
+    void reportEvents({
+      events: [
+        {
+          type: 'org.created',
+          data: {
+            kilo_user_id: userId,
+            organization_id: organization.id,
+            role: 'owner',
+            plan: organization.plan ?? null,
+            in_free_trial: organization.free_trial_end_at != null,
+          },
+        },
+      ],
+    });
+  }
+
+  return organization;
 }
 
 export async function addUserToOrganization(
@@ -235,7 +255,18 @@ export async function addUserToOrganization(
     })
     .onConflictDoNothing();
 
-  return (result.rowCount ?? 0) > 0;
+  const added = (result.rowCount ?? 0) > 0;
+  if (added) {
+    void reportEvents({
+      events: [
+        {
+          type: 'org.member_added',
+          data: { kilo_user_id: userId, organization_id: organizationId, role },
+        },
+      ],
+    });
+  }
+  return added;
 }
 
 export async function removeUserFromOrganization(
@@ -285,6 +316,19 @@ export async function removeUserFromOrganization(
             previous_role: membership.role,
           },
         });
+
+      void reportEvents({
+        events: [
+          {
+            type: 'org.member_removed',
+            data: {
+              kilo_user_id: userId,
+              organization_id: organizationId,
+              role: membership.role,
+            },
+          },
+        ],
+      });
     }
 
     return result;
@@ -496,7 +540,7 @@ export async function acceptOrganizationInvite(
   inviteToken: string
 ): Promise<AcceptInviteResult> {
   try {
-    return await db.transaction(async tx => {
+    const result = await db.transaction(async tx => {
       // Find and lock the invitation to prevent race conditions
       const [invitation] = await tx
         .select()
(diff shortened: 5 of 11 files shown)


jrf0110:
@kilocode-bot review please?

apps/web/src/lib/user.ts
```diff
@@ -752,6 +812,8 @@ export async function softDeleteUser(userId: string) {
   const originalEmail = user.google_user_email;
   const originalAppStoreAccountToken = user.app_store_account_token;
 
+  void reportEvents({ events: [{ type: 'user.deleted', data: { kilo_user_id: userId } }] });
```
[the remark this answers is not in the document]
jrf0110: Fixed: moved the `reportEvents` call for `user.deleted` to *after* the `db.transaction(...)` completes. This ensures the event is only fired once the user has actually been deleted — precondition failures (active subscription, live KiloClaw instance) will no longer result in phantom `user.deleted` events.

apps/web/src/lib/kilo-pass/cancel-and-refund.ts
```diff
@@ -234,6 +235,21 @@ export async function cancelAndRefundKiloPassForUser({
...
 
+  if (!user.blocked_reason) {
+    void reportEvents({
+      events: [
+        {
+          type: 'user.blocked',
+          data: {
+            kilo_user_id: userId,
+            reason,
+            actor_email: adminKiloUserId,
```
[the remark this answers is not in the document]
jrf0110: Fixed: changed `actor_email: adminKiloUserId` to `actor_email: null`. The function only receives `adminKiloUserId` (a UUID), not the admin's email, so passing the UUID was incorrect. Passing `null` is the safe stopgap as suggested.

apps/web/src/lib/abuse/bulkBlock.ts
```diff
@@ -59,6 +60,15 @@ export async function bulkBlockUsers(
     .where(inArray(kilocode_users.id, valid))
     .returning({ id: kilocode_users.id });
 
+  if (updated.length > 0) {
+    void reportEvents({
+      events: updated.map(u => ({
+        type: 'user.blocked' as const,
+        data: { kilo_user_id: u.id, reason, actor_email: blockedByKiloUserId ?? null },
```
jeanduplessis: [WARNING] `actor_email` receives `blockedByKiloUserId`, but this bulk-block route passes `user.id` into that argument, so emitted events carry admin UUIDs rather than actor emails.

Trace: `POST /app/admin/api/abuse/bulk-block` calls `bulkBlockUsers(..., user.id)`, then this changed payload forwards that ID as `actor_email`. Same PR already fixed equivalent UUID-as-email behavior in cancel-and-refund by using `null`.

Impact: bulk `user.blocked` events store malformed actor attribution, which can break downstream email-based grouping, audit interpretation, or abuse heuristics.

apps/web/src/lib/user.ts
```diff
@@ -305,13 +306,50 @@ export async function findUserByEmail(email: string): Promise<User | undefined>
...
+    | 'has_validation_stytch'
+    | 'has_validation_novel_card_with_hold'
+  >,
   eventType: 'signup' | 'signin',
   provider: AuthProviderId,
   requestHeaders?: Headers
 ) {
   if (!requestHeaders) return;
+
+  const [authProviderRows, membershipRows] = await Promise.all([
```
jeanduplessis: [WARNING] `fireAuthEvent` is now `async`, but these new DB reads run before `reportAuthEvent` reaches its fail-open `fetchAbuseService` wrapper. Callers intentionally discard the promise with `void fireAuthEvent(...)`.

Trace: either query in this `Promise.all` can reject, causing `fireAuthEvent` itself to reject before `reportAuthEvent` is invoked. Existing caller behavior does not await or catch that rejection.

Impact: transient auth-path DB failures can silently drop auth telemetry and surface detached promise rejections, which contradicts the PR's stated non-blocking/fire-and-forget behavior.

apps/web/src/lib/organizations/organizations.ts
```diff
@@ -597,6 +641,23 @@ export async function acceptOrganizationInvite(
         role: invitation.role,
       });
     });
+
+    if (result.success) {
```
jeanduplessis: [WARNING] This emits `org.member_added` for every successful invite acceptance, including the existing-member branch above that only marks the invite accepted and does not insert a membership.

Trace: duplicate/retry acceptance returns `successResult(...)` when membership already exists, so `result.success` remains true and reaches this new event block.

Impact: duplicate invite flows fabricate membership-add transitions in the abuse event stream even though organization membership did not change.

apps/web/src/lib/kilo-pass/stripe-handlers-subscription-events.ts
```diff
@@ -123,4 +133,19 @@ export async function handleKiloPassSubscriptionEvent(params: {
       }
     }
   });
+
+  void reportEvents({
+    events: [
+      {
+        type: 'billing.kilo_pass_changed',
+        occurred_at: subscription.created * 1000,
```
jeanduplessis: [WARNING] `occurred_at` uses `subscription.created`, which is original Stripe subscription creation time, not timestamp for later webhook-driven status/tier/streak changes handled here.

Trace: this handler processes updates and ended transitions long after creation; linked issue guidance says `occurred_at` should represent underlying business event time when known, otherwise be omitted.

Impact: downstream consumers see later Kilo Pass changes backdated to signup time, which can corrupt ordering, freshness, and state-transition analysis.

apps/web/src/routers/admin-router.ts
```diff
@@ -508,6 +511,19 @@ export const adminRouter = createTRPCRouter({
           .set(blockMetadata)
           .where(eq(kilocode_users.id, input.userId));
 
+        void reportEvents({
```
jeanduplessis: [WARNING] Event emission is unconditional after the update statement, but update result is not inspected to confirm a real block/unblock transition occurred.

Trace: retrying the same admin action, submitting stale UI state, or targeting a missing user still reaches `reportEvents(...)`. Other changed block paths gate emission on returned updated rows.

Impact: phantom `user.blocked` / `user.unblocked` events can enter downstream state history without matching database state transitions.

apps/web/src/lib/stripe.ts
```diff
@@ -791,6 +812,17 @@ export async function processStripePaymentEventHook(event: Stripe.Event) {
 
     case 'charge.dispute.created': {
       const dispute = event.data.object;
+
+      void reportEvents({
+        events: [
+          {
+            type: 'stripe.charge.dispute.created',
+            occurred_at: event.created * 1000,
+            data: { id: event.id, type: event.type, customer: dispute.charge },
```
jeanduplessis: [WARNING] This assigns `dispute.charge` into field named `customer`. The surrounding code immediately treats same value as `chargeId`, confirming it is charge identity, not customer identity.

Trace: other new Stripe events pass actual `...customer` values into `customer`, and issue schema describes this field as customer-shaped routing data.

Impact: dispute events can be mis-correlated or fail customer-based routing in abuse processing because `customer` contains `ch_...` / charge-object data.

jeanduplessis (changes requested):
Found six concrete event-contract issues in this PR:

- bulk block reports admin UUIDs as `actor_email`
- async auth enrichment can reject outside existing fail-open handling
- duplicate invite acceptance emits false `org.member_added`
- Kilo Pass changes use subscription creation time as event time
- admin block/unblock emits events without confirming a real mutation
- dispute-created Stripe payload stores charge identity under `customer`

These affect event correctness, retry behavior, and downstream abuse-service attribution/state ordering.

apps/web/src/lib/abuse/bulkBlock.ts
jrf0110: Fixed: changed `actor_email: blockedByKiloUserId` to `actor_email: null`. The function only receives a user UUID for the admin actor, not an email, so passing it as `actor_email` was incorrect. Using `null` matches the same fix already applied in the cancel-and-refund path.

apps/web/src/lib/user.ts
jrf0110: Fixed: wrapped the `Promise.all` in `.catch(() => null)` so any DB query failure returns `null` instead of rejecting `fireAuthEvent`. The `authProviderRows` and `membershipRows` fall back to empty arrays on failure, so `reportAuthEvent` is still called with the core auth data, preserving the existing fail-open behavior.

apps/web/src/lib/organizations/organizations.ts
jrf0110: Fixed: added a `membershipInserted` boolean to the transaction's success result. The existing-member branch returns `membershipInserted: false`; the new-membership branch returns `membershipInserted: true`. The `org.member_added` event is now only emitted when `result.membershipInserted` is true, preventing false events on duplicate/retry invite acceptances.

apps/web/src/lib/kilo-pass/stripe-handlers-subscription-events.ts
jrf0110: Fixed: removed the `occurred_at` field entirely. This handler processes status transitions (updates, ended states) that can occur long after subscription creation, so using `subscription.created` was incorrect. Per the guidance, when the underlying business event time is not known, `occurred_at` should be omitted.

apps/web/src/routers/admin-router.ts
jrf0110: Fixed: changed the update to use `.returning({ id: kilocode_users.id })` and wrapped the `reportEvents` call in `if (updated.length > 0)`. This matches the pattern already used in `bulkBlockUsers` and other block paths — events are only emitted when the DB confirms a real mutation occurred.

apps/web/src/lib/stripe.ts
jrf0110: Fixed: changed `customer: dispute.charge` to `customer: dispute.customer`. The `dispute.charge` field is the charge identifier (already extracted below as `chargeId`); `dispute.customer` is the actual Stripe customer ID, which matches the routing semantics of the `customer` field used in all other Stripe event payloads in this file.

jrf0110: The previous fix used `dispute.customer`, but the Stripe `Dispute` TypeScript type does not have a `customer` property, causing CI type errors. Fixed in commit bb797c30c by removing the `customer` field from the event data entirely (consistent with how `charge.dispute.funds_withdrawn` is handled). Getting the customer ID would require expanding the charge object via an extra API call.

apps/web/src/lib/credits.ts
```diff
@@ -123,6 +124,18 @@ export async function processTopUp(
     return false;
   }
 
+  void reportEvents({
```
jeanduplessis: [WARNING] `billing.credit_purchased` can be emitted before caller-owned transactions actually commit.

Trace:
1. `processTopUp()` now fires `reportEvents(...)` as soon as its insert/update helper returns success.
2. This helper also accepts `dbOrTx`, and existing callers use it inside larger transactions, then continue with additional writes before commit.
3. If those later writes or final commit fail, this fire-and-forget request may already reach abuse service even though credited balance change rolls back.

Impact: downstream abuse state can record credit purchases that never committed. This affects transactional callers such as Kilo Pass issuance and KiloClaw settlement paths using `dbOrTx`.

apps/web/src/lib/user.ts
```diff
@@ -305,13 +306,54 @@ export async function findUserByEmail(email: string): Promise<User | undefined>
...
+      .select({
+        organization_id: organization_memberships.organization_id,
+        role: organization_memberships.role,
+        plan: organizations.plan,
+        sso_domain: organizations.sso_domain,
+        free_trial_end_at: organizations.free_trial_end_at,
+      })
+      .from(organization_memberships)
+      .innerJoin(organizations, eq(organization_memberships.organization_id, organizations.id))
+      .where(eq(organization_memberships.kilo_user_id, user.id)),
```
jeanduplessis: [WARNING] Auth enrichment includes memberships from soft-deleted organizations.

Trace:
1. This new membership query joins `organization_memberships` to `organizations`.
2. Organization deletion is soft-delete, and existing membership readers exclude deleted orgs with `isNull(organizations.deleted_at)`.
3. This query does not apply that filter, so later auth events can report org memberships app logic no longer treats as active.

Impact: abuse service receives stale `org_memberships`, plan, SSO, and trial metadata after an organization has been deleted.

apps/web/src/routers/admin-router.ts
```diff
@@ -503,10 +506,26 @@ export const adminRouter = createTRPCRouter({
...
 
-        await db
+        const updated = await db
           .update(kilocode_users)
           .set(blockMetadata)
-          .where(eq(kilocode_users.id, input.userId));
+          .where(eq(kilocode_users.id, input.userId))
+          .returning({ id: kilocode_users.id });
+
+        if (updated.length > 0) {
```
jeanduplessis: [WARNING] `.returning()` proves row existence, not a real blocked-state transition.

Trace:
1. Event emission now runs whenever the update returns a row.
2. PostgreSQL `UPDATE ... RETURNING` still returns an existing row for repeated logical state requests, such as unblocking an already-unblocked user or retrying the same block action.
3. Other changed block paths gate on an actual transition; this path still emits another `user.blocked` / `user.unblocked` event for retries or stale submissions.

Impact: abuse history can accumulate duplicate or phantom block-state transitions without a corresponding logical state change.

jeanduplessis (approved):
Reviewed final combined diff and surrounding call paths. Event coverage mostly tracks intended flows, but three changed paths still need tightening: transactional credit events can escape before outer commit, auth enrichment can include soft-deleted org memberships, and admin block/unblock reporting still emits on idempotent retries rather than actual transitions.

apps/web/src/lib/credits.ts
jrf0110: Fixed: gated `billing.credit_purchased` event emission on `!dbOrTx`. When `processTopUp` receives a caller-owned transaction, the event is skipped because the outer transaction hasn't committed yet — the credit change could still roll back. When `processTopUp` owns the transaction (the default path), the event fires after the transaction has already committed, so it's safe. Callers passing `dbOrTx` (Kilo Pass issuance, KiloClaw settlement) can add their own post-commit event emission if needed.

apps/web/src/lib/user.ts
jrf0110: Fixed: added `isNull(organizations.deleted_at)` to the membership query's where clause. This matches the pattern used by other membership readers in the codebase and ensures soft-deleted orgs are excluded from auth event enrichment data.

apps/web/src/routers/admin-router.ts
jrf0110: Fixed: replaced `.returning()` + row-count check with a pre-read of current blocked state and a transition check. The handler now queries the user's current `blocked_reason` before the update, computes `wasBlocked`, and only emits the event when `isBlocking !== wasBlocked` (i.e., there's an actual state transition). Idempotent retries and stale submissions no longer produce phantom events.

apps/web/src/lib/credits.ts
jrf0110: Fixed: replaced the `!dbOrTx` guard with `after()` deferral. When `dbOrTx` is provided, the event is now deferred via Next.js `after()` which runs after the response is sent (and the caller's transaction has committed). When `processTopUp` owns the transaction, the event fires immediately since the commit already happened. This preserves event coverage for Kilo Pass issuance, KiloClaw settlement, and invoice-paid paths while still preventing premature emission.

apps/web/src/routers/admin-router.ts
jrf0110: Fixed: wrapped both the SELECT and UPDATE in a `db.transaction()` with `.for('update')` on the SELECT to acquire a row lock. This eliminates the TOCTOU race — concurrent admin requests for the same user will serialize on the row lock, so only one can observe the pre-update state and emit the event. The transition detection (wasBlocked vs isBlocking) now runs atomically within the transaction.

PR #3294, opened 2026-05-18
fix(db): add query performance indexes

RSO:
## Summary

Adds targeted database indexes for Supabase-reported hot queries:

- `microdollar_usage_metadata(session_id, created_at)` for session usage aggregation with a created-at lower bound.
- `app_builder_projects(git_repo_full_name, git_platform_integration_id)` for app builder push webhook project lookups.

Both migrations use `CREATE INDEX CONCURRENTLY` outside Drizzle's transaction wrapper.

## Verification

No manual runtime verification was run; these are database-only index migrations reviewed against the reported query predicates.

## Visual Changes

N/A

## Reviewer Notes

The index shapes intentionally differ from the single-column Supabase suggestions where the full query predicates made composite indexes a better fit. The suggested `security_analysis_owner_state(owned_by_user_id)` index was not added because an existing partial unique index already covers that lookup.


RSO:
CleanShot 2026-05-18 at 11 07 46@2x
CleanShot 2026-05-18 at 11 08 08@2x

markijbema (approved):

PR #3295, opened 2026-05-18
perf(db): drop bloated unused indexes

RSO:
## Summary

Drops three large low-value indexes from the Drizzle schema and adds a migration to reclaim the space with concurrent, idempotent drops:

- `public.stytch_fingerprints::idx_fingerprint_data`: 3958 MB, 0 scans.
- `public.code_indexing_manifest::IDX_code_indexing_manifest_file_hash`: 1024 MB, 0 scans.
- `public.free_model_usage::idx_free_model_usage_user_created_at`: 11 GB, 109 scans, 36,056 tuples read.

Research context:

- Supabase `pg_stat_user_indexes` showed the two first indexes have had zero scans since `pg_stat_database.stats_reset = 2025-02-21 10:16:42.066025+00`.
- Repo search showed `stytch_fingerprints.fingerprint_data` is stored but not queried, and `code_indexing_manifest.file_hash` is selected/aggregated but not filtered through the standalone `file_hash` index.
- The free-model user index matches one per-user rate-limit query, but that path only runs for server-side free-model products coming from Cloudflare IPs (`cloud-agent`, `code-review`, `app-builder`). Production stats show very low use relative to the 11 GB write/space cost, so this removes it rather than replacing it.
- The migration uses `DROP INDEX CONCURRENTLY IF EXISTS` to avoid write-blocking table locks while reclaiming roughly 16 GB of index storage.

## Verification

- Queried Supabase production index statistics through `supabase db query --linked`.
- Inspected the repo query paths for `stytch_fingerprints`, `code_indexing_manifest`, and `free_model_usage` to confirm whether replacement indexes were warranted.

## Visual Changes

N/A

## Reviewer Notes

`pnpm drizzle generate` produced the schema snapshot/journal update. The generated drop migration was adjusted to use the existing concurrent-index migration pattern (`COMMIT`, concurrent operation, `BEGIN`) because these indexes are large production objects.

The bootstrap migration verifier was not run because Docker API access was not granted in this session. If per-user free-model rate-limit traffic grows substantially, revisit the table design; the current table has no feature/product discriminator that would allow a more targeted replacement index.

### DIFF
--- a/packages/db/src/migrations/0133_fearless_lethal_legion.sql
+++ b/packages/db/src/migrations/0133_fearless_lethal_legion.sql
@@ -1,0 +1,5 @@
+COMMIT;--> statement-breakpoint
+DROP INDEX CONCURRENTLY IF EXISTS "IDX_code_indexing_manifest_file_hash";--> statement-breakpoint
+DROP INDEX CONCURRENTLY IF EXISTS "idx_free_model_usage_user_created_at";--> statement-breakpoint
+DROP INDEX CONCURRENTLY IF EXISTS "idx_fingerprint_data";--> statement-breakpoint
+BEGIN;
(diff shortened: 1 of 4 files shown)


markijbema (approved):
Please check https://app.kilo.ai/admin/free-model-usage before and after this PR

PR #3298, opened 2026-05-18
fix(code-reviews): retry mkdir failures

alex-alecu:
## Summary

- Treat Cloud Agent workspace directory setup failures as retryable infrastructure failures and destroy the bad sandbox before callers retry.
- Preserve sandbox startup failures as the `cause` on `WrapperNotReadyError` so existing sandbox-500 recovery can detect `ProcessExitedBeforeReadyError`-style failures.
- Keep code-review retries marker-driven and one-shot; plain prose errors still do not trigger retry behavior.

## Verification

No manual verification performed; this change is backend error propagation and retry classification only.

## Visual Changes

N/A

## Reviewer Notes

- Sandbox recovery is intentionally structural: `httpStatus: 500` causes are classified, while generic HTTP 500 message text alone is not.
- Wrapper startup cause preservation lets the existing prepare-session/code-review retry marker path handle sandbox startup 500s without a new retry policy.

### DIFF
--- a/services/cloud-agent-next/src/execution/orchestrator.ts
+++ b/services/cloud-agent-next/src/execution/orchestrator.ts
@@ -26,7 +26,7 @@ import { withDORetry } from '../utils/do-retry.js';
 import { normalizeAgentMode } from '../schema.js';
 import { buildImagePromptParts, downloadImagePromptParts } from './image-prompt-parts.js';
 import { withTimeout } from '@kilocode/worker-utils';
-import { withSandboxInternalServerErrorRecovery } from '../sandbox-recovery.js';
+import { withPreparationInfrastructureRecovery } from '../sandbox-recovery.js';
 
 /** Maximum time allowed for workspace preparation (resume, init, fast path). */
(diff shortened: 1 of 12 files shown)


services/cloud-agent-next/src/execution/orchestrator.ts
```diff
@@ -200,7 +200,7 @@ export class ExecutionOrchestrator {
     };
 
     const { prepared, wrapperClient, kiloSessionId, fileParts } =
-      await withSandboxInternalServerErrorRecovery(
+      await withPreparationInfrastructureRecovery(
```
eshurakov, line 203: Why to rename this and then duplicate a method? withSandboxInternalServerErrorRecovery is now dead code.

Shouldn't we add another error handling case to the existing method instead?

eshurakov (changes requested):

services/cloud-agent-next/src/execution/orchestrator.ts
alex-alecu: Done in 0c72ce3.

eshurakov (approved):

PR #3301, opened 2026-05-18
fix(kiloclaw) user location best effort sync

St0rmz1:
## Summary

Saving a location from the Morning Briefing Settings UI was routed through `kiloclaw.updateConfig`, which calls `provisionInstance` and then runs the DO's heavy `provision()` flow under a per-user provision lock. Inside that flow, a userLocation change triggers two sequential gateway calls (`writeUserProfile` then `updateMorningBriefingUserLocation`) on top of the rest of the provision work. Any stall in that chain trips the Vercel function timeout and surfaces as a 500 in the browser. Updating interests already had its own thin path (`updateBriefingInterests` to a focused DO method to a single gateway call) and consistently completes in roughly 500ms.

This change mirrors the interests shape for user location:

* New DO method `updateUserLocation` that updates DO state, writes USER.md, and best-effort syncs the morning-briefing plugin's config.json.
* New worker route `POST /api/platform/morning-briefing/user-location`.
* New `KiloClawInternalClient.updateUserLocation`.
* New tRPC mutation `kiloclaw.updateUserLocation` (and org variant) with the same admin gate as `updateBriefingInterests`.
* `useKiloClaw` and `useOrgKiloClaw` expose the mutation and invalidate `getStatus` on success.
* `MorningBriefingLocationEditor` in SettingsTab uses the new mutation.

Onboarding's initial location set still flows through `updateConfig` / `provision` because it is bundled with the rest of the initial config; only the post-provisioning settings edit moves off that path. The defensive try/catch around the plugin sync inside `provision()` is kept as a belt and suspenders measure for the onboarding path.

Two issues caught in review of the focused method:

1. **Safe retry.** The DO method originally persisted the new location to durable state before awaiting `writeUserProfile`. If the profile write failed, the caller saw an error but a subsequent retry short circuited on the `previous === input` check and reported success, leaving USER.md stale forever. Reordered so the required gateway write happens before DO state is persisted.
2. **Stopped instance behavior.** The old `updateConfig` path supported edits on stopped instances by gating `writeUserProfile` on `status === 'running'` while still updating DO state. The new method threw `Instance is not running` instead, which regressed stopped instance saves since the Settings editor stays visible after a stop. Now both gateway calls are skipped when stopped and DO state is persisted; the next boot picks up the new value via `KILOCLAW_USER_LOCATION`.

## Verification

* [x] Existing kiloclaw vitest suite passes (491 tests, up from 487)
* [x] Four new DO unit tests in `describe('updateUserLocation')` cover stopped persistence with no gateway calls, required write rollback, best-effort plugin sync failure, and the no-op short circuit
* [x] Typecheck, lint, format all clean across kiloclaw and web
* [ ] Manually save a new Location from Settings on a running prod instance and confirm the tRPC call returns success quickly without a 500
* [ ] Manually save a new Location while the instance is stopped and confirm DO state updates with no gateway calls
* [ ] Manually clear the Location and confirm the toast text appears and the editor returns to the empty state

## Visual Changes

N/A. No UI markup changes; only the wiring under `MorningBriefingLocationEditor.handleSave` and `handleClear` changed.

## Reviewer Notes

* `provision()` still accepts `userLocation` because onboarding bundles it with the rest of the initial config. The defensive try/catch around `updateMorningBriefingUserLocation` inside `provision()` is intentional for the onboarding path.
* The new DO method intentionally short circuits on `previous === input` only after the required gateway write succeeds, so a retry after a transient `writeUserProfile` failure will actually retry instead of falsely reporting success.
* Stopped instance behavior matches the prior `provision()` semantics: DO state moves, file and plugin state catch up on next boot via the env var path.

### DIFF
--- a/services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
+++ b/services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
@@ -3924,32 +3924,45 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
    * new value regardless. Failing the user-facing save on a transient
    * plugin write-queue stall is worse than silently degrading to "takes
    * effect on next deploy."
+   *
+   * When the instance is stopped, persist DO state only and skip gateway
+   * calls — the next boot picks up the new location through the
+   * KILOCLAW_USER_LOCATION env-var path. This mirrors what the old
+   * `updateConfig`/provision path did (it gated `writeUserProfile` on
+   * `status === 'running'` but still updated DO state regardless).
+   *
+   * Ordering matters: do not persist the new location to DO state until
+   * the required gateway write has succeeded. If we persisted first and
+   * `writeUserProfile` failed, a retry would short-circuit on the
+   * `previous === input.userLocation` check and report success while
+   * USER.md remained stale.
    */
   async updateUserLocation(input: { userLocation: string | null }) {
     await this.loadState();
-    if (this.s.status !== 'running') {
-      throw new Error('Instance is not running');
-    }
     const previous = this.s.userLocation ?? null;
     if (input.userLocation === previous) {
       return { ok: true, userLocation: previous };
     }
 
+    if (this.s.status === 'running') {
+      await gateway.writeUserProfile(this.s, this.env, {
+        userLocation: input.userLocation,
+      });
+    }
+
     this.s.userLocation = input.userLocation;
     await this.ctx.storage.put({ userLocation: input.userLocation });
 
-    await gateway.writeUserProfile(this.s, this.env, {
-      userLocation: input.userLocation,
-    });
-
-    try {
-      await gateway.updateMorningBriefingUserLocation(this.s, this.env, {
-        userLocation: input.userLocation,
-      });
-    } catch (err) {
-      doWarn(this.s, 'updateMorningBriefingUserLocation failed', {
-        error: toLoggable(err),
-      });
+    if (this.s.status === 'running') {
+      try {
+        await gateway.updateMorningBriefingUserLocation(this.s, this.env, {
+          userLocation: input.userLocation,
+        });
+      } catch (err) {
+        doWarn(this.s, 'updateMorningBriefingUserLocation failed', {
+          error: toLoggable(err),
+        });
+      }
     }
 
     return { ok: true, userLocation: input.userLocation };
--- a/services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
+++ b/services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
@@ -979,26 +979,27 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
         userTimezone,
         userLocation,
       });
-      // Also propagate userLocation to the morning-briefing plugin's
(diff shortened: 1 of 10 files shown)


services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
```diff
@@ -3905,6 +3906,68 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
...
+    if (this.s.status === 'running') {
+      await gateway.writeUserProfile(this.s, this.env, {
+        userLocation: input.userLocation,
+      });
+    }
+
+    this.s.userLocation = input.userLocation;
+    await this.ctx.storage.put({ userLocation: input.userLocation });
+
+    if (this.s.status === 'running') {
```
jeanduplessis: WARNING: stopped clears can return success without clearing the plugin's persisted location override.

Trace:
1. `updateUserLocation({ userLocation: null })` persists DO state here, but this `status === 'running'` gate skips the plugin `user-location` write while stopped.
2. A prior running save can already have written `config.json.userLocation` inside the plugin volume.
3. The plugin reads stored `config.json.userLocation` before falling back to `KILOCLAW_USER_LOCATION`, so restarting after a stopped clear can still use the old stored override.

Impact: user clears Location while the instance is stopped, receives success, then Local News can continue using the old location after restart instead of clearing/falling back.

services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
```diff
@@ -3905,6 +3906,68 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
...
+   * USER.md remained stale.
+   */
+  async updateUserLocation(input: { userLocation: string | null }) {
+    await this.loadState();
+    const previous = this.s.userLocation ?? null;
+    if (input.userLocation === previous) {
+      return { ok: true, userLocation: previous };
+    }
+
+    if (this.s.status === 'running') {
```
jeanduplessis: WARNING: this gate treats `starting` and `restarting` like `stopped`, which creates a lost-update race.

Trace:
1. `startAsync()` and restart flow move state to `starting` / `restarting` before background deployment finishes.
2. Startup builds runtime env from DO state once; if this mutation arrives after that env snapshot but before status becomes `running`, this branch skips USER.md and plugin writes while still persisting DO state and returning success.
3. Current boot then completes with old env, old USER.md, and old plugin config until another retry or redeploy.

Impact: a save during start/restart can show a success toast but fail to affect the next briefing on the newly running machine.

jeanduplessis (changes requested):
Reviewed final combined diff, PR body, unresolved review threads, KiloClaw controller/data-model specs, changed files, and surrounding caller/callee paths.

Found two warning-level correctness gaps in the new non-running `updateUserLocation` path:
1. Clearing Location while stopped can leave an older plugin `config.json.userLocation` override active after restart.
2. Saves during `starting` / `restarting` can race after runtime env construction, return success, and miss the current boot.

No unresolved GitHub review threads were active. No changed PostHog tracking loss or duplication verified. Added tests cover several direct DO branches, but not these two edge-state paths.

jeanduplessis (approved):

PR #3304, opened 2026-05-18
feat(kilo-chat): attachments backend (R2 + presigned URLs + plugin)

iscekic:
## Summary

Adds first-class Kilo Chat attachment support across the shared contract, Worker backend, controller proxy, plugin, and mobile renderer.

- Extends shared Kilo Chat schemas/events/webhooks with attachment content blocks, attachment init/read URL contracts, and bot capabilities.
- Adds R2-backed attachment storage to the kilo-chat Worker using presigned PUT/GET URLs, per-conversation Durable Object attachment rows, upload validation, canonical stored metadata, and cleanup on edit/delete/conversation destroy/orphan sweep.
- Exposes user and bot attachment routes and relays them through the kiloclaw controller.
- Updates the kiloclaw Kilo Chat plugin to advertise attachment capability, upload outbound media through R2, and download inbound attachment webhooks into OpenClaw media buffers with size caps and failure fallbacks.
- Renders attachment blocks in mobile chat bubbles as paperclip filename rows.

## Verification

Manual verification not performed; this PR primarily adds backend/API plumbing and plugin integration paths.

## Visual Changes

Not captured. Mobile message bubbles now render attachment blocks as a paperclip with filename.

## Reviewer Notes

Focus areas: R2 credential/binding setup, attachment authorization boundaries, presigned URL header requirements, Durable Object attachment cleanup paths, and plugin inbound/outbound media failure behavior.


services/kilo-chat/src/do/conversation-do.ts
```diff
@@ -1166,11 +1319,194 @@ export class ConversationDO extends DurableObject<Env> {
     };
   }
 
-  destroyAndReturnMembers(): DestroyResult {
+  initAttachment(params: InitAttachmentParams): InitAttachmentResult {
+    if (
+      typeof params.size !== 'number' ||
+      !Number.isFinite(params.size) ||
+      params.size < 0 ||
```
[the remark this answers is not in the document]
iscekic: No longer applies. `initAttachment` now validates through `initAttachmentInternalSchema`, derived from `attachmentInitRequestSchema`, before any DB write. The shared attachment schemas now use `nonnegative()` for `size`, so `size: 0` is a single consistent contract across HTTP, DO RPC, and content blocks.

services/kilo-chat/src/util/presigner.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export async function mintPutUrl(
+  params: Cfg & {
+    key: string;
+    contentType: string;
+    expiresSeconds: number;
+  }
+): Promise<{ url: string; headers: Record<string, string> }> {
+  const url = new URL(`${r2Origin(params.accountId)}/${params.bucket}/${params.key}`);
```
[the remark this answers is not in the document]
iscekic: No longer applies for the current call path. `params.key` is the R2 key returned by `buildAttachmentR2Key`, which percent-encodes each caller-controlled ID segment before storage/use. The remaining prefix is trusted Worker configuration (currently empty in `wrangler.jsonc`), and `presigner` is only called with those generated stored keys.

services/kilo-chat/src/do/conversation-do.ts
```diff
@@ -548,20 +666,64 @@ export class ConversationDO extends DurableObject<Env> {
...
+      }
+      if (row.status !== 'pending') {
+        return { ok: false, code: 'conflict', error: 'Attachment is already linked' };
+      }
+      if (row.uploader_id !== params.senderId) {
+        return { ok: false, code: 'forbidden', error: 'Attachment uploader does not match sender' };
+      }
+      attachmentRows.push(row);
+    }
+    const uploadError = await this.validateUploadedAttachmentObjects(attachmentRows);
```
jeanduplessis, line 702: WARNING: this can still double-link one uploaded attachment under overlapping sends.

Trace:
1. Request A reads row as `pending`, then awaits `validateUploadedAttachmentObjects(...)`.
2. That await crosses external R2 I/O, so request B can enter same DO method, read same row as `pending`, and pass same validation.
3. Both requests later insert messages and update attachment by `id` only; second update overwrites `message_id`.

Impact: duplicate/retried create-message requests can commit two messages that reference one attachment row. Later edit/delete cleanup follows only last `message_id`, so another committed message can keep metadata pointing at an object that gets purged.

services/kilo-chat/src/do/conversation-do.ts
```diff
@@ -1166,11 +1371,169 @@ export class ConversationDO extends DurableObject<Env> {
...
+      return { ok: false, code: 'forbidden', error: 'Forbidden' };
+    }
+
+    const now = Date.now();
+    const dedupeCutoff = now - INIT_DEDUPE_WINDOW_MS;
+    const existing = this.db
+      .select()
+      .from(attachments)
+      .where(
+        and(
```
jeanduplessis: WARNING: 30-second dedupe uses attachment metadata as idempotency identity, so distinct files can collapse.

Trace:
1. Pending init rows dedupe on uploader, filename, size, MIME type, and time window.
2. Public init contract carries no client idempotency key or content hash.
3. Two legitimate files with same metadata before first link get same `attachmentId` and `r2Key`; later PUT writes replace earlier bytes at that key.

Impact: users or bots cannot reliably send two same-metadata attachments in quick succession. Stored attachment identity can silently point at later bytes instead of first upload.

services/kilo-chat/src/do/conversation-do.ts
```diff
@@ -1166,11 +1371,169 @@ export class ConversationDO extends DurableObject<Env> {
...
     const membersCopy = info.members;
+
+    // Walk every R2 object under this conversation's prefix and delete it.
+    // This covers attachments whose DB rows we'll wipe below as well as any
+    // orphaned objects (e.g. uploads that completed after the row was
+    // expired by the orphan sweeper).
+    const prefix = `${this.keyPrefix}attachments/${info.id}/`;
+    let cursor: string | undefined;
+    do {
+      const listed = await this.env.MEDIA_BUCKET.list({ prefix, cursor });
```
jeanduplessis, line 1557: WARNING: destroy sweep can miss uploads that complete after prefix listing.

Trace:
1. Init returns a presigned PUT URL valid for 900 seconds.
2. Destroy lists/deletes objects currently present under conversation prefix, then deletes attachment rows.
3. If client got PUT URL before destroy but uploads after this `list(...)`, no object is seen during sweep; row is gone, but URL remains valid.

Impact: conversation/sandbox cleanup can leave orphaned attachment bytes in R2 after destroy, contradicting full cleanup expectation and removing future row/alarm path that would sweep them.

jeanduplessis (approved):
Found three attachment lifecycle issues worth fixing before merge:

- overlapping create-message calls can link one uploaded attachment into multiple committed messages;
- metadata-only init dedupe can merge distinct same-metadata files;
- conversation destroy can miss bytes uploaded through still-valid presigned PUT URLs after prefix scan.

Reviewed final combined diff, attachment cleanup/retry/fallback paths, active review threads, PR body claims, PostHog touch points, and new tests.

services/kilo-chat/src/do/conversation-do.ts
iscekic: Fixed in 3e7bee70. The UPDATE now requires `status='pending'` and the transaction aborts (returning a `conflict` to the caller) if a concurrent createMessage already claimed the row. Cheaper than `blockConcurrencyWhile` since it doesn't serialize unrelated traffic.

services/kilo-chat/src/do/conversation-do.ts
iscekic: Fixed in 6004c1d6. Dropped the metadata-tuple dedupe entirely and added an optional `idempotencyKey` to `attachmentInitRequestSchema`. When present, repeated inits from the same uploader (within `INIT_DEDUPE_WINDOW_MS`) return the same attachment id; when absent, every init mints a fresh attachment, so two genuinely distinct files with identical filename/size/mime no longer collide. New `idempotency_key` column + `(uploader_id, idempotency_key, created_at)` index ship in migration 0003. Plugin client passes the key through.

services/kilo-chat/src/do/conversation-do.ts
iscekic: Accepting the risk. 23d9ff00 adds an inline comment near the destroy `list()` loop documenting the gap and the lowest-effort follow-up (shorten `PUT_URL_TTL_SECONDS`) if it ever shows up in R2 usage.

PR #3312, opened 2026-05-18
feat(security-agent): move manual workflows into workers

jeanduplessis:
## Summary
Security Agent manual sync, dismissal, and analysis work now enter durable Worker-backed command paths, while callback delivery stays generic in Cloud Agent Next and finalization remains owned by Security Auto Analysis.

### Why this change is needed
Manual Security Agent actions can trigger multi-step GitHub, queue, Cloud Agent, and callback work. Keeping that orchestration on synchronous web paths makes request latency, retries, ownership checks, and terminal state recovery harder to control. The security callback fast path also made `cloud-agent-next` aware of one product-specific downstream consumer, which blurred service boundaries.

### How this is addressed
- Route manual sync, dismissal, and analysis requests through internal Worker clients that return durable queue-admission state instead of implying inline completion.
- Add Security Sync Worker support for manual sync and dismissal queue messages, including ownership enforcement and downstream Dependabot writeback.
- Add Security Auto Analysis Worker support for manual analysis admission, scheduled owner dispatch, callback finalization, sandbox extraction, triage-only completion, auto-dismiss handling, and queue-state transitions.
- Keep `cloud-agent-next` callback delivery generic: callers provide ordinary HTTP callback targets, and Security Auto Analysis now configures Worker callback ingress while retaining the web route as rollback-compatible ingress.
- Centralize callback lifecycle settlement and stale queue reconciliation so duplicate deliveries, already-terminal findings, and recovered queue rows converge without downgrading durable finding state.
- Share automatic-analysis eligibility and Dependabot dismissal target parsing across Worker/web paths so unknown severity and malformed upstream metadata behave consistently.
- Refresh Security Agent UI state after queued async work so findings, counts, and sync state converge after Worker consumers apply updates.
- Add unit and integration coverage across web handlers, Worker queue consumers, callback recovery, dismissal, launch, policy helpers, and query transitions.

### Architecture
```mermaid
flowchart LR
  UI[Security Agent UI] --> WEB[tRPC handlers]
  WEB --> MSC[Manual sync client]
  WEB --> MDC[Manual dismiss client]
  WEB --> MAC[Manual analysis client]

  MSC --> SSIN[Security Sync Worker ingress]
  MDC --> SSIN
  SSIN --> SSQ[security-sync-jobs queue]
  SSCRON[Scheduled sync cron] --> SSQ
  SSQ --> SSC[Security Sync Worker consumer]
  SSC --> GH[GitHub Dependabot API]
  SSC --> DB[(Postgres security state)]

  MAC --> AAIN[Security Auto Analysis Worker ingress]
  AAIN --> MAQ[manual analysis command queue]
  AACRON[Auto-analysis owner dispatcher] --> AOQ[owner analysis queue]
  MAQ --> AAC[Auto Analysis Worker consumers]
  AOQ --> AAC
  AAC -->|triage-only| DB
  AAC -->|sandbox needed| CAN[Cloud Agent Next]

  CAN --> CAQ[Cloud Agent callback delivery queue]
  CAQ --> CBT[HTTP callback target]
  CBT --> AAWEB[Worker callback ingress]
  CBT -. rollback route .-> WEBHOOK[Legacy web callback route]
  AAWEB --> CBQ[auto-analysis callback queue]
  WEBHOOK --> CBQ
  CBQ --> CBF[Auto Analysis callback finalizer]
  CBF --> DB
  CBF --> GH

  DB --> UI
```

## Human Verification
- Local review pass completed across security, logic, types, data, resources, style, and React focus areas; no actionable critical or warning findings remained after re-check.
- Security Sync Worker coverage remains represented in the branch for dismissal, dispatch, and unknown-severity eligibility behavior.
- Security Auto Analysis Worker tests validated manual starts, callbacks, launch flow, and queue/database reconciliation behavior.
- React Doctor diff scan completed with PASS analyzer status after the latest Security Agent UI changes.

### Human Reviewer Flags
- Architectural change: manual web actions now return durable queue admission instead of synchronous terminal results.
- Architectural boundary change: `cloud-agent-next` no longer knows about Security Auto Analysis delivery mode or service binding; callback destination ownership moves back to the caller.
- Callback routing now targets Security Auto Analysis Worker ingress in configured deployments; the web callback route stays compatibility-only for rollback routing.
- Queue consumers re-check finding ownership before cross-tenant-sensitive mutation paths.
- Callback finalization owns delayed-result recovery, terminal queue transitions, stale-row healing, and optional auto-dismiss writeback.

### Code Reviewer Agent
<details>

- Review worker trust boundaries around `services/security-sync/src/dismiss.ts` and `services/security-auto-analysis/src/manual-analysis.ts`.
- Review callback/result recovery path across Cloud Agent Next and Security Auto Analysis Worker for ID contract alignment.
- Review generic callback delivery in `services/cloud-agent-next/src/callbacks/delivery.ts` plus Security Auto Analysis target construction in `services/security-auto-analysis/src/launch.ts`.
- Review shared policy/parser boundaries in `packages/worker-utils/src/security-auto-analysis-policy.ts` and `packages/worker-utils/src/dependabot-dismissal-target.ts`.
- Review UI queued-admission refresh behavior in Security Agent React state containers.

</details>


apps/web/src/components/security-agent/SecurityAgentContext.tsx
```diff
@@ -122,6 +132,27 @@ export function SecurityAgentProvider({ organizationId, children }: SecurityAgen
...
+    []
+  );
+
+  const refreshAcceptedQueueMutation = useCallback(() => {
+    void queryClient.invalidateQueries();
+    for (const delay of ACCEPTED_QUEUE_REFRESH_DELAYS_MS) {
+      const timer = window.setTimeout(() => {
+        void queryClient.invalidateQueries();
+      }, delay);
+      acceptedQueueRefreshTimersRef.current.push(timer);
```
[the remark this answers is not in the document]
jeanduplessis: Fixed. Fired delayed-refresh timers now remove their own IDs from the ref, so mounted sessions no longer retain completed timers while unmount cleanup still cancels pending work.

services/security-auto-analysis/src/callbacks.ts
```diff
@@ -0,0 +1,416 @@
...
+      findingId: params.findingId,
+      toStatus: 'completed',
+      failureCode: 'SKIPPED_NO_LONGER_ELIGIBLE',
+      errorMessage: null,
+    });
+    await clearAnalysisStatus(params.db, params.findingId);
+    return { status: disposition };
+  }
+
+  if (params.payload.status === 'completed') {
```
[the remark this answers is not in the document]
jeanduplessis: Removed. Completed payloads are already routed to the completed finalizer before this function is reached, so the failed-finalizer guard was dead code and only obscured the terminalization path.

apps/web/src/components/security-agent/SecurityAgentPageClient.tsx
```diff
@@ -57,10 +58,31 @@ export function SecurityAgentPageClient({ organizationId }: SecurityAgentPageCli
...
+    []
+  );
+
+  const refreshAcceptedQueueMutation = useCallback(() => {
+    void queryClient.invalidateQueries();
+    for (const delay of ACCEPTED_QUEUE_REFRESH_DELAYS_MS) {
+      const timer = window.setTimeout(() => {
+        void queryClient.invalidateQueries();
+      }, delay);
+      acceptedQueueRefreshTimersRef.current.push(timer);
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in parallel with the provider path. Each timeout removes itself from the tracked timer list after it fires, preventing retained IDs across repeated accepted mutations.

jrf0110:
I'm noticing a growing number of services that use the `INTERNAL_API_SECRET` as a shared secret for cloud-to-service communication. While that does make secret rotation a little easier, it also increases the surface area of damage if this secret is compromised.

I'd recommend doing what we do for the abuse service, which is to protect your worker with Cloudflare Access, create a Cloudflare Access Service Token for authenticating with that application, and share that specific token with the cloud service to authenticate with your worker.

apps/web/src/components/security-agent/manual-analysis-admission-copy.test.ts
```diff
@@ -0,0 +1,12 @@
...
+
+describe('manualAnalysisAdmissionCopy', () => {
+  test('describes manual analysis as queued admission', () => {
+    expect(manualAnalysisAdmissionCopy).toEqual({
+      successTitle: 'Analysis queued',
+      failureTitle: 'Failed to queue analysis',
+      pendingLabel: 'Queueing',
+    });
+  });
+});
```
RSO, line 10: https://janpaulposma.nl/#duplicationjs

apps/web/src/components/security-agent/SecurityAgentContext.tsx
```diff
@@ -122,6 +125,30 @@ export function SecurityAgentProvider({ organizationId, children }: SecurityAgen
...
+    for (const delay of ACCEPTED_QUEUE_REFRESH_DELAYS_MS) {
+      const timer = window.setTimeout(() => {
+        void queryClient.invalidateQueries();
+        acceptedQueueRefreshTimersRef.current = acceptedQueueRefreshTimersRef.current.filter(
+          pendingTimer => pendingTimer !== timer
+        );
+      }, delay);
+      acceptedQueueRefreshTimersRef.current.push(timer);
+    }
+  }, [queryClient]);
```
RSO: 1) Shouldn't we return a function in the `useEffect` to clean-up any lingering timeouts upon unmount?
2) I don't get the intent of this logic, any time you call `refreshAcceptedQueueMutation` you schedule 3 new timeouts to invalidate the queries? Why does that accumulate, instead of canceling the old timeouts and starting new ones (e.g., debounce)?

I feel like there's a smell here, but I'm not really sure what we're trying to do in the first place.

apps/web/src/components/security-agent/SecurityAgentPageClient.tsx
```diff
@@ -57,10 +59,34 @@ export function SecurityAgentPageClient({ organizationId }: SecurityAgentPageCli
...
+    for (const delay of ACCEPTED_QUEUE_REFRESH_DELAYS_MS) {
+      const timer = window.setTimeout(() => {
+        void queryClient.invalidateQueries();
+        acceptedQueueRefreshTimersRef.current = acceptedQueueRefreshTimersRef.current.filter(
+          pendingTimer => pendingTimer !== timer
+        );
+      }, delay);
+      acceptedQueueRefreshTimersRef.current.push(timer);
+    }
+  }, [queryClient]);
```
RSO: Seeing how this is exactly the same as the logic in `SecurityAgentConttext`, should we abstract it to a hook?

packages/worker-utils/src/security-auto-analysis-policy.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (severity === 'low') return LOW_SEVERITY_RANK;
+  return null;
+}
+
+function getMaxSeverityRank(minSeverity: AutoAnalysisMinSeverity): AutoAnalysisSeverityRank {
+  if (minSeverity === 'critical') return 0;
+  if (minSeverity === 'high') return 1;
+  if (minSeverity === 'medium') return 2;
+  return LOW_SEVERITY_RANK;
+}
```
RSO: 1) These two methods feel practically the same?
2) Why is `LOW_SEVERITY_RANK` a constant and the rest not?
3) Shouldn't this just be an object?

```ts
const SEVERITY_RANKS = {
  critical: 0,
  high: 1,
  medium: 2,
  low: 3,
}

const rank = SEVERITY_RANKS[severity];
```

services/cloud-agent-next/src/server.ts
```diff
@@ -325,7 +325,7 @@ export default {
 
     return app.fetch(request, env, ctx);
   },
-  async queue(batch: MessageBatch<unknown>, _env: Env): Promise<void> {
+  async queue(batch: MessageBatch<unknown>): Promise<void> {
```
RSO: Why is this change in here?

services/security-auto-analysis/src/db/queries.ts
```diff
@@ -54,10 +54,21 @@ function parseSecurityConfig(config: unknown): SecurityAgentConfig {
...
   };
+
+  if (parsed.data.model_slug !== undefined) {
+    if (parsed.data.triage_model_slug === undefined) {
+      resolvedConfig.triage_model_slug = undefined;
+    }
+    if (parsed.data.analysis_model_slug === undefined) {
+      resolvedConfig.analysis_model_slug = undefined;
+    }
+  }
```
RSO, line 70: What's the thinking here? What's the relationship between `config` and `config.data`? It seems like you're defining some relationship, but only when some other things are undefined?

apps/web/src/components/security-agent/SecurityAgentPageClient.tsx
RSO: Actually, is this component still used?

apps/web/src/lib/security-agent/router/shared-handlers.ts
RSO: I think this is no longer used?

services/security-auto-analysis/src/launch.test.ts
```diff
@@ -0,0 +1,433 @@
...
+      expect.objectContaining({
+        claim: expect.objectContaining({ source: 'manual', claimToken: 'manual-claim-token' }),
+        outcome: {
+          type: 'sandbox-running',
+          cloudAgentSessionId: 'agent-session',
+          kiloSessionId: 'ses-123',
+        },
+      })
+    );
+    expect(setFindingRunning).not.toHaveBeenCalled();
```
RSO: This is the only place where this method is called in the whole codebase, so I guess it makes sense that it will never be called

services/security-auto-analysis/src/launch.test.ts
```diff
@@ -0,0 +1,433 @@
...
+    expect(triageSecurityFinding).toHaveBeenCalledTimes(1);
+    expect(setFindingPending).toHaveBeenCalledWith({}, finding.id, null);
+    expect(transitionAnalysisStartLifecycle).toHaveBeenCalledWith(
+      {},
+      expect.objectContaining({
+        claim: expect.objectContaining({ source: 'manual', claimToken: 'manual-claim-token' }),
+        outcome: expect.objectContaining({ type: 'triage-only-completed' }),
+      })
+    );
+    expect(setFindingCompleted).not.toHaveBeenCalled();
```
RSO: Same here, this is the only place where this method is referenced, so it's dead code.

services/security-auto-analysis/src/launch.test.ts
```diff
@@ -0,0 +1,433 @@
...
+    expect(setFindingFailed).not.toHaveBeenCalled();
```
RSO: Same here

RSO (changes requested):
Still working on the review, but I'm pretty sure that this PR would leak the `INTERNAL_API_TOKEN` so I'm requesting changes already.

services/security-auto-analysis/src/launch.ts
```diff
@@ -104,16 +106,60 @@ type StartSecurityAnalysisParams = {
...
+  internalApiSecret: string
+): {
+  url: string;
+  headers: { 'X-Internal-Secret': string };
+} {
+  if (env.SECURITY_ANALYSIS_CALLBACK_ROUTING_MODE === 'web') {
+    const baseUrl = env.SECURITY_ANALYSIS_CALLBACK_WEB_BASE_URL.replace(/\/$/, '');
+    return {
+      url: `${baseUrl}/api/internal/security-analysis-callback/${findingId}`,
+      headers: { 'X-Internal-Secret': internalApiSecret },
```
RSO: The `callbackTarget` is readable by anyone with access to the cloud agent session. We should not use the internal API secret as the straight-up access token, but should instead use a similar pattern as `deriveBotCallbackToken`, where we sign the token using the internal secret.

services/security-auto-analysis/src/index.ts
```diff
@@ -28,6 +33,10 @@ async function timingSafeEqual(a: string, b: string): Promise<boolean> {
   return nodeTSE(new Uint8Array(digestA), new Uint8Array(digestB));
 }
 
+function workerRouteEnabled(value: string | undefined): boolean {
+  return value !== 'false';
+}
```
RSO: I'm not convinced this is the right abstraction. I'd either opt for having a `const MANUAL_ANALYSIS_COMMAND_ROUTING_ENABLED = env.MANUAL_ANALYSIS_COMMAND_ROUTING_ENABLED !== "false"`, or just inlining the `!== "false"` check.

services/security-auto-analysis/src/callbacks.ts
```diff
@@ -0,0 +1,429 @@
...
+  finding: CallbackFindingState,
+  payload: SecurityAnalysisCallbackPayload
+): CallbackDisposition {
+  const sessionMismatch =
+    (payload.cloudAgentSessionId &&
+      finding.session_id &&
+      payload.cloudAgentSessionId !== finding.session_id) ||
+    (payload.kiloSessionId &&
+      finding.cli_session_id &&
+      payload.kiloSessionId !== finding.cli_session_id);
```
RSO, line 73: This is going to sound a little sloppy, but according to my bot, `setFindingPending` clears session IDs, meaning that this check no longer matches. That seems to mean there's a short timing window after calling `setFindingPending` and a callback on return when the session isn't marked as stale, which could cause an unexpected result. I'm not 100% sure, but it sounds plausible, so I wanted to write it down here.

services/security-sync/src/index.ts
```diff
@@ -1,64 +1,238 @@
...
+}
+
 function jsonResponse(body: unknown, status = 200): Response {
   return new Response(JSON.stringify(body), {
     status,
     headers: { 'Content-Type': 'application/json' },
   });
 }
 
+function workerRouteEnabled(value: string | undefined): boolean {
```
RSO: Second time I see this helper 😝 still not a fan

apps/web/src/components/security-agent/manual-analysis-admission-copy.test.ts
jeanduplessis: Fixed locally. Test no longer duplicates full copy object; it now asserts intent for queued success, failed queue, and pending queue labels.

apps/web/src/components/security-agent/SecurityAgentContext.tsx
jeanduplessis: Cleanup already added. Delayed invalidations are intentional because queue admission is eventually consistent after Worker write. Timers self-remove, unmount clears leftovers, and independent accepted actions keep independent refresh windows.

apps/web/src/components/security-agent/SecurityAgentPageClient.tsx
jeanduplessis: Addressed by removing `SecurityAgentPageClient`. Shared state now lives in `SecurityAgentContext` provider, so duplicated logic target is gone.

packages/worker-utils/src/security-auto-analysis-policy.ts
jeanduplessis: Fixed locally. Replaced parallel conditionals with `SEVERITY_RANKS` and `MIN_SEVERITY_MAX_RANKS` maps.

services/cloud-agent-next/src/server.ts
jeanduplessis: Outdated. Change was removed from current diff; no `cloud-agent-next` server change remains.

services/security-auto-analysis/src/db/queries.ts
jeanduplessis: Clarified locally with comment. This preserves legacy unified `model_slug` fallback instead of masking it with split model defaults when stored config predates split triage/analysis model fields.

apps/web/src/components/security-agent/SecurityAgentPageClient.tsx
jeanduplessis: Correct. Component was unused and is now removed with export/reference cleanup.

apps/web/src/lib/security-agent/router/shared-handlers.ts
jeanduplessis: Correct. Removed unused `getGitHubToken` dependency from shared handler deps and router implementations.

services/security-auto-analysis/src/launch.test.ts
jeanduplessis: Retry path is reachable from retry buttons through manual analysis start Worker. Test covers fallback when sandbox-only retry has no prior triage, preventing retry from skipping triage incorrectly.

services/security-auto-analysis/src/launch.test.ts
jeanduplessis: Not dead code. Production path reaches this when triage needs no sandbox and suggested action is `dismiss`; Worker auto-dismisses after durable completion. Test covers that side effect.

services/security-auto-analysis/src/launch.test.ts
jeanduplessis: Same path as above: triage-only auto-dismiss is production behavior, not test-only code.

services/security-auto-analysis/src/launch.ts
jeanduplessis: Addressed. Callback target no longer carries internal API secret. It uses scoped `X-Callback-Token`, derived from callback secret plus finding and claim token. Test asserts `X-Internal-Secret` is absent.

services/security-auto-analysis/src/index.ts
jeanduplessis: Fixed locally. Removed helper and inlined explicit route disable checks in `security-auto-analysis`.

services/security-auto-analysis/src/callbacks.ts
jeanduplessis: Good catch. Fixed with attempt-token guard. Callback URL carries attempt token, callback auth is scoped to finding plus claim token, and finalizer rejects missing/mismatched active claim before extraction. Session IDs are no longer sole stale guard.

services/security-sync/src/index.ts
jeanduplessis: Fixed locally. Removed helper and inlined route disable checks in `security-sync`.

services/security-sync/src/dismiss.ts
```diff
@@ -0,0 +1,132 @@
...
+    })
+    .where(eq(security_findings.id, finding.id));
+
+  await params.db.insert(security_audit_log).values({
+    owned_by_organization_id: params.message.owner.organizationId ?? null,
+    owned_by_user_id: params.message.owner.userId ?? null,
+    actor_id: params.message.actor.id,
+    actor_email: params.message.actor.email ?? null,
+    actor_name: params.message.actor.name ?? null,
+    action: SecurityAuditLogAction.FindingDismissed,
```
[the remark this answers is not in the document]
jeanduplessis: Addressed. Finding update and audit insert now run in one DB transaction after GitHub PATCH. Regression test covers rollback behavior.

apps/web/src/components/security-agent/manual-analysis-admission-copy.test.ts
RSO: I still don't see the point in having this test. It's like setting a `const FOO = "bar"` and then adding the following test: `expect(FOO).toMatch(/bar/)`. What's the point? It's just duplicating the original code, which I think is clear enough. Any time you change point A, you'll have to change test B.

packages/worker-utils/src/security-auto-analysis-policy.ts
```diff
@@ -0,0 +1,74 @@
...
+  low: LOW_SEVERITY_RANK,
+} as const satisfies Record<string, AutoAnalysisSeverityRank>;
+
+type KnownSeverity = keyof typeof SEVERITY_RANKS;
+
+const MIN_SEVERITY_MAX_RANKS = {
+  critical: SEVERITY_RANKS.critical,
+  high: SEVERITY_RANKS.high,
+  medium: SEVERITY_RANKS.medium,
+  all: LOW_SEVERITY_RANK,
```
RSO: What does `all` mean, really? It doesn't sound like it's `LOW_SEVERITY`.

services/security-auto-analysis/src/launch.test.ts
RSO: It sounds like you're disagreeing with me, but at the same time it seems like the code got updated?

services/security-auto-analysis/src/launch.test.ts
RSO: Same here: it sounds like you disagree, but the code path seems to have disappeared?

services/security-auto-analysis/src/launch.test.ts
RSO: Same here

services/security-sync/src/index.ts
```diff
@@ -1,34 +1,133 @@
...
 import { getWorkerDb } from '@kilocode/db/client';
 import { agent_configs } from '@kilocode/db/schema';
 import { eq, and, isNotNull, or } from 'drizzle-orm';
 import { syncOwner } from './sync';
+import { processSecurityFindingDismissal } from './dismiss';
+
+const SecuritySyncOwnerSchema = z
+  .object({
+    organizationId: z.string().uuid().optional(),
+    userId: z.string().uuid().optional(),
```
RSO: userId is not always a UUID. Old users have `oauth:google:googleID`

packages/worker-utils/src/security-auto-analysis-policy.ts
jeanduplessis: Fixed. Kept persisted/API value `all`, but renamed internal rank mapping to `ALL_SEVERITIES_MAX_RANK` so it reads as “all severities allowed” instead of “same thing as low severity.” I’ll consider a broader refactor of this threshold naming in a follow-up PR.

services/security-sync/src/index.ts
jeanduplessis: Fixed. `userId` now validates as non-empty text instead of UUID-only, matching `kilocode_users.id` and owned-by user columns. Added regression coverage for legacy `oauth:google:...` IDs through manual sync, queue processing, and dismissal commands.

services/security-auto-analysis/src/launch.test.ts
jeanduplessis: Same correction here: my previous reply conflated current behavior with old helper references. Current code no longer uses the old completed/running helpers; tests now assert durable lifecycle transitions through `transitionAnalysisStartLifecycle` and cover the triage-only auto-dismiss behavior directly.

services/security-auto-analysis/src/launch.test.ts
jeanduplessis: You’re right. My previous reply conflated current behavior with old helper references. The commented assertion was on a stale test-only setFindingCompleted mock. Current code no longer uses that helper; tests now assert the production path through transitionAnalysisStartLifecycle, with separate coverage for triage-only auto-dismiss.

services/security-auto-analysis/src/launch.test.ts
jeanduplessis: You’re right. My previous reply conflated current behavior with old helper references. The commented assertion was on a stale test-only setFindingRunning mock. Current code no longer uses that helper; tests now assert the production path through transitionAnalysisStartLifecycle, with coverage for retry fallback.

apps/web/src/components/security-agent/SecurityAgentContext.tsx
jeanduplessis: Agreed. Replaced the timeout fan-out with explicit bounded polling. After an accepted Worker command, the provider now runs one owner-scoped poll loop, invalidates targeted Security Agent queries every 3s for up to 18s, stops early once pending/running state appears, and clears interval/timeout on unmount. Pushed in `40c22d3ce`.

apps/web/src/components/security-agent/SecurityAgentContext.tsx
```diff
@@ -122,6 +142,96 @@ export function SecurityAgentProvider({ organizationId, children }: SecurityAgen
...
+      .some(([, data]) => listFindingsDataHasActiveAnalysis(data));
+  }, [isOrg, organizationId, queryClient, trpc]);
+
+  const pollAcceptedQueueMutation = useCallback(() => {
+    clearAcceptedQueuePoll();
+    invalidateAcceptedQueueQueries();
+
+    const intervalId = window.setInterval(() => {
+      invalidateAcceptedQueueQueries();
+      if (cachedListFindingsHasActiveAnalysis()) {
```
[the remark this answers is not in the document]
jeanduplessis: Good catch. Fixed in `07b5b4602`: the accepted-command poll now tracks whether it has ever seen pending/running state, keeps polling through that active phase, and only stops once active analysis clears or the bounded timeout expires.

RSO (approved):
Approving, but with a PR of this size, I think it's fair to assume that something will break, so let's apply some caution when we merge it.

PR #3323, opened 2026-05-19
feat(web): add collab onboarding

RSO:
## Summary
- Adds an admin-only `/collab` onboarding flow mockup for connecting collaboration, code, and issue services.
- Consolidates service selection into a single multi-select step with non-blocking warnings when chat or code platforms are missing.
- Adds a design-only authorization walkthrough and prefetches organization data to avoid the initial workspace loading waterfall.

## Verification
- Browser-smoked `/collab` as an admin fake user through workspace selection, service selection, and `/collab/authorize`.
- Browser-smoked non-admin access to `/collab` and confirmed it returns 404.
- Browser-smoked `/bot` and confirmed it no longer resolves.

## Visual Changes

https://github.com/user-attachments/assets/11d10da8-f0e9-44e4-9511-07fb00b6b08f

## Reviewer Notes
- The authorization step is still design-only and simulates OAuth success/skip behavior.
- `/collab` and `/collab/authorize` are intentionally gated to `user.is_admin === true` and return 404 for non-admin users.


apps/web/src/app/collab/_components/BotWizard.tsx
```diff
@@ -0,0 +1,237 @@
...
+    if (!canAdvance) return;
+    if (isWorkspaceStep) {
+      setStepIndex(1);
+      return;
+    }
+    if (!hasChatPlatform) {
+      setMissingPlatformWarning('chat');
+      return;
+    }
+    if (!hasCodePlatform) {
```
markijbema, line 77: is this coupling with the ordering of the pages? If you switch the order of the pages, do you need to switch tihs as well? feels a bit tricky

markijbema (approved):

apps/web/src/app/collab/_components/BotWizard.tsx
RSO: We only have one page at this point, so it's not a problem

PR #3324, opened 2026-05-19
feat(kiloclaw): expire organization trial compute safely

jeanduplessis:
## Summary
Organization-managed KiloClaw instances now follow the approved hard-expiry lifecycle when a trial loses entitlement.

### Why this change is needed
Organizations can outlive the old UI-only trial boundary while still retaining managed compute. Enforcement needs one shared entitlement rule, fail-closed creation/revival paths, recoverable suspension before deletion, and operator visibility for rollout risk.

### How this is addressed
- Adds a shared organization entitlement policy package and applies it across web redirects, mutation guards, worker provisioning, and billing lifecycle decisions.
- Suspends hard-expired unenitled organization compute, sends organization-specific lifecycle notices, grants a fresh seven-day destruction grace, and recovers access safely if entitlement returns.
- Extends warning/destruction sweeps, async resume recovery, platform stop/start reasons, and telemetry so organization lifecycle outcomes stay separate from personal KiloClaw flows.
- Adds a read-only rollout inventory command that reports would-suspend and exempt org-managed rows before enforcement rollout.
- Documents the hard-expiry contract in the team seat-billing and KiloClaw billing specs.

## Human Verification
- Validated entitlement policy behavior with focused package and web tests for trial status, paid-seat states, and exemption precedence.
- Validated organization provisioning, suspension, warning, recovery, destruction, and inventory paths with focused billing-worker and web test suites.
- Validated platform request compatibility for org entitlement denials plus organization stop/start lifecycle reasons.

### Human Reviewer Flags
- Policy/spec change: hard-expired unenitled organization trials now become enforcement-eligible after the approved three-day soft-expiry window.
- Architecture change: entitlement policy moved into `@kilocode/organization-entitlement` so web, worker, and reporting paths share one classifier.
- Rollout trade-off: first enforcement grants a fresh seven-day destruction grace instead of backdating deletion from the original trial deadline.

### Code Reviewer Agent
<details>

- Focus worker review on `organization_trial_expiry` queue paging, entitlement rechecks before warning/destruction, and recovery state reset before async restart.
- Notification fanout intentionally deduplicates associated users with owner/billing-manager recipients and keeps organization CTAs separate from personal `/claw` paths.
- Inventory command is read-only and emits IDs/boundaries rather than organization names or email addresses.

</details>


services/kiloclaw-billing/src/lifecycle.ts
```diff
@@ -2627,6 +2812,437 @@ export async function processTrialExpiryPage(
...
+  }
+
+  if (!row.organization_id) {
+    logSkippedSubscriptionRow('Skipping organization trial expiry for personal row', row, {
+      reason: 'personal_instance',
+    });
+    return;
+  }
+
+  const entitlement = classifyOrganizationEntitlement({
```
pandemicsyn, line 3170: some robot flagged items when i grilled it about the lifecycle flow:


>processOrganizationTrialExpiryPage fetches up to 50 candidate rows, then processOrganizationTrialExpiryRow decides suspension from that snapshot and performs stopInstanceForEnforcement plus an unconditional subscription update. If entitlement returns while the page is being processed, the stale row can still be stopped and moved to canceled with a destruction deadline. That creates avoidable false suspensions and lifecycle emails for organizations that are already entitled again.

>Perform a fresh entitlement lookup just before stop/suspend, or make the suspension update conditional on a current entitlement check in the same database decision path. The worker should skip rather than mutate when the organization no longer qualifies at execution time.

services/kiloclaw-billing/src/lifecycle.ts
```diff
@@ -2627,6 +2812,437 @@ export async function processTrialExpiryPage(
...
+      organizationId: row.organization_id,
+    });
+    return;
+  }
+
+  await stopInstanceForEnforcement(env, context, row, 'organization_trial_expiry');
+
+  const destructionDeadline = new Date(Date.now() + DESTRUCTION_GRACE_DAYS * MS_PER_DAY);
+  const before = await getSubscriptionById(database, row.id);
+  const [updated] = await database
```
pandemicsyn, line 3197: >The new organization suspension path persists status: 'canceled', suspended_at, and destruction_deadline before sending suspension emails at services/kiloclaw-billing/src/lifecycle.ts:3108. If delivery throws, the page-level catch records an error, but the row no longer matches the active-unsuspended expiry candidate query, so the notification cannot be retried on the next run. 

> The same pattern exists for organization destroyed emails: destruction state is finalized and the deadline cleared before sendOrganizationKiloClawLifecycleNotifications at services/kiloclaw-billing/src/lifecycle.ts:3592, leaving no retry path after a send failure. This violates the newly documented email retry contract for organization lifecycle notifications.

services/kiloclaw-billing/src/lifecycle.ts
```diff
@@ -2892,15 +3520,23 @@ async function runInstanceDestructionSweep(
...
-        logSkippedSubscriptionRow(
-          'Skipping instance destruction for organization-managed row',
-          row,
-          {
-            reason: 'organization_managed',
-            organizationId: row.organization_id,
-          }
-        );
-        continue;
+        const entitlement = classifyOrganizationLifecycleEntitlement(row, new Date(now));
```
pandemicsyn: This seems pretty edge case:

>runInstanceDestructionSweep classifies organization entitlement from the batch row selected earlier at services/kiloclaw-billing/src/lifecycle.ts:3450, then immediately proceeds to destroyInstanceForEnforcement at services/kiloclaw-billing/src/lifecycle.ts:3546. If seats are purchased or an exemption is applied after candidate selection but before that row is processed, the code never re-reads current entitlement and can permanently destroy an instance that should instead be recovered. This conflicts with the new documented requirement to re-evaluate organization entitlement before destruction.

pandemicsyn (approved):
pre-approving, but in going back and forth with the 🤖 about this pr, i think we did stumble on a couple of issues in kiloclaw-billing/src/lifecycle.ts

services/kiloclaw-billing/src/lifecycle.ts
jeanduplessis: Fixed in f187e312a. The org trial expiry row now gets reloaded by subscription id immediately before enforcement, then entitlement is classified from fresh organization and seat-purchase state. If entitlement returned or the row is no longer active/live/unsuspended, the worker skips without stopping or mutating.

PR #3325, opened 2026-05-19
feat(gateway): route and persist model experiments

markijbema:
## Summary

Implements backend model-experiment routing and attribution persistence for preview/provider-funded traffic.

- Adds experiment routing for dedicated preview public IDs. `getProvider` checks a Redis-backed experimented-public-ID membership set, then loads the selected experiment, variants, and current variant versions from Postgres for preview traffic only.
- Routes active experiments directly to the configured partner upstream via the shared `buildDirectProvider` path. Paused experiments return local model-unavailable behavior instead of falling through to default providers.
- Keeps experiment API keys encrypted at rest. The chosen variant key is decrypted per request, immediately before building the direct upstream provider.
- Captures the canonical transformed upstream request body for experimented requests, stores prompt bodies content-addressed in R2, and persists experiment attribution after the microdollar usage row is written.
- Preserves billing correctness by making experiment attribution best-effort: prompt storage or attribution failures are reported but do not roll back usage accounting.
- Requires `BYOK_ENCRYPTION_KEY` at config load and adds `R2_EXPERIMENT_PROMPTS_BUCKET_NAME` for prompt blob storage.

## Verification

- Sent two-variant experiment requests against a local instance and confirmed `model_experiment_request` rows land 1:1 with `microdollar_usage`.
- Confirmed `client_request_id` is populated from `x-kilo-request`.
- Confirmed experiment traffic uses the configured partner `upstream.base_url` rather than OpenRouter/Vercel.
- Hot-swapped a variant version and confirmed subsequent routing resolves the new `variant_version_id` from Postgres while older attribution rows still point at the prior version.
- Paused an experiment and confirmed requests to that public ID return local model-unavailable behavior.
- Ran targeted tests for experiment picker and persistence.

## Visual Changes

N/A

## Reviewer Notes

- Redis is only used as a fast membership pre-check for experimented public IDs. Admin mutations recompute `EXPERIMENTED_PUBLIC_IDS_REDIS_KEY`; if Redis is missing, corrupt, or unavailable, gateway treats that as “no preview IDs.”
- Full experiment routing data is loaded from Postgres for preview traffic, matching the custom LLM pattern and avoiding a second Redis payload cache.
- `persistExperimentAttribution` is intentionally best-effort and runs after the microdollar usage write. Billing remains the source of truth.
- Prompt bodies are stored in R2 by sha256 hash. When prompt storage fails, the attribution row still lands with `__failed__`.
- Recommended review order: routing entry points (`route.ts`, `get-provider.ts`); variant selection and fail-closed behavior (`pick-variant.ts`, `pick-variant.types.ts`); billing and policy guarantees (`processUsage.ts`, `llm-proxy-helpers.ts`); attribution persistence (`persist.ts`, `experiment-prompts.ts`); tests (`pick-variant.test.ts`, `persist.test.ts`, related provider/request tests).


markijbema:
Addressed review feedback in the latest pushes:

- **#1 (billing not zeroed) — fixed.** Added `provider_funded` flag on `MicrodollarUsageContext`; `processTokenData` now zeroes `cost_mUsd` / `cacheDiscount_mUsd` while preserving `market_cost`.
- **#2 (org data-collection bypass) — fixed.** Refuse experiment routing with `dataCollectionRequiredResponse()` when the org has `data_collection: 'deny'`. Anonymous and non-org users continue to route normally.
- **#3 (warm cache plaintext keys) — fixed.** Cache now stores ciphertext exclusively. Decryption + the `BYOK_ENCRYPTION_KEY` missing-key check moved into the per-pick path so warm caches no longer serve plaintext after key rotation/removal. Also: only the chosen variant's key is decrypted per request.
- **#4 (stale membership cache) — deferred to follow-up.** Documented in `.plans/experimental-models-1.md` with the agreed fix (TTL on the membership key + rely on existing DB fallback).
- **#5 (PII erasure) — out of scope per plan.** Plan defines an experiment-specific opt-in retention policy distinct from the default `microdollar_usage_metadata` soft-delete; the locking test lives in Phase 6 alongside the model-specific opt-in copy that's a prerequisite to the first real partner experiment.
- **#6 (UTF-16 cap) — fixed.** New `truncateToUtf8Bytes` walks back to a UTF-8 codepoint boundary; `buildExperimentPromptCapture` now measures with `Buffer.byteLength` and truncates by bytes.
- **#7 (tests) — added.** Two new test files (27 cases) cover the picker, fail-closed paths (paused → not-found, missing allocation subject → unavailable, paused via cache invalidation), hot-swap preserving the slot but serving the new RC, weighted distribution, historical attribution via old version rows, and the UTF-8 truncation edge cases. Full `pnpm --filter web test` is now 373 suites / 5658 passing.

Each fix is a separate commit for review.

markijbema:
Addressed the request-level data-collection follow-up:

- The previous fix #2 only blocked org-level `settings.data_collection: 'deny'`. A request-level opt-out via `provider.data_collection: 'deny'` or `provider.zdr: true` would still route to the experiment and capture the prompt to R2, violating the caller's stated intent.
- New unconditional check right after `getProvider` returns: `if (experiment && !isFreePromptTrainingAllowed(body.provider)) return dataCollectionRequiredResponse()`. Covers anonymous, BYOK, and non-org authenticated users — anyone who set the field on the request body.
- Reuses the existing `isFreePromptTrainingAllowed` predicate (already used by the kilo-exclusive free-model gate) so the opt-out semantics are consistent across the gateway.
- Added a 7-case unit test for the predicate so future regressions can't silently allow capture for opted-out callers.

Commit: 33b3baf65

markijbema:
Extracted the buildDirectProvider refactor as a standalone prep PR: #3330.

Either order of merge works:
- If #3330 lands first, this PR will deduplicate the commit on rebase via git's patch-id matching.
- If this PR lands first, #3330 has nothing to do (its diff is already in main); the chore can just be closed.

Refactor is pure and well-tested by the existing custom_llm coverage (83 tests). No behavioral change to the live kilo-internal/... path.

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -469,7 +496,7 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
   if (!isAnonymousContext(user) && !bypassAccessCheck) {
     const { balance, settings, plan } = await balanceAndSettingsPromise;
 
-    if (balance <= 0 && !isFreeModel(originalModelIdLowerCased) && !userByok) {
+    if (!skipBalanceCheck && balance <= 0 && !isFreeModel(originalModelIdLowerCased) && !userByok) {
```
Comment (markijbema) [inline apps/web/src/app/api/openrouter/[...path]/route.ts]: balance <= 0 should still be the first conndition

apps/web/src/lib/ai-gateway/experiments/persist.ts
```diff
@@ -0,0 +1,214 @@
...
+/**
+ * Build a bounded capture of the canonical post-`transformRequest` upstream
+ * request body, split into the leading system message and the remainder.
+ *
+ * - chat_completions: extracts a single leading `role === 'system'` message
+ *   into `systemPromptContent` and serializes the remainder
+ *   (`{...body, messages: nonSystem}`) into `requestBodyContent`.
+ * - messages / responses: leaves the full body in `requestBodyContent` and
+ *   leaves `systemPromptContent` null. The provider-specific system field
+ *   (e.g. anthropic `body.system`) is captured as part of the body blob.
```
markijbema: are we also saving what type of request it was?

markijbema: should we even normalize like this? or should we log a more raw request?

apps/web/src/lib/ai-gateway/experiments/pick-variant.ts
```diff
@@ -0,0 +1,487 @@
...
+      } else {
+        return ids.includes(publicId);
+      }
+    }
+  } catch {
+    // captureException already invoked by the redis helper
+  }
+
+  try {
+    const rows = await db
```
markijbema: we fall through to db but dont cache; either we update cache here or we dont fall through, this is silly

apps/web/src/lib/ai-gateway/experiments/pick-variant.ts
```diff
@@ -0,0 +1,487 @@
...
+  if (totalWeight <= 0) {
+    captureMessage('Experiment total weight is non-positive', {
+      level: 'error',
+      tags: { source: 'model-experiments' },
+      extra: { experimentId: exp.experimentId },
+    });
+    return { status: 'unavailable' };
+  }
+
+  const seed = `model_exp_${exp.experimentId}_${subjectPick.subject}_${subjectPick.value}`;
```
markijbema: is subjectPick the user here?

apps/web/src/lib/ai-gateway/experiments/pick-variant.ts
```diff
@@ -0,0 +1,487 @@
...
+        experimentId: exp.experimentId,
+        variantId: v.variantId,
+        variantVersionId: v.variantVersionId,
+        upstream: { ...v.upstream, api_key: apiKey },
+        allocationSubject: subjectPick.subject,
+      };
+    }
+  }
+
+  // Unreachable: getRandomNumber returns [0, totalWeight). Fail closed.
```
markijbema: if this is unreachable why is there code here

apps/web/src/lib/ai-gateway/experiments/pick-variant.ts
```diff
@@ -0,0 +1,460 @@
...
+/** TTL for the per-public-id experiment cache. The cache holds CIPHERTEXT
+ *  (encrypted api keys + the upstream blob); decryption happens per-pick
+ *  inside `pickModelExperimentVariant`, so this TTL only governs how stale
+ *  the variant set / current-version-per-variant resolution can be, not
+ *  how long plaintext keys can live in Redis. */
+const PER_PUBLIC_ID_CACHE_TTL_SECONDS = 600;
+
+let warnedMissingEncryptionKey = false;
+
+function warnMissingEncryptionKeyOnce() {
```
markijbema: remove this code path, making alternative pr ensuring this key is always available

apps/web/src/lib/ai-gateway/experiments/pick-variant.ts
```diff
@@ -0,0 +1,460 @@
...
+ * Reads from a JSON-encoded array in Redis. On Redis error, falls back to a
+ * single Postgres query for that public id. If both fail, returns
+ * `unavailable` so the caller can fail closed instead of routing as
+ * unexperimented traffic.
+ */
+export async function isPublicIdExperimented(
+  publicId: string
+): Promise<true | false | 'unavailable'> {
+  try {
+    const cached = await redisGet(EXPERIMENTED_PUBLIC_IDS_REDIS_KEY);
```
markijbema: should we use createCachedFetch

apps/web/src/lib/ai-gateway/providers/openrouter/types.test.ts
```diff
@@ -0,0 +1,37 @@
+import { describe, expect, it } from '@jest/globals';
```
RSO: This feels unrelated to the PR? You don't touch the file, but you do add tests for it.

apps/web/src/lib/organizations/organization-models.ts
```diff
@@ -40,6 +41,7 @@ export async function getAvailableModelsForOrganization(
 
   filteredModels.push(...(await getDirectByokModelsForOrganization(organizationId)));
   filteredModels.push(...(await listAvailableCustomLlms(organizationId)));
+  filteredModels.push(...(await listAvailableExperimentModels()));
```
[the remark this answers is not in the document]
RSO: @markijbema this sounds similar to the concern I linked to you on Slack. Are you sure this is "resolved"?

apps/web/src/lib/redis-keys.ts
```diff
@@ -39,7 +39,7 @@ export const botIdentityRedisKey = (platform: string, teamId: string, userId: st
 /**
  * Set of public_model_ids that have a routing-relevant model_experiment row
  * (status IN 'active' | 'paused'). Used by `getProvider` as a fast pre-check
- * before fetching the per-public-id experiment payload.
+ * before loading the experiment routing payload from Postgres.
```
RSO: Such an interesting edit here

RSO (approved):

apps/web/src/lib/redis-keys.ts
markijbema: i think there were more changes here earlier, and i reduced the diff here

PR #3332, opened 2026-05-19
feat(web): connect collab OAuth flow

RSO:
## Summary

- Connect the collab authorization wizard to real OAuth entry points for Slack, Discord, Linear, GitHub, and GitLab.
- Add signed relative return paths so OAuth callbacks can resume the multi-service collab setup flow safely.
- Normalize GitHub callback redirects through the configured app URL so callback host aliases such as 0.0.0.0 do not leak into app redirects.

## Verification

- Not manually verified end-to-end; the GitHub App install callback requires the external GitHub App flow and local app configuration.

## Visual Changes

N/A

## Reviewer Notes

- Review return-path handling across OAuth callbacks, especially error-path query parameters.
- GitHub App install redirects now use APP_URL as the redirect base instead of preserving the incoming callback host.

### DIFF
--- a/apps/web/src/app/api/integrations/github/callback/route.ts
+++ b/apps/web/src/app/api/integrations/github/callback/route.ts
@@ -29,6 +29,9 @@ import { isOrganizationMember } from '@/lib/organizations/organizations';
 import { PLATFORM } from '@/lib/integrations/core/constants';
 import { botPlatforms } from '@/lib/bot/platforms';
 
+const appendQueryParam = (path: string, queryParam: string): string =>
+  `${path}${path.includes('?') ? '&' : '?'}${queryParam}`;
+
 function htmlPage(title: string, message: string, status = 200): Response {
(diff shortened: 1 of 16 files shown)


apps/web/src/app/api/integrations/github/callback/route.ts
```diff
@@ -245,22 +254,24 @@ export async function GET(request: NextRequest) {
         });
 
         // Redirect back to integrations page with pending approval status
-        const redirectPath =
-          owner.type === 'org'
+        const redirectPath = returnTo
+          ? appendQueryParam(returnTo, 'github_pending_approval=true')
+          : owner.type === 'org'
```
markijbema: I dont' really like that we add the queory param thrice. why not do the null check and the ternary here, and then after that add the queryparam unconditionally?

markijbema (approved):

PR #3347, opened 2026-05-19
fix(usage): resolve user filter labels

catrielmuller:
## Summary
- Resolve all user IDs returned by the usage filter suggestion dropdown before rendering labels, so owners can select users by email/name instead of raw IDs.
- Extend the org-user resolver to map legacy `oauth/<provider>:<account>` usage IDs through linked auth providers while preserving the existing org-membership PII guard.

## Verification
N/A (not manually verified).

## Visual Changes
N/A

## Reviewer Notes
Unresolvable IDs still fall back to the raw usage key if they do not map to a current org member or linked legacy auth provider.

### DIFF
--- a/apps/web/src/routers/usage-analytics-router.ts
+++ b/apps/web/src/routers/usage-analytics-router.ts
@@ -1,6 +1,6 @@
 import { TRPCError } from '@trpc/server';
 import * as z from 'zod';
-import { and, eq, inArray } from 'drizzle-orm';
+import { and, eq, inArray, or } from 'drizzle-orm';
 import { baseProcedure, createTRPCRouter, type TRPCContext } from '@/lib/trpc/init';
 import { readDb } from '@/lib/drizzle';
 import { getEnvVariable } from '@/lib/dotenvx';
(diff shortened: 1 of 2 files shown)


apps/web/src/routers/usage-analytics-router.ts
```diff
@@ -1067,12 +1099,77 @@ export const usageAnalyticsRouter = createTRPCRouter({
...
+            lookup.id,
+          ])
+        );
+        const legacyConditions = legacyLookups.map(lookup =>
+          and(
+            eq(user_auth_provider.provider, lookup.parsed.provider),
+            eq(user_auth_provider.provider_account_id, lookup.parsed.providerAccountId)
+          )
+        );
+        const legacyWhere = or(...legacyConditions);
```
jrf0110, line 1133: This makes it such that the query can change/grow in size based on the data. That's not so great for observability (e.g. this query has n distinct permutations) or caching. Is it not possible to do a `user_id in $1` style query  here?

apps/web/src/routers/usage-analytics-router.ts
jrf0110, line 1074: Why is this function taking a list of user ID's as input rather than just organization id? Couldn't we just query microdollar_usage for the list of users who had activity in requested period? Or you could go straight to memberships table?

jrf0110 (approved):

apps/web/src/routers/usage-analytics-router.ts
catrielmuller: we don't use microdollar_usage from pg any more for this section.

PR #3348, opened 2026-05-19
feat(kiloclaw): add Composio calendar onboarding

pandemicsyn:
## Summary

- Add Composio-managed Google Calendar onboarding for personal and organization KiloClaw instances, including popup-based Connect Link approval, callback verification, status endpoints, and OpenClaw instance credential wiring.
- Add encrypted owner-scoped Composio identity storage so managed identities are reused per Kilo user or per Kilo user+organization context across new instances after destroy/create cycles.
- Track per-instance Composio source directly on `kiloclaw_instances` and store the verified Google Calendar connected account on the owner-scoped Composio identity, avoiding a separate instance config table.
- Rehydrate previously connected managed Composio CLI credentials during provision/reprovision so new OpenClaw instances keep using the same Composio org/key, while BYO Composio credentials remain an explicit manual override.
- Use the same agent-owned Composio user/org/project context as the setup-script/CLI path for Connect Link and status calls, resolving the consumer project before session creation and refreshing older managed identities that are missing consumer project context.
- Update onboarding and settings UI to show Kilo-managed Composio state, keep OAuth approval in a correlated popup flow, and preserve the advanced manual credential path.

## Verification

- Manually verified the Composio popup approval flow returns to KiloClaw onboarding and reflects the connected state.
- Manually verified the Composio settings page distinguishes Kilo-managed Composio from manual credentials.
- [ ] Add before/after screenshots for the new tools onboarding step and settings indicator.

## Visual Changes

| Before | After |
| ------ | ----- |
| Existing onboarding flow without a dedicated Composio tools connection step. | New Google Calendar tools step with popup approval, connected state, and BYO credential guidance. |
| Settings only reflected raw Composio secrets. | Settings shows configured/Kilo-managed state on the Composio row when managed Composio is active. |

## Reviewer Notes

- Manual Composio credentials intentionally take precedence over Kilo-managed credentials for the current OpenClaw instance; delayed managed callbacks must not overwrite manual setup.
- Managed Composio identities are owner-scoped and lifecycle-tracked (`pending` → `active` → `revoked`); only active identities with stored credentials and resolved consumer-project context are used for rehydration.
- Composio control-plane calls use the managed identity’s user/org/project context and must not log secrets, Connect Links, or OAuth material.
- The popup callback flow is correlated by attempt id and escapes inline payloads before embedding them in the callback page.
- After rebasing onto latest `main`, incoming `0135`–`0137` DB migrations remain unchanged and the generated Composio schema lands as first-shipment migration `0138_brief_valkyrie`; no Composio behavior in this PR has shipped yet.


packages/db/src/migrations/0138_rapid_blob.sql
```diff
@@ -0,0 +1,10 @@
+DROP INDEX "UQ_kiloclaw_composio_identities_active_user";--> statement-breakpoint
+DROP INDEX "UQ_kiloclaw_composio_identities_active_org_user";--> statement-breakpoint
+ALTER TABLE "kiloclaw_composio_identities" ALTER COLUMN "composio_agent_key_encrypted" DROP NOT NULL;--> statement-breakpoint
+ALTER TABLE "kiloclaw_composio_identities" ALTER COLUMN "composio_user_api_key_encrypted" DROP NOT NULL;--> statement-breakpoint
+ALTER TABLE "kiloclaw_composio_identities" ALTER COLUMN "composio_org_id" DROP NOT NULL;--> statement-breakpoint
+ALTER TABLE "kiloclaw_composio_identities" ADD COLUMN "status" text DEFAULT 'pending' NOT NULL;--> statement-breakpoint
```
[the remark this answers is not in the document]
pandemicsyn: This has only been live like a day and it wasn't advertised (burie din settings) and pending would be desirable for these.

packages/db/src/migrations/0138_brief_valkyrie.sql
```diff
@@ -0,0 +1,32 @@
...
+CREATE UNIQUE INDEX "UQ_kiloclaw_composio_identities_current_user" ON "kiloclaw_composio_identities" USING btree ("user_id") WHERE "kiloclaw_composio_identities"."owner_type" = 'user' AND "kiloclaw_composio_identities"."revoked_at" IS NULL;--> statement-breakpoint
+CREATE UNIQUE INDEX "UQ_kiloclaw_composio_identities_current_org_user" ON "kiloclaw_composio_identities" USING btree ("organization_id","user_id") WHERE "kiloclaw_composio_identities"."owner_type" = 'organization_user' AND "kiloclaw_composio_identities"."revoked_at" IS NULL;--> statement-breakpoint
+CREATE INDEX "IDX_kiloclaw_composio_identities_user" ON "kiloclaw_composio_identities" USING btree ("user_id");--> statement-breakpoint
+CREATE INDEX "IDX_kiloclaw_composio_identities_organization" ON "kiloclaw_composio_identities" USING btree ("organization_id");--> statement-breakpoint
+ALTER TABLE "kiloclaw_instances" ADD CONSTRAINT "kiloclaw_instances_composio_config_source_check" CHECK ("kiloclaw_instances"."composio_config_source" IS NULL OR "kiloclaw_instances"."composio_config_source" IN ('managed', 'manual'));
```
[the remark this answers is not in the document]
pandemicsyn: Leaving this as generated SQL for now. The repo does not currently use the NOT VALID / VALIDATE CONSTRAINT pattern for added CHECK constraints afaik, including prior migrations on kiloclaw_instances, and this migration is adding a brand-new nullable column whose existing ~80k rows all satisfy the constraint as NULL. The validation scan should therefore be simple and bounded,

apps/web/src/hooks/useKiloClaw.ts
```diff
@@ -17,6 +17,16 @@ export function useKiloClawConfig() {
   return useQuery(trpc.kiloclaw.getConfig.queryOptions());
 }
 
+export function useKiloClawComposioOnboardingStatus(enabled = true) {
+  const trpc = useTRPC();
+  return useQuery(
+    trpc.kiloclaw.getComposioOnboardingStatus.queryOptions(undefined, {
+      enabled,
+      refetchInterval: enabled ? 15_000 : false,
```
St0rmz1: This query is enabled for the entire `ClawOnboardingFlow` mount, so the 15s `refetchInterval` keeps running through the email and interests steps too. Once an instance and an active managed identity exist, each `getComposioOnboardingStatus` poll makes a worker `getConfig` call plus a Composio `listComposioConnectedAccounts` call, so this is steady third-party API traffic on steps where the tools status is not used.

Consider gating the interval on the tools step being active, e.g. only set `refetchInterval` when `renderStep === 'tools'` or `composioPopupPending` is true. The popup flow already refetches on focus, visibilitychange, and post-message, so the background interval is mostly redundant outside that step.

apps/web/src/routers/kiloclaw-router.ts
```diff
@@ -849,6 +857,32 @@ const patchBotIdentitySchema = z.object({
...
+  popup: z.boolean().optional(),
+  attemptId: z.string().min(1).max(100).optional(),
+});
+
+function composioSecretsPatchSource(
+  secrets: Record<string, string | null>
+): 'upsert_manual' | 'clear' | 'none' {
+  const touchedEntries = Object.entries(secrets).filter(
+    ([key]) => key === 'composioUserApiKey' || key === 'composioOrg'
+  );
```
St0rmz1: The `clear` branch requires both `composioUserApiKey` and `composioOrg` to be present and null. If a patch sends only one of the two as `null` (partial clear), `touchedEntries.length` is 1, this check is skipped, the `some(value !== null)` check is also false, and the function falls through to `'none'`. So a partial clear leaves `composio_config_source` untouched, while a partial set still returns `'upsert_manual'`. That asymmetry can leave an instance marked `manual` after its manual credentials were partially removed, which then blocks managed connect via the `CONFLICT` guard in `createComposioGoogleCalendarLink`.

If the API contract guarantees both keys are always cleared together, this is fine but worth a comment documenting that assumption. Otherwise, handle a partial clear explicitly.

apps/web/src/lib/kiloclaw/composio-onboarding.ts
```diff
@@ -0,0 +1,274 @@
...
+}): string {
+  const url = new URL('/api/integrations/composio/callback', APP_URL);
+  url.searchParams.set('returnTo', params.returnTo);
+  if (params.organizationId) url.searchParams.set('organizationId', params.organizationId);
+  if (params.popup) url.searchParams.set('popup', '1');
+  if (params.attemptId) url.searchParams.set('attemptId', params.attemptId);
+  return url.toString();
+}
+
+export async function applyManagedComposioCredentials(params: {
```
St0rmz1: This function is exported but has no callers anywhere in the branch (`git grep` finds only this definition). The live credential-application paths are `buildComposioProvisionSecrets` and `completeManagedComposioGoogleCalendarConnection`. Recommend removing it to avoid confusion about which path is authoritative, or add a comment if it is a deliberate seam for an upcoming change.

apps/web/src/lib/kiloclaw/composio-identities.ts
```diff
@@ -0,0 +1,263 @@
...
+
+export async function getActiveManagedComposioIdentity(
+  scope: ComposioOwnerScope
+): Promise<DecryptedComposioIdentity | null> {
+  const row = await findActiveComposioIdentity(scope);
+  return row ? decryptComposioIdentity(row) : null;
+}
+
+export async function ensureManagedComposioIdentity(
+  scope: ComposioOwnerScope
```
St0rmz1, line 219: This module has no test file. The callback route, `composio-client`, and `composio-onboarding` all have tests, but `composio-identities.ts` does not, and the onboarding tests mock `getActiveManagedComposioIdentity`, so `ensureManagedComposioIdentity` is never exercised. Its logic is non-trivial: lock acquisition, the `needsComposioIdentityRefresh` refresh path, pending-reservation reuse, idempotency-key recovery via `signupComposioAgentIdentity({ idempotencyKey: pending.id })`, and the two-step pending → active write. This is the module that decides whether a managed identity is reused or re-provisioned, so a regression here means duplicate Composio signups or stale project context. Recommend adding unit tests covering the active fast path, the refresh path, and the reservation-reuse path.

apps/web/src/lib/kiloclaw/composio-onboarding.ts
```diff
@@ -0,0 +1,274 @@
...
+  const accounts = await listComposioConnectedAccounts({
+    auth,
+    userId: identity.consumerUserId,
+  });
+  const connected = accounts.some(
+    account => account.id === params.connectedAccountId && account.status === 'ACTIVE'
+  );
+  if (!connected) return false;
+
+  const sandboxConfigSource = await getComposioInstanceConfigSource(params.instance.id);
```
St0rmz1, line 169: This `manual` check and the `patchSecrets` write below it are separate statements with the worker call in between, so the guard is not atomic. If a user saves manual credentials in the window between this read and `patchSecrets`, the managed callback can still overwrite the just-saved manual credentials, which contradicts the stated invariant that delayed managed callbacks must not overwrite manual setup. The window requires a concurrent manual save during the OAuth callback, so it is unlikely and the next status check self-corrects, hence low severity. A short comment acknowledging the residual race, or moving the source check to be re-read closer to the write, would help future readers.

St0rmz1 (approved):

PR #3356, opened 2026-05-20
feat(admin): add code review date interval

alex-alecu:
## Why

Admins could only look at code review data in fixed day ranges. They need a way to check a smaller or exact time period when looking into a problem.

## What changed

The code review dashboard still starts on Last 7 days, and it now has a Date interval option with start and end times. The custom interval starts with the same seven-day window, shows a clear message when the times are missing or out of order, and blocks refresh and export until the range is valid. Code review data requests now accept time-based ranges, keep the end time outside the result set, and reject broken ranges before they run. CSV download names also stay clean when those time values are included.

## How to test

1. Open the admin Code Review Telemetry page and confirm Date Range shows Last 7 days and Date interval.
2. Select Date interval and confirm both time fields are filled with a seven-day window.
3. Set the start time after the end time and confirm the warning appears while Refresh and Export CSV are disabled.
4. Enter a valid time range and confirm the warning clears and the actions become available.
5. Export rows for a range with data and confirm the downloaded CSV name uses safe date and time text.

### DIFF
--- a/apps/web/src/app/admin/code-reviews/page.tsx
+++ b/apps/web/src/app/admin/code-reviews/page.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useState, useEffect, useCallback, useMemo } from 'react';
+import { useState, useCallback, useMemo } from 'react';
 import { format, subDays } from 'date-fns';
 import AdminPage from '@/app/admin/components/AdminPage';
 import { BreadcrumbItem, BreadcrumbPage } from '@/components/ui/breadcrumb';
@@ -169,22 +169,17 @@ export default function CodeReviewsPage() {
   const isIntervalDraftInvalid = rangeType === 'interval' && intervalValidation.interval === null;
   const { activeInterval, intervalDraft } = dateIntervalState;
   const { startDate, endDate } = activeInterval;
+  const hasPendingIntervalDraft =
+    rangeType === 'interval' &&
+    intervalValidation.interval !== null &&
+    (intervalValidation.interval.startDate !== startDate || intervalValidation.interval.endDate !== endDate);
 
-  useEffect(() => {
-    if (rangeType !== 'interval' || !intervalValidation.interval) return;
-
+  const handleApplyInterval = () => {
     const nextInterval = intervalValidation.interval;
-    setDateIntervalState(current => {
-      if (
-        current.activeInterval.startDate === nextInterval.startDate &&
-        current.activeInterval.endDate === nextInterval.endDate
-      ) {
-        return current;
-      }
+    if (!nextInterval) return;
 
-      return { ...current, activeInterval: nextInterval };
-    });
-  }, [intervalValidation.interval, rangeType]);
+    setDateIntervalState(current => ({ ...current, activeInterval: nextInterval }));
+  };
 
   const handleRangeTypeChange = (nextRangeType: RangeType) => {
     setRangeType(nextRangeType);
@@ -420,6 +415,14 @@ export default function CodeReviewsPage() {
                     aria-invalid={isIntervalDraftInvalid}
                   />
                 </div>
+                <Button
+                  variant="outline"
+                  size="sm"
+                  onClick={handleApplyInterval}
+                  disabled={!hasPendingIntervalDraft}
+                >
+                  Apply interval
+                </Button>
               </div>
               <p
                 id="code-review-date-interval-feedback"
@@ -432,7 +435,9 @@ export default function CodeReviewsPage() {
               >
                 {isIntervalDraftInvalid
                   ? intervalValidation.error
-                  : "Times use this browser's local timezone."}
+                  : hasPendingIntervalDraft
+                    ? "Apply the interval to update telemetry. Times use this browser's local timezone."
+                    : "Times use this browser's local timezone."}
               </p>
             </div>
           )}
--- a/apps/web/src/app/admin/code-reviews/page.tsx
+++ b/apps/web/src/app/admin/code-reviews/page.tsx
@@ -37,7 +37,7 @@ const breadcrumbs = (
   </BreadcrumbItem>
 );
 
-type RangeType = '7d' | '30d' | '90d';
+type RangeType = '7d' | 'interval';
 type OwnershipType = 'all' | 'personal' | 'organization';
 type RetryAccountingModeType = 'final_outcome' | 'all_attempts';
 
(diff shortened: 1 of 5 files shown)


apps/web/src/app/admin/code-reviews/page.tsx
```diff
@@ -78,22 +162,47 @@ export default function CodeReviewsPage() {
...
+    const nextInterval = intervalValidation.interval;
+    setDateIntervalState(current => {
+      if (
+        current.activeInterval.startDate === nextInterval.startDate &&
+        current.activeInterval.endDate === nextInterval.endDate
+      ) {
+        return current;
+      }
+
+      return { ...current, activeInterval: nextInterval };
```
alex-alecu: As the user types in the datetime-local inputs, every transient valid value writes through to `activeInterval` here, which changes `filterParams` and triggers refetches across all 7 admin queries (overview/daily/performance/waitTime/cancellation/error/segmentation) on every keystroke that produces a parseable datetime — there is no debounce or `staleTime`. The PR description says refresh and export are blocked until the range is valid, but in practice the queries refetch automatically while the admin is still picking a time. Was that intentional, or should `activeInterval` only update on an explicit Refresh / Apply?

apps/web/src/app/admin/code-reviews/page.tsx
```diff
@@ -78,22 +162,47 @@ export default function CodeReviewsPage() {
...
+    setRangeType(nextRangeType);
+    if (nextRangeType !== '7d') return;
 
-    const daysMap: Record<RangeType, number> = {
-      '7d': 7,
-      '30d': 30,
-      '90d': 90,
-    };
+    const nextPresetInterval = createTrailingSevenDayIntervalState().activeInterval;
+    setDateIntervalState(current => ({ ...current, activeInterval: nextPresetInterval }));
```
alex-alecu: Switching to `7d` resets `activeInterval` but leaves `intervalDraft` pointing at the old datetime-local values. When the user toggles back to `interval`, the effect at line 173 sees the still-valid stale draft and immediately overwrites the 7d preset with those old timestamps. Should this also reset `intervalDraft` to match the new preset so toggling back does not silently snap the dashboard back to old times?

apps/web/src/app/admin/code-reviews/page.tsx
```diff
@@ -53,10 +53,94 @@ type SelectedOrg = {
...
+  { value: 'interval', label: 'Date interval' },
+] satisfies { value: RangeType; label: string }[];
+
+function toDatetimeLocalInput(date: Date): string {
+  return format(date, "yyyy-MM-dd'T'HH:mm");
+}
+
+function roundToDatetimeLocalMinute(date: Date): Date {
+  const minuteDate = new Date(date);
+  minuteDate.setSeconds(0, 0);
```
alex-alecu, line 90: Could this exclusive preset end use the actual current instant or a rounded-up minute? Rounding down here means `Last 7 days` omits reviews created during the current partial minute.

apps/web/src/app/admin/code-reviews/page.tsx
```diff
@@ -53,10 +53,94 @@ type SelectedOrg = {
...
+function formatDateIntervalLabel(value: string): string {
+  return format(new Date(value), 'yyyy-MM-dd HH:mm');
+}
+
 export default function CodeReviewsPage() {
   const [rangeType, setRangeType] = useState<RangeType>('7d');
-  const [startDate, setStartDate] = useState<string>('');
-  const [endDate, setEndDate] = useState<string>('');
+  const [dateIntervalState, setDateIntervalState] = useState<DateIntervalState>(
+    createTrailingSevenDayIntervalState
```
alex-alecu, line 145: Should Refresh rebuild this `7d` interval before refetching? Because `endDate` is captured here once, `Last 7 days` stops advancing after the page loads and Refresh cannot show newer reviews.

apps/web/src/routers/admin-code-reviews-router.ts
```diff
@@ -77,17 +77,38 @@ const attemptErrorCategoryExpr = sql<string>`CASE
...
+}
+
+const intervalOrderValidation = {
+  message: 'Start date must be before end date',
+  path: ['endDate'],
+};
+
+const FilterSchema = z
+  .object(FilterSchemaShape)
+  .refine(hasAscendingDateInterval, intervalOrderValidation);
```
alex-alecu: Should this validation cap the interval length? The custom range now lets the UI send windows far beyond the old 90-day presets, and these endpoints run aggregate telemetry queries across the whole span.

apps/web/src/app/admin/code-reviews/page.tsx
```diff
@@ -192,7 +304,7 @@ export default function CodeReviewsPage() {
 
   // Handle CSV export
   const handleExport = useCallback(async () => {
-    if (isExporting || !startDate || !endDate) return;
+    if (isExporting || isIntervalDraftInvalid || !startDate || !endDate) return;
```
alex-alecu: Is this empty-date branch still reachable? `activeInterval` is initialized with generated timestamps and every later update keeps both fields populated.

apps/web/src/app/admin/code-reviews/page.tsx
alex-alecu: Fixed in ccc907b4a.

apps/web/src/app/admin/code-reviews/page.tsx
alex-alecu: Fixed in e5c8ac721.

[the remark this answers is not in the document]
alex-alecu: Fixed in e5c8ac721.

apps/web/src/app/admin/code-reviews/page.tsx
alex-alecu: Fixed in 75238a59b.

apps/web/src/app/admin/code-reviews/page.tsx
alex-alecu: Fixed in b46766f3b.

apps/web/src/routers/admin-code-reviews-router.ts
alex-alecu: Fixed in 1b2132fae.

apps/web/src/app/admin/code-reviews/page.tsx
alex-alecu: Fixed in 9fdc70d2c.

apps/web/src/app/admin/code-reviews/page.tsx
```diff
@@ -53,10 +53,97 @@ type SelectedOrg = {
...
+
+function validateDateIntervalDraft(draft: DateIntervalDraft): DateIntervalValidation {
+  const start = parseDatetimeLocal(draft.startInput);
+  const end = parseDatetimeLocal(draft.endInput);
+
+  if (!start || !end) {
+    return { interval: null, error: 'Choose valid start and end times.' };
+  }
+
+  if (start.getTime() >= end.getTime()) {
```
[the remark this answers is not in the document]
alex-alecu: It's fine for admin-only dashboard

markijbema (approved):

PR #3357, opened 2026-05-20
fix(code-reviews): drain stuck pending queues

alex-alecu:
## Why

Pending code reviews could wait a long time after a slot opened. They only woke up when another event arrived, so some reviews stayed parked even though work could continue.

## What changed

This adds a scheduled queue sweep that wakes owners with reviews ready to try again. It reserves capacity safely before the slow work starts, so overlapping events do not take more slots than allowed. Old queued claims can recover again, and old running reviews are only shown as a warning signal, not failed automatically. The admin code review page now shows a live queue health card so support can see waiting work right away.

## How to test

1. Open `/admin/code-reviews` as an admin and confirm `Current queue health` appears above the older telemetry cards.
2. Create or seed a pending review for an owner with free review capacity, then call the cron endpoint with the configured bearer secret and confirm the review moves toward dispatch.
3. Seed a queued review older than five minutes and confirm the scheduled drain can pick it up again.
4. Seed a running review older than ninety minutes and confirm it is shown in queue health but is not marked failed by the drain.


apps/web/src/lib/code-reviews/dispatch/dispatch-pending-code-review-owners.ts
```diff
@@ -0,0 +1,114 @@
...
+  if (!botUserId) {
+    errorExceptInTest('[dispatchPendingCodeReviewOwners] Code-review bot user not found', {
+      organizationId: candidate.id,
+    });
+    captureMessage('Bot user missing for organization code review drain', {
+      level: 'error',
+      tags: { source: 'dispatch-pending-code-review-owners' },
+      extra: { organizationId: candidate.id },
+    });
+    return null;
```
alex-alecu: Can this path recover when bot creation failed during config save? Those org reviews can still be queued by the existing webhook fallback, but this sweep keeps selecting the owner and returning here, so its backlog never drains.

apps/web/src/routers/admin-code-reviews-router.ts
```diff
@@ -156,6 +160,68 @@ function buildBaseConditions(input: FilterInput): SQL[] {
...
+              ${cloud_agent_code_reviews.created_at}
+            ) < ${staleRunningCutoff}
+        )`,
+        owners_with_waiting_reviews_count: sql<number>`COUNT(DISTINCT CASE
+          WHEN ${cloud_agent_code_reviews.owned_by_organization_id} IS NOT NULL
+            THEN CONCAT('org:', ${cloud_agent_code_reviews.owned_by_organization_id}::text)
+          ELSE CONCAT('user:', ${cloud_agent_code_reviews.owned_by_user_id})
+        END) FILTER (WHERE ${waitingOwnerCondition})`,
+      })
+      .from(cloud_agent_code_reviews);
```
alex-alecu, line 252: Can this aggregate query restrict the base scan to the live rows used by these metrics? Without a `WHERE`, loading or refreshing this admin card also scans completed, failed, and cancelled review history, which grows independently of the live queue.

apps/web/src/lib/code-reviews/db/code-reviews.ts
```diff
@@ -161,6 +179,110 @@ export async function getCodeReviewById(reviewId: string): Promise<CloudAgentCod
...
+          ON active_work.owner_type = reconsiderable_work.owner_type
+          AND active_work.owner_id = reconsiderable_work.owner_id
+        LEFT JOIN ${kilocode_users}
+          ON reconsiderable_work.owner_type = 'user'
+          AND ${kilocode_users.id} = reconsiderable_work.owner_id
+      )
+      SELECT owner_type, owner_id
+      FROM capacity_candidates
+      WHERE active_count < capacity_limit
+      ORDER BY oldest_reconsiderable_at ASC, owner_type ASC, owner_id ASC
```
alex-alecu, line 253: Could this minute cron avoid grouping and sorting the whole dispatchable owner set before applying the 100-owner cap? The limit comes after both CTE aggregates and this order, so a large pending or active backlog still has to be evaluated every tick.

apps/web/src/lib/code-reviews/db/code-reviews.ts
```diff
@@ -161,6 +179,110 @@ export async function getCodeReviewById(reviewId: string): Promise<CloudAgentCod
...
+            AND COALESCE(
+              ${cloud_agent_code_reviews.started_at},
+              ${cloud_agent_code_reviews.updated_at},
+              ${cloud_agent_code_reviews.created_at}
+            ) >= ${staleRunningCutoff}
+          )
+          OR (
+            ${cloud_agent_code_reviews.status} = 'queued'
+            AND ${cloud_agent_code_reviews.updated_at} >= ${staleQueuedCutoff}
+          )
```
alex-alecu, line 223: This `active_work` CTE encodes the same "currently consuming a slot" rule as `reservePendingReviewsForDispatch` in `dispatch-pending-reviews.ts` (running with the `started_at`/`updated_at`/`created_at` COALESCE vs `staleRunningCutoff`, plus queued with `updated_at >= staleQueuedCutoff`). If only one side gets edited later (e.g. a new recovery state, a different staleness rule), can the cron pre-filter surface owners the dispatcher immediately rejects, or skip owners it would have served? Worth a single shared SQL helper?

apps/web/src/routers/admin-code-reviews-router.ts
```diff
@@ -156,6 +160,68 @@ function buildBaseConditions(input: FilterInput): SQL[] {
...
+    );
+    const staleQueuedCutoff = staleQueuedCodeReviewCutoffSql();
+    const staleRunningCutoff = staleRunningCodeReviewCutoffSql();
+    const waitingOwnerCondition = sql`(
+      ${cloud_agent_code_reviews.status} = 'pending'
+      OR (
+        ${cloud_agent_code_reviews.status} = 'queued'
+        AND ${cloud_agent_code_reviews.updated_at} < ${staleQueuedCutoff}
+      )
+    )`;
```
alex-alecu, line 219: This `waitingOwnerCondition` ("pending or queued-but-stale") is a third copy of the dispatcher's "what should we be picking up" predicate — the others are in `reservePendingReviewsForDispatch` candidate SELECT and the `reconsiderable_work` CTE in `code-reviews.ts`. Since the card description claims it's a "live dispatch snapshot", could it silently start showing a different population than the drain actually picks up if any of these drifts? Worth sharing one SQL helper across all three call sites?

apps/web/src/lib/code-reviews/dispatch/dispatch-pending-reviews.ts
```diff
@@ -207,12 +261,12 @@ export async function tryDispatchPendingReviews(owner: Owner): Promise<DispatchR
...
       owner,
       dispatched,
-      total: pendingReviews.length,
+      total: reservations.length,
     });
 
     return {
       dispatched,
-      pending: pendingReviews.length - dispatched,
+      pending: reservations.length - dispatched,
```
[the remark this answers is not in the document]
alex-alecu: Fixed in abb6338c2.

apps/web/src/lib/code-reviews/dispatch/dispatch-pending-code-review-owners.ts
```diff
@@ -0,0 +1,114 @@
...
+    });
+    return { status: 'failed' };
+  }
+}
+
+export async function dispatchPendingCodeReviewOwners(): Promise<DispatchPendingCodeReviewOwnersSummary> {
+  const candidates = await listDispatchableCodeReviewOwnerCandidates({ limit: OWNER_SCAN_LIMIT });
+  const limit = pLimit(OWNER_DISPATCH_CONCURRENCY);
+  const outcomes = await Promise.all(
+    candidates.owners.map(candidate => limit(async () => await drainOwner(candidate)))
```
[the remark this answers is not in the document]
alex-alecu: Fixed in 395834f36.

apps/web/src/lib/code-reviews/dispatch/dispatch-pending-code-review-owners.ts
alex-alecu: Fixed in a9cf55d62.

apps/web/src/routers/admin-code-reviews-router.ts
alex-alecu: Fixed in 134c66820.

apps/web/src/lib/code-reviews/db/code-reviews.ts
alex-alecu: Fixed in fee6591f4.

apps/web/src/lib/code-reviews/db/code-reviews.ts
alex-alecu: Fixed in e70fbf8a4.

apps/web/src/routers/admin-code-reviews-router.ts
alex-alecu: Fixed in 262bea245.

alex-alecu:
Manual test passed.

Tested:
- Opened `/admin/code-reviews` as a local admin; the `Current queue health` card rendered and reflected seeded pending, stale queued, and stale-running queue state before and after refresh.
- Called `/api/cron/dispatch-pending-code-reviews` without auth (`401`) and with the configured bearer secret (`200`); the authorized drain considered 2 owners and dispatched 2 reviews.

Verified:
- A pending review and a queued claim older than five minutes were re-reserved and handed to the local code-review worker, which accepted both dispatches.
- Persisted queue state recorded replacement reservation IDs and attempt rows, while a running review older than ninety minutes stayed `running`; queue-health counts dropped pending/stale queued to zero while retaining the stale-running diagnostic.

eshurakov (approved):

PR #3361, opened 2026-05-20
fix(cloud-agent): accept scoped callback tokens

jeanduplessis:
## Summary
- Add resource-scoped HMAC callback-token verification backed by dedicated `CALLBACK_TOKEN_SECRET` material.
- Accept `X-Callback-Token` on security analysis, code-review status, auto-fix PR, and webhook execution callback receivers while preserving legacy shared-secret auth for existing producers.
- Move same-web bot callback HMAC signing and verification to the callback secret with legacy receiver compatibility, and add callback auth regression coverage.

## Verification
N/A - no manual browser or end-to-end callback exercise performed.

## Visual Changes
N/A

## Reviewer Notes
- Deploy-safe phase 1 of 2: cross-service Cloud Agent callback producers keep their legacy headers in this PR. The follow-up producer PR switches them to token-only callback metadata after these receivers land.
- Deploy requires provisioning `CALLBACK_TOKEN_SECRET` for web plus the webhook receiver Secrets Store bindings before token-authenticated callbacks are exercised.
- Bot callback producer and receiver ship in the same web artifact; the receiver accepts both old internal-secret HMACs and new callback-secret HMACs for already-issued bot sessions.


services/code-review-infra/src/code-review-orchestrator.ts
```diff
@@ -35,6 +36,24 @@ function callbackUrlForAttempt(apiUrl: string, reviewId: string, attemptId?: str
   return url.toString();
 }
 
+async function callbackTargetForAttempt(
+  apiUrl: string,
+  reviewId: string,
+  attemptId: string | undefined,
+  internalApiSecret: string
+): Promise<{ url: string; headers: { 'X-Callback-Token': string } }> {
+  return {
```
alex-alecu: Can this retain the legacy header, or otherwise guarantee the web token-verifying receiver deploys first? If this Worker deploys first, callbacks from this path now send only `X-Callback-Token`, while the existing receiver accepts only `X-Internal-Secret`, so terminal updates can return 401 and leave reviews in progress.

jeanduplessis: Two separate PRs to ensure the rollout order of receivers first and then producers using the new token approach.

apps/web/src/lib/config.server.ts
```diff
@@ -39,6 +39,7 @@ export const OPENAI_API_KEY = getEnvVariable('OPENAI_API_KEY');
 export const INCEPTION_API_KEY = getEnvVariable('INCEPTION_API_KEY');
 export const EXA_API_KEY = getEnvVariable('EXA_API_KEY');
 export const INTERNAL_API_SECRET = getEnvVariable('INTERNAL_API_SECRET');
+export const CALLBACK_TOKEN_SECRET = getEnvVariable('CALLBACK_TOKEN_SECRET');
```
markijbema, line 42: Why do we fall back to empty string here, instead of failing by using requireEnv like we do in constants.ts; I would prefer to use that here as well for important secrets unless we have a very good reason for not doing so, as an accidental empty token sounds dangerous

jeanduplessis: Agreed. Will update.

alex-alecu (approved):

PR #3373, opened 2026-05-20
feat(mobile): add Kilo Chat attachments

iscekic:
## Summary

- Adds Kilo Chat attachment support to the mobile app, including native camera/photo-library/file pickers, permission recovery prompts, attachment previews, upload progress/retry/remove controls, and attachment-aware message submission.
- Renders and opens chat attachments on mobile with image thumbnails, full-screen previews, native sharing, bounded cache filenames, and cleanup of materialized files after sharing.
- Preserves attachments through reply/edit flows, supports attachment-only messages, hides copy for attachment-only content, and shows attachment filenames in reply previews.
- Refactors shared attachment helpers into `@kilocode/kilo-chat` / `@kilocode/kilo-chat-hooks` so web and mobile share file-size formatting, edit-content attachment preservation, and XHR upload behavior while keeping upload internals private.

## Verification

- Not manually verified; this update only refreshed the PR description after reading the changed files.

## Visual Changes

N/A — no screenshots captured while updating the description.

## Reviewer Notes

- Please review mobile native attachment edge cases closely: permission-denied recovery, upload retry/removal, cache filename bounding, and share cleanup.
- Web attachment upload behavior now uses the shared XHR upload helper; reviewer focus should include parity with the previous web implementation and the public API boundary.


packages/kilo-chat-hooks/src/use-attachment-queue.ts
```diff
@@ -128,6 +135,84 @@ export type PerformUpload = (
   opts: { onProgress: (fraction: number) => void; signal: AbortSignal }
```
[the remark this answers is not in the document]
iscekic: Fixed in `a227435ea` (`fix(kilo-chat): keep xhr upload internals private`). `XhrUploadOutcome` and `mapXhrUploadResultToOutcome` are no longer exported from `use-attachment-queue`, the public barrel now has a regression test for this, and the XHR behavior remains covered through `createXhrPerformUpload`.

apps/mobile/src/components/kilo-chat/message-attachment-open.ts
```diff
@@ -0,0 +1,76 @@
...
+  url,
+  attachmentId,
+  filename,
+}: {
+  url: string;
+  attachmentId: string;
+  filename: string;
+}): Promise<string> {
+  const { Directory, File, Paths } = await import('expo-file-system');
+  const directory = new Directory(Paths.cache, 'kilo-chat-attachments');
```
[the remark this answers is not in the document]
iscekic: Fixed in `6b28be716` (`fix(mobile): clean up shared attachment cache files`). `shareRemoteAttachment` now materializes the downloaded cache file as a cleanup-capable object and deletes it in a `finally` after the share attempt, with regression coverage for both successful shares and share failures.

apps/mobile/src/components/kilo-chat/message-attachment.tsx
```diff
@@ -0,0 +1,169 @@
...
+      if (!url) {
+        throw new Error(getAttachmentOpenErrorMessage());
+      }
+
+      await shareRemoteAttachment({
+        url,
+        attachmentId: block.attachmentId,
+        filename: block.filename,
+      });
+    } catch {
```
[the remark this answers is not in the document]
iscekic: I checked the installed `expo-sharing@55.0.19` implementation and did not make a code change here. On iOS, `SharingModule.swift` resolves the promise when the share sheet is dismissed without action (`type == nil && !completed`). On Android, `SharingModule.kt` resolves on `OnActivityResult` without checking result code. So normal user cancellation is not demonstrated as a rejection in this version, and the suggested `AbortError` branch would not address a proven path. The existing catch remains for real refetch/download/share failures.

apps/mobile/src/components/kilo-chat/message-attachment-open.ts
```diff
@@ -0,0 +1,185 @@
...
+}
+
+export async function shareMaterializedAttachment(
+  attachment: MaterializedAttachment,
+  shareFile: (uri: string) => Promise<void> = shareLocalFile
+): Promise<void> {
+  try {
+    await shareFile(attachment.uri);
+  } finally {
+    attachment.delete();
```
jeanduplessis, line 99: **WARNING:** Cache cleanup can race Android share-target reads.

`shareAsync()` settling does not prove Android receiver already consumed shared bytes. In `expo-sharing@55.0.19`, Android passes original file through a `FileProvider` URI and resolves JS promise from activity-result handling; `ACTION_SEND` has no target-consumption result. This `finally` then deletes backing file immediately.

Trace:
1. `shareRemoteAttachment()` materializes cache file, then delegates to `shareMaterializedAttachment()`.
2. `shareMaterializedAttachment()` always runs `attachment.delete()` immediately after `shareFile()` settles.
3. Android share targets may open granted URI lazily after chooser/result callback, so file can disappear before target reads it.

Impact: Android shares can deliver missing/broken attachments for receivers that defer URI reads, especially slower apps or larger files.

apps/mobile/src/components/kilo-chat/message-attachment-picker.ts
```diff
@@ -0,0 +1,124 @@
...
+        fileName: asset.fileName,
+        mimeType: asset.mimeType,
+        fileSize: asset.fileSize,
+      })
+    )
+  );
+}
+
+export async function pickLibraryImages(): Promise<AddFileInput[]> {
+  const permission = await ImagePicker.requestMediaLibraryPermissionsAsync();
```
jeanduplessis: **WARNING:** Image-only library picker is blocked by a permission gate that current SDK picker path does not require.

This path launches `launchImageLibraryAsync()` with `mediaTypes: ['images']`. Expo SDK 55 documents media-library permission as required for `launchImageLibraryAsync()` only on iOS 10; manual preflight is called out for specific video flows, not this image-only path. Returning before launch means denial of broader photo permission disables a still-valid system picker flow.

Trace:
1. `pickLibraryImages()` requests media-library permission before launching picker.
2. `!permission.granted` shows Settings alert and returns `[]`.
3. Image-only picker launch never runs, even on platforms where chooser access works without that permission grant.

Impact: Modern iOS/Android users who deny or cannot grant media-library permission cannot attach library images despite system picker support.

jeanduplessis (approved):
Reviewed final combined diff and PR-head context. Found two warning-level correctness issues:

1. Android share cleanup deletes materialized file before target app is guaranteed to consume it.
2. Image-only photo-library selection is unnecessarily blocked after media-library permission denial.

PR body otherwise matches implementation, unresolved review threads are clear, and no PostHog tracking path changed in this diff.

apps/mobile/src/components/kilo-chat/message-attachment-open.ts
iscekic: Fixed in 83719ff0d. Successful Android shares now keep the materialized cache file available after shareAsync returns; immediate deletion is still preserved for failed shares and non-Android successful shares. Added a regression test covering the Android success case.

apps/mobile/src/components/kilo-chat/message-attachment-picker.ts
iscekic: Fixed in 0ff6d0a85. Removed the media-library permission preflight from the image-only library picker path so launchImageLibraryAsync can use the platform picker directly. Camera still keeps its permission gate. Updated the picker test to assert library launch proceeds without requesting broad photo-library permission.

PR #3374, opened 2026-05-20
fix(kiloclaw): connect Composio before provision

pandemicsyn:
## Summary

- KiloClaw already delays first provision until the user has selected or skipped location, this pushes that to cover composio. Managed or manual Composio setup now completes before initial provision instead of racing after the machine is already booting. Without this - the instance is provisioned BEFORE composio creds are added and so the instance requires another boot to pick up fly env changes.

## Visual Changes

| Before | After |
| ------ | ----- |
| Tools step could connect Composio after provisioning had already started, which could require a later restart before CLI login took effect. | Tools step completes managed/manual Composio setup before initial provision so first boot can consume the chosen credentials. |
| Popup success could surface before the durable connected-account status was confirmed. | Popup flow remains pending until status refetch confirms the managed connection is durably connected. |

## Reviewer Notes

- Provisioning now starts after Tools connect/save/skip, before Email; legacy non-Composio Calendar retains its existing active-instance sequencing.
- Manual Composio credentials remain a real pre-provision path and are passed directly into the initial provision request.
- Managed Connect Link creation, callback completion, and provision now coordinate through the same personal/org provision locks to avoid first-boot races.
- `skipIncompleteManagedComposioConnection` is only used by the explicit Tools skip path so users can proceed without waiting on an abandoned managed-connect attempt.

### DIFF
--- a/apps/web/src/app/(app)/claw/components/ClawOnboardingFlow.tsx
+++ b/apps/web/src/app/(app)/claw/components/ClawOnboardingFlow.tsx
@@ -172,6 +172,68 @@ function createComposioConnectAttemptId(): string {
   return `${Date.now()}-${Math.random().toString(36).slice(2)}`;
 }
 
+const ONBOARDING_DRAFT_STORAGE_PREFIX = 'kiloclaw:onboarding-draft:';
+
+type OnboardingDraft = {
+  userId: string;
+  botIdentity: BotIdentity;
+  userLocation: string | null;
+};
+
+function onboardingDraftStorageKey(organizationId?: string): string {
+  return `${ONBOARDING_DRAFT_STORAGE_PREFIX}${organizationId ?? 'personal'}`;
+}
+
+function readOnboardingDraft(
+  organizationId: string | undefined,
+  userId: string
+): OnboardingDraft | null {
+  if (typeof window === 'undefined') return null;
+  try {
+    const raw = window.sessionStorage.getItem(onboardingDraftStorageKey(organizationId));
+    if (!raw) return null;
+    const parsed: unknown = JSON.parse(raw);
+    if (typeof parsed !== 'object' || parsed === null) return null;
+    const record = parsed as Record<string, unknown>;
+    if (record.userId !== userId) return null;
+    const botIdentity = record.botIdentity;
+    if (typeof botIdentity !== 'object' || botIdentity === null) return null;
+    const identity = botIdentity as Record<string, unknown>;
+    if (
+      typeof identity.botName !== 'string' ||
+      typeof identity.botNature !== 'string' ||
+      typeof identity.botVibe !== 'string' ||
+      typeof identity.botEmoji !== 'string'
+    ) {
+      return null;
+    }
+    const userLocation = record.userLocation;
+    if (userLocation !== null && typeof userLocation !== 'string') return null;
+    return {
+      userId,
+      botIdentity: {
+        botName: identity.botName,
+        botNature: identity.botNature,
+        botVibe: identity.botVibe,
+        botEmoji: identity.botEmoji,
+      },
+      userLocation,
+    };
+  } catch {
+    return null;
+  }
+}
+
+function writeOnboardingDraft(organizationId: string | undefined, draft: OnboardingDraft): void {
+  if (typeof window === 'undefined') return;
+  window.sessionStorage.setItem(onboardingDraftStorageKey(organizationId), JSON.stringify(draft));
+}
+
+function clearOnboardingDraft(organizationId?: string): void {
+  if (typeof window === 'undefined') return;
+  window.sessionStorage.removeItem(onboardingDraftStorageKey(organizationId));
+}
+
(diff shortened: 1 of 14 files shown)


apps/web/src/app/(app)/claw/components/ClawOnboardingFlow.tsx
```diff
@@ -756,21 +911,27 @@ function ClawOnboardingFlowInner({
...
+            if (!flowState.instanceStatus && currentUser?.id) {
+              writeOnboardingDraft(organizationId, {
+                userId: currentUser.id,
+                botIdentity: identity,
+                userLocation: weatherLocation?.location ?? null,
+              });
+            }
             setOnboardingStep('tools');
           } else if (hasCalendarStep) {
+            if (!flowState.instanceStatus) provisionInstance(weatherLocation?.location);
```
St0rmz1, line 931: `claw_create_instance_clicked` is now only captured inside `startProvisionForCreateFlow`. The calendar-only and email-only flows provision directly here (and at line 934), and the later `advanceToEmail` -> `startProvisionForCreateFlow` call early-returns because `createSetupStarted` is already true, so its capture never fires. Result: those two flows emit no `claw_create_instance_clicked` event at all, which will silently skew the onboarding funnel.

Suggest capturing the event wherever a first provision actually starts, or moving the capture into `provisionInstance` itself gated on it being a first provision, so there is a single source of truth. A test asserting the event fires for the tools, calendar-only, and email-only flows would catch regressions.

apps/web/src/app/(app)/claw/components/ClawOnboardingFlow.tsx
```diff
@@ -873,14 +1043,21 @@ function ClawOnboardingFlowInner({
...
           posthog?.capture('claw_setup_tools_completed', { connected: true, skipped: false });
           advanceToEmail();
         }}
         onSaveManualCredentials={credentials => {
           posthog?.capture('claw_setup_tools_manual_credentials_save_clicked');
+          if (flowState.instanceStatus === null) {
+            posthog?.capture('claw_setup_tools_manual_credentials_saved', {
+              provision_input: true,
+            });
+            advanceToEmail(credentials);
```
St0rmz1, line 1058: When no instance exists yet, manual credentials go straight into the provision `secrets` field via `advanceToEmail(credentials)`, skipping the `patchSecrets` mutation below. `patchSecrets` validates catalog secrets against their pattern and `maxLength` (the "Composio user API keys start with uak_" check, per-field length cap); the provision path does none of that, and the `secrets` schema in `kiloclaw-router.ts` is `z.record(z.string(), z.string())` with no `.max()` on values.

So a typo'd key gets immediate, specific feedback on the post-provision path but silently misconfigures Composio in the sandbox on this path, where provision just succeeds. Suggest running the same catalog validation (pattern + `maxLength`) on the manual credentials before provisioning, ideally inside `buildComposioProvisionSecrets`' manual branch so both paths share it, and adding `.max(N)` to the `secrets` value schema.

apps/web/src/app/(app)/claw/components/ClawOnboardingFlow.tsx
```diff
@@ -386,22 +480,34 @@ function ClawOnboardingFlowInner({
...
-      setComposioPopupPendingState(false);
       setOnboardingStep('tools');
       void composioStatus.refetch();
 
       if (message.result === 'success') {
-        posthog?.capture('claw_setup_tools_composio_completed');
-        toast.success('Google Calendar connected');
+        composioPopupAwaitingConfirmationRef.current = true;
+        clearComposioPopupConfirmationTimeout();
+        composioPopupConfirmationTimeoutRef.current = window.setTimeout(() => {
```
St0rmz1, line 489: Edge case worth noting: if the confirmation timeout fires at 45s (sets `composioPopupPending` to false, shows the error toast) and then `composioStatus` flips to `connected` shortly after, the success effect early-returns because `composioPopupPending` is now false. The user sees a "could not confirm" error even though the connection succeeded, while the Tools step still renders as connected.

Low priority given the generous window. If you want to close it, have a later `connected` status supersede or clear the prior error toast.

apps/web/src/app/(app)/claw/components/ClawOnboardingFlow.tsx
```diff
@@ -731,6 +871,21 @@ function ClawOnboardingFlowInner({
     );
   }
 
+  function startProvisionForCreateFlow(
+    composioManualCredentials?: { composioUserApiKey: string; composioOrg: string },
```
St0rmz1, line 878: Nit: `provisionInstance`, `advanceToEmail`, and `startProvisionForCreateFlow` thread three positional optionals (`userLocation?`, `composioManualCredentials?`, `skipIncompleteManagedComposioConnection?`) through their signatures. An options object would make call sites self-documenting and remove the chance of passing arguments in the wrong slot. Not blocking.

St0rmz1 (approved):

PR #3376, opened 2026-05-20
fix(kiloclaw): stop repeated subscription-suspended emails from being triggered

pandemicsyn:
## Summary

A user received the "Your KiloClaw Subscription Has Ended" email many times over the course of a few days. Two bugs combined into a self-sustaining loop:

1. **The Worker proxy could still wake stopped instances.** All four legacy proxy branches (`/i/:instanceId/*`, host-based, cookie-routed, default catch-all) gated on `runtimeId` only, never on `status === 'running'`. A stopped Fly machine retains its `flyMachineId`, so the Worker still forwarded to `{flyAppName}.fly.dev` with `fly-force-instance-id`, which triggers Fly Proxy's autostart and silently woke instances we deliberately suspended for billing reasons. The machine would then wake and the controller would perform its ready checkin. 

2. **`deliverChatWebhook` (the WorkerEntrypoint RPC called by our chat dispatcher was a fifth proxy site with the same bypass** — it issued its own `fetch(targetUrl)` rather than going through `proxyThroughTarget`, so an inbound chat message to a suspended instance also triggered Fly autostart.

3. **`completeAutoResumeIfReady` cleared suspension state for canceled subscriptions.** It treated any non-null `suspended_at` as a pending auto-resume and called `clearAutoResumeState`, which deletes the `claw_suspended_subscription` row from `kiloclaw_email_log` and nulls `suspended_at` / `destruction_deadline`. Per `.specs/kiloclaw-billing.md §1132.1`, auto-resume completion fires only when a subscription transitions to `active`. A `canceled`-and-period-ended subscription never satisfies that, but the readiness callback (fired after the proxy-driven Fly wakeup) was wiping the once-per-lifecycle email-log dedupe gate every cycle.

The fixes:

- Add a strict `status === 'running'` gate to all four proxy branches in `services/kiloclaw/src/index.ts` AND to `deliverChatWebhook`. Anything else returns `409 { error: "Instance not running" }` (chat webhook throws). Document the contract on `proxyThroughTarget` so future contributors don't reintroduce the bypass. The chat-webhook throw is a load-bearing error string pattern-matched by `isDefiniteUnreachable` in `services/kilo-chat`, so this PR also extends that classifier in lock-step (with a regression test) — without it, suspended bots would show stale-online in chat platform UIs until staleness inference catches up.
- Require `subscription.status === 'active'` in `completeAutoResumeIfReady` before invoking `clearAutoResumeState`. Stale ready callbacks for `canceled` or `past_due` subscriptions become no-ops.

No schema, migration, or rollout changes. Both fixes are tightenings of existing checks; no happy-path semantics change.

Spec context: `.specs/kiloclaw-billing.md` Email Notifications §1118.1 ("each notification at most once per user per lifecycle event") and Auto-Resume on Payment Recovery §1132.1 ("transitions to active while the subscription's instance is suspended").

## Verification

- [ ] Local Worker test suite exercises the four 409-on-stopped branches plus the `deliverChatWebhook` stopped-throw case and asserts `fetch` is never called against the Fly target.
- [ ] Local web lifecycle test asserts no clear for `status='canceled'` and `status='past_due'`, and asserts the clear still runs for `status='active'` with pending markers.
- [ ] No end-to-end manual run against a real Fly app; the originally-affected instance is already self-destroyed and reproducing the loop in dev would require a 1+ hour billing-sweep cadence. Post-deploy, query AE for `reconcile.sync_status` events with `status='stopped'` over 24h — should drop to ~0 since nothing is waking stopped machines anymore.
- [ ]

## Visual Changes

N/A

## Reviewer Notes

- **Risk**: a user mid-page-load whose subscription has just expired now sees a 409 from the proxy instead of a slow autostart. The response includes `hint: "Start it from the dashboard."` and the dashboard already shows the stopped state. Previous behavior would have charged Fly for the wake then re-stopped the instance ~1 hour later anyway — net better for the user.
- **Where to focus review**: the four `status !== 'running'` gates in `services/kiloclaw/src/index.ts` (search the diff) plus the `deliverChatWebhook` gate. The default-catch-all branch uses `status && status !== 'running'` to avoid false-positives on the `null` status path that the existing `!runtimeId` 404 already handles; the other three branches read directly from `getStatus()` and use the unconditional form.
- **Cross-service change**: this PR also touches `services/kilo-chat/src/services/bot-status-request.ts` to extend the `isDefiniteUnreachable` classifier with the new `is not running` substring match. The `deliverChatWebhook` doc-comment explicitly requires lock-step updates to that classifier; landing them together avoids a window where chat dispatchers misclassify suspended bots as transient.
- **Pre-existing**: `clearAutoResumeState`'s "no active instance found" branch (`apps/web/src/lib/kiloclaw/instance-lifecycle.ts:330-339`) still wipes `suspended_at` regardless of status when readiness fires for a destroyed instance — different code path, rare in practice. The worker-side `clearAutoResumeState` copy in `services/kiloclaw-billing/src/lifecycle.ts:564-614` is currently safe by call-site invariant (only reachable post-renewal where `status='active'`); documented in deviations log for future readers.
- **Test inversion**: the existing `instance-lifecycle.test.ts:202` test was inverted — previously it asserted that a `canceled`+`suspended_at` subscription triggers `clearAutoResumeState`, which encoded the bug. The replacement asserts the active-recovery case, plus two new cases (canceled no-clear, past_due no-clear).

### DIFF
--- a/apps/web/src/lib/kiloclaw/instance-lifecycle.ts
+++ b/apps/web/src/lib/kiloclaw/instance-lifecycle.ts
@@ -340,6 +340,7 @@ export async function completeAutoResumeIfReady(
 
   const [subscription] = await db
     .select({
+      status: kiloclaw_subscriptions.status,
       suspended_at: kiloclaw_subscriptions.suspended_at,
       auto_resume_requested_at: kiloclaw_subscriptions.auto_resume_requested_at,
       auto_resume_retry_after: kiloclaw_subscriptions.auto_resume_retry_after,
@@ -355,18 +356,29 @@ export async function completeAutoResumeIfReady(
     )
     .limit(1);
 
-  const hadPendingResume = !!(
+  // Per .specs/kiloclaw-billing.md §1132 (Auto-Resume on Payment Recovery),
+  // auto-resume completion fires when a subscription "transitions to active
+  // while the subscription's instance is suspended". A canceled or past_due
+  // subscription has not transitioned to active and MUST NOT have its
+  // suspension state cleared by a stale instance-ready callback — otherwise
+  // a Fly-Proxy-driven wakeup of a stopped machine would silently delete
+  // the once-per-lifecycle email-log entries (§1118.1) and re-enable the
+  // hourly subscription-expiry sweep to send another suspension email.
+  const hasPendingResumeState = !!(
     subscription?.suspended_at ||
     subscription?.auto_resume_requested_at ||
     subscription?.auto_resume_retry_after ||
     (subscription?.auto_resume_attempt_count ?? 0) > 0
   );
+  const isPendingResume = subscription?.status === 'active' && hasPendingResumeState;
 
-  if (!hadPendingResume) {
+  if (!isPendingResume) {
     logInfo('Instance ready without pending async auto-resume state', {
       user_id: kiloUserId,
       instance_id: targetInstance.id,
       sandbox_id: sandboxId,
+      subscription_status: subscription?.status ?? null,
+      has_suspended_at: subscription?.suspended_at != null,
     });
     return { instanceId: targetInstance.id, resumeCompleted: false };
   }
--- a/apps/web/src/lib/kiloclaw/instance-lifecycle.ts
+++ b/apps/web/src/lib/kiloclaw/instance-lifecycle.ts
@@ -132,12 +132,50 @@ async function clearAutoResumeState(
     logMessage: string;
     changeLogReason: string;
     logFields?: Record<string, unknown>;
+    /**
+     * When set, the transactional clear is gated on the subscription still
+     * being in `status='active'` at lock time. Used by the genuine
+     * auto-resume completion path to close a TOCTOU window: the
+     * subscription status read in `completeAutoResumeIfReady` happens
+     * outside this transaction, and a concurrent transition (credit-renewal
+     * sweep flipping back to past_due, user cancellation, subscription
+     * expiry) between the precondition and this transaction would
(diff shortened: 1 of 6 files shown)


apps/web/src/lib/kiloclaw/instance-lifecycle.ts
```diff
@@ -181,12 +219,23 @@ async function clearAutoResumeState(
     }
   });
 
+  if (skippedNoActiveSubscription) {
```
jeanduplessis, line 222: Warning: this guarded skip is still reported upstream as completed auto-resume. `clearAutoResumeState` returns early here, but `completeAutoResumeIfReady` still returns `{ resumeCompleted: true }`; `apps/web/src/app/api/internal/kiloclaw/instance-ready/route.ts` then logs `event: 'resume_completed'`, `outcome: 'completed'`.

Trace: the new `requireActiveSubscription` lock path bails before delete/update/change-log work, while caller-side completion signal stays unchanged.

Impact: active-to-non-active race now emits both a skipped lifecycle log and a completed readiness log even though no resume state was cleared, so post-deploy diagnosis can misclassify guarded no-ops as successful recoveries.

jeanduplessis (commented):
Reviewed final combined diff plus billing lifecycle specs, proxy branches, ready-callback caller, and changed tests. Found one non-blocking reporting mismatch: the new transactional no-op race guard still propagates as `resumeCompleted: true`, which makes readiness telemetry say completion occurred when mutation was intentionally skipped.

jeanduplessis (approved):

PR #3378, opened 2026-05-21
feat(kiloclaw): advertise controller capabilities

pandemicsyn:
## Summary

- Add an explicit controller capability contract to `GET /_kilo/version` with `apiVersion` and sorted `capabilities`.
- Thread the optional capability fields through the Worker schema, Instance DO return types, platform response, and web API type.
- Add a controller endpoint capability registry, including conditional Kilo Chat capabilities only when Kilo Chat routes are registered.
- Update the KiloClaw controller spec so capability hints are documented as advisory route/behavior support, not mutation-time guarantees.

## Verification

- Manual verification was not performed; this is a controller/API contract and schema-threading change with no UI behavior changes.

## Visual Changes

N/A

## Reviewer Notes

- Existing CalVer-gated UI behavior is intentionally unchanged.
- Old controllers remain compatible because `apiVersion` and `capabilities` are optional.
- Worker validation now rejects malformed, duplicate, or unsorted capability arrays.
- Capability hints are advisory; future mutation paths still need old-controller and runtime fallback handling.

### DIFF
--- a/.specs/kiloclaw-controller.md
+++ b/.specs/kiloclaw-controller.md
@@ -588,9 +588,27 @@ The response fields are:
 | `commit`          | string         | Yes      | Controller build commit hash                                                       |
 | `openclawVersion` | string \| null | Yes      | Installed openclaw version                                                         |
 | `openclawCommit`  | string \| null | Yes      | Installed openclaw commit hash                                                     |
+| `apiVersion`      | number         | No       | Controller response/protocol envelope shape version                                |
+| `capabilities`    | string[]       | No       | Sorted capability hints for controller routes/behaviors registered in this build   |
 | `gateway`         | object \| null | Yes      | Supervisor stats (same as `/_kilo/gateway/status`), null if supervisor not created |
 | `controllerState` | object         | No       | Current controller lifecycle state, present when state ref is wired                |
 | `kiloChatHealth`  | object         | No       | Kilo Chat health probe summary, present when Kilo Chat probing is active           |
+
+Version capability hint rules:
+
+1. `apiVersion`, when present, describes the version response/protocol
+   envelope shape only. It MUST NOT be used as the sole proof that a
+   specific endpoint exists when a named capability is available.
+2. `capabilities`, when present, MUST be a sorted, duplicate-free array
+   of stable lowercase dot/kebab strings.
+3. A capability MUST only be advertised when the corresponding route or
+   behavior is registered in the running controller build.
+4. Capability hints are advisory. Clients MUST keep normal mutation-time
+   handling for old routes, auth failures, validation failures,
+   conflicts, and runtime errors.
+5. Missing `capabilities` on a controller with a non-null version means
+   capability support is unknown and SHOULD fail closed for newly
+   capability-gated features.
 
 #### Gateway (bearer token)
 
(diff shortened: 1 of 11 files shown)


services/kiloclaw/controller/src/endpoint-capabilities.test.ts
```diff
@@ -0,0 +1,30 @@
+import { describe, expect, it } from 'vitest';
+import {
+  CONTROLLER_ENDPOINT_CAPABILITIES,
+  KILO_CHAT_ENDPOINT_CAPABILITIES,
+  getControllerEndpointCapabilities,
+} from './endpoint-capabilities';
+
+describe('getControllerEndpointCapabilities', () => {
```
St0rmz1, line 10: **Moderate, Testing:** Nothing here validates that registry entries conform to the Worker's acceptance rules. The Worker schema (`ControllerVersionResponseSchema` in `gateway-controller-types.ts`) enforces `/^[a-z][a-z0-9]*(?:[.-][a-z][a-z0-9]*)*$/` on every capability, but this test only checks sort order and uniqueness.

If a non-conforming capability (camelCase, underscore, etc.) is later added to `CONTROLLER_ENDPOINT_CAPABILITIES`, controller CI stays green, but the Worker's `safeParse` rejects the whole array. `callGatewayController` then throws `GatewayControllerError(502)`, and `getControllerVersion` rethrows it since the error is not an unknown-route error. One malformed string takes down the entire `/_kilo/version` response: `version`, `commit`, `gateway` stats, and `controllerState`.

Suggestion: add a test asserting every entry in `CONTROLLER_ENDPOINT_CAPABILITIES` and `KILO_CHAT_ENDPOINT_CAPABILITIES` matches that same regex, so drift is caught at controller build time.

services/kiloclaw/controller/src/endpoint-capabilities.ts
```diff
@@ -0,0 +1,60 @@
...
+  'kilo-chat.attachments',
+] as const;
+
+export type ControllerEndpointCapability =
+  | (typeof CONTROLLER_ENDPOINT_CAPABILITIES)[number]
+  | (typeof KILO_CHAT_ENDPOINT_CAPABILITIES)[number];
+
+export function getControllerEndpointCapabilities(options?: {
+  includeKiloChatCapabilities?: boolean;
+}): string[] {
```
St0rmz1: **Nit, Design:** This builds its result from the typed `as const` arrays but returns `string[]`, discarding the `ControllerEndpointCapability` union. Returning `ControllerEndpointCapability[]` would preserve the type for future callers.

services/kiloclaw/controller/src/routes/health.test.ts
```diff
@@ -257,6 +265,19 @@ describe('GET /_kilo/version', () => {
     const body = (await resp.json()) as { gateway: null };
     expect(body.gateway).toBeNull();
   });
+
+  it('includes Kilo Chat capabilities only when the route set is registered', async () => {
```
St0rmz1, line 270: **Nit, Testing:** This covers the positive case (Kilo Chat capabilities present when registered) but there is no route-level test that they are absent by default. `endpoint-capabilities.test.ts` covers the negative case at the function level, so this is minor; a one-line assertion that `body.capabilities` does not contain `kilo-chat.attachments` without the flag would close the gap.

.specs/kiloclaw-controller.md
```diff
@@ -588,10 +588,49 @@ The response fields are:
...
 | `gateway`         | object \| null | Yes      | Supervisor stats (same as `/_kilo/gateway/status`), null if supervisor not created |
 | `controllerState` | object         | No       | Current controller lifecycle state, present when state ref is wired                |
 | `kiloChatHealth`  | object         | No       | Kilo Chat health probe summary, present when Kilo Chat probing is active           |
 
+Version capability hint rules:
+
+1. `apiVersion`, when present, describes the version response/protocol
+   envelope shape only. It MUST NOT be used as the sole proof that a
+   specific endpoint exists when a named capability is available.
+2. `capabilities`, when present, MUST be a sorted, duplicate-free array
```
jeanduplessis: nit (non-blocking): Do we want to specify how it should be sorted? And should the sort order be stable across versions?

jeanduplessis: Maybe something like ...
```
`capabilities`, when present, MUST be sorted in ascending ASCII
lexicographic order by capability string and MUST be duplicate-free.
Clients MUST treat the array as a set; ordering is canonicalization only
and carries no feature semantics.
```

jeanduplessis (approved):

PR #3379, opened 2026-05-21
refactor(integrations): unify OAuth integration routes

RSO:
## Summary

- Unified standard OAuth integrations behind dynamic `/api/integrations/[platform]/connect` and `/api/integrations/[platform]/callback` routes while keeping GitHub App and Google OAuth on their special routes.
- Replaced Slack/Linear/Discord `getOAuthUrl` tRPC initiation with shared connect paths, and unified user/org platform detail pages behind shared dynamic UI.
- Preserved OAuth resume behavior, including sign-in callback allowlisting, `returnTo` redirects, encoded callback errors, and self-hosted GitLab credential handling via an authenticated POST/Redis handoff instead of query-string secrets.

## Verification

- Not manually verified in browser; OAuth provider flows require external app credentials and callback setup.

## Visual Changes

N/A

## Reviewer Notes

- Review the OAuth redirect and `returnTo` paths closely, especially unauthenticated connect resumes and platform-specific success/error query parameters.
- GitLab self-hosted OAuth uses POST for `clientSecret` handoff; the GET connect route intentionally no longer consumes `clientSecret`.


apps/web/src/lib/integrations/slack-service.ts
```diff
@@ -220,11 +217,11 @@ export async function upsertSlackInstallation({
...
 
   // For org integrations, get a model that respects org access policy.
-  // For user integrations, use the Slack-specific default model
+  // For user integrations, use the shared bot default model.
   const defaultModel =
     owner.type === 'org'
-      ? await getDefaultAllowedModel(owner.id, SLACK_DEFAULT_MODEL)
-      : SLACK_DEFAULT_MODEL;
+      ? await getDefaultAllowedModel(owner.id, DEFAULT_BOT_MODEL)
+      : DEFAULT_BOT_MODEL;
```
alex-alecu, line 224: Is this Slack default model change intended? New personal Slack installations now default to `kilo-auto/frontier` through `DEFAULT_BOT_MODEL` instead of the previous `kilo-auto/free`, which can change credit/cost behavior for Slack bot usage.

apps/web/src/components/integrations/SlackIntegrationDetails.tsx
```diff
@@ -150,19 +147,9 @@ export function SlackIntegrationDetails({
     }
   }, [success, error]);
```
alex-alecu, line 149: Can the removed `getOAuthUrl` tRPC procedures stay as compatibility shims for one deploy? Existing pages loaded before this rollout still call them when the user clicks Connect/Re-install; this new redirect path only helps freshly loaded clients.

apps/web/src/lib/integrations/oauth/platforms/gitlab-connect.ts
```diff
@@ -0,0 +1,196 @@
...
+          ? buildGitLabDetailCallbackPath(organizationId)
+          : undefined
+      );
+    }
+
+    const searchParams = request.nextUrl.searchParams;
+    const instanceUrl = searchParams.get('instanceUrl') || undefined;
+    const returnToParam = searchParams.get('returnTo') || undefined;
+    const returnTo = returnToParam ? validateReturnPath(returnToParam) : null;
+
```
alex-alecu, line 74: Could this path keep a short-lived compatibility branch for authenticated GETs with `clientId`/`clientSecret`? Old client bundles still send self-hosted GitLab OAuth credentials as query params, and this now drops them before building the OAuth URL, so those starts fail with `oauth_init_failed` during rollout.

apps/web/src/lib/integrations/slack-service.ts
RSO: Yeah, this was an oversight that I decided to fix: `auto/free` is not a great experience for the bot.

apps/web/src/components/integrations/SlackIntegrationDetails.tsx
RSO: That's a very specific and small window. I don't think it's worth the extra overhead of splitting off that change.

RSO:
@alex-alecu I addressed your comments, and made changes for the `clientSecret` through GET parameter case!

alex-alecu (approved):

PR #3380, opened 2026-05-21
CloudAgent - Gate Dev Container Sessions With Release Toggle

eshurakov:
## Summary
- Gate personal and organization Cloud Agent dev-container sessions with the `cloud-agent-devcontainer` release toggle.
- Move the dev-container control into the profile picker, persist the browser preference, and surface an enabled indicator beside the picker.
- Add router and preference coverage for toggle authorization and browser storage behavior.

## Verification

- [x] Verified manually

## Visual Changes

Screenshot 2026-05-21 at 09 49 06

## Reviewer Notes
- Development mode keeps the existing local bypass; production requests now require the release toggle instead of admin status.

### DIFF
--- a/apps/web/src/app/(app)/cloud/(agent)/page.tsx
+++ b/apps/web/src/app/(app)/cloud/(agent)/page.tsx
@@ -1,5 +1,5 @@
 import { getUserFromAuthOrRedirect } from '@/lib/user.server';
-import { isFeatureFlagEnabled, isReleaseToggleEnabled } from '@/lib/posthog-feature-flags';
+import { isFeatureFlagEnabled } from '@/lib/posthog-feature-flags';
 import { NewSessionPanel } from '@/components/cloud-agent-next/NewSessionPanel';
 import { CloudSessionsPage } from '@/components/cloud-agent/CloudSessionsPage';
 
@@ -8,7 +8,7 @@ export default async function PersonalCloudPage() {
   const isDevelopment = process.env.NODE_ENV === 'development';
   const useNextAgent = isDevelopment || (await isFeatureFlagEnabled('cloud-agent-next', user.id));
   const isDevcontainerAvailable =
-    isDevelopment || (await isReleaseToggleEnabled('cloud-agent-devcontainer', user.id));
+    isDevelopment || (await isFeatureFlagEnabled('cloud-agent-devcontainer', user.id));
 
   if (useNextAgent) {
     return <NewSessionPanel isDevcontainerAvailable={isDevcontainerAvailable} />;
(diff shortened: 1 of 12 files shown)


apps/web/src/app/(app)/cloud/(agent)/page.tsx
```diff
@@ -7,9 +7,11 @@ export default async function PersonalCloudPage() {
   const user = await getUserFromAuthOrRedirect('/users/sign_in?callbackPath=/cloud');
   const isDevelopment = process.env.NODE_ENV === 'development';
   const useNextAgent = isDevelopment || (await isFeatureFlagEnabled('cloud-agent-next', user.id));
+  const isDevcontainerAvailable =
+    isDevelopment || (await isFeatureFlagEnabled('cloud-agent-devcontainer', user.id));
```
RSO: In the past we noticed that fetching feature flags can be quite slow. Do you have an idea on the impact on overall performance of the page now that you await 2 feature flags sequentially in production? Should we `Promise.all`, or rely on the hook instead?

RSO (approved):

apps/web/src/routers/organizations/organization-cloud-agent-next-router.ts
```diff
@@ -192,10 +193,14 @@ export const organizationCloudAgentNextRouter = createTRPCRouter({
     .input(PrepareSessionInput)
     .output(basePrepareSessionNextOutputSchema)
     .mutation(async ({ ctx, input }) => {
-      if (input.devcontainer && !ctx.user.is_admin) {
+      if (
+        input.devcontainer &&
+        process.env.NODE_ENV !== 'development' &&
+        !(await isFeatureFlagEnabled('cloud-agent-devcontainer', ctx.user.id))
```
markijbema: can we keep this check centralized? (dev or featureflag)?

markijbema (approved):
I'd appreciate not duplicating the check as it can get out of sync. if we need to, lplease link with comments pointing to the other location

apps/web/src/app/(app)/cloud/(agent)/page.tsx
eshurakov: I've removed the cloud agent next flag, we actually don't need it

apps/web/src/routers/organizations/organization-cloud-agent-next-router.ts
eshurakov: Centralized this in `isFeatureFlagEnabledOrDevelopment`. The Cloud landing pages and both Cloud Agent Next routers now share the same development-or-feature-flag decision path, with helper coverage for the development short-circuit and production PostHog lookup.

PR #3382, opened 2026-05-21
fix(kiloclaw): align billing recovery with funding

jeanduplessis:
## Summary
Credit-funded KiloClaw recovery now routes users to credits, while Stripe-backed subscriptions keep Stripe-facing payment labels.

### Why this change is needed
Pure-credit renewal failures were shown with an Update Payment action that tried to open Stripe billing, even though those subscriptions have no Stripe funding path. Hybrid KiloClaw rows also store `payment_source = credits` while Stripe still owns payment collection, which made subscription summaries look credit-funded when users are still paying through Stripe.

### How this is addressed
- Route pure-credit past-due banner recovery to `/credits` with credit-specific copy and an Add Credits CTA.
- Keep Stripe portal recovery behavior for Stripe-funded and hybrid subscriptions.
- Prefer Stripe-funding presence over raw payment source when rendering customer-facing payment summaries.
- Add regression coverage for both failed credit-renewal recovery and hybrid payment-source labeling.

## Human Verification
- `NODE_ENV=test pnpm --filter web exec jest --runTestsByPath "src/components/subscriptions/helpers.test.ts" "src/app/(app)/claw/components/billing/billing-types.test.ts"`

### Human Reviewer Flags
- Hybrid subscriptions intentionally render as `Stripe` when a Stripe subscription ID exists, even if local `payment_source` is `credits`; this reflects who collects payment from user.

### DIFF
--- a/apps/web/src/app/(app)/claw/components/billing/BillingBanner.tsx
+++ b/apps/web/src/app/(app)/claw/components/billing/BillingBanner.tsx
@@ -1,5 +1,7 @@
 'use client';
 
+import React from 'react';
+import Link from 'next/link';
 import {
   AlertCircle,
   AlertTriangle,
@@ -148,6 +150,19 @@ function getBannerContent(state: ClawBannerState, billing: ClawBillingStatus) {
         action: 'reactivate' as const,
       };
     case 'subscription_past_due':
+      if (
+        billing.subscription &&
+        !billing.subscription.hasStripeFunding &&
+        billing.subscription.paymentSource === 'credits'
+      ) {
+        return {
+          title: 'Payment failed — action required',
+          message:
+            'Your credit balance is insufficient for the next renewal. Add credits to avoid service interruption.',
+          cta: 'Add Credits',
+          action: 'add_credits' as const,
+        };
+      }
       return {
         title: 'Payment failed — action required',
         message:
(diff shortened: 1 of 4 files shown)


apps/web/src/app/(app)/claw/components/billing/BillingBanner.tsx
```diff
@@ -148,6 +150,19 @@ function getBannerContent(state: ClawBannerState, billing: ClawBillingStatus) {
         action: 'reactivate' as const,
       };
     case 'subscription_past_due':
+      if (
+        billing.subscription &&
+        !billing.subscription.hasStripeFunding &&
+        billing.subscription.paymentSource === 'credits'
```
St0rmz1, line 156: **Low, Functional:** This branch requires `paymentSource === 'credits'` in addition to `!hasStripeFunding`. `paymentSource` is typed `'stripe' | 'credits' | null`, so a past-due subscription with no Stripe funding but a `null` `paymentSource` falls through to the Stripe-portal recovery branch and still gets the "Update Payment" CTA that opens a Stripe portal with no funding path. That is the bug this PR fixes, just for a narrower set of rows.

Since `!hasStripeFunding` already means there is no Stripe portal to route to, consider dropping the `paymentSource === 'credits'` half of the condition. If it is kept intentionally, please confirm `paymentSource` is always non-null for credit-funded subscriptions, and a test covering `paymentSource: null` would document that.

apps/web/src/app/(app)/claw/components/billing/billing-types.test.ts
```diff
@@ -162,6 +165,30 @@ describe('KiloClaw billing display helpers', () => {
   });
 });
 
+describe('BillingBanner credit renewal recovery', () => {
```
St0rmz1, line 168: **Low, Testing:** This new test only covers the pure-credit recovery path. There is no assertion that a past-due subscription with `hasStripeFunding: true` still renders "Update Payment" and does not render the `/credits` link. The new branch is a guard that diverts a subset of past-due subscriptions, so a future change to the condition could silently divert Stripe-funded subscriptions to `/credits` with nothing to catch it.

Consider adding a sibling test with `hasStripeFunding: true` (and the hybrid `paymentSource: 'credits'` case) asserting the banner still shows "Update Payment" and no `href="/credits"`.

St0rmz1 (approved):

apps/web/src/app/(app)/claw/components/billing/BillingBanner.tsx
jeanduplessis: Leaving this guard unchanged. Paid billing status only serializes a subscription when Stripe funding exists or `payment_source` is `credits`, and KiloClaw paid-row invariants forbid a null payment source. The nullable transport type could be tightened separately, but that state should not reach this banner.

apps/web/src/app/(app)/claw/components/billing/billing-types.test.ts
jeanduplessis: Accepted. I added BillingBanner regression coverage for a hybrid Stripe-funded past-due row (`paymentSource: 'credits'`, `hasStripeFunding: true`) so `Update Payment` remains visible and `/credits` stays absent.

PR #3394, opened 2026-05-21
feat(kiloclaw) briefing first chat message

St0rmz1:
## Summary

When a user finishes KiloClaw onboarding, they now land in a chat conversation titled "Today's briefing" that already holds their first morning briefing as a message from the bot, in place of the default empty welcome bubble. Returning users who are not going through onboarding still go straight to chat as before.

The flow: once the new instance is ready, the onboarding wizard asks the worker to start the briefing. The morning briefing plugin creates the conversation and posts a loading message, then returns the conversation id quickly so the wizard can route the user straight into it. Generation runs in the background; when it finishes, the loading message is replaced with the briefing. While generation runs, the bot shows a typing indicator so the wait feels alive.

The first briefing is tailored for a brand new user:

- It opens with a short welcome instead of a dated header.
- It skips the Kilo Chat activity section, since a new user has no chat history.
- It is not sent to Telegram, Discord, or Slack. It stays in the chat as the user's first message there.
- Each unconnected source in the "Connect more" list links to the Settings page, with the correct path for both personal and organization instances.

Reliability:

- The whole briefing is one chat message (one bubble), split across multiple text blocks so it stays under the per block size limit.
- If the gateway restarts while generation is running, or generation fails, a later trigger resumes it rather than leaving the loading message stranded.
- Overlapping triggers are serialized, so a user never ends up with two "Today's briefing" conversations.
- If the trigger fails, onboarding falls back to the plain chat redirect and never gets stuck.

The briefing is posted directly as the bot with no agent involved, so the content is exactly what was generated.

## Verification

- [x] Completed onboarding on a fresh personal instance and confirmed the user lands in a "Today's briefing" conversation, the briefing renders as a single message, and the bot typing indicator shows while it generates.

The organization settings link, the restart resume path, and the skip of external channel delivery were covered by the automated suites and code review, and were not exercised by hand.

## Visual Changes

| Before | After |
| ------ | ----- |
|        |       |

## Reviewer Notes

The main new surface is the chat write path in the morning briefing plugin (creating the conversation, posting and editing the bot message) and the worker routing that triggers it from the onboarding wizard.


services/kiloclaw/plugins/kiloclaw-morning-briefing/src/index.ts
```diff
@@ -1684,6 +1759,184 @@ async function resolveDateKeyForOffset(
...
+  }
+
+  const writeClient = createKiloChatWriteClient();
+  if (!writeClient.configured) {
+    throw new Error(
+      `Kilo Chat is not available for the onboarding briefing: ${writeClient.reason}`
+    );
+  }
+
+  const conversationId = await writeClient.createConversation(ONBOARDING_BRIEFING_TITLE);
```
pandemicsyn, line 1957: The idempotency record is written only after both conversation creation and the loading message succeed. A crash/restart or write failure after createConversation or sendTextMessage but before writeJsonFile leaves no persisted conversationId, so the next trigger creates another "Today's briefing" conversation. This violates the “never two conversations” guarantee.

services/kiloclaw/plugins/kiloclaw-morning-briefing/src/kilo-chat-write-client.ts
```diff
@@ -0,0 +1,239 @@
...
+export function toTextContentBlocks(text: string): Array<{ type: 'text'; text: string }> {
+  if (text.length <= KILO_CHAT_TEXT_BLOCK_MAX) {
+    return [{ type: 'text', text }];
+  }
+  const blocks: Array<{ type: 'text'; text: string }> = [];
+  for (let i = 0; i < text.length; i += KILO_CHAT_TEXT_BLOCK_MAX) {
+    blocks.push({ type: 'text', text: text.slice(i, i + KILO_CHAT_TEXT_BLOCK_MAX) });
+  }
+  return blocks;
+}
```
pandemicsyn, line 76: toTextContentBlocks enforces the 8k text-block limit but not Kilo Chat’s 20-block message limit. A briefing over 160k chars produces more than 20 blocks, causing send/edit validation failure upstream and replacing the generated briefing with fallback text.

services/kiloclaw/plugins/kiloclaw-morning-briefing/src/kilo-chat-write-client.ts
```diff
@@ -0,0 +1,239 @@
...
+ * on the controller proxy, so no new controller route is required for the
+ * writes themselves.
+ *
+ * Messages posted here are authored by the bot (`bot:kiloclaw:<sandboxId>`):
+ * the controller's bot-auth middleware sets that sender id. They are NOT
+ * posted as the user.
+ */
+
+const DEFAULT_CONTROLLER_BASE_URL = 'http://127.0.0.1:18789';
+const DEFAULT_TIMEOUT_MS = 20_000;
```
pandemicsyn, line 24: The Worker-to-controller call uses the default 30s timeout, but the plugin can spend up to 20s creating the conversation plus 20s sending the loading message. Slow but successful Kilo Chat calls can outlive the Worker request, causing onboarding to fall back to plain chat while the plugin may still create the conversation in the background.

services/kiloclaw/plugins/kiloclaw-morning-briefing/src/index.ts
```diff
@@ -1684,6 +1759,184 @@ async function resolveDateKeyForOffset(
...
+      deliverToChannels: false,
+    });
+    clearInterval(typingTimer);
+    await writeClient.stopTyping(conversationId).catch(() => {});
+    const message = buildBriefingMessage({
+      sections: result.sections,
+      statuses: result.statuses,
+      tldr: result.tldr,
+      settingsHref: record.settingsHref,
+    });
```
pandemicsyn, line 1837: This ones probably fine, not sure we want to scare folks if some integration bit failed right out of the gate?

Onboarding chat messages drop result.failures. The saved briefing includes a ## Failures section for configured-but-failing sources, but buildBriefingMessage has no failures input, so the first in-chat briefing silently omits source failure notes.

pandemicsyn (approved):

PR #3396, opened 2026-05-21
fix(code-reviews): bound cron pending drain

alex-alecu:
## Summary

Followup to https://github.com/Kilo-Org/cloud/pull/3357

- Bound the cron pending code-review drain to pending reviews created between 60 and 75 minutes ago, matching the plan's main goal while preserving direct/manual dispatch as unbounded.
- Threaded the same `created_at` recovery window through owner discovery and per-owner reservation so cron candidate selection and claiming agree.
- Removed the unrelated retry-budget schema, migration, release-claim, and fixture changes so this PR has no database migration or manual retry semantic change.

## Reviewer Notes
- Stale queued recovery remains independent of the pending `created_at` window.
- The cron window is intentionally narrow and inclusive: reviews aged 60 through 75 minutes are eligible.
- Direct calls to `tryDispatchPendingReviews(owner)` still consider all pending reviews.


apps/web/src/lib/code-reviews/dispatch/dispatch-pending-code-review-owners.ts
```diff
@@ -62,11 +69,18 @@ async function drainOwner(
   }
 }
 
-export async function dispatchPendingCodeReviewOwners(): Promise<DispatchPendingCodeReviewOwnersSummary> {
-  const candidates = await listDispatchableCodeReviewOwnerCandidates({ limit: OWNER_SCAN_LIMIT });
+export async function dispatchPendingCodeReviewOwners(
+  options: DispatchPendingCodeReviewOwnersOptions = {}
+): Promise<DispatchPendingCodeReviewOwnersSummary> {
+  const pendingCreatedAfter = options.pendingCreatedAfter ?? cronPendingCodeReviewCreatedAfterSql();
```
alex-alecu: At deploy time, are there any rows with `status = 'pending'` and `created_at < now() - interval '2 hours'`? Both `listDispatchableCodeReviewOwnerCandidates` and `reconsiderableCodeReviewWorkCondition` now require `created_at >= now() - 2h` for pending, so any stragglers older than the cutoff become invisible to every future cron run unless fresh activity happens to call `tryDispatchPendingReviews` for that exact owner. The PR's "stale queued recovery remains uncapped" guarantee covers `queued`, not `pending`.

apps/web/src/lib/code-reviews/dispatch/dispatch-constants.ts
```diff
@@ -17,11 +19,23 @@ export function staleRunningCodeReviewCutoffSql() {
...
+
 export function reconsiderableCodeReviewWorkCondition(
-  staleQueuedCutoff = staleQueuedCodeReviewCutoffSql()
+  staleQueuedCutoff = staleQueuedCodeReviewCutoffSql(),
+  pendingCreatedAfter?: SQL
 ) {
+  const pendingCondition = pendingCreatedAfter
+    ? sql`(
+        ${cloud_agent_code_reviews.status} = 'pending'
+        AND ${cloud_agent_code_reviews.created_at} >= ${pendingCreatedAfter}
```
alex-alecu: The cutoff is on `created_at`, but `resetCodeReviewForRetry` does not refresh it. If a user retriggers a `failed`/`cancelled`/`interrupted` review whose `created_at` is older than 2 hours and the immediate `tryDispatchPendingReviews(owner)` call in the retrigger router returns `dispatched: 0` (e.g. owner at capacity), the row goes back to `pending` and every subsequent cron cycle skips it under this filter. The retrigger router only accepts `failed`/`cancelled`/`interrupted` so the user has no way to nudge it again. Should this gate on `updated_at`, or should `resetCodeReviewForRetry` refresh `created_at` so manually retriggered old reviews are not stranded?

apps/web/src/lib/code-reviews/dispatch/dispatch-pending-code-review-owners.ts
alex-alecu: Fixed in f527fa0.

apps/web/src/lib/code-reviews/dispatch/dispatch-constants.ts
alex-alecu: Covered in 9ca7d6b.

apps/web/src/lib/code-reviews/dispatch/dispatch-pending-code-review-owners.ts
alex-alecu: Adjusted in 980a347. The bound is intentional — the cron drain still scans only recently active pending rows, but the gate is now on `updated_at` rather than `created_at`, so manual retries (which refresh `updated_at`) remain reachable for the next two hours even when the immediate dispatch can't run. Old idle pending rows stay outside this cron path by design; if the deploy-time backlog needs to be flushed, that should be a separate one-time recovery, not an unbounded scan in every cron tick.

apps/web/src/lib/code-reviews/dispatch/dispatch-constants.ts
alex-alecu: Re-addressed in 980a347. The pending gate now keys on `updated_at`, and `resetCodeReviewForRetry` already refreshes `updated_at` while preserving `created_at`. So when an old retried review's immediate dispatch returns `dispatched: 0` because the owner is at capacity, the row stays `pending` with a fresh `updated_at` and remains eligible for cron for the next two hours. Tests cover both the unbounded direct path and the bounded cron path with refreshed `updated_at`.

apps/web/src/lib/code-reviews/dispatch/dispatch-pending-code-review-owners.ts
alex-alecu: @kilocode-bot open a new PR to update REVIEW.md with a bigger enforcement on the initial goal - we don't want to raise issues about the initial goal unless is unsafe

jrf0110 (approved):

PR #3399, opened 2026-05-21
chore(devcontainer): add cloud toolchain container

jeanduplessis:
## Summary
Add a repo devcontainer that gives cloud contributors a reproducible local toolchain.

### Why this change is needed
Local development depends on Node 24, pnpm, Docker Compose, worker CLIs, tunneling/payment/Fly tooling, Git LFS, and tmux. Those requirements exist across setup docs and the Nix shell, but the repo had no devcontainer contract for container-based workspaces.

### How this is addressed
- Add a Debian/Node 24.14.1 devcontainer image aligned with `.nvmrc` and the pinned pnpm version.
- Install Docker-in-Docker, Compose v2, Buildx, and the repo CLI surface used for web, Worker, KiloClaw, and local-service workflows.
- Pin Bun, Fly CLI, Wrangler, Vercel, and dotenvx versions; verify Bun/Fly release archives with architecture-specific SHA-256 checksums.
- Run repo dependency install after container creation and forward common web, Grafana, Postgres, Redis, and Worker ports.

## Human Verification
- Built the image with `docker build -f .devcontainer/Dockerfile . --tag kilo-cloud-devcontainer-check`.
- Ran CLI smoke checks inside the built image for Node, pnpm, Bun, Wrangler, Vercel, dotenvx, Stripe, cloudflared, Fly, psql, GitHub CLI, 1Password CLI, Git LFS, tmux, and socat.

### Human Reviewer Flags
- Bun and Fly CLI use checksum-verified release archives so both amd64 and arm64 container builds stay reproducible.
- Vendor apt repos provide cloudflared, Stripe CLI, GitHub CLI, 1Password CLI, and PostgreSQL client tooling to keep those installs package-managed.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,116 @@
+FROM node:24.14.1-bookworm
+
+ARG BUN_VERSION=1.3.14
+ARG DOTENVX_VERSION=1.66.0
+ARG FLYCTL_VERSION=0.4.54
+ARG PNPM_VERSION=11.1.2
+ARG VERCEL_VERSION=54.3.0
+ARG WRANGLER_VERSION=4.90.1
+
+ENV COREPACK_ENABLE_DOWNLOAD_PROMPT=0
+ENV FLYCTL_INSTALL=/usr/local/fly
+ENV PNPM_HOME=/home/node/.local/share/pnpm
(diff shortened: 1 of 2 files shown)


eshurakov (approved):
I wonder if we can do without docker in docker. 

Not critical, but as it stands now it is not gonna work in cloud agent, because it is already docker in docker at that moment :-)

PR #3400, opened 2026-05-21
fix(kiloclaw): harden Composio onboarding

pandemicsyn:
## Summary

- Stabilize the managed Composio Google Calendar onboarding flow by making identity reservation idempotent without advisory locks, restoring the tools step, and keeping callback completion resilient to Composio lifecycle lag.
- Harden callback completion: accept a caller-owned connected account even before Composio reports it `ACTIVE`, but treat worker `patchSecrets` failures as real infra failures with bounded retry and no connected-account DB write on failure.
- Keep managed status grounded in durable state for the current sandbox: provisioned instances only report connected when the sandbox is marked managed and worker secrets are configured, while popup/manual-save paths use event-driven refetch instead of interval polling.
- Extract popup coordination into a dedicated `useComposioPopup` state machine that preserves `postMessage`, `localStorage`, and `BroadcastChannel` callback channels, including late callback delivery after popup-close races.

## Verification

N/A

## Visual Changes

| Before | After |
| ------ | ----- |
| Composio tools step was temporarily bypassed while the flow was broken. | Composio tools step is restored in onboarding. |
| Inbound email Continue could be changed independently of the parent loading contract. | Inbound email again waits for instance status/address setup before Continue is enabled. |
| Popup coordination lived as refs/effects inside `ClawOnboardingFlow`. | Popup behavior is unchanged visually, but is now driven by `useComposioPopup`. |

## Reviewer Notes

- Conceptual callback split: Composio lifecycle lag is accepted after ownership verification; worker secret patch failures retry 3 times and then surface honestly.
- `google_calendar_connected_account_id` is only persisted after worker secrets are patched for an existing sandbox; status also verifies the current sandbox source/secrets before reporting connected.
- Composio onboarding status no longer polls every 15 seconds; popup callback, focus/visibility, and mutation invalidation paths are expected to drive refreshes.
- Popup callback handling now accepts same-attempt success after an early `popup_closed` observation and reads the stored callback result on focus/visibility to recover missed browser events.
- Focus review on lock-free identity reservation, callback race handling, current-sandbox status checks, and popup state-machine edges such as dedupe, popup close, confirmation timeout, late redirect races, and late callback delivery.
- Existing database indexes are reused; this PR does not add migrations or change schema.


apps/web/src/lib/kiloclaw/composio-onboarding.ts
```diff
@@ -277,39 +346,32 @@ export async function getManagedComposioGoogleCalendarStatus(params: {
...
-      };
-    }
+  const connectedAccountId = identity.row.google_calendar_connected_account_id;
+  if (!connectedAccountId) {
     return { enabled: true, status: 'disconnected', connectedAccountId: null, sandboxConfigSource };
-  } catch {
-    return { enabled: true, status: 'error', connectedAccountId: null, sandboxConfigSource };
   }
+
+  if (params.instance && !params.sandboxHasComposioSecrets) {
```
St0rmz1, line 367: **Moderate, Functional:** With the 15s `refetchInterval` removed from `getComposioOnboardingStatus`, the `awaiting-confirmation` state in `useComposioPopup` relies on a single `refetchStatus()` at callback time plus `focus`/`visibilitychange` events. This guard reports `disconnected` whenever a provisioned, managed sandbox has not yet reported `sandboxHasComposioSecrets`. If that flag is not `true` on the one post-callback refetch and the user never blurs and refocuses the tab, the 45s confirmation timer expires and surfaces "Could not confirm Google Calendar" even though the connection succeeded.

The callback route writes `patchSecrets`, then the connected-account id, then `composio_config_source='managed'` before returning success, so the only remaining variable is whether `sandboxHasComposioSecrets` flips to `true` synchronously once the worker `patchSecrets` call returns. If the worker config is read through any cache or eventually-consistent path, please either keep a short poll while the popup is in `awaiting-confirmation` or have the state machine refetch on its own interval during that state.

apps/web/src/lib/kiloclaw/composio-onboarding.ts
```diff
@@ -277,39 +346,32 @@ export async function getManagedComposioGoogleCalendarStatus(params: {
...
+    return { enabled: true, status: 'disconnected', connectedAccountId: null, sandboxConfigSource };
+  }
+
+  return {
+    enabled: true,
+    status: 'connected',
```
St0rmz1, line 373: **Low, Functional:** `getManagedComposioGoogleCalendarStatus` no longer calls `listComposioConnectedAccounts`, so `connected` is now derived purely from durable state (`composio_config_source`, the stored `google_calendar_connected_account_id`, and `sandboxHasComposioSecrets`). This is reasonable and removes a network call from a hot path, but it means a Google Calendar connection revoked on Google's side keeps showing "Connected" in this surface. If revocation is expected to be handled elsewhere (worker, reconnect prompt), this is fine as scoped. Flagging it so the tradeoff is intentional rather than incidental.

apps/web/src/app/api/integrations/composio/callback/route.ts
```diff
@@ -188,24 +183,17 @@ export async function GET(request: NextRequest) {
...
-            ? { ownerType: 'organization_user', userId: user.id, organizationId }
-            : { ownerType: 'user', userId: user.id },
-          connectedAccountId,
-        });
-      }
-    );
+    const instance = organizationId
+      ? await getActiveOrgInstance(user.id, organizationId)
+      : await getActiveInstance(user.id);
+    const verified = await completeManagedComposioGoogleCalendarConnection({
```
St0rmz1, line 189: **Low, Functional:** Dropping `withKiloclawProvisionContextLock` here means the `getComposioInstanceConfigSource(...) === 'manual'` check inside `completeManagedComposioGoogleCalendarConnection` and its subsequent `patchSecrets` plus `composio_config_source='managed'` write are no longer atomic against `provision`/`updateConfig`, which still take that lock. The manual `patchSecrets` save path never took this lock even before, so the manual-versus-managed race is pre-existing and the durable source marker is the intended arbiter. Not a regression introduced here, just noting the lock no longer provides any mutual exclusion for this path.

apps/web/src/lib/kiloclaw/composio-onboarding.ts
```diff
@@ -26,12 +26,57 @@ import {
...
+
+function waitForManagedComposioPatchSecretsRetry(ms: number): Promise<void> {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+function responseExcerpt(body: string): string {
+  return body.slice(0, 500);
+}
+
+function isRetryablePatchSecretsError(error: unknown): boolean {
```
St0rmz1, line 47: **Low, Testing:** `isRetryablePatchSecretsError` returns `true` for any non-`KiloClawApiError`, and `logTerminalManagedComposioPatchSecretsError` has a distinct `status: 'transport_error'` branch. The tests cover the `KiloClawApiError` 5xx-retry and 4xx-terminal paths but not a plain `Error`/`TypeError` (fetch abort, DNS failure). Add a test that rejects `patchSecrets` with a plain `Error` and asserts the retry budget is consumed and the log records `status: 'transport_error'`, so a regression in that branch does not ship silently.

St0rmz1 (approved):

PR #3411, opened 2026-05-22
fix(model-eval): match prefixed promotion stats

lambertjosh:
## Summary

Fixes an issue where Kilo Bench sync'd data uses the full Kilo model ID format (e.g. `kilo/anthropic/sonnet-4.6`) whereas the model stats API was filtering to the OpenRouter ID format. So this changes the filtering to try OpenRouter ID first, and if that doesn't exist try the full path. 

This now allows Kilo Bench data to be integrated into the model stats. 

I went with falling back and preserving the full string, in the event we want to test the same model from different providers.(Some providers may quantize, use different prompts, or make other changes in particular for open weight models)

## Visual Changes

N/A

## Validation

Post-change inserted some local kilo-bench runs and then validated that the model-stats API contains the kilo-bench data.

## Reviewer Notes

- The fallback only strips the outer `kilo/` when the remaining ID still contains a slash, so provider IDs like `kilo/special-model` do not collapse to bare `special-model`.
- Existing ingested promotions still need manual re-sync after deployment to recompute their caches.


services/model-eval-ingest/src/sync.ts
```diff
@@ -13,6 +13,7 @@ import {
 
 const PROMOTION_PULL_LIMIT = 1000;
 const MICRODOLLARS_PER_DOLLAR = 1_000_000;
+const KILO_MODEL_PREFIX = 'kilo/';
```
RSO: Is this really the first place we define the Kilo model prefix? Shouldn't this be centralized somewhere in the code?

RSO (approved):

services/model-eval-ingest/src/sync.ts
lambertjosh: @RSO - I shifted out some more re-usable definitions and functions into `packages/worker-utils/src/kilo-model-id.ts`. I didn't find another good location, if you know of one, let me know. We can do a follow up PR to update a few other places like cloud agent to use this.

jrf0110 (approved):

PR #3423, opened 2026-05-22
feat(cloud-agent): add Gastown badge

jrf0110:
Adds a Gastown badge to SessionsList components for sessions with createdOnPlatform 'gastown'.


jrf0110:
Superseded by #3424, which targets the correct convoy head branch instead of `main`. Closing as a stray.

PR #3444, opened 2026-05-24
feat(gateway): route Qwen3.7 Max directly through Alibaba

lambertjosh:
## Summary

- Register `qwen/qwen3.7-max` as an Alibaba-direct Kilo model so requests bypass OpenRouter's single-provider route.
- Configure direct model metadata and explicit-cache pricing under the existing Alibaba discount and request-shaping path.
- Add coverage for Alibaba routing registration and Qwen3.7-Max usage/caching cost calculation.

## Verification

- Not manually tested; this is server-side provider-selection and metering logic with no interactive workflow.

## Visual Changes

N/A

## Reviewer Notes

- OpenRouter currently exposes Alibaba as the only endpoint for Qwen3.7-Max; direct routing also uses the existing Alibaba explicit-cache breakpoint handling.

### DIFF
--- a/apps/web/src/lib/ai-gateway/models.test.ts
+++ b/apps/web/src/lib/ai-gateway/models.test.ts
@@ -1,9 +1,14 @@
 import { describe, test, expect } from '@jest/globals';
-import { autoFreeModels, kiloExclusiveModels, requiresKiloDataCollection } from './models';
+import {
+  autoFreeModels,
+  findKiloExclusiveModel,
+  kiloExclusiveModels,
+  requiresKiloDataCollection,
+} from './models';
 import { isFreeModel } from './is-free-model';
 import { getInferenceProvider } from './providers/kilo-exclusive-model';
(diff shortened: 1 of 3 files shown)


lambertjosh:
@arimesser - I think we should go direct because:
1. We do not pay the 5% OR BYOK surcharge.
2. The only provider for 3.7 Max is Alibaba, so there is no value in the routing flexibility.

Can you confirm if our discout applies to 3.7 Max as well? And do a review?

lambertjosh:
Also - in earlier testing the OR cache rate was 0%, they may have fixed it but I feel more comfortable with our own implementation.

arimesser (approved):

lambertjosh:
Validated with local cloud build: 

image

lambertjosh:
Confirmed caching is working: 
image

lambertjosh:
It does look like OR fixed caching on May 24th finally, but will keep an eye that it doesn't break after this.

PR #3463, opened 2026-05-25
docs(experiments): add model experiments spec

markijbema:
## Summary

- Add `.specs/model-experiments.md` as the durable source of truth for model experiment scope, routing, lifecycle, retention, reporting, and secret-handling rules.
- Register the spec in `AGENTS.md` so future model-experiment changes read the business rules before editing the domain.

## Verification

- Not manually exercised; docs-only spec and registration change.

## Visual Changes

N/A

## Reviewer Notes

- The spec is intentionally business-rule focused and excludes implementation-plan details unless they protect an invariant.

### DIFF
--- a/.specs/model-experiments.md
+++ b/.specs/model-experiments.md
@@ -78,7 +78,7 @@ Experiment attribution rows MUST NOT store prompt content in Postgres. They stor
 
 Prompt bodies are stored as full canonical post-transform request bodies in a dedicated per-environment R2 bucket, content-addressed by lowercase sha256. There is one full-body prompt object per unique bounded body; v1 does not store a separate system-prompt object.
 
-Prompt capture is bounded analytics data. The serialized body MUST be capped before it is retained for asynchronous persistence, with deterministic valid-UTF-8 truncation and `was_truncated` recording whether truncation occurred. R2 writes are best-effort: attribution rows SHOULD still land when prompt storage fails, using a reserved sentinel.
+Prompt capture is analytics data. Implementations MAY cap the serialized body before retaining it for asynchronous persistence. If truncation is applied, it MUST be deterministic, preserve valid UTF-8, and record `was_truncated = true`. R2 writes are best-effort: attribution rows SHOULD still land when prompt storage fails, using a reserved sentinel.
 
 Valid prompt references are a 64-character lowercase sha256 hex digest or a reserved sentinel. Reserved sentinels are:
 
--- a/.specs/model-experiments.md
+++ b/.specs/model-experiments.md
@@ -1,0 +1,130 @@
+# Model Experiments
+
+## Role of This Document
+
+This spec defines the business rules and invariants for model experiments. It is the source of truth for what future implementation, refactor, admin, reporting, and retention changes must preserve. It deliberately avoids prescribing internal handler names or UI layout details except where those choices enforce a business rule.
+
+## Status
+
+Draft -- created 2026-05-25.
+
+## Conventions
+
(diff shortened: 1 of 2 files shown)


.specs/model-experiments.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Role of This Document
+
+This spec defines the business rules and invariants for model experiments. It is the source of truth for what future implementation, refactor, admin, reporting, and retention changes must preserve. It deliberately avoids prescribing internal handler names or UI layout details except where those choices enforce a business rule.
+
+## Scope
+
+Model experiments exist only to A/B test preview or otherwise experimental model checkpoints in partnership with model providers. They are not a general-purpose traffic-splitting or rollout mechanism for production models.
+
+An experimented `public_model_id` MUST be a dedicated preview or experiment id that users explicitly select. Production model ids MUST NOT be silently bucketed. Experimented ids MUST NOT be added to `kilo-auto` candidate sets, presets, or other automatic selection paths unless the spec is explicitly changed to allow that behavior.
```
jeanduplessis, line 11: **WARNING:** Scope exclusions are not enforced by the current routing/admin path.

Trace:
1. The spec adds hard guarantees that experimented IDs are dedicated preview/experiment IDs and that `kilo-internal/...` traffic is outside model-experiment routing.
2. Current admin/routing code only requires `public_model_id` to be a non-empty string; `assertActivatable` does not reject production IDs or `kilo-internal/...`; `getProvider` checks experiment membership before `kilo-internal/...` custom LLM routing.
3. An admin can create/activate an experiment for a production model ID or `kilo-internal/...` ID, after which membership causes the experiment path to run despite the new spec saying those states MUST NOT happen.

Impact: The merged spec would claim opt-in/preview-only and `kilo-internal/...` exclusion invariants that current code does not preserve. A misconfigured active experiment can silently reroute production or internal-model traffic and capture experiment attribution/prompts under a contract that says that traffic is out of scope.

.specs/model-experiments.md
```diff
@@ -0,0 +1,122 @@
...
+No other transition is valid. `completed` is terminal and historical. It MUST NOT be used as a temporary traffic-blocking state; use `paused` until the preview public id is removed or a replacement experiment is active.
+
+Routing behavior by status:
+
+- `draft`: invisible to the gateway.
+- `active`: gateway buckets and routes through the experiment.
+- `paused`: gateway returns a local model-unavailable response for the experimented public id and MUST NOT silently fall through to default model routing.
+- `completed`: historical and non-routing; removed from routing membership and eligible to coexist with a draft or active replacement for the same public id.
+
+Activation MUST validate that the experiment has at least two variants, every variant has positive weight, every variant has a current version effective at or before activation time, and no other active or paused experiment targets the same public id.
```
jeanduplessis: **WARNING:** Activation minimum conflicts with current behavior.

Trace:
1. The spec requires activation to validate at least two variants.
2. Current implementation rejects only `variants.length < 1`; the admin UI says “at least 1 variant”; and `model-experiments-router.test.ts` explicitly activates a one-variant experiment successfully.
3. The repository already permits and tests a one-variant active experiment, while the new source-of-truth spec says activation MUST reject it.

Impact: Future reviewers/implementers will rely on a false invariant. One-variant experiments can be activated today, so the experiment may not be an A/B test and cohort/reporting assumptions based on “at least two variants” can be wrong.

jeanduplessis (approved):
Left two warning comments where the proposed source-of-truth spec appears stricter than current model-experiment behavior.

.specs/model-experiments.md
markijbema: This is true; i scoped this out. I think we might want to reconsider how exactly we serve these experiments (it might make sense to combine them with a fallback for when an experiment pauses for instance, and put them behind another model (like kilo-auto/free). However, we do these changes manually so i think we should always do this carefully.

.specs/model-experiments.md
markijbema: oh nice catch, this is actually wrong. Though we started out with a/b testing i also want this to be used with purely sequential testing

PR #3465, opened 2026-05-25
feat(referrals): add Kilo Pass referrals

jeanduplessis:
## Summary
Adds the Kilo Pass Impact Advocate referral program flow, reward lifecycle, attribution safeguards, and post-launch Kilo Pass bonus behavior.

- Adds Kilo Pass-specific Impact Advocate configuration, widget/token routing, participant registration, and referral page UI.
- Records product-scoped referral and affiliate touches, then resolves Kilo Pass paid conversions with referral-priority attribution.
- Grants double-sided Kilo Pass referral rewards for eligible first monthly Stripe conversions, applies one pending reward per future eligible monthly issuance, and blocks stacking with other bonus-like issuance items.
- Queues Impact Advocate reward redemptions for Kilo Pass reward allocations so Impact reporting matches the local reward ledger without making Impact the source of truth.
- Handles Kilo Pass referral adverse-payment cancellation/review flows for refunds, fraud, and chargebacks.
- Adds operational handling for referral state, including an hourly Kilo Pass referral reward expiration cron and a local script for applying pending referral bonuses during manual testing.
- Redirects `/subscriptions/kilo-pass` visitors back to `/subscriptions` when they do not have a Kilo Pass subscription.
- Removes the hidden month-2 50% welcome promo cutoff behavior so Kilo Pass matches the production first-month-only promo copy.

## Verification
Full end-to-end manual verification of the Kilo Pass referral flow with Impact Advocate, including signup attribution, paid conversion, local reward allocation/application, SALE reporting, and Impact Advocate reward redemption retry behavior.

## Visual Changes
image
image
image
Helium 2026-05-26 17 33 27
Helium 2026-05-26 17 33 35

## Reviewer Notes
- Spec alignment: this PR implements and updates behavior governed by `.specs/impact-referrals.md`, especially Kilo Pass program separation, 30-day attribution, reward expiry, Impact redemption reporting, and first-month-only welcome bonus behavior.
- Impact redemption is reporting/reconciliation only. The local referral reward ledger remains authoritative for eligibility, application, cancellation, reversal, billing, and access.
- Operational behavior: Kilo Pass referral reward expiry uses the hourly Vercel cron at `/api/cron/kilo-pass-expire-referral-rewards`; queued affiliate/Advocate dispatch now also backfills missing Kilo Pass reward redemption rows.
- Product isolation: Kilo Pass Advocate config is treated as invalid if it reuses KiloClaw/default program or widget IDs.
- Review focus: attribution resolution for first paid Kilo Pass Stripe invoices, affiliate SALE suppression, `referral_bonus` mutual exclusion with normal bonus issuance, and Impact Advocate redemption retry/failure states.

### DIFF
--- a/apps/web/src/app/api/cron/kilo-pass-expire-referral-rewards/route.ts
+++ b/apps/web/src/app/api/cron/kilo-pass-expire-referral-rewards/route.ts
@@ -1,0 +1,45 @@
+import { timingSafeEqual } from 'node:crypto';
+import { NextResponse } from 'next/server';
+
+import { CRON_SECRET } from '@/lib/config.server';
+import { expirePendingKiloPassReferralRewards } from '@/lib/impact/kilo-pass-referrals';
+import { sentryLogger } from '@/lib/utils.server';
+
+if (!CRON_SECRET) {
+  throw new Error('CRON_SECRET is not configured in environment variables');
+}
+
+function isExpectedCronAuthorization(authHeader: string | null): boolean {
+  if (!authHeader) return false;
+
+  const authHeaderBuffer = Buffer.from(authHeader);
+  const expectedAuthBuffer = Buffer.from(`Bearer ${CRON_SECRET}`);
+  if (authHeaderBuffer.length !== expectedAuthBuffer.length) return false;
+
+  return timingSafeEqual(authHeaderBuffer, expectedAuthBuffer);
+}
+
+export async function GET(request: Request) {
+  const authHeader = request.headers.get('authorization');
+  if (!isExpectedCronAuthorization(authHeader)) {
+    sentryLogger(
+      'cron',
+      'warning'
+    )(
+      'SECURITY: Invalid CRON job authorization attempt: ' +
+        (authHeader ? 'Invalid authorization header' : 'Missing authorization header')
+    );
+    return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
+  }
+
+  const summary = await expirePendingKiloPassReferralRewards();
+
+  return NextResponse.json(
+    {
+      success: true,
+      summary,
+      timestamp: new Date().toISOString(),
+    },
+    { status: 200 }
+  );
+}
--- a/apps/web/src/lib/impact/kilo-pass-referrals.ts
+++ b/apps/web/src/lib/impact/kilo-pass-referrals.ts
@@ -1,0 +1,975 @@
+import 'server-only';
+
+import { addMonths } from 'date-fns';
+import { and, asc, count, eq, inArray, isNull, lte, ne, sql } from 'drizzle-orm';
+
+import { db, type DrizzleTransaction } from '@/lib/drizzle';
+import {
+  IMPACT_ACTION_TRACKER_IDS,
+  buildSalePayload,
+  hashEmailForImpact,
+  isImpactConfigured,
+} from '@/lib/impact';
+import { isImpactAdvocateConfigured } from '@/lib/impact/advocate';
+import {
+  dispatchImpactConversionReportById,
+  queueImpactAdvocateRewardRedemption,
+  resolveWinningAttributionTouch,
+  type AdverseReferralPaymentReason,
+} from '@/lib/impact/kiloclaw-referrals';
+import { hashNormalizedEmailForDeletionTombstone } from '@/lib/impact/referral';
+import { logImpactReferralDebug } from '@/lib/impact/debug';
+import { KILO_PASS_TIER_CONFIG } from '@/lib/kilo-pass/constants';
+import {
+  deleted_user_email_tombstones,
+  impact_advocate_participants,
+  impact_attribution_touches,
+  impact_conversion_reports,
+  impact_referral_conversions,
+  impact_referral_reward_decisions,
+  impact_referral_rewards,
+  impact_referrals,
+  kilo_pass_issuances,
+  kilo_pass_subscriptions,
+  kilocode_users,
+  user_affiliate_attributions,
+  type ImpactAttributionTouch,
+} from '@kilocode/db/schema';
+import {
+  ImpactAdvocateProgramKey,
+  ImpactAttributionTouchType,
+  ImpactConversionReportState,
+  ImpactReferralBeneficiaryRole,
+  ImpactReferralDecisionOutcome,
+  ImpactReferralPaymentProvider,
+  ImpactReferralProduct,
+  ImpactReferralRewardKind,
+  ImpactReferralRewardStatus,
+  ImpactReferralWinningTouchType,
+  KiloPassCadence,
+  KiloPassWelcomePromoEligibilityReason,
+  type KiloPassTier,
+} from '@kilocode/db/schema-types';
+
+type DatabaseClient = typeof db | DrizzleTransaction;
+
+export type KiloPassPaidConversionDisposition = {
+  shouldEnqueueAffiliateSale: boolean;
+  winningTouchType: 'referral' | 'affiliate' | 'none';
+  conversionId: string | null;
+  disqualificationReason: string | null;
+};
+
+export type KiloPassAdverseReferralPaymentSummary = {
+  conversionId: string | null;
+  canceledRewards: number;
+  reviewRequiredRewards: number;
+};
+
+export type KiloPassReferralRewardExpirationSummary = {
+  expiredRewards: number;
+};
+
+export const KILO_PASS_REFERRER_REWARD_CAP = 5;
+const KILO_PASS_REFERRAL_REWARD_PERCENT = 0.5;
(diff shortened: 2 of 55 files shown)


apps/web/src/routers/kilo-pass-router.test.ts
```diff
@@ -1449,25 +1557,25 @@ describe('kiloPassRouter', () => {
...
     });
 
-    it('returns isEligibleForFirstMonthPromo=false after the promo cutoff', async () => {
-      freezeKiloPassClock(PROMO_OFFER_EXPIRED_TEST_TIME);
+    it('keeps first-month promo eligibility after the first-month promo check', async () => {
+      freezeKiloPassClock(LATER_FIRST_MONTH_PROMO_TEST_TIME);
 
       const user = await insertTestUser({
-        google_user_email: 'kilo-pass-promo-expired-no-sub@example.com',
+        google_user_email: 'kilo-pass-promo-after-first-month promo-no-sub@example.com',
```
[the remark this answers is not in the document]
jeanduplessis: Fixed by removing the accidental space from this test email fixture.

apps/web/src/routers/kilo-pass-router.test.ts
```diff
@@ -1449,25 +1557,25 @@ describe('kiloPassRouter', () => {
...
       expect(result.subscription).toBeNull();
     });
 
     it('keeps isEligibleForFirstMonthPromo=true for a never-subscribed user', async () => {
-      freezeKiloPassClock(PROMO_OFFER_ACTIVE_TEST_TIME);
+      freezeKiloPassClock(FIRST_MONTH_PROMO_TEST_TIME);
 
       const user = await insertTestUser({
-        google_user_email: 'kilo-pass-promo-cutoff-still-eligible@example.com',
+        google_user_email: 'kilo-pass-promo-first-month promo check-still-eligible@example.com',
```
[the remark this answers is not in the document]
jeanduplessis: Fixed by removing the accidental spaces from this test email fixture.

apps/web/src/routers/kilo-pass-router.ts
```diff
@@ -164,6 +198,12 @@ type StripeManagedKiloPassSubscription = KiloPassSubscriptionState & {
   stripeSubscriptionId: string;
 };
```
St0rmz1, line 204: **Moderate, Design:** This duplicates `KILO_PASS_REFERRER_REWARD_CAP` from `apps/web/src/lib/impact/kilo-pass-referrals.ts`. The first is the value enforced when granting rewards; this one is the value the referral page shows the user via `getReferralRewardSummary` (`reached: grantedRewards >= KILO_PASS_REFERRER_REWARD_CAP`).

If product later changes the cap to 10, it is easy to update one site and miss the other. The UI would then say "Cap reached at 10" while the backend silently still enforces 5, or vice versa.

Suggestion: export the constant from `lib/impact/kilo-pass-referrals.ts` (or a shared constants module) and import it here so a single edit propagates.

apps/web/src/lib/impact/kilo-pass-referrals.ts
```diff
@@ -0,0 +1,966 @@
...
+      affiliateTouchId: resolution.affiliateTouch?.id ?? null,
+      referralTouchId: resolution.referralTouch?.id ?? null,
+    });
+
+    // Preserve pre-touch-model affiliate attribution without allowing an expired scoped touch
+    // to bypass the referral attribution window.
+    if (
+      resolution.winner === 'none' &&
+      touches.length === 0 &&
+      (await hasHistoricalImpactAffiliateAttribution({ userId: params.userId, database: tx }))
```
St0rmz1, line 555: **Moderate, Functional:** When a Kilo Pass paid conversion has no Kilo Pass-scoped touches but the user has any `user_affiliate_attributions` row with `provider = 'impact'`, this branch records the conversion as `winningTouchType = 'affiliate'` with `shouldEnqueueAffiliateSale: true` and a Kilo Pass affiliate SALE is reported to Impact.

`user_affiliate_attributions` is not product-scoped, so a user whose only historical Impact attribution came from a KiloClaw signup will still trigger a Kilo Pass affiliate SALE on their first Kilo Pass paid conversion. The KiloClaw equivalent (`kiloclaw-referrals.ts`) does not have this fallback, so the asymmetry is real, not a parity issue. The commit `feat(referrals): preserve historical Kilo Pass affiliate sales` suggests this is intentional, but it is worth a sanity check: if Impact contracts expect Kilo Pass SALEs to be driven by Kilo Pass touches only, this over-attributes to affiliates whose touch predates the Kilo Pass program.

Suggestion: confirm with the affiliate-payments owner that "any historical impact attribution -> Kilo Pass affiliate SALE" is contractually correct. If it is not, narrow `hasHistoricalImpactAffiliateAttribution` to attributions that have any Kilo Pass linkage, or record the conversion without enqueueing the affiliate SALE in this branch.

apps/web/src/app/api/cron/kilo-pass-expire-referral-rewards/route.ts
```diff
@@ -0,0 +1,35 @@
...
+import { sentryLogger } from '@/lib/utils.server';
+
+if (!CRON_SECRET) {
+  throw new Error('CRON_SECRET is not configured in environment variables');
+}
+
+export async function GET(request: Request) {
+  const authHeader = request.headers.get('authorization');
+  const expectedAuth = `Bearer ${CRON_SECRET}`;
+  if (authHeader !== expectedAuth) {
```
St0rmz1: **Low, Security:** `authHeader !== expectedAuth` is not timing-constant. An attacker who can hit this endpoint and measure response time could in theory probe the cron secret byte by byte. For a Vercel-fronted internal cron the practical risk is low.

This is the same pattern as the other cron routes, so it is a codebase-wide convention rather than something new in this PR. Up to you whether to harden here or track it separately, but flagging it because the cron secret protects the only auth boundary on this endpoint.

Suggestion: compare with `crypto.timingSafeEqual(Buffer.from(authHeader ?? ''), Buffer.from(expectedAuth))` after a length check to avoid the timing-leak.

apps/web/src/lib/impact/kiloclaw-referrals.ts
```diff
@@ -155,25 +158,30 @@ function hasAcceptedTrackingValue(touch: ImpactAttributionTouch): boolean {
 }
```
St0rmz1, line 159: **Low, Functional:** Validity is now computed from `touched_at + IMPACT_REFERRAL_TOUCH_VALIDITY_MS` and the stored `touch.expires_at` is no longer consulted. The kiloclaw test at `kiloclaw-referrals.test.ts` had to be updated to align `touched_at` with `expires_at` because the old expiration data became irrelevant.

Today the writer always stamps `expires_at = touched_at + 30 days`, so the two are equivalent in practice. But the column is now effectively dead: any future writer who stamps a non-30-day expiry (a test fixture, a one-off migration, a feature flag) will be silently overridden by the 30-day constant.

Suggestion: either drop `expires_at` from `impact_attribution_touches` (and remove the schema field) so there is one source of truth, or keep using the stored value here. Leaving both in but only honoring one is the worst of both worlds. At minimum a comment on `IMPACT_REFERRAL_TOUCH_VALIDITY_MS` noting it overrides any stored expiry would help future readers.

St0rmz1 (approved):

apps/web/src/routers/kilo-pass-router.ts
jeanduplessis: Fixed by exporting the Kilo Pass referrer cap from the reward-granting module and importing that same constant in the router summary, so UI and grant enforcement share one source of truth.

apps/web/src/lib/impact/kilo-pass-referrals.ts
jeanduplessis: Kept the historical affiliate fallback intentionally: the affiliate spec treats the provider attribution row as the SALE eligibility gate for attributed Kilo Pass users, including degraded historical rows without a click ID. The fallback is still blocked when any product-scoped touch exists, so expired Kilo Pass touches cannot bypass the 30-day referral/affiliate window. I added a code comment documenting that boundary.

apps/web/src/app/api/cron/kilo-pass-expire-referral-rewards/route.ts
jeanduplessis: Fixed by switching this cron auth check to a length-checked timingSafeEqual comparison.

apps/web/src/lib/impact/kiloclaw-referrals.ts
jeanduplessis: Added an inline note at the conversion-time validity check explaining that expires_at is denormalized metadata and qualification intentionally follows the referral spec's touched_at + 30 * 24h UTC rule.

PR #3467, opened 2026-05-25
fix(code-reviews): gracefully handle blocked models

alex-alecu:
## Why

Some code reviews failed when a team blocked the selected model. The review should stop cleanly and tell the user to choose a model their team allows.

## What changed

Code reviews now treat blocked models as cancelled instead of failed. GitHub and GitLab show a clear message that the selected model is not allowed for the team. Health checks and admin reports count these runs as cancellations, not system errors. Old saved failures with the same message are handled the same way.

## How to test

1. Start a code review with a model that is not allowed for the team.
2. Confirm the PR or MR status is cancelled.
3. Confirm the summary says the selected model is not allowed for the team and points to code review settings.
4. Confirm admin reporting and health checks do not count it as a system failure.

### DIFF
--- a/apps/web/src/app/api/code-reviews/up/route.test.ts
+++ b/apps/web/src/app/api/code-reviews/up/route.test.ts
@@ -14,6 +14,7 @@ import { eq } from 'drizzle-orm';
 import { GET } from './route';
 
 const REPO = `test-org/code-review-up-${Date.now()}`;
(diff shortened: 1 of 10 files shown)


alex-alecu:
If a model is not available anymore, we will disable code-reviews for that org / individual until they choose an available model. I will handle `model_not_found` in a different PR together with other misconfigs that fall in the same bucket and have the same outcome.

PR #3471, opened 2026-05-25
feat(cloud-agent-next): reject unavailable models before admission

eshurakov:
## Summary

- Adds lightweight personal and organization model-validation API contracts so Cloud Agent can check selected Kilo model availability against the effective catalog before dispatch.
- Architecturally, `cloud-agent-next` now performs prompt-only model preflight at prepare, start, and send admission boundaries, failing unavailable selections synchronously and treating unavailable validation infrastructure as retryable failure.
- Aligns dispatched model normalization and extends the fake gateway/E2E harness to prove rejected models create no sandbox work or chat-completion dispatch.

## Verification

- [x] Confirmed accepted models complete a cold turn plus hot follow-up turns through both the unified and legacy API surfaces using the deterministic local fake gateway.
- [x] Confirmed an unavailable initial model is rejected before sandbox creation or chat-completion dispatch through both API surfaces using the fake validation gateway.
- [x] Confirmed an unavailable initial model is rejected before sandbox creation through both API surfaces when validation routes through the local Next.js gateway.
- [x] Confirmed unavailable follow-up sends reject without an added chat-completion dispatch for both API surfaces; the gated legacy turn completes after release with no remaining fake-gateway waiters.

## Reviewer Notes

- Focus on catalog-source resolution in `services/cloud-agent-next/src/model-validation.ts`, including official versus override/token-routed validation and scoped `401` fallback behavior.
- Prompt-only enforcement is intentional; command-valued prepared sessions remain outside this change.
- The preflight adds an availability request before accepted prompt work and fails closed when availability cannot be established.


services/cloud-agent-next/src/model-validation.ts
```diff
@@ -0,0 +1,247 @@
...
+  organizationId: string | undefined
+): string {
+  const trimmed = baseURL.replace(/\/+$/, '');
+  let modelsBaseUrl: string;
+  if (organizationId) {
+    modelsBaseUrl = trimmed.includes('/api/organizations/')
+      ? trimmed
+      : trimmed.endsWith('/api')
+        ? `${trimmed}/organizations/${organizationId}`
+        : `${trimmed}/api/organizations/${organizationId}`;
```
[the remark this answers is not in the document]
eshurakov: Fixed in `974b45399`: organization IDs inserted into constructed override validation paths now use `encodeURIComponent`, with regression coverage for path and query delimiters in both organization URL forms.

services/cloud-agent-next/src/session/model-preflight.ts
```diff
@@ -0,0 +1,61 @@
...
+  );
+}
+
+export async function preflightPreparedInitialPromptModel(
+  input: StoredSessionPreflightInput
+): Promise<void> {
+  const metadata = await requireSessionMetadata(input);
+  const turn = metadata.initialMessage?.turn;
+  if (turn?.type === 'command') return;
+  if (turn?.type !== 'prompt' && metadata.initialMessage?.prompt === undefined) return;
```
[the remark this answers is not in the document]
eshurakov: Clarified in `974b45399`: incomplete legacy prepared metadata is now covered by a characterization test and documented as intentionally deferring to the existing admission-time `No prompt provided` error.

services/cloud-agent-next/src/router/handlers/session-send.ts
```diff
@@ -29,21 +30,27 @@ const sendMessageHandler = protectedProcedure
...
+          prompt: input.message.prompt,
+          images: input.message.images,
         },
-        { env: ctx.env, userId: ctx.userId, botId: ctx.botId }
-      );
+        agent: input.agent,
+        finalization: input.finalization,
+      };
+      const admissionContext = { env: ctx.env, userId: ctx.userId, botId: ctx.botId };
+      await preflightExistingPromptModel({
```
alex-alecu, line 48: Could this preflight run only for a new admission? A retry with an already admitted `message.id` currently reaches model validation before `queueMessage` can return its stored ack, so model removal or a validation outage turns an idempotent retry into an error.

services/cloud-agent-next/src/model-validation.ts
```diff
@@ -0,0 +1,248 @@
...
+    method: 'POST',
+    headers: requestHeaders(context),
+    body: JSON.stringify({ modelId }),
+  });
+  if (!response) return { type: 'validation-unavailable', source: 'official' };
+  if (response.status === 401 && (context.token || context.organizationId)) {
+    return validateFromOfficialSource(env, modelId, anonymousCatalogContext(context.feature));
+  }
+  if (response.status === 403) return { type: 'access-denied', source: 'official' };
+  if (!response.ok) return { type: 'validation-unavailable', source: 'official' };
```
alex-alecu, line 107: Can a missing validation route be handled during rollout before failing closed? `apps/web` and this Worker deploy independently, so a Worker-first rollout maps a 404 from the new route to `SERVICE_UNAVAILABLE` and blocks official-source prompt admissions.

apps/web/src/app/api/openrouter/models/validate/route.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  const feature = validateFeatureHeader(request.headers.get(FEATURE_HEADER));
+  const auth = await tryGetUserFromAuth();
+  try {
+    const models = await getEnhancedOpenRouterModels();
+    if (!Array.isArray(models.data)) {
+      throw new Error('Model catalog returned invalid data');
+    }
+    const byokModels = auth?.user ? await getDirectByokModelsForUser(auth.user.id) : [];
```
alex-alecu: Could the BYOK and experiment catalog reads run in parallel here? This route is now on every prompt admission attempt, so serial independent reads add their latency before work can be queued.

eshurakov: Fixed in `01a08128b`: the personal validation route now loads BYOK and experiment model catalogs concurrently with `Promise.all`, with coverage for concurrent lookup startup.

services/cloud-agent-next/src/router/handlers/session-send.ts
eshurakov: Fixed in `01a08128b`: prompt retries with an already durable `message.id` now return the stored admission acknowledgement before model preflight; only requests without an existing admission proceed through validation before queueing.

services/cloud-agent-next/src/model-validation.ts
eshurakov: Fixed in `01a08128b`: a `404` from an official compact validation route falls back to the existing same-scope GET catalog during rollout. Override-selected sources remain fail-closed.

eshurakov: Follow-up in `feab35479`: simplified rollout handling per discussion. An official validation-route `404` now skips the model check without fetching a fallback catalog; override-selected validation routes still fail closed.

alex-alecu (approved):

PR #3478, opened 2026-05-26
fix(gitlab): restore glab oauth mode

alex-alecu:
## Summary
- Restore GitLab command-line OAuth mode when the agent gives `glab` a managed GitLab token.
- Keep user-set OAuth mode values unchanged and cover the queued review path that hit the 401.

## Verification
1. After deployment, run a GitLab code review for an OAuth-connected project.
2. Confirm the review note is posted or updated without a 401.
3. Run a follow-up review on the same session and confirm it publishes too.

## Visual Changes
N/A

## Reviewer Notes
This fixes the missing bearer-mode flag only. It does not change the separate project-token author issue or add proof that the note was published.

### DIFF
--- a/services/cloud-agent-next/src/session-service.ts
+++ b/services/cloud-agent-next/src/session-service.ts
@@ -847,6 +847,7 @@ export class SessionService {
     githubToken?: string;
     gitUrl?: string;
     gitToken?: string;
+    gitlabTokenManaged?: boolean;
     upstreamBranch?: string;
     branchName?: string;
     envVars?: Record<string, string>;
@@ -875,6 +876,7 @@ export class SessionService {
       githubToken: options.githubToken,
       gitUrl: options.gitUrl,
       gitToken: options.gitToken,
+      gitlabTokenManaged: options.gitlabTokenManaged,
       platform: options.platform,
       envVars: options.envVars,
     };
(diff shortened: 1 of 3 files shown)


services/cloud-agent-next/src/session-service.ts
```diff
@@ -1116,6 +1116,9 @@ export class SessionService {
     // Set GITLAB_TOKEN for GitLab repos, respecting user overrides
     if (gitToken && effectivePlatform === 'gitlab' && !baseEnvVars.GITLAB_TOKEN) {
       envVars.GITLAB_TOKEN = gitToken;
+      if (baseEnvVars.GLAB_IS_OAUTH2 === undefined) {
+        envVars.GLAB_IS_OAUTH2 = 'true';
```
alex-alecu, line 1124: Could this be limited to managed GitLab tokens? For legacy `type: git` sessions, `effectivePlatform` is inferred from a GitLab URL while `gitToken` is the stored arbitrary token, so marking it OAuth can make `glab` authenticate a PAT or project token in bearer mode and return 401.

alex-alecu: Fixed in d18d71b5f.

markijbema (approved):

PR #3490, opened 2026-05-26
fix(gitlab): restore review bot attribution

alex-alecu:
## Summary

GitLab code reviews could post comments as the connected person instead of the review bot after a review session was saved or reused. Reviews now keep a safe pointer to the bot credential and load the current bot credential when work starts or continues, without copying the secret into session data. Continued reviews are limited to the same GitLab project and server, and use the correct sign-in mode for the bot token.

- `prepareReviewPayload` already creates/gets the GitLab Project Access Token (PrAT), and before this PR it passed it as `gitToken`.
- But for `platform: 'gitlab'`, `cloud-agent-next` treats GitLab as a managed provider and does not persist/use that raw `gitToken` in the prepared session metadata.
- When the workspace is actually prepared, it resolves GitLab credentials through `getGitLabToken`, which returns the connected user/integration token, not the project bot token.
- So the first cloud-agent run can already post as the connected person, even though earlier GitLab API lookups in `apps/web` used the PrAT correctly.

### Why not use the GitHub flow directly

- GitHub has a durable installation identity. The session can store/resolve `githubInstallationId`, and GitHub App tokens naturally post as the bot/app.
- GitLab bot attribution depends on a per-project Project Access Token stored under the GitLab integration metadata.
- A repo URL/user/org is not enough to safely infer the right token, especially with multiple GitLab integrations or self-hosted GitLab instances.
- Storing the raw PrAT in session metadata would be worse. This PR stores only `{ integrationId, projectId }` and resolves the current token at runtime.

### Why so many files changed

The core change is not huge; the file count is mostly because the credential has to cross multiple typed/service boundaries safely.

## Verification

1. Start a GitLab code review on a test merge request and confirm the comment author is the Kilo Code Review Bot.
2. Push another commit to that merge request, run the next review, and confirm the new comment still comes from the bot.
3. Reconnect the GitLab integration to a different test server or project and confirm an old session is not reused to post to the new review.

## Reviewer Notes

The session stores only a project-token reference; the current bot token is read at run time. If that reference is missing, points at another GitLab server, or cannot be read safely, the review fails instead of posting as the connected person.


alex-alecu:
Manual test passed.

Tested:
- Local GitLab review persistence flow with a current review, previous completed review, GitLab integration, and project-token metadata.
- git-token-service dev RPC lookup through the HTTP test worker for the GitLab code-review token reference.
- cloud-agent-next prepare/session credential flow for GitLab code-review sessions, including the missing-project-token edge case.
- Durable Object metadata path for GitLab code-review token references and continuation updates.

Verified:
- Persisted review rows retain the GitLab integration ID, numeric project ID, and previous session for same-project continuation.
- Project-token metadata is present for the GitLab project without needing to expose the raw token in verification output.
- The credential source resolves as the code-review project access token and does not fall back to the connected-user GitLab token.
- Missing project-token lookup fails safely with no managed-token fallback.

apps/web/src/lib/code-reviews/db/code-reviews.ts
```diff
@@ -1444,20 +1444,37 @@ export async function cancelSupersededReviewsForPR(
...
+ * GitLab continuation requires the exact integration and project identity so
+ * equivalent project paths on separate GitLab instances never share sessions.
  */
 export async function findPreviousCompletedReview(
   repoFullName: string,
   prNumber: number,
   excludeSha: string,
-  platform: string = 'github'
+  platform: string = 'github',
+  gitLabScope?: GitLabReviewContinuationScope
```
RSO: Shouldn't this be ` {platform: "github" } | { platform: "gitlab", gitLabScope: GitLabReviewContinuationScope }`? To indicate that gitLabScope is non-optional when platform == gitlab?

apps/web/src/lib/code-reviews/db/code-reviews.ts
```diff
@@ -1444,20 +1444,37 @@ export async function cancelSupersededReviewsForPR(
...
 /**
  * Finds the most recent completed review for the same PR with a different SHA.
  * Used for incremental reviews: returns the previous HEAD SHA so the agent
  * can diff against it instead of re-reviewing the entire PR.
  * Also returns session_id (nullable) so the caller can derive both the
  * incremental diff base and the session continuation target from a single row.
+ * GitLab continuation requires the exact integration and project identity so
+ * equivalent project paths on separate GitLab instances never share sessions.
  */
 export async function findPreviousCompletedReview(
```
RSO, line 1460: It feels like we're making the wrong abstraction here by pushing the `if (platform === gitlab)` logic down. Also, because of the broken type, there now is a logical path where `platform===gitlab` and `gitLabScope === undefined`, but it sounds like we never want that to happen.

RSO, line 1460: Actually, come to think of it. Why are we not always sending the `platform_integration_id` and `projectId`? They should be unique enough for GitHub as well, right? That would remove the branch here and at the call site and clean up the code tremendously.

packages/worker-utils/src/cloud-agent-next-client.ts
```diff
@@ -26,6 +31,7 @@ export type CloudAgentPrepareSessionInput = {
   githubToken?: string;
   gitUrl?: string;
   gitToken?: string;
+  gitlabCodeReviewTokenRef?: GitLabCodeReviewTokenRef;
```
RSO: This also feels like it's the wrong abstraction. To now have code-review logic in the CloudAgent Next call site. Why is this not part of the callbackTarget?

alex-alecu:
Closing and reopening to check if the latest commits will appear

apps/web/src/lib/code-reviews/db/code-reviews.ts
alex-alecu: Fixed in `1dc4dd0ea`: the missing GitLab continuation identity case is now handled in `prepareReviewPayload`; the DB helper no longer exposes a callable GitLab-without-scope path or `sql false` fallback.

apps/web/src/lib/code-reviews/db/code-reviews.ts
alex-alecu: Fixed in `1dc4dd0ea`: GitLab continuation now uses a discriminated `ReviewContinuationScope`, so the GitLab variant requires exact `integrationId` and `projectId` at the type boundary.

apps/web/src/lib/code-reviews/db/code-reviews.ts
alex-alecu: Fixed in `1dc4dd0ea`: exact integration/project filtering remains GitLab-specific because GitHub historical reviews do not store a numeric `platform_project_id`, and GitHub `owner/repo` is already globally qualified. The invalid optional GitLab path has been removed.

packages/worker-utils/src/cloud-agent-next-client.ts
alex-alecu: Fixed in `05b10167b`: `gitlabCodeReviewTokenRef` was removed from Cloud Agent prepare/update payloads. GitLab code-review token selection now occurs inside `git-token-service` from runtime repository context, while `callbackTarget` remains callback-routing data only.

services/git-token-service/src/gitlab-runtime-token-resolver.ts
```diff
@@ -0,0 +1,193 @@
...
+  );
+  const qualifiedCandidates = evaluations.flatMap(evaluation =>
+    evaluation.status === 'qualified' ? [evaluation.candidate] : []
+  );
+
+  if (qualifiedCandidates.length > 1) {
+    return { success: false, reason: 'ambiguous_integration' };
+  }
+
+  const tokenFailure = evaluations.find(evaluation => evaluation.status === 'token_failed');
```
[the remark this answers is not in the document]
alex-alecu: Fixed in d9d8906.

alex-alecu:
Manual GitLab code-review attribution verification passed.

What I exercised locally:
- Sent a GitLab Merge Request webhook through `/api/webhooks/gitlab` using a fake GitLab API/repo and fake LLM.
- Confirmed the webhook queued and completed a GitLab code review: `e1274052-7525-4d4f-b675-007a1a8a5262`.
- Confirmed bot-attributed GitLab API operations used the project access token, including `/api/v4/user`, commit statuses, MR award emoji, notes, discussions, and MR lookups.
- Confirmed runtime token resolution for `createdOnPlatform: code-review` returned the project token with `glabIsOAuth2: false`.
- Confirmed removing the project token returns `success: false, reason: no_project_token` and does not fall back to the connected-user token.

Local-only setup used a fake HTTPS GitLab server, a bare test repo, and the deterministic fake LLM; no production services were used.

apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
```diff
@@ -461,7 +471,7 @@ export async function prepareReviewPayload(
       platform === 'gitlab'
         ? {
             // GitLab: use full git URL for cloning
-            gitUrl: `${gitlabInstanceUrl || 'https://gitlab.com'}/${review.repo_full_name}.git`,
+            gitUrl: `${(gitlabInstanceUrl || 'https://gitlab.com').replace(/\/+$/, '')}/${review.repo_full_name}.git`,
```
RSO: Is there some new branch in the code where `gitlabInstanceUrl` can suddenly contain a trailing slash, or is this an unrelated fix? Either way the replace feels a bit out-of-place here 🤔, is this the only place where we construct a gitlab URL with this variable?

services/cloud-agent-next/src/router/handlers/session-prepare.ts
```diff
@@ -369,7 +368,9 @@ const updateSessionHandler = internalApiProtectedProcedure
       );
       const stub = ctx.env.CLOUD_AGENT_SESSION.get(doId);
 
-      const result = await stub.tryUpdate({ callbackTarget: input.callbackTarget });
+      const result = await stub.tryUpdate({
+        callbackTarget: input.callbackTarget,
+      });
```
RSO: It seems to me that nothing changed in this file other than formatting and comments?

services/cloud-agent-next/src/router/schemas.ts
```diff
@@ -665,7 +665,7 @@ export const SendMessageInput = z
 
 /**
  * Input schema for updateSession endpoint.
- * Retained only for rewriting callbackTarget on session continuations.
+ * Retained for internal callback target updates before continued executions.
```
RSO: Only change in this file is a comment, seems unnecessary?

services/cloud-agent-next/test/helpers/session-setup.ts
```diff
@@ -99,7 +99,6 @@ export function groupedRegisterSessionInput(input: TestRegisterSessionInput): Re
       ? {
           type: 'gitlab',
           url: input.gitUrl,
-          token: input.gitToken,
```
RSO, line 105: This feels like a weird, isolated change. Why is this needed?

services/git-token-service/src/gitlab-lookup-service.ts
```diff
@@ -7,6 +7,8 @@ import {
 } from '@kilocode/db/schema';
 import { eq, and, isNull, isNotNull, sql } from 'drizzle-orm';
 
+const DEFAULT_GITLAB_INSTANCE_URL = 'https://gitlab.com';
```
RSO: It feels like I saw the string `https://gitlab.com` in other places in this PR already, so maybe this constant should be hoisted up somewhere so that it can be reused?

RSO (approved):
Code looks good to me, but I'd love it if @eshurakov can check out the git-token-service stuff, to be sure.

apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
alex-alecu: Fixed in 6f167d5.

services/cloud-agent-next/src/router/handlers/session-prepare.ts
alex-alecu: Fixed in 2bf5aa7.

services/cloud-agent-next/src/router/schemas.ts
alex-alecu: Fixed in 5ff0e92.

services/cloud-agent-next/test/helpers/session-setup.ts
alex-alecu: No code change here. GitLab session records do not keep the token on the repository object now; the helper stores it when the session becomes ready.

services/git-token-service/src/gitlab-lookup-service.ts
alex-alecu: Fixed in 73f4bf5.

PR #3504, opened 2026-05-26
feat(gateway): enforce abuse rule actions

jrf0110:
## Summary

Makes our gateway enforce abuse rules-engine actions. We keep the normal request path fast by using Redis as a lightweight action cache (only identities with a cached blocking/quarantine action wait on a fresh abuse check).

- `nothing`: no gateway enforcement; cached as `none`.
- `log`: no gateway enforcement, but the action is cached so future requests stay on the nonblocking path.
- `rate-limit`: return a request-local `429 rate_limit_exceeded` before upstream.
- `block`: return a request-local `403 abuse_blocked` before upstream.
- `quarantine-1`: add 2s of artificial latency, then continue normally.
- `quarantine-2`: add 6s of artificial latency, then continue normally.
- `quarantine-3`: add 6s of artificial latency and, for non-BYOK requests, rewrite to an auto-free model when one is available; BYOK and no-candidate cases are latency-only.

## Verification
- Ran the local abuse worker at `http://localhost:5173` and the cloud web server at `http://localhost:3000` with `ABUSE_SERVICE_URL=http://localhost:5173`, backed by local Postgres and Redis.
- Created temporary active abuse rules through `/trpc/rules.create` for `block`, `rate-limit`, `quarantine-1`, `quarantine-2`, `quarantine-3`, and `log`, each matching a unique `request.user_agent` marker.
- Sent direct `/api/classify` requests to confirm each temporary rule resolved to the expected `rules_engine.resolved_action` and matched the expected rule ID.
- Sent gateway requests to `/api/openrouter/v1/chat/completions` and confirmed the Redis-gated behavior: the first request classified/cached, and the second request enforced cached blocking actions.
- Confirmed `block` returned `403 abuse_blocked`, `rate-limit` returned `429 rate_limit_exceeded`, `quarantine-1` added about 2s latency, `quarantine-2` added about 6s latency, `quarantine-3` added about 6s latency, and `log` continued without enforcement.
- Seeded local OpenRouter model metadata, restarted the web server, and confirmed non-BYOK `quarantine-3` rewrote to a free Nemotron model before proxying.
- Deleted all temporary smoke-test abuse rules through `/trpc/rules.delete`, cleared Redis abuse classification cache keys, removed the seeded OpenRouter metadata cache, and restarted the web server to clear in-memory metadata.

## Visual Changes
N/A

## Reviewer Notes
- `block` remains request-local and does not mutate `kilocode_users.blocked_reason`.
- Quarantine latency is currently 2s for level 1 and 6s for levels 2/3.
- Quarantine-3 degrades to latency-only if no auto-free candidate is available for the request API kind.
- Redis stores only the last rules-engine action string (`block`, `rate-limit`, `quarantine-*`, `log`, or `none`) with no expiry, keyed by the abuse identity shape for users/fingerprints.


apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -461,6 +464,66 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
     return forbiddenFreeModelResponse(fraudHeaders);
   }
 
+  const classifyResult = await awaitClassifyAbuse(classifyPromise);
```
Comment (jrf0110) [inline apps/web/src/app/api/openrouter/[...path]/route.ts]: We need to be extra sure about this change. Blocking upstream requests on the abuse service is a big deal. I need to make sure that a failed abuse request doesn't fail the gateway request

apps/web/src/lib/ai-gateway/abuse-service.ts
```diff
@@ -189,8 +200,222 @@ export type AbuseClassificationResponse = {
...
+    value === 'log' ||
+    value === 'rate-limit' ||
+    value === 'quarantine-1' ||
+    value === 'quarantine-2' ||
+    value === 'quarantine-3' ||
+    value === 'block'
+  );
+}
+
+function parseCachedRulesEngineResult(raw: string): RulesEngineClassificationResult | null {
```
jrf0110: We should probably use zod for this

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -461,6 +464,66 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
...
+    providerResult = await getProvider({
+      requestedModel: originalModelIdLowerCased,
+      request: requestBodyParsed,
+      user,
+      organizationId,
+      taskId,
+      clientIp: ipAddress ?? null,
+      machineId: machineIdHeader,
+    });
+    if (providerResult.kind === 'not-found') {
```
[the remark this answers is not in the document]
Comment (jrf0110) [reply apps/web/src/app/api/openrouter/[...path]/route.ts]: Fixed. The not-found and unavailable model-override provider failure paths now apply the quarantine delay before returning, and there is a regression test covering this path.

apps/web/src/app/api/openrouter/[...path]/route.ts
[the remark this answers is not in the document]
Comment (jrf0110) [reply apps/web/src/app/api/openrouter/[...path]/route.ts]: Addressed by the Redis cache gate. Classification no longer blocks every request. It only blocks when Redis has a cached blocking action for that abuse identity; nonblocking/missing cache entries run classification in parallel with upstream and update Redis when classification returns.

Comment (jrf0110) [reply apps/web/src/app/api/openrouter/[...path]/route.ts]: Addressed by changing the enforcement flow to be Redis cache-gated. The gateway now always only blocks on the Redis lookup first. If Redis is unavailable, the request proceeds normally. Abuse classification only blocks the upstream request when the cached rules-engine action is already blocking; if that blocking refresh fails, the cached decision is used. Otherwise classification runs in parallel with upstream and only updates the cache for the next request.

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -461,6 +475,81 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
...
+
+    provider = providerResult.provider;
+    userByok = providerResult.userByok;
+    bypassAccessCheck = providerResult.bypassAccessCheck;
+    skipProviderPin = providerResult.skipProviderPin;
+    skipKiloExclusiveModelSettings = providerResult.skipKiloExclusiveModelSettings;
+    experiment = providerResult.experiment;
+
+    if (!provider.supportedChatApis.includes(requestBodyParsed.kind)) {
+      return apiKindNotSupportedResponse(
```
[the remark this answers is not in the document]
Comment (jrf0110) [reply apps/web/src/app/api/openrouter/[...path]/route.ts:570]: Fixed. The unsupported-API-kind branch inside the quarantine-3 model-override path now applies the same delay before returning `apiKindNotSupportedResponse`, and I added a focused regression test for that path.

apps/web/src/app/api/openrouter/[...path]/route.test.ts
```diff
@@ -0,0 +1,240 @@
...
+    organizationId: undefined,
+  });
+  mockedGetBalanceAndOrgSettings.mockResolvedValue({
+    balance: 1000,
+    settings: undefined,
+    plan: undefined,
+  });
+}
+
+function classifyResult(action: 'block' | 'rate-limit' | 'quarantine-3' | null) {
```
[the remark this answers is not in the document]
Comment (jrf0110) [reply apps/web/src/app/api/openrouter/[...path]/route.test.ts]: Fixed. Added regression coverage for quarantine-1 and quarantine-2 that verifies the configured delay is applied, upstream is eventually called, and the model is not rewritten.

apps/web/src/lib/ai-gateway/abuse-service.ts
jrf0110: Fixed. Replaced the manual cached rules-engine parsing with Zod schemas for the action enum and cached classification result, and derive the TypeScript types from those schemas.

jrf0110:
Addressed the outstanding review summary items: the quarantine-3 unsupported-API-kind model override path now applies delay before returning, and cached rules-engine JSON now validates through Zod. Also replied to and resolved all review threads.

apps/web/src/lib/ai-gateway/abuse-service.ts
```diff
@@ -189,8 +218,183 @@ export type AbuseClassificationResponse = {
...
+    );
+  } catch (error) {
+    console.warn('Failed to write cached rules-engine classification', {
+      identityKey: args.identityKey,
+      error,
+    });
+  }
+}
+
+export function isRulesEngineBlockingAction(action: AbuseRuleAction | null | undefined): boolean {
```
jrf0110, line 351: We should add a comment explaining what this function is for and what a "blocking" means in the context

jrf0110: Fixed in 74f4bdd3e. Added a doc comment explaining that a "blocking" action means a cached rules-engine action is severe enough for the gateway to wait for a fresh abuse classification before contacting the upstream model. Verified with `pnpm format` and `pnpm --filter web exec jest --runTestsByPath "src/app/api/openrouter/[...path]/route.test.ts"`.

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -385,7 +395,7 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
 
   // Use new shared helper for fraud & project headers
   const { fraudHeaders, projectId } = extractFraudAndProjectHeaders(request);
-  const providerResult = await getProvider({
+  let providerResult = await getProvider({
```
Comment (jrf0110) [inline apps/web/src/app/api/openrouter/[...path]/route.ts]: I'd like to see some more explainer comments on these changes since the flow can be sort of unintuitive

Comment (jrf0110) [reply apps/web/src/app/api/openrouter/[...path]/route.ts]: Fixed in 1a727c15e. Added explainer comments around the unintuitive parts of the flow: resolving provider context before abuse enforcement, starting classification early but only awaiting it behind the cache gate, falling back to cached enforcement when a blocking refresh fails, and rerunning provider resolution after quarantine-3 rewrites the model. Verified with `pnpm format` and `pnpm --filter web exec jest --runTestsByPath "src/app/api/openrouter/[...path]/route.test.ts"`.

apps/web/src/lib/ai-gateway/abuse-service.ts
```diff
@@ -24,6 +24,35 @@ import {
...
+import { getOpenRouterModels } from '@/lib/ai-gateway/providers/gateway-models-cache';
+import { redisGet, redisSet } from '@/lib/redis';
+import { abuseRulesClassificationRedisKey } from '@/lib/redis-keys';
+import type { FraudDetectionHeaders } from '@/lib/utils';
+import { z } from 'zod';
+
+const CLASSIFY_ABUSE_TIMEOUT_MS = 2000;
+const QUARANTINE_1_LATENCY_MS = 2000;
+const QUARANTINE_2_LATENCY_MS = 6000;
+const ABUSE_RULES_CLASSIFICATION_CACHE_TTL_SECONDS = 60;
```
jrf0110: This TTL should be higher. Maybe more like 1hr

jrf0110: Actually, now that I think about it... maybe we shouldn't have a ttl at all since this means that even blocked users just have to wait for the ttl before making their next request. Since this key space only grows with the number of distinct users, that ought to be okay

apps/web/src/lib/ai-gateway/abuse-service.ts
```diff
@@ -189,8 +218,187 @@ export type AbuseClassificationResponse = {
...
+
+export async function cacheRulesEngineClassification(args: {
+  identityKey: string;
+  rulesEngine: RulesEngineClassificationResult | undefined;
+}): Promise<void> {
+  if (!args.rulesEngine) return;
+  try {
+    await redisSet(
+      abuseRulesClassificationRedisKey(args.identityKey),
+      JSON.stringify(args.rulesEngine),
```
jrf0110: We pay for redis bandwidth and get/write operations here are in a very hot path. We should reduce the amount of data we're writing to this key to the absolute minimum (which I guess is simply a boolean indicating whether or not we should block on the abuse service?)

jrf0110: Fixed in 98dc6b10d. The Redis value is now only the last resolved action string (`block`, `rate-limit`, `quarantine-*`, `log`, or `none`) instead of the full rules-engine JSON. I kept the action instead of a boolean so the gateway can still enforce the correct cached response if a blocking refresh fails. Verified with `pnpm format`, `pnpm --filter web exec jest --runTestsByPath "src/app/api/openrouter/[...path]/route.test.ts"`, and `scripts/typecheck-all.sh --changes-only`.

apps/web/src/lib/ai-gateway/abuse-service.ts
jrf0110: Fixed in 98dc6b10d. Removed the Redis TTL entirely so a cached blocking/quarantine action keeps forcing fresh abuse classification instead of expiring after a minute. Also updated the PR notes to reflect that the action cache has no expiry. Verified with `pnpm format`, `pnpm --filter web exec jest --runTestsByPath "src/app/api/openrouter/[...path]/route.test.ts"`, and `scripts/typecheck-all.sh --changes-only`.

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -255,7 +265,7 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
     }
   }
 
-  const originalModelIdLowerCased = requestBodyParsed.body.model.toLowerCase();
+  let originalModelIdLowerCased = requestBodyParsed.body.model.toLowerCase();
```
Comment (chrarnoldus) [inline apps/web/src/app/api/openrouter/[...path]/route.ts]: I doubt this name makes anymore.

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -385,7 +395,9 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
 
   // Use new shared helper for fraud & project headers
   const { fraudHeaders, projectId } = extractFraudAndProjectHeaders(request);
-  const providerResult = await getProvider({
+  // Resolve the initial provider before abuse enforcement because abuse needs
+  // provider/BYOK context, and quarantine-3 may later rewrite these values.
+  let providerResult = await getProvider({
```
Comment (chrarnoldus) [inline apps/web/src/app/api/openrouter/[...path]/route.ts]: but do we need to mutate the original variable? can we have a new one `const initialProviderResultForAbuseService` so we don't confuse the two?

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -402,14 +414,12 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
...
-    skipProviderPin,
-    skipKiloExclusiveModelSettings,
-    experiment,
-  } = providerResult;
+  let provider = providerResult.provider;
+  let userByok = providerResult.userByok;
+  let bypassAccessCheck = providerResult.bypassAccessCheck;
+  let skipProviderPin = providerResult.skipProviderPin;
+  let skipKiloExclusiveModelSettings = providerResult.skipKiloExclusiveModelSettings;
+  let experiment = providerResult.experiment;
```
Comment (chrarnoldus) [inline apps/web/src/app/api/openrouter/[...path]/route.ts]: seems it might be cleaner to not unpack these fields

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -461,6 +481,89 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
...
+      request_id: classifyResult.request_id,
+    });
+  }
+  if (rulesEngineDecision.response) {
+    return rulesEngineDecision.response;
+  }
+  if (rulesEngineDecision.modelOverride) {
+    // Quarantine-3 rewrites non-BYOK requests to an auto-free candidate, so the
+    // provider and derived policy flags must be resolved again for that model.
+    requestBodyParsed.body.model = rulesEngineDecision.modelOverride;
```
Comment (chrarnoldus) [inline apps/web/src/app/api/openrouter/[...path]/route.ts:524]: Can we log very loudly in Vercel/Axiom, microdollar_usage, o11y service that this is happening? We don't want to end up having people debug issues that are caused by a model downgrade on our end.

Comment (jrf0110) [reply apps/web/src/app/api/openrouter/[...path]/route.ts:524]: I will add more logging, but if a user complains, it's possible to see what rule matches they are currently a part of and which ones they previously held by viewing their profile in the abuse service:

image

apps/web/src/app/api/openrouter/[...path]/route.ts
Comment (jrf0110) [reply apps/web/src/app/api/openrouter/[...path]/route.ts]: I agree. I was debating on whether or not to figure out a cleaner override mechanism. I'll make it better

apps/web/src/app/api/openrouter/[...path]/route.ts
Comment (jrf0110) [reply apps/web/src/app/api/openrouter/[...path]/route.ts]: Yeah similar to https://github.com/Kilo-Org/cloud/pull/3504#discussion_r3326493673 - I'll try cleanup this deconstruction

apps/web/src/app/api/openrouter/[...path]/route.ts
Comment (jrf0110) [reply apps/web/src/app/api/openrouter/[...path]/route.ts]: Fixed in 9ee9bf74a. Renamed `originalModelIdLowerCased` to `effectiveModelIdLowerCased` so the name reflects that quarantine-3 can rewrite the model before the rest of the route uses it. Verified with `pnpm format`, `pnpm --filter web exec jest --runTestsByPath "src/app/api/openrouter/[...path]/route.test.ts"`, `scripts/typecheck-all.sh --changes-only`, and `pnpm --filter web typecheck`.

apps/web/src/app/api/openrouter/[...path]/route.ts
Comment (jrf0110) [reply apps/web/src/app/api/openrouter/[...path]/route.ts]: Fixed in 9ee9bf74a. The first provider lookup is now immutable as `initialProviderResultForAbuseService`, and quarantine-3 uses a separate `quarantineProviderResult` before updating the effective provider context. Verified with `pnpm format`, `pnpm --filter web exec jest --runTestsByPath "src/app/api/openrouter/[...path]/route.test.ts"`, `scripts/typecheck-all.sh --changes-only`, and `pnpm --filter web typecheck`.

apps/web/src/app/api/openrouter/[...path]/route.ts
Comment (jrf0110) [reply apps/web/src/app/api/openrouter/[...path]/route.ts]: Fixed in 9ee9bf74a. Removed the mutable destructured provider fields and kept a single `effectiveProviderContext` instead, so downstream checks, provider-specific transforms, upstream routing, metrics, logging, and response rewriting all read from the same effective context after any quarantine-3 override. Verified with `pnpm format`, `pnpm --filter web exec jest --runTestsByPath "src/app/api/openrouter/[...path]/route.test.ts"`, `scripts/typecheck-all.sh --changes-only`, and `pnpm --filter web typecheck`.

apps/web/src/app/api/openrouter/[...path]/route.ts
Comment (jrf0110) [reply apps/web/src/app/api/openrouter/[...path]/route.ts:524]: Fixed in 9ee9bf74a. Added a loud `console.warn("SECURITY: Abuse quarantine-3 model override applied", ...)` when the downgrade happens, including user/identity, abuse request id, resolved action, matched rule ids, original and overridden model/provider, BYOK state, feature, and project id so it is visible in Vercel/Axiom logs. Verified with `pnpm format`, `pnpm --filter web exec jest --runTestsByPath "src/app/api/openrouter/[...path]/route.test.ts"`, `scripts/typecheck-all.sh --changes-only`, and `pnpm --filter web typecheck`.

apps/web/src/lib/ai-gateway/abuse-service.ts
```diff
@@ -189,8 +219,187 @@ export type AbuseClassificationResponse = {
...
+export async function getCachedRulesEngineAction(
+  identityKey: string
+): Promise<CachedRulesEngineAction | null> {
+  try {
+    const raw = await redisGet(abuseRulesClassificationRedisKey(identityKey));
+    if (!raw) return null;
+    const action = parseCachedRulesEngineAction(raw);
+    return action !== undefined ? { identityKey, action } : null;
+  } catch (error) {
+    console.warn('Failed to read cached rules-engine action', { identityKey, error });
```
RSO, line 324: We should probably set up an anomaly detection rule in Axiom for this.

apps/web/src/lib/ai-gateway/abuse-service.ts
```diff
@@ -189,8 +219,187 @@ export type AbuseClassificationResponse = {
...
+};
+
+export type RulesEngineActionDecision = {
+  action: AbuseRuleAction | null;
+  delayMs: number;
+  modelOverride: string | null;
+  response: NextResponse<unknown> | null;
+};
+
+export function sleepForRulesEngineAction(ms: number): Promise<void> {
```
RSO, line 242: Maybe we should add a `console.info` here as well, might be helpful if anybody is looking at the logs trying to see what's going on.

RSO (approved):

apps/web/src/lib/ai-gateway/abuse-service.ts
chrarnoldus: added

PR #3508, opened 2026-05-26
feat(kiloclaw): add agent config CRUD routes

pandemicsyn:
## Summary

- Add authenticated, capability-advertised controller endpoints for listing, reading, creating, updating, and deleting configured OpenClaw agents.
- Introduce a narrow agent-config boundary: reads and focused model/default-setting patches are controller-native with optimistic conflict protection, while create/delete delegate lifecycle behavior to the installed OpenClaw CLI.
- Serialize controller-originated lifecycle and patch mutations per config path to avoid lost updates, document the unverified filesystem disposition of CLI deletes, and enable typechecking for standalone controller package changes.

## Verification

- No additional interactive manual verification was performed; this change affects machine-side controller endpoints and was exercised through the controller container validation flow during implementation rather than through a user-facing UI.

Smoke test log:

```
--- agent config CRUD endpoints ---
PASS: trash command omitted from packaged image (got )
PASS: create configured agent (got crud-smoke)
PASS: created agent leaves OpenClaw config valid (got True)
PASS: list includes created agent (got True)
PASS: patch updates created agent model (got kilocode/kilo-auto/balanced)
PASS: delete reports unverified filesystem disposition (got unverified)
PASS: deleted agent is no longer configured (got False)
PASS: packaged no-trash baseline retains deleted workspace (got 1)
```

## Visual Changes

N/A

## Reviewer Notes

- Delete intentionally returns `filesystemDisposition: \"unverified\"`: v1 delegates to `openclaw agents delete --force --json` without installing `trash-cli`, so the API does not promise file retention or cleanup.
- Native PATCH mutations use config-level etag/pre-write checks; controller-originated create/delete/PATCH operations also share local serialization, while external-process races remain bounded by OpenClaw's optimistic write behavior.
- `GET /_kilo/config/agents/:agentId` returns normalized state only; raw OpenClaw config remains behind the existing authenticated config-read endpoint.


services/kiloclaw/controller/src/openclaw-agent-config.ts
```diff
@@ -0,0 +1,552 @@
...
+  const normalized = trimmed.toLowerCase();
+  if (VALID_AGENT_ID.test(trimmed)) {
+    return normalized;
+  }
+  return (
+    normalized
+      .replace(INVALID_AGENT_ID_CHARS, '-')
+      .replace(LEADING_DASHES, '')
+      .replace(TRAILING_DASHES, '')
+      .slice(0, 64) || DEFAULT_AGENT_ID
```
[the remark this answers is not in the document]
pandemicsyn: No controller-side change here: this function intentionally mirrors the normalization contract in the pinned OpenClaw implementation (`src/routing/session-key.ts:92-109`), which uses the same trailing-dash-removal then `slice(0, 64)` sequence. Changing only the controller would allow its normalized IDs to diverge from IDs that OpenClaw creates/resolves through the CLI. If stricter canonicalization is desired, it needs to land upstream in OpenClaw first and be adopted here with the image dependency update.

services/kiloclaw/controller/src/routes/config-agents.ts
```diff
@@ -0,0 +1,156 @@
...
+      const result = await deps.updateDefaults(parsed.data);
+      return c.json({ ok: true, etag: result.snapshot.etag, defaults: result.defaults });
+    } catch (error) {
+      return respondError(c, error);
+    }
+  });
+
+  app.delete('/_kilo/config/agents/:agentId', async c => {
+    try {
+      const deleted = await deps.serializeMutation(() => deps.deleteViaCli(c.req.param('agentId')));
```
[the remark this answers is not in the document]
pandemicsyn: No change needed for the pinned CLI boundary: `openclaw agents delete` normalizes its input internally before checking `main` or finding the configured agent (`src/commands/agents.commands.delete.ts:34-53`). Passing the route parameter to the CLI preserves OpenClaw-owned deletion semantics and `DELETE .../Main` resolves through that normalization path.

pandemicsyn:
## Smoke test evidence

Executed the packaged-image controller entrypoint smoke flow on this PR branch, including `0ce37e74a` (`test(kiloclaw): validate created agent config in smoke test`).

Relevant CRUD / OpenClaw validation output:

```text
--- agent config CRUD endpoints ---
PASS: trash command omitted from packaged image (got )
PASS: create configured agent (got crud-smoke)
PASS: created agent leaves OpenClaw config valid (got True)
PASS: list includes created agent (got True)
PASS: patch updates created agent model (got kilocode/kilo-auto/balanced)
PASS: delete reports unverified filesystem disposition (got unverified)
PASS: deleted agent is no longer configured (got False)
PASS: packaged no-trash baseline retains deleted workspace (got 1)

=== Results: 19 passed, 0 failed ===
```

`openclaw config validate --json` is used after real CLI-backed creation to confirm the resulting persisted OpenClaw config is schema-valid without running repair/migration behavior.

services/kiloclaw/controller/src/openclaw-agent-cli.ts
```diff
@@ -0,0 +1,179 @@
...
+}
+
+export async function createAgentViaCli(
+  body: BasicAgentCreateBody,
+  deps: OpenClawAgentCliDeps = defaultDeps
+): Promise<CreateAgentCliResult> {
+  const args = [
+    'agents',
+    'add',
+    body.name,
```
St0rmz1, line 168: **Moderate, Security:** `body.name`, `body.workspace`, `body.agentDir`, `body.model`, and bindings are placed into the `openclaw` argv with no validation against leading `--`, and there is no `--` separator before the positional name. `execFile` is shell-safe, but the openclaw CLI's argv parser is not: a bearer-authenticated caller can post `name: "--help"`, `name: "--non-interactive"`, `model: "--config=/path"`, or `workspace: "--something"` and have those values reinterpreted as flags.

Net effect: the "basic create" endpoint becomes a way to invoke arbitrary openclaw flags beyond what `BasicAgentCreateBodySchema` appears to permit. The actual blast radius depends on which flags openclaw exposes (alternate config paths, debug toggles, output destinations), but it expands the surface of this endpoint past its documented contract.

Suggestion: reject leading `--` in the relevant zod fields, or invoke the CLI with a `--` separator before positionals (e.g. `['agents', 'add', '--workspace=' + workspace, ..., '--', name]`) if the openclaw CLI honors it. The schema fix is the cheapest; the `--` separator is the most robust.

services/kiloclaw/controller/src/openclaw-agent-config.ts
St0rmz1, line 218: **Low, Design:** `requireAgentId` only rejects an empty-after-trim input. Any non-empty path segment that collapses to empty under `normalizeAgentId` (e.g. `@@@`, `!!!`, `----`) falls through the `|| DEFAULT_AGENT_ID` at line 218 and silently becomes `main`. That means `GET /_kilo/config/agents/@@@` and `PATCH /_kilo/config/agents/----` operate on the `main` agent without any signal to the caller that their input was rewritten. DELETE is unaffected because it delegates to the CLI's own normalization.

Mostly a footgun: a client typo or encoding bug could mutate the wrong agent and get a 200 back.

Suggestion: in `requireAgentId`, when the input is non-empty but `normalizeAgentId` produces the `DEFAULT_AGENT_ID` fallback from collapse (not from a literal `main` input), throw `invalid_agent_id` (400). Or move the empty-collapse check into `requireAgentId` and have `normalizeAgentId` return `""` in that case so the caller can distinguish.

services/kiloclaw/controller/src/openclaw-agent-config.ts
```diff
@@ -0,0 +1,552 @@
...
+  const configPath = options.configPath ?? OPENCLAW_CONFIG_PATH;
+  let raw: string;
+  try {
+    raw = fs.readFileSync(configPath, 'utf8');
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error);
+    throw new AgentConfigError(
+      500,
+      'agent_config_read_failed',
+      `Failed to read config: ${message}`
```
St0rmz1: **Low, Security:** `fs.readFileSync` errors include the absolute path (e.g. `ENOENT: no such file or directory, open '/root/.openclaw/openclaw.json'`). That message is wrapped into the `AgentConfigError` and returned verbatim to the client through `respondError`. Other 500 paths in the controller log internally and return a generic body; this one diverges.

Low impact since the path is already implied by the spec, but it is a consistency and info-disclosure nit worth fixing.

Suggestion: return a fixed `Failed to read agent config` to the client and `console.error` the underlying message, matching the fallback branch of `respondError`.

St0rmz1 (approved):

services/kiloclaw/controller/src/openclaw-agent-cli.ts
pandemicsyn: Fixed in `8072d48be`. `BasicAgentCreateBodySchema` now rejects option-like free-form CLI values beginning with `-` for `name`, `model`, and `bindings`, with regression coverage. `workspace` and `agentDir` remain absolute-path-only inputs, so they cannot be parsed as command options.

services/kiloclaw/controller/src/openclaw-agent-config.ts
pandemicsyn: Fixed in `8072d48be`. Config read failures now log the underlying filesystem error internally and return a fixed `Failed to read agent config` typed error to callers. Regression coverage asserts that filesystem path details are not exposed in the returned error.

services/kiloclaw/controller/src/openclaw-agent-config.ts
pandemicsyn: Fixed in `8072d48be`. `requireAgentId` now rejects non-empty native resource identifiers that normalize to the reserved `main` fallback (for example `@@@` and `----`) while still accepting literal/case-varied `main`. Regression tests cover the rejected collapse cases.

PR #3509, opened 2026-05-27
fix(kiloclaw) billing clear detached destruction deadline

St0rmz1:
## Summary

The billing instance destruction sweep had two production bugs that left it functionally idle. Both are fixed here.

**Bug 1: detached subscriptions starved the queue.** The sweep selects
candidates ordered by `destruction_deadline ASC, id ASC` with `LIMIT
INSTANCE_DESTRUCTION_BATCH_SIZE`. When a row has `instance_id IS NULL`
(its instance was destroyed via another path) the loop logged
"missing_instance_id" and continued without clearing the deadline, so
the same handful of detached rows pinned the head of the queue on
every cron and every row behind them was starved. About 50 detached
rows blocked 25k real overdue rows for over a month.

The fix is a new helper
`clearDetachedSubscriptionDestructionDeadlineBestEffort` that batches
all detached subscription IDs collected during the loop and runs a
single SELECT plus guarded UPDATE plus changelog INSERT inside one
database transaction. The UPDATE guards (`instance_id IS NULL` and
`destruction_deadline IS NOT NULL`) make a concurrent reattach or a
second clear a silent skip. The transaction wrap is important for
audit: if the INSERT throws the UPDATE rolls back, the rows stay
detached with their deadlines, and the next sweep retries the whole
pair atomically. Without the transaction, a transient INSERT failure
would erase the only signal that the cleanup ever ran and the audit
history would be permanently lost.

A small race guard is also added on the existing happy path UPDATE
(`isNotNull(destruction_deadline)` plus an `if (updated)` around the
changelog write) so a concurrent clear from the new code path does
not produce a phantom `instance_destroyed` audit entry with identical
before and after states.

**Bug 2: throughput barely matched inflow.** With the queue unblocked,
the sweep processes 50 rows in about 6 minutes. Each destroy takes 7
to 8 seconds through Fly plus Durable Object cleanup, so the rate was
around 200 destroys per hour. Observed inflow runs close to that, so
the queue was hovering at break even or slightly growing. The batch
size constant is raised to 75. Each row's wall clock budget stays
well inside the Cloudflare queue consumer's 15 minute ceiling. A
comment on the constant explains the math and warns that going higher
requires parallelizing the destroy loop first.

## Verification

The bug only manifests when the candidate query returns rows with
`instance_id IS NULL` at the head of the queue. Reproducing this in
a clean test database is not feasible because the bug requires real
historical detached rows to pin the head, which the fixed code itself
clears on the first sweep. Manual verification before merge therefore
relies on the regression test plus log inspection on the deployed
worker.

After the deploy:

- [ ] Watch the billing worker logs for the next two cron firings.
      Confirm the `sweep_completed` event for `instance_destruction`
      shows `sweep3_instance_destruction` greater than zero and
      `errors` of zero. Before the fix this value was zero on every
      sweep.
- [ ] Run a SQL count of overdue rows where `instance_id IS NULL` and
      confirm it stays at or near zero. The new cleanup keeps it
      drained automatically rather than allowing buildup.

## Visual Changes

N/A. Server side billing worker change. No UI surface area.

| Before | After |
|---|---|
|  |  |

## Reviewer Notes

Focus areas:

- `clearDetachedSubscriptionDestructionDeadlineBestEffort` runs after
  the candidate loop closes, with the collected IDs as input. The
  helper returns early on empty input, so a sweep with no detached
  rows pays no cost.
- The transaction makes the UPDATE plus changelog INSERT atomic. The
  outer try and catch logs a rollback but does not rethrow, so a
  single bulk cleanup failure does not abort the rest of the sweep.
  Every other row in the same sweep still has its bookkeeping written
  normally.
- The bump from 50 to 75 is a config change. The wall clock math is
  in the comment above the constant. The existing happy path test was
  updated to assert the new value.
- The regression test seeds one detached row, runs the sweep, and
  asserts the transaction scoped UPDATE and INSERT fired, no destroy
  POST went out to the kiloclaw worker, and the existing skip log is
  still emitted.

### DIFF
--- a/services/kiloclaw-billing/src/lifecycle.ts
+++ b/services/kiloclaw-billing/src/lifecycle.ts
@@ -727,7 +727,14 @@ async function insertLifecycleChangeLogBestEffort(
  *   - `instance_id IS NULL` — only clear rows that are still detached.
  *   - `destruction_deadline IS NOT NULL` — skip rows already cleared.
  * Rows that match neither guard are silently skipped (no changelog entry).
- * Changelog entries for all cleared rows are written in a single bulk INSERT.
+ *
+ * The SELECT + UPDATE + changelog INSERT run inside a single transaction so
+ * the audit record cannot be lost. If the changelog INSERT fails the UPDATE
+ * rolls back; the rows remain detached with their deadlines, and the next
+ * sweep re-discovers them and retries the whole pair atomically. Without
+ * this, a transient INSERT failure would erase the only signal
+ * (`destruction_deadline IS NOT NULL`) that the cleanup ever ran — there
+ * would be no future sweep to reconstruct the missing audit history.
  */
 async function clearDetachedSubscriptionDestructionDeadlineBestEffort(
   database: WorkerDb,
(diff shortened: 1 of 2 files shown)


services/kiloclaw-billing/src/lifecycle.ts
```diff
@@ -3644,6 +3751,8 @@ async function runInstanceDestructionSweep(
             reason: 'missing_instance_id',
           }
         );
+        // Bulk-cleared after the loop — see detachedSubscriptionIds below.
+        detachedSubscriptionIds.push(row.id);
```
jeanduplessis: WARNING: This collector is reached only after `isSoftDeletedUserEmail(row.email)` returns false. `softDeleteUser` retains `kiloclaw_subscriptions` and rewrites the email to `deleted+${userId}@deleted.invalid`, so an overdue detached retained row is still selected by the ordered batch but skips before its ID can be added here. Its `destruction_deadline` remains non-null, and enough soft-deleted detached rows can continue pinning the bounded queue and starving overdue live rows behind them. Please clear the marker for detached soft-deleted rows without re-enabling user-facing side effects, and add a regression case for that retained-row state.

jeanduplessis (changes requested):
Found one concrete edge-state gap in the detached subscription cleanup. Soft-deleted users retain subscription rows, but their anonymized email causes the destruction sweep to continue before the new detached-row collection path, leaving expired `destruction_deadline` markers eligible to pin the bounded queue. No unresolved review threads were present, and no PostHog path changes were introduced by this PR.

jeanduplessis (approved):

PR #3516, opened 2026-05-27
feat(edit): add /api/edit/completions endpoint

markijbema:
## Summary

- Adds a new proxy endpoint at `/api/edit/completions` for edit completion requests, backed by Inception's Mercury Edit endpoint.
- Mirrors the existing `/api/fim/completions` auth, balance, organization policy, BYOK, upstream proxy, and usage logging flow while validating Inception's single-user-message edit contract.
- Tags usage as `edit_completions`, reports cached input token discounts, and returns `unsupported_edit_model` for unsupported model prefixes.

## Verification

- Manual review of the route against the existing `/api/fim/completions` route to confirm the auth, balance, org-policy, BYOK, proxying, and usage logging flow remains aligned.
- Manual review that the public route, error type, API kind, logs, tests, and PR metadata consistently use `edit` terminology.

## Visual Changes

N/A — backend-only.

## Reviewer Notes

- Only Inception models prefixed with `inception/` are supported for now.
- Inception's `/v1/edit/completions` endpoint accepts exactly one `role: "user"` message and does not support streaming, so the route rejects unsupported message shapes and `stream: true` before proxying.
- Future work: route through the Kilo gateway instead of calling Inception directly, and add other edit providers once they expose an equivalent endpoint.

### DIFF
--- a/apps/web/src/lib/ai-gateway/llm-proxy-helpers.ts
+++ b/apps/web/src/lib/ai-gateway/llm-proxy-helpers.ts
@@ -708,7 +708,7 @@ export function extractNextEditPromptInfo(body: { messages: NextEditMessage[] })
   const totalLength = body.messages.reduce((sum, m) => sum + (m.content?.length ?? 0), 0);
   return {
     system_prompt_prefix: '', // /v1/edit/completions bakes its system prompt in server-side
-    system_prompt_length: totalLength,
+    system_prompt_length: totalLength, // total message content length; there is no explicit system prompt
     user_prompt_prefix: content.slice(0, 100),
   };
 }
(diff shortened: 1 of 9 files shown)


apps/web/src/lib/ai-gateway/llm-proxy-helpers.ts
```diff
@@ -696,6 +696,148 @@ export function countAndStoreFimUsage(
...
+
+type EditMessage = { role: string; content: string };
+
+export function extractEditPromptInfo(body: { messages: EditMessage[] }): PromptInfo {
+  const lastUser = [...body.messages].reverse().find(m => m.role === 'user');
+  const content = lastUser?.content ?? '';
+  const totalLength = body.messages.reduce((sum, m) => sum + (m.content?.length ?? 0), 0);
+  return {
+    system_prompt_prefix: '', // /v1/edit/completions bakes its system prompt in server-side
+    system_prompt_length: totalLength, // total message content length; there is no explicit system prompt
```
chrarnoldus: doesn't 0 make more sense then? not that it matters.

chrarnoldus (approved):

PR #3523, opened 2026-05-27
fix(github): repair renamed installation login metadata

eshurakov:
## Summary

- Keep GitHub App integrations usable after account renames by synchronizing current installation identity from rename webhooks and manual refreshes.
- Repair stale login metadata during git-token lookup misses with throttling and fail-closed ambiguity handling, then mint repository-scoped tokens.
- Architecturally, the web integration path performs proactive reconciliation while the token service provides defensive repair for existing stale records.

## Verification

- [x] Verified locally

## Visual Changes

N/A

## Reviewer Notes

- Review authorization/ambiguity handling, webhook retry ordering, and the 10-minute repair cooldown.

### DIFF
--- a/apps/web/src/lib/integrations/db/platform-integrations.ts
+++ b/apps/web/src/lib/integrations/db/platform-integrations.ts
@@ -178,6 +178,21 @@ export async function updateRepositoriesForIntegration(
     .where(eq(platform_integrations.id, integrationId));
 }
 
+export async function updateIntegrationAccountIdentity(
+  integrationId: string,
+  platformAccountId: string,
+  platformAccountLogin: string
+) {
+  await db
+    .update(platform_integrations)
+    .set({
+      platform_account_id: platformAccountId,
+      platform_account_login: platformAccountLogin,
+      updated_at: new Date().toISOString(),
+    })
+    .where(eq(platform_integrations.id, integrationId));
+}
+
 /**
  * Suspends a platform integration
  */
--- a/services/git-token-service/src/github-token-service.ts
+++ b/services/git-token-service/src/github-token-service.ts
@@ -79,6 +79,7 @@ export class GitHubTokenService {
       if (cooldownMarker !== null) {
         return null;
       }
+      // Cool down failed attempts too, so repeated lookup misses cannot hammer GitHub.
       await this.env.TOKEN_CACHE.put(cooldownKey, new Date().toISOString(), {
         expirationTtl: INSTALLATION_LOGIN_REFRESH_TTL_SECONDS,
       });
--- a/services/git-token-service/src/github-token-service.ts
+++ b/services/git-token-service/src/github-token-service.ts
@@ -24,7 +24,7 @@ const CACHE_KEY_PREFIX = 'gh-token:';
 const MIN_TTL_SECONDS = 60;
 const EXPIRY_BUFFER_MS = 5 * 60 * 1000;
 const INSTALLATION_LOGIN_REFRESH_CACHE_KEY_PREFIX = 'gh-installation-login-refresh:v1:';
-const INSTALLATION_LOGIN_REFRESH_TTL_SECONDS = 10 * 60;
+const INSTALLATION_LOGIN_REFRESH_TTL_SECONDS = 15 * 60;
 
 const GitHubInstallationAccountSchema = z.object({
   account: z.object({
--- a/services/git-token-service/src/github-token-service.ts
+++ b/services/git-token-service/src/github-token-service.ts
@@ -1,4 +1,5 @@
 import { createAppAuth } from '@octokit/auth-app';
+import { Octokit } from '@octokit/rest';
 import * as z from 'zod';
 
 type Token = z.infer<typeof Token>;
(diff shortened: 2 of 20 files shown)


services/git-token-service/src/installation-lookup-service.ts
```diff
@@ -32,11 +37,93 @@ export type InstallationLookupFailure = {
...
+export function buildInstallationLookupQuery(db: WorkerDb, params: FindInstallationParams) {
+  const [repoOwner = ''] = params.githubRepo.split('/');
+  return buildAuthorizedInstallationsQuery(db, params, repoOwner).limit(2);
+}
+
+export function buildInstallationRefreshCandidatesQuery(
+  db: WorkerDb,
+  params: FindInstallationParams
+) {
+  return buildAuthorizedInstallationsQuery(db, params);
```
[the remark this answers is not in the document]
eshurakov: Fixed in `925b9eee6`: `buildInstallationRefreshCandidatesQuery` now caps the defensive repair scan at 10 authorized candidates, with a query regression test. This bounds DB rows and upstream calls per token request.

services/git-token-service/src/github-token-service.ts
```diff
@@ -61,6 +69,74 @@ export class GitHubTokenService {
...
+    installationId: string,
+    appType: GitHubAppType = 'standard'
+  ): Promise<string | null> {
+    const cooldownKey = `${INSTALLATION_LOGIN_REFRESH_CACHE_KEY_PREFIX}${appType}:${installationId}`;
+    if (this.env.TOKEN_CACHE) {
+      const cooldownMarker = await this.env.TOKEN_CACHE.get(cooldownKey);
+      if (cooldownMarker !== null) {
+        return null;
+      }
+      await this.env.TOKEN_CACHE.put(cooldownKey, new Date().toISOString(), {
```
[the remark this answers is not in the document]
eshurakov: Kept this as an intentional attempt cooldown rather than a success cooldown: this fallback runs on lookup misses, and immediate retries during GitHub/auth failures would amplify upstream traffic. Added an inline explanation and regression test covering failed-attempt throttling in `925b9eee6`.

services/git-token-service/src/index.ts
```diff
@@ -67,27 +67,68 @@ export class GitTokenRPCEntrypoint extends WorkerEntrypoint<CloudflareEnv> {
...
+          return { success: false, reason: 'no_installation_found' };
+        case 'database_not_configured':
+        case 'invalid_repo_format':
+        case 'no_installation_found':
+        case 'invalid_org_id':
+          return { success: false, reason: installation.reason };
+      }
+    }
+
+    const [, repoName] = params.githubRepo.split('/');
```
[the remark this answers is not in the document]
eshurakov: Retaining this guard intentionally. The successful lookup establishes the runtime invariant, but the guard also narrows `repoName` to `string` at the RPC/token-mint boundary without a non-null assertion or cast; removing it would trade a cheap defensive check for weaker expression of that boundary.

apps/web/src/lib/integrations/db/platform-integrations.ts
```diff
@@ -178,6 +178,21 @@ export async function updateRepositoriesForIntegration(
...
+  platformAccountLogin: string
+) {
+  await db
+    .update(platform_integrations)
+    .set({
+      platform_account_id: platformAccountId,
+      platform_account_login: platformAccountLogin,
+      updated_at: new Date().toISOString(),
+    })
+    .where(eq(platform_integrations.id, integrationId));
```
jrf0110, line 193: Do we need to add `and suspended_at is null`?

eshurakov: No guard is needed here. `installation_target.renamed` synchronizes identity metadata only and does not clear `suspended_at` or reactivate the integration. Updating it while suspended keeps the installation identity current for a later unsuspend; suppressing the write would retain stale login metadata. This also matches the existing `installation_repositories` synchronization path, which runs before the generic suspended-event skip.

services/git-token-service/src/github-token-service.ts
```diff
@@ -61,6 +69,75 @@ export class GitHubTokenService {
...
+
+    try {
+      const numericId = this.validateInstallationId(installationId);
+      const credentials = this.getCredentials(appType);
+      const auth = createAppAuth({
+        appId: credentials.appId,
+        privateKey: credentials.privateKey,
+      });
+      const { token } = await auth({ type: 'app' });
+      const response = await fetch(`https://api.github.com/app/installations/${numericId}`, {
```
jrf0110: Octokit doesn't provide a a way to get the installation? Why do we have to raw-fetch?

services/git-token-service/src/installation-lookup-service.ts
```diff
@@ -66,89 +174,70 @@ export class InstallationLookupService {
...
   }
+
+  async updateAccountLogin(integrationId: string, accountLogin: string): Promise<void> {
+    await this.getDb()
+      .update(platform_integrations)
+      .set({
+        platform_account_login: accountLogin,
+        updated_at: new Date().toISOString(),
+      })
+      .where(eq(platform_integrations.id, integrationId));
```
jrf0110: I think we need another check for `suspended_at is null`

services/git-token-service/src/index.ts
```diff
@@ -67,27 +67,76 @@ export class GitTokenRPCEntrypoint extends WorkerEntrypoint<CloudflareEnv> {
...
+        candidate.githubAppType
+      );
+      if (
+        !refreshedAccountLogin ||
+        refreshedAccountLogin.toLowerCase() === candidate.accountLogin?.toLowerCase()
+      ) {
+        continue;
+      }
+
+      await this.installationLookupService.updateAccountLogin(
```
jrf0110: If this update doesn't actually update any rows, should we detect that and at least log that something went wrong?

jrf0110 (approved):

services/git-token-service/src/installation-lookup-service.ts
eshurakov: I think its okay to update suspended row if that will happen somehow. But it shouldn't happen because we're fetching only active installations

services/git-token-service/src/github-token-service.ts
eshurakov: `cloudflare-git-token-service` intentionally uses only `@octokit/auth-app`: it provides app JWT/installation-token authentication, but an authenticated REST client requires adding `@octokit/core` or `@octokit/request`. This is one bounded GET followed by Zod validation, so using native Workers `fetch` avoids a new production dependency and keeps the upstream response boundary explicit.

services/git-token-service/src/index.ts
eshurakov: Fixed in `435640395`: `updateAccountLogin` now uses `returning` to report whether a row changed. The RPC path emits the repair success log only after a persisted update and emits a safe structured warning when the target row disappeared; regression coverage verifies both behavior and omitted login values.

services/git-token-service/src/github-token-service.ts
eshurakov: Follow-up in `435640395`: switched this lookup to `@octokit/rest` `apps.getInstallation`, matching the existing GitHub adapter SDK pattern in `apps/web`.

PR #3542, opened 2026-05-27
feat(kiloclaw): validate openclaw config file saves

pandemicsyn:
## Summary

- Today the file editor saves every permitted file, including `openclaw.json`, through the generic `files/write` controller endpoint, which performs no installed OpenClaw schema validation. This change adds validation before an upgraded instance saves `openclaw.json`, preserving the user's changes and presenting an explicit **Save anyway** override when validation warns.
- Route `openclaw.json` saves through a dedicated `files/write-openclaw-config` endpoint only when the running controller advertises that capability. Capability gating lets not-yet-upgraded controllers keep their existing generic-save behavior; the dedicated endpoint is necessary so a controller rolled back after capability detection cannot ignore a new option on the generic endpoint and silently perform an unvalidated write.
- Validate the submitted bytes by writing a protected `0600` sibling candidate file under `/root/.openclaw` and invoking `openclaw config validate --json` with `OPENCLAW_CONFIG_PATH` set to that candidate. OpenClaw therefore validates the pending save rather than the live `openclaw.json`, while placing the file in the same directory preserves ordinary relative-include resolution.
- Carry the new write result through the Worker/Durable Object boundary and harden the controller/editor flow with ETag rechecks, protected staging artifacts, bounded diagnostics, save-race protection, and conflict-safe override UX; update the controller contract, smoke coverage, and KiloClaw changelog.

## Verification

- Manual browser verification was not performed in this session because the new dialog requires an upgraded running KiloClaw instance exposing the new controller capability.
- [ ] Add manual verification details for valid-save, invalid-warning, and **Save anyway** behavior on an upgraded instance.

## Visual Changes

SCR-20260527-ooyy

## Reviewer Notes

- Rollout behavior is intentional: controllers without `files.write-openclaw-config` continue using legacy generic `files/write`; capable controllers switch `openclaw.json` to the validated path.
- The dedicated route is a mutation-time safety boundary, not just an API convenience. If the UI sees the capability and the machine is then rolled back before Save, the dedicated call fails as unsupported instead of being accepted by an old generic writer that would ignore validation semantics and write immediately.
- Validation mechanics: the controller writes `.openclaw.kiloclaw-validation-candidate.json` next to the live config with mode `0600`, executes `openclaw config validate --json` with `OPENCLAW_CONFIG_PATH` targeting that staged file, parses the bounded JSON result, rechecks validated bytes before committing, and cleans up staged artifacts in `finally`. The internal staging filename is hidden from the generic file browser/read/write surface.
- Diagnostics are returned only through authenticated config management, and an explicit **Save anyway** may still be rejected or restored by OpenClaw during reload/startup.

### DIFF
--- a/services/kiloclaw/controller/src/openclaw-config-validation.ts
+++ b/services/kiloclaw/controller/src/openclaw-config-validation.ts
@@ -143,29 +143,46 @@ function unavailableIssue(message: string): OpenclawConfigValidationResult {
   };
 }
 
-function referencesTargetConfig(value: unknown, configPath: string): boolean {
-  if (Array.isArray(value)) {
-    return value.some(entry => referencesTargetConfig(entry, configPath));
+function unexpectedValidationFailure(error: unknown): OpenclawConfigValidationResult {
+  switch (errorCode(error)) {
+    case 'ENOSPC':
+      return unavailableIssue('There is not enough disk space to validate this configuration.');
+    case 'EACCES':
+    case 'EPERM':
+      return unavailableIssue('OpenClaw cannot access the temporary validation file.');
+    case 'EEXIST':
+      return unavailableIssue('Configuration validation is already in progress. Try saving again.');
+    default:
+      return unavailableIssue('OpenClaw configuration validation could not be started.');
   }
-  if (typeof value !== 'object' || value === null) {
-    return false;
-  }
+}
 
-  for (const [key, nestedValue] of Object.entries(value)) {
-    if (key === '$include') {
-      const includes = Array.isArray(nestedValue) ? nestedValue : [nestedValue];
-      if (
-        includes.some(
-          include =>
-            typeof include === 'string' &&
-            path.resolve(path.dirname(configPath), include) === path.resolve(configPath)
-        )
-      ) {
-        return true;
-      }
+function referencesTargetConfig(value: unknown, configPath: string): boolean {
+  const pending: unknown[] = [value];
+  while (pending.length > 0) {
+    const current = pending.pop();
+    if (Array.isArray(current)) {
+      for (const entry of current) pending.push(entry);
+      continue;
     }
-    if (referencesTargetConfig(nestedValue, configPath)) {
-      return true;
+    if (typeof current !== 'object' || current === null) {
+      continue;
+    }
+
+    for (const [key, nestedValue] of Object.entries(current)) {
+      if (key === '$include') {
+        const includes = Array.isArray(nestedValue) ? nestedValue : [nestedValue];
+        if (
+          includes.some(
+            include =>
+              typeof include === 'string' &&
+              path.resolve(path.dirname(configPath), include) === path.resolve(configPath)
+          )
+        ) {
+          return true;
+        }
+      }
+      pending.push(nestedValue);
     }
   }
   return false;
@@ -231,7 +248,7 @@ export async function validateOpenclawConfigCandidate(
     return unavailableIssue('OpenClaw could not validate this configuration.');
   } catch (error) {
     console.error('[openclaw-config-validation] Validation failed unexpectedly:', errorCode(error));
-    return unavailableIssue('OpenClaw configuration validation could not be started.');
+    return unexpectedValidationFailure(error);
   } finally {
     try {
       deps.removeFile(stagePath);
--- a/services/kiloclaw/controller/src/openclaw-config-validation.ts
+++ b/services/kiloclaw/controller/src/openclaw-config-validation.ts
@@ -1,0 +1,230 @@
+import { execFile } from 'node:child_process';
+import fs from 'node:fs';
+import path from 'node:path';
+import { z } from 'zod';
+
+const VALIDATION_TIMEOUT_MS = 30_000;
+const VALIDATION_MAX_OUTPUT_BYTES = 1_048_576;
+const VALIDATION_MAX_ISSUES = 20;
+const VALIDATION_MAX_TEXT_LENGTH = 500;
+export const VALIDATION_STAGE_FILENAME = '.openclaw.kiloclaw-validation-candidate.json';
+
+export function isOpenclawValidationArtifactPath(relativePath: string): boolean {
+  return (
+    relativePath === VALIDATION_STAGE_FILENAME ||
+    relativePath.startsWith(`${VALIDATION_STAGE_FILENAME}.`)
+  );
+}
+
+const ValidationIssueSchema = z.object({
+  path: z.string().optional(),
+  message: z.string(),
+  allowedValues: z.array(z.string()).optional(),
+});
+
+const ValidationOutputSchema = z.object({
+  valid: z.boolean(),
+  issues: z.array(ValidationIssueSchema).optional(),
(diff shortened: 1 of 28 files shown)


services/kiloclaw/controller/src/routes/files.ts
```diff
@@ -725,8 +738,73 @@ export function registerFileRoutes(app: Hono, expectedToken: string, rootDir: st
...
+    if (typeof result !== 'string') {
+      return c.json(
+        { error: result.error, ...(result.code && { code: result.code }) },
+        result.status
+      );
+    }
+
+    const hasEtagConflict = () => {
+      if (!body.etag) return false;
+      const currentContent = fs.readFileSync(result, 'utf-8');
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in f809cec43. An unreadable or removed config during ETag verification now returns a conflict.

services/kiloclaw/controller/src/openclaw-config-validation.ts
```diff
@@ -0,0 +1,230 @@
...
+    }
+    if (result.data.issues && result.data.issues.length > 0) {
+      return {
+        valid: false,
+        reason: 'invalid',
+        issues: normalizeIssues(result.data.issues, stagePath),
+      };
+    }
+    return unavailableIssue('OpenClaw could not validate this configuration.');
+  } catch {
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in f809cec43. Unexpected validation failures now log only a safe error-code classification.

services/kiloclaw/controller/src/routes/files.ts
```diff
@@ -708,8 +722,7 @@ export function registerFileRoutes(app: Hono, expectedToken: string, rootDir: st
 
     if (body.etag) {
```
St0rmz1: **Moderate, Functional:** On a controller that advertises `files.write-openclaw-config`, this legacy `/_kilo/files/write` handler still accepts `body.path === 'openclaw.json'` and writes without `serializeAgentConfigMutation`. The dedicated handler below queues per `configPath` and runs validation; this one does neither, so a legacy or alternate client can interleave a plain write of `openclaw.json` with an in-flight validate-then-write or with `mutateAgentConfig`. The post-validation ETag recheck in the dedicated handler will detect the resulting race and return 409, but that is detection rather than prevention, and the spec change in this PR frames the dedicated route as "a mutation-time safety boundary, not just an API convenience."

Suggestion: when this controller advertises the capability, either reject `path === 'openclaw.json'` here (forcing callers onto the validated path) or route the `openclaw.json` branch through `serializeAgentConfigMutation({ configPath: result })` so it queues against the validated writer and agent-config mutations.

services/kiloclaw/controller/src/routes/files.ts
```diff
@@ -725,8 +738,77 @@ export function registerFileRoutes(app: Hono, expectedToken: string, rootDir: st
...
+              { code: 'file_etag_conflict', error: 'File was modified externally' },
+              409
+            );
+          }
+        }
+
+        try {
+          backupFile(result, rootDir);
+        } catch (err) {
+          console.warn('[files] Failed to create backup, proceeding with write:', err);
```
St0rmz1: **Low, Security:** This `console.warn` logs the raw `err` from `backupFile`. The validation module elsewhere in this PR is careful to log only `errorCode(err)` so absolute paths and subprocess output stay out of logs, and the new spec point 5 says "Responses and logs MUST NOT expose staging paths or unrestricted subprocess output." This call is asymmetric with that stance: if `backupFile` ever surfaces the openclaw config or backup path in its error, it lands here in plaintext.

Suggestion: log via the same `errorCode(err)` helper (or another scrub) used in `openclaw-config-validation.ts`. The same applies to the matching `console.warn` in the legacy `/files/write` handler above.

services/kiloclaw/controller/src/openclaw-config-validation.ts
```diff
@@ -0,0 +1,243 @@
...
+      return {
+        valid: false,
+        reason: 'invalid',
+        issues: normalizeIssues(result.data.issues, stagePath),
+      };
+    }
+    return unavailableIssue('OpenClaw could not validate this configuration.');
+  } catch (error) {
+    console.error('[openclaw-config-validation] Validation failed unexpectedly:', errorCode(error));
+    return unavailableIssue('OpenClaw configuration validation could not be started.');
```
St0rmz1: **Low, Functional:** All non-timeout staging failures collapse to the same user-facing message: `"OpenClaw configuration validation could not be started."`. The dialog already differentiates `invalid` vs `validation-unavailable`, but a user repeatedly hitting `ENOSPC`, `EACCES`, or a stale staging file from an interrupted prior run sees the same message they'd see for "openclaw binary missing." That makes self-diagnosis impossible.

Suggestion: map a small set of recognized codes (`ENOSPC`, `EACCES`, `EEXIST`) to distinct bounded messages so the dialog can hint at retry vs. report. Keep server logging on `errorCode(error)` only.

services/kiloclaw/controller/src/openclaw-config-validation.ts
```diff
@@ -0,0 +1,243 @@
...
+
+function unavailableIssue(message: string): OpenclawConfigValidationResult {
+  return {
+    valid: false,
+    reason: 'validation-unavailable',
+    issues: [{ path: '', message }],
+  };
+}
+
+function referencesTargetConfig(value: unknown, configPath: string): boolean {
```
St0rmz1: **Low, Design:** This recursive walk visits every node in the parsed config to detect `$include` self-reference, with no depth cap and via native recursion. Realistically OpenClaw configs aren't deep enough to blow the stack, and `JSON.parse` already imposes some bounds, but a user-supplied candidate of pathological depth would still be processed unbounded here on every save attempt.

Suggestion: add a depth limit (e.g., 64) or convert to iterative traversal with an explicit worklist for defense in depth.

St0rmz1 (approved):

services/kiloclaw/controller/src/routes/files.ts
pandemicsyn: Fixed in caf0f79d2. Generic writes resolving to openclaw.json now share the config mutation queue, including in-root aliases.

services/kiloclaw/controller/src/routes/files.ts
pandemicsyn: Fixed in caf0f79d2. Backup failures now log only safe error-code metadata.

services/kiloclaw/controller/src/openclaw-config-validation.ts
pandemicsyn: Fixed in caf0f79d2. Common staging failures now return bounded actionable validation-unavailable messages.

services/kiloclaw/controller/src/openclaw-config-validation.ts
pandemicsyn: Fixed in caf0f79d2. Self-include inspection now uses iterative traversal with regression coverage for deep candidates.

PR #3555, opened 2026-05-28
fix(cloud-agent-next): recover leaked sandbox wrappers safely

eshurakov:
## Summary

- Prevent shared Cloudflare sandbox disk exhaustion from leaked wrapper processes by durably leasing physical wrappers, fencing launch/reuse/deletion on verified cleanup, and retrying teardown with persisted backoff.
- Introduce the Cloudflare-only `AgentSandbox` seam as the provider boundary for physical discovery, stop, keep-alive, and capacity operations; runtime, supervisor, deletion, interrupt, terminal, and router flows now converge on that lifecycle rather than performing ad-hoc sandbox teardown.
- Make cold workspace admission and rolling deployment handling fail-safe: low capacity triggers conservative stale cleanup plus recheck, the new capacity-admission recovery gate is restricted to explicit unusable-filesystem evidence, and legacy wrappers are identified through observable environment markers. Pre-existing recovery destruction for preparation infrastructure failures remains unchanged.

## Verification

- [x] Automated smoke test
- [x] Manual cloud agents test in a local browser

### DIFF
--- a/services/cloud-agent-next/src/agent-sandbox/cloudflare/cloudflare-agent-sandbox.test.ts
+++ b/services/cloud-agent-next/src/agent-sandbox/cloudflare/cloudflare-agent-sandbox.test.ts
@@ -1,0 +1,572 @@
+import { describe, expect, it, vi } from 'vitest';
+import type { Env, SandboxInstance } from '../../types.js';
+import type { SessionMetadata } from '../../persistence/session-metadata.js';
+import { WrapperClient } from '../../kilo/wrapper-client.js';
+import { WRAPPER_VERSION } from '../../shared/wrapper-version.js';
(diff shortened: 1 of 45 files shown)


eshurakov:
Addressed the actionable warm-bootstrap progress finding in `078978f90`: ordinary bootstrap now reports the `kilo_server` startup step on both warm and cold workspaces, with regression coverage for warm delivery.

I did not change the missing-metadata deletion branch. The public delete route short-circuits when session metadata is absent, and the Durable Object cannot reconstruct the provider sandbox or verify physical wrapper absence without that metadata. Retaining a non-`none` lease in that corrupted/orphaned state is intentionally fail-closed rather than erasing the only durable cleanup signal.

chrarnoldus (approved):

PR #3591, opened 2026-05-29
Add indication when model is retiring

chrarnoldus:
CleanShot 2026-05-29 at 11 48 07


apps/web/src/lib/ai-gateway/providers/openrouter/index.ts
```diff
@@ -79,6 +79,14 @@ function formatName(model: OpenRouterModel, preferredIndex: number) {
   const ageDays = (Date.now() / 1_000 - model.created) / (24 * 3600);
   const isNew = preferredIndex >= 0 && ageDays >= 0 && ageDays < 7;
   if (isNew) return model.name + ' (new)';
+  if (model.expiration_date) {
```
marius-kilocode, line 82: Shall we add a test?

marius-kilocode (approved):

PR #3602, opened 2026-05-29
feat(admin): redesign organizations table with entitlements/usage tabs

alexkgold:
Brings a lot more data into admin organizations table and splits out trial organizations from others.

image
image
image

## Summary

- Splits `/admin/organizations` into Entitlements and Usage tabs and reuses the same component for `/admin/organizations/trials`, with a renamed header, a Trial-specific create button label, and Usage as the default tab. The Organizations page no longer shows a Create button at all.
- Surfaces new operational signals (SSO, Provider/Model controls, Data Privacy, GitHub/GitLab/Slack integrations, Auto Top-Up, Stripe link in Links) by extending the admin `list` query with derived booleans and JSONB checks; the row-level Stripe link, Pylon link, and View Org link now live together in a Links cell.
- Centralizes the Stripe subscription status registry under `lib/admin/stripe-subscription-statuses` so the filter dropdown, badge, and TRPC input enum share one source of truth, extracts a `hasPlatformIntegrationSql` helper for the github/gitlab/slack EXISTS subqueries, fixes a count-query bug that broke the new Stripe Status filter, and nulls `subscription_amount_usd` for non-billable subscription statuses so the Subscription column doesn't display the price of a churned plan as if it were current MRR.
- Broadens the org-dashboard seed so every new column has variety (Stripe customer IDs, auto top-up, SSO domains, settings overrides, github/gitlab/slack integrations across active/pending/suspended).

## Verification

- [x] Re-seeded with `pnpm dev:seed app:org-dashboard` and walked `/admin/organizations` (Entitlements + Usage tabs) and `/admin/organizations/trials` (Trial Organizations header, Usage tab default, Create Org Trial button).
- [x] Confirmed the Stripe Status filter applies on `/admin/organizations` without crashing the count query (regression test added: `list — stripe_status filter` in `organization-admin-router.test.ts`).
- [x] Confirmed Tier Features pills, Integrations pills, Auto Top-Up On/Off, KiloClaw counts, and Stripe/Pylon/View Org links render with seed variety.
- [x] Verified `git check-ignore` still ignores arbitrary `.env.foo` while preserving committed `.env.local.example`, `.envrc`, and `apps/web/.env.test`.

## Visual Changes

| Before | After |
|---|---|
|  |  |

## Reviewer Notes

- The `latestSubscriptions` derived table no longer filters on `subscription_status = 'active'` so the new \`latest_stripe_status\` can populate; \`subscription_amount_usd\` is masked to \`NULL\` when the latest status is not in \`('active','trialing','past_due')\` to preserve the previous \"current paying value\" semantics.
- \`OrganizationsTable\` callsite ergonomics changed: instead of \`showCreateButton\` + \`createButtonLabel\`, callers pass an optional \`create={{ label }}\` config so a button label can never be rendered without a wired click target.
- \`countQuery\` now joins \`latestSubscriptions\` because \`finalWhereCondition\` may reference \`latest_subscriptions.subscription_status\`. A regression test exercises this path.
- The \`.gitignore\` change replaces the broad \`.env*\` rule with \`.env\` + \`.env.*\` plus \`!\` overrides for \`.env.example\`, \`.env.*.example\`, \`.env.test\`, and \`.envrc\` so committed env example/test files keep being trackable.


jrf0110 (approved):

apps/web/src/app/admin/components/OrganizationsTable.tsx
```diff
@@ -69,332 +94,210 @@ export function OrganizationsTable() {
...
-      </Button>{' '}
-    </>
+  const handleTabChange = useCallback(
+    (tab: TableVariant) => {
+      updateUrl({ ...sharedParams(), tab, page: '1' });
+    },
+    [sharedParams, updateUrl]
   );
 
+  const buttons = create ? (
```
evanjacobson, line 302: Problem: OrganizationsTable now only renders the create button when a create prop is passed, but /admin/organizations/page.tsx still calls <OrganizationsTable /> without that prop, so the main page loses the “Create Organization” action.

Solution: preserve the old default by either:
- passing create={{ label: 'Create Organization' }} from /admin/organizations/page.tsx, or
- defaulting create inside OrganizationsTable to { label: 'Create Organization' }.

apps/web/src/routers/organizations/organization-admin-router.ts
```diff
@@ -754,19 +713,32 @@ export const organizationAdminRouter = createTRPCRouter({
...
+        )
         .orderBy(orderCondition)
         .limit(limit)
         .offset((page - 1) * limit);
 
-      // Get total count using the same filtering logic
+      // Get total count using the same filtering logic. Must mirror baseQuery's
+      // joins so finalWhereCondition references (e.g. latestSubscriptions for the
+      // stripe_status filter) resolve.
       const countQuery = db
```
evanjacobson: Problem: the pagination count query always joins the latestSubscriptions windowed subquery, even though it only needs that join when stripe_status is filtering by latest subscription status. This adds avoidable historical subscription-table work to normal list requests.

Solution: build the count query without latestSubscriptions by default, and add that join only when stripe_status is present.

evanjacobson (changes requested):

apps/web/src/app/admin/components/OrganizationsTable.tsx
alexkgold: @evanjacobson -- this was an intentional change. Since all orgs are started as trials, I intentionally moved the create trial button to that page and removed it from the org page.

apps/web/src/routers/organizations/organization-admin-router.ts
alexkgold: @kilocode-bot -- please address

alexkgold: Addressed in 3d3db8d26 — count query now only joins `latestSubscriptions` when `stripe_status` is set; the no-filter path skips the join entirely. Added regression coverage for both branches in `organization-admin-router.test.ts`.

evanjacobson (approved):

PR #3603, opened 2026-05-29
fix(kiloclaw): derive upgrade banner rollout subject

evanjacobson:
## Summary

The upgrade banner could evaluate rollout eligibility with a different subject than the actual KiloClaw restart path, so users could see stale or incorrect upgrade availability. This change makes the banner resolve the rollout subject server-side from the authoritative instance row, matching the same legacy-vs-instance-keyed sandbox behavior used by `restartMachine({ imageTag: 'latest' })`.

It also keeps anonymous/latest-version lookups separate from instance-scoped rollout lookups, so callers cannot spoof Early Access state while existing non-instance consumers still get the public `:latest` result.

<details>
<summary>Details</summary>

- Centralizes image rollout subject derivation for legacy user-keyed rows and `ki_` instance-keyed rows.
- Resolves Early Access and rollout subject together from the instance row on the Worker side.
- Updates personal and org banner paths to pass active instance rows through the instance-scoped lookup.
- Renames the selector boundary to `rolloutSubject` so the internal API matches the new semantics.
- Adds coverage for anonymous lookups, missing Hyperdrive fallback, server-derived rollout subjects, and DO restart behavior.

</details>

## Verification

No manual testing performed; this is a backend rollout-selection and internal API-contract change with no visual/manual flow changes.

## Visual Changes

N/A

## Reviewer Notes

The important review path is consistency between the banner endpoint and `restartMachine({ imageTag: 'latest' })`: both should bucket by `userId` for legacy sandbox IDs and by instance UUID for `ki_` sandbox IDs. The HTTP query params remain backward-compatible; only the internal selector option was renamed to reflect that it now receives a rollout subject rather than always receiving an instance ID.


services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
```diff
@@ -1933,8 +1934,8 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
       throw Object.assign(new Error('Cannot apply pin: instance has no userId'), { status: 404 });
     }
 
-    const resolvedInstanceId = instanceId ?? this.s.userId;
-    await this.resolveImageStateForPin(imageTag, this.s.userId, resolvedInstanceId, {
+    const rolloutSubject = instanceId ?? this.s.userId;
```
pandemicsyn: Must-fix: applyPinnedVersion(null, instanceId) always uses the DB row id as the rollout subject because the platform route passes instanceId into this RPC. For legacy user-keyed instances, restartMachine({ imageTag: 'latest' }) buckets by imageRolloutSubjectFromSandboxId(this.s.sandboxId, this.s.userId), so clearing a pin can write a different trackedImageTag than the actual upgrade path and leave DB/DO state inconsistent when there is no follow-up restart or that restart fails. Derive the rollout subject from loaded DO state here, for example imageRolloutSubjectFromSandboxId(this.s.sandboxId, this.s.userId), and add a legacy sandbox regression test for clearing a pin.

pandemicsyn (changes requested):
Found one must-fix issue in the pin-clear rollout subject path.

pandemicsyn (approved):

PR #3611, opened 2026-05-31
fix(kiloclaw): replace provision locks with registry admission

pandemicsyn:
## Summary

- Replace the unsafe Vercel/Postgres advisory-lock provisioning flow with durable `KiloClawRegistry` admission reservations so fresh creates are serialized before provider allocation.
- The change addresses recurring production failures where a successful provision logged `Failed to release provision context lock`, then repeated same-context `kiloclaw.provision` retries hung for about 120 seconds without reaching the Worker. Axiom showed this pattern again on 2026-05-31, including eight timeout-shaped retries after one successful create.
- Add atomic unresolved-reservation enforcement, canonical active/subscription reconciliation, completion repair, destroy fencing, and alarm-retry cleanup so partially successful or failed provisions cannot create duplicate infrastructure or revive destroyed routes.
- Remove the web advisory-lock helper and map Worker admission/finalization conflicts into user-facing tRPC conflicts while preserving existing-instance update and subscription-recovery behavior.

## Visual Changes

N/A

## Reviewer Notes

- This is a cross-service correctness change: `KiloClawRegistry` is now a fresh-provision admission boundary rather than best-effort routing metadata.
- The Durable SQLite migration adds non-routable `provision_reservations` state with a partial unique unresolved-admission guard.
- The Worker/web cutover is intentionally shipped together under the accepted assumption that Worker deployment completes before the slower Vercel deployment becomes routable; monitor Axiom for new `provision_in_progress`, `provision_completion_pending`, and 120-second timeout signals after rollout.


services/kiloclaw/src/routes/platform.ts
```diff
@@ -1275,16 +1495,49 @@ platform.post('/provision', async c => {
...
       });
+      if (instanceMarkedDestroyed) {
+        await withResolvedDORetry(
+          c.env,
+          userId,
+          provisionedInstanceId,
+          stub => stub.allowProvisionReservationReleaseOnFinalize(),
+          'allowProvisionReservationReleaseOnFinalize'
+        ).catch(releaseSignalError => {
+          instanceMarkedDestroyed = false;
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in 4c7fc4def.

apps/web/src/routers/kiloclaw-router.ts
```diff
@@ -575,6 +571,27 @@ function getKiloClawApiErrorPayload(err: KiloClawApiError): { message?: string;
   }
 }
 
+function handleProvisionError(err: unknown): never {
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in 4c7fc4def.

services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
```diff
@@ -2511,6 +2531,113 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
...
+              {
+                registryKey,
+                doKeysTried: legacyDoKeys,
+                activeEntryCount: entries.length,
+              }
+            );
+          }
+        }
+      }
+      if (!releaseProvisionReservation) {
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in 4c7fc4def.

services/kiloclaw/src/routes/platform.ts
```diff
@@ -1195,11 +1379,47 @@ platform.post('/provision', async c => {
...
+      const activeAfterProvision = await getActiveProvisionContextInstance(c.env, userId, orgId);
+      if (activeAfterProvision?.id !== instanceId) {
+        return jsonError('Instance was destroyed during update', 409, 'instance_destroyed');
+      }
+    }
   } catch (err) {
+    if (freshReservationAdmitted && provisionRegistry) {
+      try {
+        if (freshProviderWorkStarted) {
+          await provisionRegistry.stub.failFreshProvision(
```
St0rmz1, line 1404: **Moderate, Functional:** When fresh provider work has started and then throws, this sets the reservation to `failed_requires_reconciliation`. That status counts as unresolved in the partial unique index, so it blocks every subsequent `beginFreshProvision` for this user.

For an ordinary transient provider failure (a Fly 500 or timeout) no Postgres row and no subscription get created, so the conflict-repair path at the top of the route finds no subscribed active instance to repair and returns `provision_in_progress`. The user is then locked out of retrying. The only automatic recovery is the instance DO's stale-provision cleanup alarm eventually destroying the half-provisioned DO, which releases the reservation via `finalizeDestroyedInstance`.

Two things worth confirming: that the stale-cleanup alarm reliably releases these, and on what timescale. Consider whether a failure before any infrastructure is created should `releaseFreshProvision` (retryable) rather than `failFreshProvision` (fail closed). Fail closed is right once infra may exist, but the window before the first resource is allocated is safely retryable.

services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
```diff
@@ -2511,6 +2532,126 @@ export class KiloClawInstance extends DurableObject<KiloClawEnv> {
...
+      const registryKeys = [`user:${userId}`];
+      if (orgId) registryKeys.push(`org:${orgId}`);
+
+      for (const registryKey of registryKeys) {
+        const registryStub = this.env.KILOCLAW_REGISTRY.get(
+          this.env.KILOCLAW_REGISTRY.idFromName(registryKey)
+        );
+        if (registryInstanceId) {
+          if (releaseAllowed) {
+            await registryStub.finalizeDestroyedInstance(
```
St0rmz1, line 2593: **Low, Functional:** This loops both `user:${userId}` and `org:${orgId}` registry keys. For an org instance the old code called `destroyInstance`, which updated zero rows in the user registry (a no-op there). `finalizeDestroyedInstance` does an `insert ... onConflictDoUpdate set destroyed_at`, so it now inserts a fresh destroyed tombstone into the user registry for an instanceId that never lived there.

Not a correctness problem since instanceIds are unique and `listInstances` filters on `destroyed_at IS NULL`, but it accumulates dead rows in the personal registry on every org instance destroy. Consider finalizing only in the registry that actually owns the reservation, or skipping the upsert when no matching row exists in that registry.

services/kiloclaw/src/durable-objects/kiloclaw-registry.ts
```diff
@@ -110,7 +151,200 @@ export class KiloClawRegistry extends DurableObject<KiloClawEnv> {
...
+        .get();
+      if (!reservation) {
+        if (reservationRequired)
+          throw new Error('Provision reservation not found during completion');
+        return false;
+      }
+      if (reservation.status === 'released') {
+        throw new Error('Cannot complete a released provision reservation');
+      }
+      if (
```
St0rmz1, line 317: **Low, Functional:** On the repair path (`reservationRequired === false`), a reservation in `failed_requires_reconciliation` passes the `released` check and then skips this `reservationRequired` guard, so it gets promoted straight to `completed` and its route is published.

This is probably the intended self-heal, since it only fires when a subscribed active instance actually exists, which is strong evidence the provision really succeeded. But it quietly resolves a state that the admission path treats as requiring reconciliation, and that intent is not obvious from the code. Worth a comment here noting that repair deliberately clears a reconciliation-required reservation when canonical state proves success, or a distinct telemetry event when repair clears a failed state so it is observable.

apps/web/src/lib/kiloclaw/provision-error-handler.ts
```diff
@@ -0,0 +1,28 @@
+import { TRPCError } from '@trpc/server';
+import { UpstreamApiError } from '@/lib/trpc/init';
+import { KiloClawApiError } from './kiloclaw-internal-client';
+
+type ProvisionErrorPayload = { message?: string; code?: string };
+
+type ProvisionErrorPayloadReader = (err: KiloClawApiError) => ProvisionErrorPayload;
+
+export function handleProvisionError(err: unknown, getPayload: ProvisionErrorPayloadReader): never {
+  if (err instanceof KiloClawApiError && (err.statusCode === 409 || err.statusCode === 503)) {
```
St0rmz1: **Low, Design:** This only maps `409` and `503` responses, so the `404 instance_not_found` that the platform route can now return falls through to the raw rethrow at the bottom. In a destroy-during-update race the personal `provision` and `updateConfig` callers would surface a raw `KiloClawApiError` instead of a clean tRPC code. Worth deciding on a deliberate mapping for `instance_not_found` (likely `NOT_FOUND`, or `CONFLICT` to match the others) rather than letting it leak through.

services/kiloclaw/src/durable-objects/kiloclaw-registry.ts
```diff
@@ -206,6 +513,10 @@ export class KiloClawRegistry extends DurableObject<KiloClawEnv> {
       const instance = await getActivePersonalInstance(db, userId);
 
       if (instance) {
+        const hasSubscription = await hasSubscriptionForInstance(db, instance.id);
+        if (!hasSubscription && isInstanceKeyedSandboxId(instance.sandboxId)) {
+          return;
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in 3a4db36f3.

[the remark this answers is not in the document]
pandemicsyn: Fixed in 3a4db36f3.

services/kiloclaw/src/durable-objects/kiloclaw-instance/index.ts
pandemicsyn: Fixed in 76b4aef50.

services/kiloclaw/src/durable-objects/kiloclaw-registry.ts
pandemicsyn: Fixed in 76b4aef50.

apps/web/src/lib/kiloclaw/provision-error-handler.ts
pandemicsyn: Fixed in 76b4aef50.

services/kiloclaw/src/routes/platform.ts
pandemicsyn: Reviewed: this is intentional fail-closed behavior once provider work has started because infrastructure may already exist. The provisioned DO alarm runs on the idle cadence (30m + jitter) and destroy finalization releases the reservation after cleanup; no code change.

St0rmz1 (approved):

PR #3628, opened 2026-06-01
chore(cloud-agent-next): bump sandbox runtime packages

eshurakov:
## Summary
- Bump Cloud Agent Next Kilo CLI/SDK pins to the newest release allowed by the repository release-age policy.
- Bump Cloudflare sandbox/container SDK pins and Docker sandbox image tags while keeping package, image, and lockfile references synchronized.
- Regenerate the fallback slash-command catalog and fix its generated type ownership to avoid a circular type import.

### DIFF
--- a/services/cloud-agent-next/src/shared/default-slash-commands.generated.ts
+++ b/services/cloud-agent-next/src/shared/default-slash-commands.generated.ts
@@ -17,7 +17,7 @@ export type SlashCommandInfo = {
  *
  * Regenerate with `pnpm --filter cloud-agent-next update-default-slash-commands`.
  */
-export const DEFAULT_SLASH_COMMANDS_SOURCE = 'kilo@7.2.49';
+export const DEFAULT_SLASH_COMMANDS_SOURCE = 'kilo@7.3.12';
 
 /**
  * Default slash command catalog used when no live wrapper-reported catalog is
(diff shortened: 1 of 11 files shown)


services/cloud-agent-next/src/shared/default-slash-commands.generated.ts
```diff
@@ -17,34 +17,42 @@ export type SlashCommandInfo = {
  *
  * Regenerate with `pnpm --filter cloud-agent-next update-default-slash-commands`.
  */
-export const DEFAULT_SLASH_COMMANDS_SOURCE = 'kilo@7.2.49';
+export const DEFAULT_SLASH_COMMANDS_SOURCE = 'kilo@7.3.12';
```
jeanduplessis, line 20: Current version is 7.3.16 - worth bumping to latest?

jeanduplessis (approved):
Non-blocking: consider updating to latest patch versions of Kilo

services/cloud-agent-next/src/shared/default-slash-commands.generated.ts
jeanduplessis: Nevermind, I read the PR description again _"newest release allowed by the repository release-age policy."_

PR #3631, opened 2026-06-01
fix(kiloclaw): avoid sidebar status polling

jeanduplessis:
## Summary
KiloClaw sidebars now use a lightweight database-backed nav-state query instead of the polling status query.

### Why this change is needed
The sidebar only needs to know whether an active KiloClaw instance exists. Using the full status query for that state caused unnecessary 10-second polling and could call the KiloClaw worker from global navigation.

### How this is addressed
- Add personal and organization `getNavState` tRPC queries that read active instance state from the database.
- Switch personal and organization app sidebars to use nav state instead of KiloClaw status.
- Invalidate nav-state caches when KiloClaw mutations can change instance presence.
- Add regression coverage for absent, active, destroyed, and scope-isolated nav-state behavior.

## Human Verification
Verified polling stops and nav-state works as expected in local dev env.

### Human Reviewer Flags
- Adds a dedicated lightweight nav-state endpoint instead of reusing the richer status endpoint, trading a small API addition for lower sidebar polling cost.

### Code Reviewer Agent
<details>

- Personal nav-state tests assert the KiloClaw worker client is not constructed or queried.
- Organization nav-state tests cover org scoping and ensure personal instances do not leak into organization navigation.
- Destroyed instances are ignored in both personal and organization nav-state paths.

</details>

### DIFF
--- a/apps/web/src/routers/kiloclaw-router.ts
+++ b/apps/web/src/routers/kiloclaw-router.ts
@@ -2989,7 +2989,6 @@ export const kiloclawRouter = createTRPCRouter({
     const instance = await getActiveInstance(ctx.user.id);
     return {
       hasActiveInstance: instance !== null,
-      instanceId: instance?.id ?? null,
     };
   }),
 
--- a/apps/web/src/routers/kiloclaw-router.ts
+++ b/apps/web/src/routers/kiloclaw-router.ts
@@ -2985,6 +2985,14 @@ export const kiloclawRouter = createTRPCRouter({
     } satisfies KiloClawDashboardStatus;
   }),
 
+  getNavState: baseProcedure.query(async ({ ctx }) => {
+    const instance = await getActiveInstance(ctx.user.id);
+    return {
+      hasActiveInstance: instance !== null,
+      instanceId: instance?.id ?? null,
+    };
+  }),
+
   getDiskUsage: baseProcedure.query(async ({ ctx }) => {
     const instance = await getActiveInstance(ctx.user.id);
     if (!instance) {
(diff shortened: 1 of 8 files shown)


RSO (approved):

apps/web/src/routers/kiloclaw-router.ts
```diff
@@ -2985,6 +2985,14 @@ export const kiloclawRouter = createTRPCRouter({
     } satisfies KiloClawDashboardStatus;
   }),
 
+  getNavState: baseProcedure.query(async ({ ctx }) => {
+    const instance = await getActiveInstance(ctx.user.id);
+    return {
+      hasActiveInstance: instance !== null,
+      instanceId: instance?.id ?? null,
```
St0rmz1: **Low, Design:** `getNavState` returns `instanceId: instance?.id ?? null` unconditionally, but the adjacent `getStatus` deliberately guards the same field: `instanceId: workerInstanceId(instance) ? instance.id : null`, with a comment that returning a legacy instance's row UUID makes the frontend/gateway resolve the wrong DO. No live bug today since the sidebars only read `hasActiveInstance`, but the field is public and asserted in tests, so a future caller could use it to resolve the worker DO and hit exactly that legacy failure. Either drop `instanceId` from the nav-state payload (no consumer uses it, and it keeps the response minimal) or apply the same `workerInstanceId(instance)` guard so the two endpoints agree. Same applies to `organization-kiloclaw-router.ts` getNavState.

jeanduplessis: Fixed in 19ff3d839. Removed the unused instanceId field from both personal and organization getNavState payloads and updated regression expectations. Nav state now exposes only hasActiveInstance, avoiding ambiguous DB-row versus worker-routing identity semantics.

PR #3635, opened 2026-06-01
feat(cloud-agent-next): show context window usage

eshurakov:
## Summary

- Show Cloud Agent Next context-window usage below the composer with a CLI-style compact label such as `239.1K (24%)`, while preserving exact token counts in the tooltip and accessible label.
- Derive usage from the latest eligible root-session assistant response using input, output, reasoning, cache-read, and cache-write tokens rather than cumulative session spend.
- Add a Cloud Agent Next-specific exact-model-ID context-length map from the organization catalog and fail closed when provider identity, runtime payloads, or catalog metadata cannot be trusted.
- Architecturally, expose context usage as a derived session-manager atom and keep the context-length projection separate from shared model-picker options.

## Verification

- [x] Verified the changes running locally.

## Visual Changes

Screenshot 2026-06-01 at 13 06 36

## Reviewer Notes

- Usage intentionally follows CLI informational semantics and is not a compaction-threshold meter.
- Runtime model IDs are joined to exact catalog IDs only; the UI hides the metric instead of guessing aliases or capacities.

### DIFF
--- a/apps/web/src/components/cloud-agent-next/CloudChatPage.tsx
+++ b/apps/web/src/components/cloud-agent-next/CloudChatPage.tsx
@@ -37,6 +37,8 @@ import {
 } from './terminal-tabs';
 import { isMessageStreaming } from './types';
 import { useOrganizationModels } from './hooks/useOrganizationModels';
+import { ContextUsageIndicator } from './ContextUsageIndicator';
+import { resolveContextWindow } from './model-context-lengths';
 import { useSlashCommandSets } from '@/hooks/useSlashCommandSets';
 import { useCelebrationSound } from '@/hooks/useCelebrationSound';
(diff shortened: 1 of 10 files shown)


jeanduplessis (approved):
Approved. Reviewed the final combined diff and surrounding context (root-session filtering in `messagesListAtom`, `messageIds` chronological ordering via `insertSorted`, `AssistantMessage.tokens` shape, and the exact-ID catalog join). Logic, nullish/fallback, streaming zero-output, session-switch reset, and catalog conflict-dedup edge states all resolve correctly and are well covered by tests. No PostHog/analytics paths are affected. No blocking issues; one minor non-blocking test-clarity nit (unescaped `.` / brittle markup assertion in `context-usage-indicator.test.ts`).

PR #3636, opened 2026-06-01
fix(kilo-pass): centralize bonus eligibility decisions

RSO:
## Summary
- Centralize Kilo Pass bonus eligibility and rounding in shared decision helpers used by both usage-triggered issuance and `kiloPass.getState` projections.
- Read the initial welcome-promo eligibility decision from the subscription's first issuance so month-2 reused-card subscriptions fall back to the normal ramp while eligible first-month/new-card subscriptions keep their 50% bonus display.
- Update renewal-row logic to prefer the server-projected next bonus when there is no scheduled plan change, with regression fixtures for production-shaped issuance history.

## Verification
Not manually verified; no browser validation performed because this change is primarily API-state and projection logic.

## Visual Changes
N/A

## Reviewer Notes
Review focus: the new shared decision helper is now the source of truth for monthly bonus eligibility, while scheduled-change UI rows still use local projection because server `nextBonusCreditsUsd` only describes the active subscription. Stripe subscriptions only enforce welcome-promo eligibility when an initial settled-payment decision exists, preserving legacy/null issuance behavior.

### DIFF
--- a/apps/web/src/components/profile/kilo-pass/KiloPassActiveSubscriptionCard.test.ts
+++ b/apps/web/src/components/profile/kilo-pass/KiloPassActiveSubscriptionCard.test.ts
@@ -23,6 +23,7 @@ function buildSubscription(
     startedAt: null,
     refillAt: null,
     nextBillingAt: null,
+    nextBonusCreditsUsd: 9.5,
     currentPeriodBaseCreditsUsd: 19,
     currentPeriodUsageUsd: 0,
     currentPeriodBonusCreditsUsd: 0,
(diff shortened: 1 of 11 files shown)


apps/web/src/components/profile/kilo-pass/KiloPassActiveSubscriptionCard.test.ts
```diff
@@ -23,6 +23,7 @@ function buildSubscription(
     startedAt: null,
     refillAt: null,
     nextBillingAt: null,
+    nextBonusCreditsUsd: 9.5,
```
jeanduplessis: SUGGESTION (non-blocking): This arbitrary `9.5` default now feeds every buildSubscription() caller. In computeRefillRowModel, when scheduledChange is null the model returns this server-projected value and skips the local computeYearlyCadenceMonthlyBonusUsd / computeMonthlyCadenceBonusPercent branch. Existing no-scheduled-change tests ("pluralizes day vs days", "yearly pending cancellation: active_until…") don't assert bonusUsd so they still pass, but they now exercise the server-projection branch rather than the local computation they set up. Since 9.5 maps to no real tier bonus, consider defaulting to null (keeping untargeted tests on the local path) and setting nextBonusCreditsUsd explicitly only in the test that asserts it.

Trace: fixture default (test.ts:26) -> computeRefillRowModel serverProjectedBonusUsd (logic.ts) -> local bonus branch skipped when scheduledChange is null.
Impact: no breakage; minor readability/coverage-intent nit only.

jeanduplessis (approved):

apps/web/src/components/profile/kilo-pass/KiloPassActiveSubscriptionCard.test.ts
RSO: Applied in e74dee034. `buildSubscription()` now defaults `nextBonusCreditsUsd` to `null`, so untargeted no-scheduled-change tests stay on the local computation path, and the server-projection path is explicit in the dedicated test.

Verification: `pnpm format` and `pnpm test -- apps/web/src/components/profile/kilo-pass/KiloPassActiveSubscriptionCard.test.ts`.

PR #3651, opened 2026-06-01
feat(kiloclaw) "one click" install of a ClawByte into KiloClaw instance

St0rmz1:
## Summary

Adds an install flow for ClawBytes so a user can go from a byte in the catalog to running it in their KiloClaw chat with one confirmation.

- New route at `claw/install/[source]/[slug]` renders a confirmation card showing the byte title and description with Confirm Install and Cancel actions. The page fetches and verifies on the server, and dispatches only on an explicit Confirm click through a POST mutation, so loading the URL by itself never runs anything.
- Byte content is fetched from a registry of approved sources (today `byte` maps to the catalog `data.json`). The host always comes from the registry rather than user input, the slug is percent encoded into one path segment, and redirects are not followed.
- Payloads are signed with Ed25519. The app verifies the signature against a pinned public key, checks the key id, enforces an age window, and confirms the signed slug matches the requested slug. Only the signed fields (slug, title, description, prompt) are modelled, so unsigned content can never be shown.
- Access is checked before any fetch. A user without an active subscription is routed to set one up, so fetching and verifying a byte happens only for paid users.
- Dispatch is bound to the exact payload the user reviewed. The Confirm click sends the reviewed signature, the server fetches and verifies again, and refuses if the signature changed since the page rendered.
- Each install opens its own conversation (a new `forceNewConversation` option on the shared message primitive) instead of appending to the user's most recent chat, and then navigates straight into that conversation.
- The app reaches the chat service over an authenticated internal HTTP route (`POST /internal/v1/post-message-as-user`) guarded by a shared secret, because the app and the chat service run on different platforms and cannot use a service binding.
- If the user has no instance yet, they are routed to provisioning. Install intent is not persisted across that flow, since the install link is repeatable.

## Verification

Tested locally against a running app, catalog, chat service, and a provisioned instance.

- [x] Paid user with an instance: open the install link, click Confirm Install, land in a new conversation with the byte prompt sent and a reply.
- [x] Signature binding: a changed signature is refused with a clear message instead of dispatching.
- [x] No instance: routed to provisioning rather than dispatching.
- [x] Logged out: bounced to sign in, then returned to the confirmation page.
- [x] Unknown slug: shows a 404.
- [x] Loading the URL by itself never dispatches; the Confirm click is required.

## Visual Changes

| Before | After |
|---|---|
| No install page existed | Confirmation card: ClawByte badge with icon, byte title, the line "This ClawByte installs a skill to:" followed by the description, and Confirm Install and Cancel buttons |

## Reviewer Notes

Suggested focus areas:

- Signature verification and the bound dispatch in `apps/web/src/lib/kiloclaw/install.ts` and `install-dispatch.ts`.
- The access check that runs before any byte fetch in the install page.
- The shared secret auth on the internal chat route.
- The source registry as the only place a new origin can be added.


apps/web/src/lib/kiloclaw/install-dispatch.ts
```diff
@@ -0,0 +1,211 @@
...
+  // prompt); if it no longer matches, the byte changed since the confirmation
+  // page rendered, so refuse rather than run a prompt the user didn't approve.
+  if (payload.signature !== expectedSignature) {
+    throw new TRPCError({
+      code: 'CONFLICT',
+      message: 'This byte changed since you reviewed it. Please reload and confirm again.',
+    });
+  }
+
+  const instance = await deps.getActiveInstance(userId);
```
pandemicsyn, line 127: Must-fix: This dispatch picks `getActiveInstance(userId)` after generic `clawAccessProcedure` access, but those are not bound to the same billing anchor. In an inconsistent state, such as a current subscription anchored to a destroyed or different row while an orphaned active row remains, the gate can pass and this sends a prompt into an unentitled runtime. Require access for this exact instance before resolving sandbox/chat, for example with `requireKiloClawAccessAtInstance(userId, instance.id)`, or reuse the existing personal anchor resolution and fail closed on mismatch.

apps/web/src/lib/kiloclaw/kilo-chat-internal-client.ts
```diff
@@ -0,0 +1,110 @@
...
+  if (!INTERNAL_API_SECRET) {
+    throw new Error(
+      'INTERNAL_API_SECRET is not configured — cannot authenticate to kilo-chat internal routes'
+    );
+  }
+
+  const url = `${getKiloChatBaseUrl()}/internal/v1/post-message-as-user`;
+  let res: Response;
+  try {
+    res = await fetch(url, {
```
pandemicsyn, line 81: Suggested/nit: This secret-bearing fetch uses a base URL from `NEXT_PUBLIC_KILO_CHAT_URL` and follows redirects by default. A misconfigured or redirecting destination can send the internal API key and prompt outside kilo-chat. Prefer a server-only internal URL or allowlisted origin, and set `redirect: 'error'` before attaching the key.

apps/web/src/app/(app)/claw/install/[source]/[slug]/InstallClient.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  return (
+    <div className="mx-auto flex min-h-[60vh] w-full max-w-lg items-center px-6 py-12">
+      <Card className="w-full text-left">
+        <CardHeader>
+          <Badge variant="secondary" className="w-fit gap-1.5">
+            <KiloClawbsterIcon className="h-4 w-auto" />
+            {sourceLabel}
+          </Badge>
+          <CardTitle className="mt-2 text-xl">{payload.title}</CardTitle>
```
Comment (pandemicsyn) [inline apps/web/src/app/(app)/claw/install/[source]/[slug]/InstallClient.tsx]: Suggested/nit: `payload.title` and `payload.description` are external signed strings and can contain long unbroken tokens. On narrow screens, they can overflow this fixed-width card instead of reflowing. Add `break-words` or `overflow-wrap:anywhere` to both text blocks.

apps/web/src/app/(app)/claw/install/[source]/[slug]/InstallClient.tsx
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  const busy = install.isPending || navigating;
+
+  return (
+    <div className="mx-auto flex min-h-[60vh] w-full max-w-lg items-center px-6 py-12">
+      <Card className="w-full text-left">
+        <CardHeader>
+          <Badge variant="secondary" className="w-fit gap-1.5">
+            <KiloClawbsterIcon className="h-4 w-auto" />
```
Comment (pandemicsyn) [inline apps/web/src/app/(app)/claw/install/[source]/[slug]/InstallClient.tsx:89]: Suggested/nit: This badge icon is decorative, but the SVG is exposed to assistive tech without a label. Mark this use `aria-hidden="true"` and `focusable="false"` by passing those props through `KiloClawbsterIcon`, or hide the SVG at this call site.

pandemicsyn (changes requested):

pandemicsyn (approved):

apps/web/src/lib/kiloclaw/install-dispatch.ts
St0rmz1: Fixed. `dispatchInstallFromSource` now calls `requireKiloClawAccessAtInstance(userId, instance.id)` right after resolving the active instance, before resolving the sandbox or dispatching. It fails closed with FORBIDDEN/NOT_FOUND when the resolved instance is not entitled, so a passing `clawAccessProcedure` gate no longer lets a prompt reach an unentitled runtime in an inconsistent billing state. Added a unit test asserting it refuses and never dispatches when the per-instance check throws.

apps/web/src/lib/kiloclaw/kilo-chat-internal-client.ts
St0rmz1: Fixed. The internal fetch now sets `redirect: 'error'`, so the key cannot be forwarded via a 3xx. `getKiloChatBaseUrl` also enforces an origin allowlist (`isAllowedKiloChatOrigin`) and refuses before the key is attached unless the resolved origin is the deployed kilo-chat origin (`https://chat.kiloapps.io`) or loopback for local dev. A unit test covers an off-allowlist origin being refused with no fetch. Moving off the `NEXT_PUBLIC_` URL to a server-only variable is tracked as a follow-up, since it touches env provisioning across all environments.

apps/web/src/app/(app)/claw/install/[source]/[slug]/InstallClient.tsx
Comment (St0rmz1) [reply apps/web/src/app/(app)/claw/install/[source]/[slug]/InstallClient.tsx]: Fixed. Added `break-words` to both the title and description so long unbroken tokens reflow inside the card instead of overflowing on narrow screens.

apps/web/src/app/(app)/claw/install/[source]/[slug]/InstallClient.tsx
Comment (St0rmz1) [reply apps/web/src/app/(app)/claw/install/[source]/[slug]/InstallClient.tsx:89]: Fixed. `KiloClawbsterIcon`'s svg now sets `aria-hidden="true"` and `focusable="false"`, so the decorative badge icon is hidden from assistive tech.

PR #3671, opened 2026-06-02
feat(cloud-agent): add Kilo SDK session facade

eshurakov:
### Why

Cloud Agent sessions need a stable, authenticated SDK surface so external clients can attach without depending on sandbox topology or exposing internal wrapper details. The facade gives `@kilocode/sdk/v2` consumers a narrow public contract for owned root sessions while preserving durable admission, runtime fencing, and cold-history access when a wrapper is unavailable.

### What was done

- Add an authenticated `/kilo` facade backed by a per-user `UserKiloFacade` Durable Object. It supports owned root-session listing, detail and message reads, async text prompts with optional Kilo agent/model selection, abort, global SSE, and session-scoped SSE.
- Keep reads live-first through the sandbox wrapper `/kilo-proxy` path, with a persisted `session-ingest` fallback for session detail and transcript reads. Cold transcript pages use native cursors, bounded materialization, and omission metadata for oversized or forward-compatible unsupported items.
- Route `prompt_async` and abort through `CloudAgentSession` instead of directly to the wrapper so admission and interruption remain durable across cold runtimes.
- Add a fenced wrapper global-feed producer path: the wrapper converts runtime SSE into an authenticated WebSocket feed, and the facade filters, virtualizes, and fans events back out as public SSE.
- Keep the public projection intentionally narrow: child sessions remain hidden, unsupported SDK routes return `501`, typed outer routing fields are virtualized, and nested owner-visible payload content is preserved.

### High-level SDK facade architecture

```mermaid
sequenceDiagram
  autonumber
  participant SDK as @kilocode/sdk/v2 client
  participant Worker as cloud-agent-next Worker
  participant Facade as UserKiloFacade DO
  participant SessionDO as CloudAgentSession DO
  participant Wrapper as sandbox wrapper
  participant Runtime as in-process Kilo SDK server
  participant Ingest as session-ingest / SessionIngestDO

  SDK->>Worker: Authenticated /kilo/* request
  Worker->>Facade: Route with authenticated user context

  rect rgb(235, 245, 255)
    Note over SDK,Ingest: Live-first session detail and transcript reads
    Facade->>Wrapper: GET /kilo-proxy/session/:id[/message]
    Wrapper->>Runtime: Forward SDK read
    alt Runtime is available
      Runtime-->>Wrapper: SDK-compatible response
      Wrapper-->>Facade: Return live response
    else Runtime is unavailable
      Facade->>Ingest: Read persisted snapshot or transcript page
      Ingest-->>Facade: Bounded cold projection with native cursor metadata
    end
    Facade-->>SDK: Projected public response via Worker
  end

  rect rgb(245, 245, 235)
    Note over SDK,Wrapper: Durable mutations
    SDK->>Facade: POST prompt_async or abort via Worker
    Facade->>SessionDO: Admit prompt or interrupt execution
    opt Runtime is ready
      SessionDO->>Wrapper: Dispatch admitted work
    end
    Facade-->>SDK: Accepted or idempotent response via Worker
  end

  rect rgb(240, 250, 240)
    Note over SDK,Runtime: Fenced public event stream
    Runtime-->>Wrapper: /global/event SSE
    Wrapper->>Worker: Authenticated producer WebSocket upgrade
    Worker->>SessionDO: Validate producer identity
    SessionDO-->>Worker: Producer accepted
    Worker->>Facade: Forward producer socket
    Wrapper-->>Worker: Runtime event frames
    Worker-->>Facade: Forward event frames
    Facade-->>SDK: Filtered and virtualized public SSE
  end
```

### Architecture decision

**Decision:** expose Cloud Agent sessions through a narrow authenticated `/kilo` facade that implements the relevant `@kilocode/sdk/v2` contract, rather than introducing a Cloud Agent-specific client API.

**Context:** existing UI clients already integrate with the Kilo SDK for session reads, conversation flows, async prompts, aborts, and event consumption. Cloud Agents provide similar capabilities, but their internal lifecycle differs: sessions can move between live sandbox runtimes and persisted history, mutations need durable admission, and wrapper topology must remain private.

**Rationale:** implementing the relevant Kilo SDK surface gives those clients a familiar integration path with minimal additional client-side work. The facade adapts the Cloud Agent lifecycle behind that contract: reads are live-first with bounded persisted fallback, prompts and aborts continue through `CloudAgentSession`, and public events are filtered and virtualized before reaching SDK consumers. The initial facade intentionally supports only the subset needed by external clients; unsupported SDK routes return `501` rather than implying broader compatibility.

**Alternatives considered:**

- **Create a Cloud Agent-specific client API.** This could mirror the internal Cloud Agent lifecycle more directly, but UI clients would need a second integration for substantially the same conversation flows. It would also create another protocol to document, maintain, and evolve alongside the Kilo SDK.
- **Expose sandbox wrappers or runtime SDK servers directly.** This would reduce facade code, but it would couple clients to sandbox topology and runtime availability. It would also bypass the stable ownership boundary, durable mutation path, persisted-history fallback, and public event filtering provided by the facade.

**Consequences:** Cloud Agents gain an adaptation layer that must preserve SDK-compatible behavior across live and cold sessions. In return, existing Kilo SDK clients can integrate with Cloud Agent-backed sessions without adopting a parallel protocol, while the internal runtime remains private and independently evolvable.

## Verification

- [x] Exercised the authenticated facade locally through `@kilocode/sdk/v2`, including session attach and reads, async chat, abort, and event consumption.
- [x] Built a quick VS Code extension proof of concept against the facade and verified that an extension can attach to a Cloud Agent-backed SDK session and use the conversation flow.

## Visual Changes

N/A

## Reviewer Notes

- Public SDK access is limited to user-owned Cloud Agent-backed root sessions with current organization access. Child sessions remain hidden.
- Cold transcript pages use an 8 MiB aggregate budget, a page size up to 100, and at most 128 persisted message-row scans. Oversized rows are skipped internally; if a native cursor cannot safely represent progress, the facade returns `413` rather than an ambiguous continuation.
- Global-feed producer fencing is connection-oriented: identity is validated at WebSocket upgrade and facade acceptance time, and newer wrapper connections replace older sockets. Frames are not revalidated against `CloudAgentSession` after socket acceptance.
- Deploy `session-ingest` before `cloud-agent-next`; the facade consumes additive session-ingest RPCs. The `cloud-agent-next` deployment must include the Wrangler `v5` migration for the SQLite-backed `UserKiloFacade` Durable Object.
- Existing warm wrappers do not expose the new `/kilo-proxy` or global-feed producer paths. Recycle or drain wrappers during rollout so live SDK reads and SSE delivery consistently use the new wrapper behavior.


services/cloud-agent-next/src/kilo-facade/user-kilo-facade.ts
```diff
@@ -0,0 +1,1634 @@
...
+}
+
+export { publicCloudAgentDirectory };
+
+function isKiloSdkSessionInfo(value: unknown, kiloSessionId: string): value is KiloSdkSessionInfo {
+  return (
+    isRecord(value) &&
+    value.id === kiloSessionId &&
+    typeof value.slug === 'string' &&
+    typeof value.projectID === 'string' &&
```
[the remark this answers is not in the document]
eshurakov: Keeping the per-frame validation intentionally. The session DO can clear or replace the wrapper runtime fence after WebSocket upgrade while the facade socket remains connected, and the facade has no push revocation path for that transition. Caching validation for the socket lifetime would allow stale producers to continue publishing until reconnect. The throughput tradeoff is real, but a behavior-preserving optimization needs explicit revocation across every runtime-fence invalidation path, so that is separate follow-up scope.

services/cloud-agent-next/src/kilo-facade/user-kilo-facade.ts
```diff
@@ -0,0 +1,1634 @@
...
+    'KILO_SESSION_SNAPSHOT_PENDING',
+    'Cloud Agent Kilo session snapshot is not available yet'
+  );
+  response.headers.set('Retry-After', '1');
+  return response;
+}
+
+function unsupportedStructuredPathsResponse(entity: 'session' | 'messages'): Response {
+  return facadeError(
+    502,
```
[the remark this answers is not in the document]
eshurakov: Fixed locally. `Response | unknown` collapses back to `unknown` in TypeScript, so `readRequestJson` now returns a discriminated `{ success, value | response }` result and the caller must handle the error branch before prompt parsing. Added regressions for malformed JSON, declared oversized bodies, and streamed oversized bodies before balance or admission side effects.

services/session-ingest/src/dos/kilo-sdk-materialization.ts
```diff
@@ -0,0 +1,829 @@
...
+): number {
+  return left.identity.time - right.identity.time || compareId(left.identity.id, right.identity.id);
+}
+
+function compareId(left: string, right: string): number {
+  return left < right ? -1 : left > right ? 1 : 0;
+}
+
+function compareKiloSdkPart(left: Record<string, unknown>, right: Record<string, unknown>): number {
+  const idSchema = z.object({ id: z.string() });
```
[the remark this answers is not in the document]
eshurakov: Fixed in `87988bbc9`. Replaced the comparator-local Zod schema allocation with equivalent direct string narrowing, which preserves the fail-soft ordering behavior while avoiding both repeated schema construction and repeated `safeParse` work during sort comparisons.

services/cloud-agent-next/src/kilo-facade/user-kilo-facade.ts
```diff
@@ -0,0 +1,1575 @@
...
+        'Content-Type': 'text/event-stream',
+        'Cache-Control': 'no-cache, no-transform',
+        'X-Accel-Buffering': 'no',
+        'X-Content-Type-Options': 'nosniff',
+      },
+    });
+  }
+
+  async webSocketMessage(ws: WebSocket, message: string | ArrayBuffer): Promise<void> {
+    this.handleGlobalFeedMessage(ws, message);
```
jeanduplessis, line 1431: **[WARNING]:** Revalidate producer fence before forwarding each frame, or add explicit facade socket revocation for every runtime-fence invalidation path.

`webSocketMessage()` trusts socket attachment for full connection lifetime. Upgrade path validates `wrapperRunId`, `wrapperGeneration`, and `wrapperConnectionId`, but later frames call `handleGlobalFeedMessage()` without checking current `CloudAgentSession` runtime state.

Trace:
1. Producer socket is valid during HTTP upgrade.
2. Session runtime identity later clears or rotates during supersede, failure, interrupt, or keep-warm transitions.
3. Those transitions do not push revocation into `UserKiloFacade`.
4. Old socket can still submit a frame before replacement socket closes it, and handler broadcasts because root `sessionID` still matches attached source.

Impact: Public SSE consumers can receive stale duplicate or out-of-order assistant, idle, or session events from superseded runtime.

This matches earlier thread rationale: facade has no push revocation path after runtime-fence changes. Removing per-frame validation needs equivalent explicit revocation across every invalidation path.

services/cloud-agent-next/test/e2e/sdk-basic-chat.ts
```diff
@@ -0,0 +1,577 @@
...
+}
+
+function assertProjectedWakeEvent(
+  event: GlobalEvent['payload'],
+  sessionID: string,
+  action: string
+): void {
+  assert.equal(event.type, 'message.updated', `${action} did not carry a projected assistant`);
+  if (event.type !== 'message.updated') return;
+  assertProjectedAssistant(
```
jeanduplessis: **[SUGGESTION]:** Align local SDK E2E wake assertion with final nested-payload preservation contract.

`assertProjectedWakeEvent()` requires nested `message.updated.properties.info.path.cwd/root` to equal public virtual directory. Final facade intentionally rewrites only outer event `directory`; PR body says nested owner-visible payload content is preserved. Focused runtime coverage also expects nested `/workspace/private/...` assistant paths unchanged.

Trace:
1. Native wake frame reaches facade with private nested assistant path.
2. Facade preserves nested payload and virtualizes outer routing directory only.
3. This helper calls `assertProjectedAssistant()`, then later `assertSafeProjection()` rejects same valid payload.

Impact: Documented manual command `pnpm --filter cloud-agent-next exec tsx test/e2e/sdk-basic-chat.ts` can fail on valid wake events, creating false-negative local acceptance signal.

jeanduplessis (approved):
Assuming stale-producer fencing is addressed, overall design looks like a good starting point. Stable /kilo facade keeps sandbox and wrapper topology out of public SDK contract, while existing session DO remains lifecycle authority. That feels like right boundary for shipping quickly without locking us into current runtime implementation.

A few points may be worth considering before external consumers start depending on behavior:
- Event delivery semantics: current SSE feed appears best-effort, without replay after disconnect. That seems reasonable for UI updates if clients reconcile through session.get() and session.messages() after reconnect. If reliable automation feeds are expected, replay cursors and persisted events may be worth thinking through sooner since that would be more structural.
- Projection contract: outer routing directory is virtualized, while nested native payload fields may still include sandbox-local values. Preserving those fields is a reasonable v1 choice, but it may help to describe nested payload content as opaque so clients do not build dependencies on runtime topology.
- Cold-history contract: keeping cursors opaque and omission states explicit should preserve flexibility. It may be useful to avoid documenting storage ordering or cursor encoding as public behavior.

Some technical debt seems acceptable for first version, with a few areas worth watching as usage grows:
- UserKiloFacade is taking on several coordination responsibilities. That is fine initially, though stable policies such as event projection and producer fencing may become good extraction points later.
- User-scoped facade DO may become hot for users with many active sessions or subscribers. Metrics around subscriber count, event throughput, and slow-consumer eviction should make it clear whether internal sharding is ever needed.
- Warm and cold reads have separate implementations. Shared contracts and current tests are a solid base; a small conformance suite could help prevent behavior drift as SDK surface expands.

None of these look like reasons to hold the PR for a broader redesign. Main thing is keeping v1 transport details from becoming accidental public guarantees, so implementation remains easy to evolve after shipping.

services/cloud-agent-next/test/e2e/sdk-basic-chat.ts
eshurakov: Fixed locally. Wake-event E2E assertions now validate the correlated assistant event without applying HTTP transcript projection rules to nested SSE payload content. Removed the whole-event private-path rejection for scoped and global wake events; the outer global directory virtualization remains covered by the event predicate.

services/cloud-agent-next/src/kilo-facade/user-kilo-facade.ts
eshurakov: Keeping connect-time validation intentionally for v1. The residual gap is limited to best-effort SSE event integrity: an already-authenticated producer for the same owned root session can emit stale duplicate or out-of-order frames after its runtime fence changes and before replacement closes the socket. The facade attachment and root session checks still prevent cross-user or cross-session publication, and clients reconcile authoritative state through session reads. Per-frame DO RPC validation would add cost to every streamed event; explicit lifecycle revocation is broader follow-up work if reliable automation semantics become required.

PR #3674, opened 2026-06-02
fix(ai-gateway): use binary reasoning for mistral medium

chrarnoldus:
## Summary
- Return `REASONING_VARIANTS_BINARY` for `mistral-medium-3-5` models so they expose instant and thinking modes.
- Increase the shared binary `thinking` effort to `high` for models that use that variant set.

## Verification
- Not run (per request).

## Visual Changes
- N/A

## Reviewer Notes
- The `thinking` effort change applies to every model using `REASONING_VARIANTS_BINARY`.

### DIFF
--- a/apps/web/src/lib/ai-gateway/providers/model-settings.ts
+++ b/apps/web/src/lib/ai-gateway/providers/model-settings.ts
@@ -16,7 +16,7 @@ import { ReasoningEffortSchema } from '@kilocode/db/schema-types';
 
 export const REASONING_VARIANTS_BINARY = {
   instant: { reasoning: { enabled: false, effort: 'none' } },
-  thinking: { reasoning: { enabled: true, effort: 'medium' } },
+  thinking: { reasoning: { enabled: true, effort: 'high' } },
 } as const;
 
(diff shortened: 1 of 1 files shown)


apps/web/src/lib/ai-gateway/providers/model-settings.ts
```diff
@@ -16,7 +16,7 @@ import { ReasoningEffortSchema } from '@kilocode/db/schema-types';
 
 export const REASONING_VARIANTS_BINARY = {
   instant: { reasoning: { enabled: false, effort: 'none' } },
-  thinking: { reasoning: { enabled: true, effort: 'medium' } },
+  thinking: { reasoning: { enabled: true, effort: 'high' } },
```
chrarnoldus, line 19: this is just here so some value gets preserved in this path: https://github.com/Kilo-Org/kilocode/blob/5e7e158503ab6eab9a7de932b4dbc8b6bdbbf6b1/packages/opencode/src/kilocode/provider-options.ts#L25

mistral documents that you should set high, I doubt it matters for OpenRouter, but I think it's fine to change

marius-kilocode (approved):

PR #3676, opened 2026-06-02
feat(admin/orgs): trial-active filter, sortable entitlements, member-count exclusions, rounded dollar formatting

alexkgold:
image
image

## Summary

- Adds a third trial-tab filter (default ON) — **Trial active** — that filters to orgs whose \`free_trial_end_at\` is in the future. Backend gains a \`trial_ending_in_future\` input; URL/handler/reset/active-badge plumbing matches the existing \`Usage > 0\` and \`Users > 1\` filters.
- Makes every Entitlements-tab column sortable: Plan, Kilo Pass, Stripe Status, Subscription. Schema sortable-fields enum extended; sort expressions wired for derived columns (latest subscription, kilo_pass_tier subquery).
- \`member_count\` (and the \`Users > 1\` HAVING / \`member_count\` sort) now excludes \`role = 'billing_manager'\` memberships and \`is_bot = true\` users via inline JOIN filters and \`count(kilocode_users.id)\`. Both \`baseQuery\` and \`countBase\` get matching joins to keep totals consistent.
- Subscription, Usage, and Balance cells render rounded to the nearest dollar with comma grouping (\`$25,000\` instead of \`$25000.00\` / \`$1234.56\`).

## Verification

- [x] Walked \`/admin/organizations\` and \`/admin/organizations/trials\`. Confirmed Entitlements columns sort. Confirmed Trial active checkbox defaults ON, toggles persist, reset clears all three trial filters.
- [x] Confirmed \`Users / Seats\` count drops billing-manager and bot members (verified via DB after toggling membership rows).
- [x] Confirmed Subscription column shows e.g. \`$25,000\` for an enterprise plan and a small monthly Teams plan rounds to the nearest dollar.

## Visual Changes

| Before | After |
|---|---|
|  |  |

## Reviewer Notes

- The membership join's \`role != 'billing_manager'\` filter and the new \`kilocode_users\` LEFT JOIN with \`is_bot = false\` are inlined into the \`ON\` clauses so \`count(kilocode_users.id)\` naturally excludes the unwanted rows. Both the page query and the count query mirror these joins.
- Sort by \`subscription_amount_usd\` reuses the same \`CASE WHEN status IN ('active','trialing','past_due')\` expression that drives the displayed value, so sort order and visible value stay aligned (rows with non-billable statuses sort as NULL).
- \`subscriptionAmountUsd.toFixed(2)\` was replaced with \`Math.round(...).toLocaleString('en-US')\` (Subscription) and \`formatMicrodollars(..., 0)\` (Usage / Balance), both of which add comma grouping via \`Intl.NumberFormat\`.

### DIFF
--- a/apps/web/src/routers/organizations/organization-admin-router.ts
+++ b/apps/web/src/routers/organizations/organization-admin-router.ts
@@ -7,8 +7,23 @@ import {
   organization_seats_purchases,
   credit_transactions,
   platform_integrations,
+  kilo_pass_subscriptions,
 } from '@kilocode/db/schema';
-import { ilike, or, asc, desc, count, eq, ne, gt, and, isNull, sql, type SQL } from 'drizzle-orm';
+import {
+  ilike,
+  or,
+  asc,
+  desc,
+  count,
+  eq,
+  ne,
+  gt,
+  and,
+  isNull,
+  inArray,
+  sql,
+  type SQL,
+} from 'drizzle-orm';
 import type { PgColumn } from 'drizzle-orm/pg-core';
 import * as z from 'zod';
 import { OrganizationsApiGetResponseSchema } from '@/types/admin';
@@ -27,6 +42,10 @@ import { TRPCError } from '@trpc/server';
 import { successResult } from '@/lib/maybe-result';
 import { reportEvents } from '@/lib/ai-gateway/abuse-service';
 import { getMostRecentSeatPurchase } from '@/lib/organizations/organization-seats';
+import {
+  ORGANIZATION_TRIAL_ACTIVE_MIN_DAYS_REMAINING,
+  ORGANIZATION_TRIAL_DURATION_DAYS,
+} from '@kilocode/organization-entitlement';
 
 const OrganizationListInputSchema = z.object({
   page: z.number().int().min(1).default(1),
@@ -58,7 +77,7 @@ const OrganizationListInputSchema = z.object({
   // Trial-tab filters: hide orgs with no recorded usage / a single member.
   has_usage: z.boolean().default(false),
   has_multiple_users: z.boolean().default(false),
-  // When true, only orgs whose free_trial_end_at is non-null and in the future.
+  // When true, only orgs whose effective trial end keeps them trial_active.
   trial_ending_in_future: z.boolean().default(false),
 });
 
@@ -575,6 +594,7 @@ export const organizationAdminRouter = createTRPCRouter({
 
       const searchTerm = search.trim();
       const sortField = sortBy;
+      const sortsByKiloPassTier = sortField === 'kilo_pass_tier';
 
       const conditions = [];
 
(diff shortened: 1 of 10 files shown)


apps/web/src/routers/organizations/organization-admin-router.ts
```diff
@@ -593,21 +607,18 @@ export const organizationAdminRouter = createTRPCRouter({
...
-        orderCondition = orderFunction(
-          sql`${organizations.total_microdollars_acquired} - ${organizations.microdollars_used}`
-        );
-      } else {
-        orderCondition = orderFunction(organizations[sortField]);
+      // Trial-tab filter: only orgs whose trial end date is in the future.
+      // Orgs with a null free_trial_end_at are excluded — a trial without an
+      // end date does not "end in the future".
+      if (trial_ending_in_future) {
+        conditions.push(sql`${organizations.free_trial_end_at} > NOW()`);
```
evanjacobson: The new "Trial active" filter only checks `free_trial_end_at > NOW()`. Elsewhere in the app, orgs without `free_trial_end_at` still count as being in trial for 14 days after creation — `classifyOrganizationEntitlement` falls back to `created_at + 14 days` in `packages/organization-entitlement/src/trial.ts:10-14`. So this filter can hide active trial orgs that never had their `free_trial_end_at` set. Consider using the same effective trial end date as the rest of the app, e.g. `COALESCE(free_trial_end_at, created_at + interval '14 days') > NOW()`.

apps/web/src/routers/organizations/organization-admin-router.ts
```diff
@@ -621,6 +632,34 @@ export const organizationAdminRouter = createTRPCRouter({
...
+        orderCondition = orderFunction(
+          sql`CASE WHEN ${latestSubscriptions.subscription_status} IN ('active','trialing','past_due') THEN ${latestSubscriptions.amount_usd}::float8 ELSE NULL END`
+        );
+      } else if (sortField === 'latest_stripe_status') {
+        orderCondition = orderFunction(latestSubscriptions.subscription_status);
+      } else if (sortField === 'kilo_pass_tier') {
+        // Sorting by kilo_pass_tier mirrors the displayed-tier subquery so the
+        // order matches what the user sees in the column.
+        orderCondition = orderFunction(
+          sql`(SELECT kps.tier FROM organization_memberships om2 JOIN kilo_pass_subscriptions kps ON kps.kilo_user_id = om2.kilo_user_id WHERE om2.organization_id = ${organizations.id} AND kps.status = 'active' ORDER BY kps.tier LIMIT 1)`
```
evanjacobson: Sorting by Kilo Pass tier runs a correlated subquery per org before pagination. That's fine on a small dataset, but as the admin org list grows this could get expensive — the subquery joins `organization_memberships` + `kilo_pass_subscriptions`, orders, and takes the top row for every candidate before `LIMIT`/`OFFSET` kick in. Worth precomputing the displayed tier in a joined/lateral subquery and sorting by that instead.

apps/web/src/routers/organizations/organization-admin-router.ts
```diff
@@ -754,12 +811,26 @@ export const organizationAdminRouter = createTRPCRouter({
...
         .from(organizations)
         .leftJoin(
           organization_memberships,
-          eq(organizations.id, organization_memberships.organization_id)
+          and(
+            eq(organizations.id, organization_memberships.organization_id),
+            ne(organization_memberships.role, 'billing_manager')
+          )
+        )
+        .leftJoin(
```
evanjacobson: The total-count query now always joins through `organization_memberships` and `kilocode_users`, even when the `has_multiple_users` filter is off. Most admin list requests don't need that join, so this adds unnecessary database work. Could build the count query with those joins only when `has_multiple_users` is enabled.

evanjacobson (commented):
Left a few inline notes on the trial filter semantics and query performance. Nothing blocking, just things worth looking at before this hits production.

evanjacobson (approved):

PR #3679, opened 2026-06-02
feat(gastown): simplify town overview header cards

### LINKED ISSUE #3678: Reduce top-bar stat cards on the Gas Town overview to only useful information
The current top section of the Gas Town list page includes four stat cards: `Towns`, `Mode`, `Core`, and `Promise`.

Image

Image

---

- `Mode` is always shown as `Live`, which is static and reflects no actual runtime state.
- `Core` just lists `MEOW · GUPP · NDI`, which provides no immediately actionable information in that context and only confuses users.
- `Promise` is a brand copy line (`Discover, don't track`) and doesn't belong in a numeric/status card group.

These cards make the header feel cluttered without giving the user decision-making value. The only card with real data is `Towns`, and it should stay visible. The rest should be removed so the page reads more clearly.

Additionally, the loading state for the town count was inconsistent. While the rest of the page showed skeleton cards during load, the `Towns` count rendered as `...`, creating a jarring mismatch with the skeleton UI pattern used elsewhere on the page.

IamCoder18:
## Summary

Updated `apps/web/src/app/(app)/gastown/TownListPageClient.tsx` to clean up the top stat card row on the Gas Town overview. Removed the `Mode`, `Core`, and `Promise` cards, which provided misleading or non-actionable information. Kept only the `Towns` count, moved the **New Town** button to sit alongside it on the right, and improved the loading state for the count.

Closes #3678

## Verification

- Visually verified `/gastown` in the dev build and confirmed the header reads correctly with the single `Towns` card and aligned CTA.

### Before

before_1
before_2

### After

after_1
after_2

## Reviewer Notes

No functional or backend changes. This is a UI-only cleanup. The removed cards had no connected state or behavior.

### DIFF
--- a/apps/web/src/app/(app)/gastown/TownListPageClient.tsx
+++ b/apps/web/src/app/(app)/gastown/TownListPageClient.tsx
@@ -82,12 +82,12 @@ export function TownListPageClient() {
             transparency: every object is clickable; every outcome is attributable.
           </p>
 
-          <div className="flex items-end gap-3">
+          <div className="flex flex-wrap items-end gap-3">
             <div className="min-w-[120px] rounded-2xl border border-white/10 bg-white/[0.03] px-4 py-3">
               <div className="text-[11px] tracking-wider text-white/40 uppercase">Towns</div>
               <div className="mt-1 text-2xl font-semibold text-white/90">
                 {townsQuery.isLoading ? (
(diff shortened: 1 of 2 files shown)


eshurakov (approved):

eshurakov:
Mirrored into #3794 from a trusted repository branch so Vercel can build it. The replacement PR preserves the commit authorship from this PR head and points back here for original community attribution/review context.

PR #3698, opened 2026-06-03
fix(web): use uuid for client-generated ids

iscekic:
## Summary
- Replaces browser `crypto.randomUUID()` calls in Cloud Agent, App Builder, and shared upload client code with the existing `uuid` package.
- Keeps generated values as UUID v4 strings for attachment paths, upload item IDs, terminal tab IDs, and profile draft IDs while avoiding runtimes where `crypto.randomUUID` is unavailable.

## Verification
N/A

## Visual Changes
N/A

## Reviewer Notes
- Automated checks run locally: `pnpm format`, focused web upload tests, web typecheck, and `git diff --check`.
- Full monorepo typecheck was skipped in favor of the targeted web package check.


RSO (approved):

RSO:
Ideally we'd have some type of compatibility layer in the lint rules that automatically detects these 🤔

PR #3702, opened 2026-06-03
fix(session-ingest): delay queue retries safely

johnnyeric:
## Summary
- Add a 60s delay before retrying failed session-ingest queue messages to reduce immediate pressure on hot Durable Objects.
- Preserve session serving order under delayed retries by scanning persisted messages/parts by `ingested_at` with SQLite `id` as a tie-breaker.
- Keep metrics ordering aligned with ingest time rather than delayed processing/insertion order.

## Validation
- `pnpm --filter cloudflare-session-ingest exec vitest run src/queue-consumer.test.ts src/dos/SessionIngestDO.test.ts src/dos/session-ingest-history-budget.test.ts`
- `pnpm --filter cloudflare-session-ingest typecheck`
- `scripts/lint-all.sh`

### DIFF
--- a/services/session-ingest/src/dos/SessionIngestDO.ts
+++ b/services/session-ingest/src/dos/SessionIngestDO.ts
@@ -1,5 +1,5 @@
 import { DurableObject } from 'cloudflare:workers';
-import { eq, ne, gt, gte, lt, and, inArray, isNotNull } from 'drizzle-orm';
+import { eq, ne, gt, gte, lt, and, or, inArray, isNull, isNotNull } from 'drizzle-orm';
 import { drizzle, type DrizzleSqliteDODatabase } from 'drizzle-orm/durable-sqlite';
 import { migrate } from 'drizzle-orm/durable-sqlite/migrator';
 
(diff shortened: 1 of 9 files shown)


iscekic (commented):
One finding from review.

services/session-ingest/src/dos/SessionIngestDO.ts
```diff
@@ -295,25 +318,31 @@ export class SessionIngestDO extends DurableObject<Env> {
...
               .from(ingestItems)
-              .where(and(eq(ingestItems.item_type, 'message'), gt(ingestItems.id, msgCursor)))
-              .orderBy(ingestItems.id)
+              .where(
+                and(
+                  eq(ingestItems.item_type, 'message'),
+                  msgCursor ? afterIngestOrderCursor(msgCursor) : undefined
+                )
+              )
+              .orderBy(ingestItems.ingested_at, ingestItems.id)
```
iscekic, line 340: This switches the stream pagination to `ORDER BY ingested_at, id`, but the DO SQLite migrations still only index `item_id` and add `ingested_at` without an index. Since `getAllStream()` paginates in batches of 10, large sessions can now repeatedly scan/sort the same table instead of walking the PK as before. Please add a SQLite migration/schema index for the new ordering, at least `(ingested_at, id)`, and verify the message/part query plans before merging.

johnnyeric: Thanks, addressing it now.

johnnyeric: Addressed it now: added a new migration, generated the files and tested that the query plan was using the new index.
``` 
QUERY PLAN
`--SCAN ingest_items USING INDEX ingest_items_ingested_at_id_idx
QUERY PLAN
`--SEARCH ingest_items USING INDEX ingest_items_ingested_at_id_idx (ingested_at>?)
QUERY PLAN
|--SEARCH ingest_items USING INDEX ingest_items_item_id_unique (item_id>? AND item_id<?)
`--USE TEMP B-TREE FOR ORDER BY
QUERY PLAN
|--SEARCH ingest_items USING INDEX ingest_items_item_id_unique (item_id>? AND item_id<?)
`--USE TEMP B-TREE FOR ORDER BY
```

iscekic (approved):

PR #3703, opened 2026-06-03
feat(ai-gateway): expose Terminal Bench model stats

lambertjosh:
## Summary
- Enrich the existing OpenRouter model catalog with optional Terminal Bench 2.0 completion and per-attempt cost metadata.
- Publish active non-internal results with at least five attempts and a complete-attempt cost; stealth models remain eligible.
- Cache the compact benchmark map for five minutes and fail open to stale or empty data so stats lookup failures never block model selection.

## Verification
- Manually verified end to end (see https://github.com/Kilo-Org/kilocode/pull/10879)

## Visual Changes
N/A

## Reviewer Notes
- Prefer deploying this catalog enrichment before the client PR, although either order is safe because the new field is optional.
- Metrics attach only to canonical OpenRouter catalog entries, with the existing safe outer `kilo/` prefix fallback.


apps/web/src/lib/model-stats/terminal-bench.ts
```diff
@@ -0,0 +1,83 @@
...
+  isActive: boolean | null;
+  isStealth: boolean;
+  benchmarks: unknown;
+};
+
+export function summarizeTerminalBench(rows: readonly Row[]): TerminalBenchSummaries {
+  const summaries = new Map<string, TerminalBenchSummary>();
+
+  for (const row of rows) {
+    if (!row.isActive || row.isStealth) continue;
```
lambertjosh: we should not publish kilo-internal models (they should not be promoted anyway) but stealth are fine to include.

lambertjosh: Agreed. Updated the summary policy to include stealth models and explicitly exclude the `kilo-internal/` custom-LLM namespace at both the query and summary boundaries. The focused policy test now covers both cases.

imanolmzd-svg (approved):

PR #3711, opened 2026-06-03
fix(mobile): stabilize Android agent sessions

iscekic:
## Summary

Stabilizes Android Cloud Agent sessions and task UI behavior. The branch improves Android picker/task surfaces, reconnects Cloud Agent streams when the app resumes, clears stale disconnect state after reconnect, shows queued prompt activity before streaming starts, keeps the agent composer above the Android keyboard, and refreshes agent session queries on native resume/session creation.

## Verification

- Not manually verified on an Android device in this environment.

## Visual Changes

N/A

## Reviewer Notes

Focus review on Android lifecycle behavior, React Query native focus/online bridging, and the shared Cloud Agent disconnect-state cleanup.

### DIFF
--- a/apps/mobile/src/app/(app)/_layout.tsx
+++ b/apps/mobile/src/app/(app)/_layout.tsx
@@ -1,4 +1,6 @@
 import { Stack } from 'expo-router';
+import { Platform, StatusBar, useWindowDimensions } from 'react-native';
+import { useSafeAreaInsets } from 'react-native-safe-area-context';
 
 import { UserWebConnectionProvider } from '@/components/agents/user-web-connection-provider';
 import { KiloChatPresenceMount } from '@/components/kilo-chat/kilo-chat-presence-mount';
@@ -8,6 +10,12 @@ import { StoreKiloPassPurchaseProvider } from '@/lib/kilo-pass/use-store-kilo-pa
 
 export default function AppLayout() {
   const colors = useThemeColors();
+  const { height } = useWindowDimensions();
+  const { top } = useSafeAreaInsets();
+  const androidTopInset = top > 0 ? top : (StatusBar.currentHeight ?? 0);
+  const androidFullSheetDetent =
+    height > 0 ? Math.max(0.5, (height - androidTopInset) / height) : 1;
+  const fullSheetDetent = Platform.OS === 'android' ? androidFullSheetDetent : 1;
 
   return (
     <UserWebConnectionProvider>
@@ -29,7 +37,7 @@ export default function AppLayout() {
                 name="agent-chat/model-picker"
                 options={{
                   presentation: 'formSheet',
-                  sheetAllowedDetents: [0.5, 1],
+                  sheetAllowedDetents: [0.5, fullSheetDetent],
                   sheetGrabberVisible: true,
                   headerShown: false,
                 }}
@@ -38,7 +46,7 @@ export default function AppLayout() {
                 name="agent-chat/repo-picker"
                 options={{
                   presentation: 'formSheet',
-                  sheetAllowedDetents: [0.5, 1],
+                  sheetAllowedDetents: [0.5, fullSheetDetent],
                   sheetGrabberVisible: true,
                   headerShown: false,
                 }}
--- a/apps/mobile/src/lib/model-picker-rows.test.ts
+++ b/apps/mobile/src/lib/model-picker-rows.test.ts
@@ -29,8 +29,15 @@ describe('buildModelPickerRows', () => {
     ]);
   });
 
-  it('filters models by name and id', () => {
-    expect(buildModelPickerRows({ models, search: 'gpt-5' })).toEqual([
+  it('filters models by name', () => {
+    expect(buildModelPickerRows({ models, search: 'Sonnet 4' })).toEqual([
+      { key: 'recommended', title: 'RECOMMENDED', type: 'header' },
+      { key: 'model:anthropic/claude-sonnet-4', model: models[0], type: 'model' },
+    ]);
+  });
+
+  it('filters models by id', () => {
+    expect(buildModelPickerRows({ models, search: 'openai/' })).toEqual([
       { key: 'all', title: 'ALL MODELS', type: 'header' },
       { key: 'model:openai/gpt-5', model: models[1], type: 'model' },
     ]);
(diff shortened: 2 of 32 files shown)


apps/web/src/lib/cloud-agent-sdk/service-state.ts
```diff
@@ -379,6 +379,9 @@ function createServiceState(config: ServiceStateConfig): ServiceState {
 
     // Clear terminated on connected
     terminated = false;
+    if (status.type === 'disconnected') {
```
jeanduplessis: **WARNING:** Viewer WebSocket reconnect clears real wrapper-disconnected state

Trace:
1. `wrapper_disconnected` normalizes to `stopped(disconnected)`, storing `{ type: 'disconnected' }` and surfacing `Connection to agent lost`.
2. `/stream` emits synthetic `connected` on every viewer WebSocket attachment and explicitly omits authoritative wrapper `sessionStatus`.
3. This changed branch resets disconnected status to idle for that synthetic viewer event. `session-manager.ts` then clears error atom and indicator.

Impact: After wrapper dies, app resume or viewer-stream reconnect can make session appear healthy and re-enable composer while wrapper remains unavailable. Recovery needs authoritative wrapper state or separate transport-disconnect state.

apps/mobile/src/app/(app)/_layout.tsx
```diff
@@ -8,6 +10,12 @@ import { StoreKiloPassPurchaseProvider } from '@/lib/kilo-pass/use-store-kilo-pa
 
 export default function AppLayout() {
   const colors = useThemeColors();
+  const { height } = useWindowDimensions();
+  const { top } = useSafeAreaInsets();
+  const androidTopInset = top > 0 ? top : (StatusBar.currentHeight ?? 0);
+  const androidFullSheetDetent =
+    height > 0 ? Math.max(0.5, (height - androidTopInset) / height) : 1;
```
jeanduplessis, line 17: **WARNING:** Android full-sheet detent subtracts top inset twice

Trace:
1. This ratio replaces full detent `1` with `(height - androidTopInset) / height`.
2. Pinned `react-native-screens@4.23.0` defaults `sheetShouldOverflowTopInset` to `false`.
3. Native Android `SheetDelegate.tryResolveSafeAreaSpaceForSheet()` already subtracts top system-bar/display-cutout inset before applying detent ratio.

Impact: Android picker sheets with positive top inset stop below native safe-area maximum, leaving avoidable extra gap at full expansion. Existing full detent `1` already respects status bar.

services/cloud-agent-next/scripts/dev-with-docker-proxy.sh
```diff
@@ -20,6 +20,18 @@ socket="${DOCKER_PROXY_SOCKET:-${TMPDIR:-/tmp}/cloud-agent-dind-${hash}.sock}"
 socket="$(printf '%s' "$socket" | sed 's:/\{1,\}:/:g')"
 export DOCKER_PROXY_SOCKET="$socket"
 
+if [ -z "${MINIFLARE_CONTAINER_EGRESS_IMAGE:-}" ]; then
+  docker_arch="$(docker info --format '{{.Architecture}}' 2>/dev/null || true)"
```
jeanduplessis: **WARNING:** ARM architecture probe ignores supported `DOCKER_SOCKET` routing

Trace:
1. Probe invokes ambient `docker info` and converts failures into empty architecture through `|| true`.
2. Existing `docker-privileged-proxy.mjs` explicitly supports valid upstream override through `DOCKER_SOCKET`.
3. With valid ARM `DOCKER_SOCKET` but stale ambient Docker CLI context, proxy remains usable while this probe skips ARM image override.

Impact: Wrangler launches known-bad amd64 egress sidecar on ARM and local sandbox startup fails with `setsockoptint: protocol not available`. Probe architecture through same socket target or fail clearly.

apps/mobile/src/lib/model-picker-rows.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    expect(buildModelPickerRows({ models, search: '' })).toEqual([
+      { key: 'recommended', title: 'RECOMMENDED', type: 'header' },
+      { key: 'model:anthropic/claude-sonnet-4', model: models[0], type: 'model' },
+      { key: 'all', title: 'ALL MODELS', type: 'header' },
+      { key: 'model:openai/gpt-5', model: models[1], type: 'model' },
+    ]);
+  });
+
+  it('filters models by name and id', () => {
+    expect(buildModelPickerRows({ models, search: 'gpt-5' })).toEqual([
```
jeanduplessis: **SUGGESTION:** Filter test cannot independently verify name and ID search

Trace:
1. Helper supports `name.includes(query) || id.includes(query)`.
2. Test query `gpt-5` appears in both fixture name and ID.
3. Test remains green if either supported branch is deleted.

Impact: Future regression can silently break one picker search mode. Use separate non-overlapping name-only and ID-only queries.

apps/mobile/src/lib/query-client-lifecycle.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    const cleanup = installQueryClientNativeLifecycle({
+      sources: native.sources,
+      managers: {
+        focus: { setFocused: vi.fn() },
+        online: { setOnline },
+      },
+    });
+
+    native.setConnectivity({ isConnected: false, isInternetReachable: false });
```
jeanduplessis, line 90: **SUGGESTION:** Nullable NetInfo fallback path lacks coverage

Trace:
1. New production logic uses `isInternetReachable ?? isConnected ?? true`.
2. Test covers only aligned `false/false` and `true/true` pairs.
3. NetInfo startup and transitions can surface nullable reachability.

Impact: Future edit can break startup or recovery behavior while suite stays green. Add `false/null`, `true/null`, and `null/null` cases.

jeanduplessis (approved):

apps/mobile/src/lib/model-picker-rows.test.ts
iscekic: Fixed in c08b2a62d. The test now uses separate non-overlapping queries: a name-only search for "Sonnet 4" and an ID-only search for "openai/".

apps/mobile/src/lib/query-client-lifecycle.test.ts
iscekic: Fixed in 46ba60742. Added coverage for false/null, true/null, and null/null NetInfo states so the isInternetReachable ?? isConnected ?? true fallback is locked down.

eshurakov (approved):

PR #3714, opened 2026-06-03
feat(kiloclaw): surface official Google Calendar OAuth in settings, retire Gog on web

St0rmz1:
## Summary

Adds the official, Google-approved KiloCode OAuth client as the way to connect Google Calendar from KiloClaw settings, and begins retiring the legacy self-service "Gog" connection on web.

- **New Google Calendar card** (Settings → Productivity). Connects via KiloCode's official OAuth client (`kilo_owned`, read-only `calendar.readonly`). Connect is a GET redirect to `/api/integrations/google/connect`; disconnect is a same-origin form POST to `/api/integrations/google/disconnect`. Connected state, account email, and a "Reconnect" state are driven by the existing `googleOAuth*` status fields.
- **Legacy Google (Gog) card demoted.** The old "Google Account" card now renders **only** for users who already have it connected (`status.googleConnected`). It shows a retirement notice pointing them to the new card and to their auto-provisioned **Inbound Email** address (shown inline only when delivery is enabled) for email, plus the existing Disconnect and Gmail Notifications controls. The self-service setup-command flow is no longer surfaced, so **new Gog setups are no longer possible on web** while existing connections keep working untouched.
- **OAuth redirect feedback.** A one-shot effect (in the settings page wrapper, so it survives the no-instance bounce to onboarding) reads the `?success=`/`?error=` params the Google routes append on redirect, toasts a message, and strips the param. Known error codes get friendly copy; any other code (e.g. `access_denied`/consent cancelled) falls back to a generic message so failures are never silent.
- **Dead-code cleanup.** Removed the now-unused web `useClawGoogleSetupCommand` hook.
- **Changelog.** Added a `/claw/changelog` entry announcing the official OAuth client and the Gog deprecation.

### Scope / non-goals
- **Web/frontend only.** No DB, migration, or rollout changes.
- The backend `getGoogleSetupCommand` tRPC procedures are **intentionally retained**: the mobile app still uses them as its only Google connection path and has no OAuth flow yet. Retiring Gog on mobile is a separate future effort.

## Verification

- [x] Connected Google Calendar via the new card in a local dev instance — OAuth round-trip succeeded and the card showed the connected account; then disconnected/removed the config.
- [x] Reviewed the legacy Gog card copy and layout (retirement notice, inline Inbound Email address, Disconnect, Gmail Notifications) by temporarily forcing it to render, since the test account had no real Gog connection.

## Visual Changes

| Before | After |
|---|---|
| Settings → Productivity showed only the "Google Account" (Gog terminal-command) card. | New "Google Calendar" OAuth card; legacy "Google Account" card hidden for users without an existing Gog connection. |

## Reviewer Notes

- OAuth redirect feedback is handled in the settings page wrapper (`ClawSettingsPage`), above the no-instance gate, so `?error=missing_instance` still surfaces even when settings bounces the user to onboarding.
- The legacy card only presents the Inbound Email alias as usable when `inboundEmailEnabled` is true (the backend rejects disabled aliases with 410), so a disconnecting user isn't pointed at a dead delivery path.
- Disconnect is a native same-origin form POST so the route's `Origin` check passes; the 303 redirect lands back on settings with a status param. The submit is guarded on the form ref so a missing ref can't permanently disable the button.
- `returnTo` is validated server-side (`isSafeGoogleOAuthReturnTo`) and org IDs are URL-encoded; error copy comes from a static map (no reflected values), so no reflected-XSS surface.
- The retained `getGoogleSetupCommand` endpoints are unreachable from web but still serve mobile — see Scope note.

### DIFF
--- a/apps/web/src/app/(app)/claw/components/SettingsTab.tsx
+++ b/apps/web/src/app/(app)/claw/components/SettingsTab.tsx
@@ -340,13 +340,16 @@ const GOOGLE_CALENDAR_FEATURES: Array<{ included: boolean; label: string }> = [
 ];
 
 // Friendly messages for the `?error=` codes the Google OAuth connect/callback/
-// disconnect routes append when they redirect back to settings. Only these
-// known codes (plus the two success values) are handled, so unrelated query
-// params on the settings page are left untouched.
+// disconnect routes append when they redirect back to settings. Known codes get
+// tailored copy; any other error value (e.g. a sanitized provider description)
+// falls back to the generic message below so failures are never silent.
+const GOOGLE_OAUTH_GENERIC_ERROR = 'Could not connect Google Calendar. Please try again.';
+
 const GOOGLE_OAUTH_ERROR_MESSAGES: Record<string, string> = {
+  access_denied: 'Google Calendar connection was cancelled.',
   missing_permissions:
     'Calendar access was not granted. Please allow calendar permission and try again.',
-  connection_failed: 'Could not connect Google Calendar. Please try again.',
+  connection_failed: GOOGLE_OAUTH_GENERIC_ERROR,
   oauth_init_failed: 'Could not start the Google connection. Please try again.',
   missing_instance: 'Your KiloClaw instance is still starting. Try again in a moment.',
   missing_code: 'Google did not return an authorization code. Please try again.',
@@ -2119,7 +2122,13 @@ export function SettingsTab({
 
     const success = searchParams.get('success');
     const error = searchParams.get('error');
-    const errorMessage = error ? GOOGLE_OAUTH_ERROR_MESSAGES[error] : undefined;
+    // Any `error=` on this route comes from the Google OAuth routes, so map
+    // known codes to friendly copy and fall back to a generic message for the
+    // rest (e.g. access_denied/cancel, or a sanitized provider description) so
+    // the failure is always surfaced and the param is always cleaned up.
+    const errorMessage = error
+      ? (GOOGLE_OAUTH_ERROR_MESSAGES[error] ?? GOOGLE_OAUTH_GENERIC_ERROR)
+      : undefined;
     const isGoogleSuccess = success === 'google_connected' || success === 'google_disconnected';
 
     if (!isGoogleSuccess && !errorMessage) return;
--- a/apps/web/src/app/(app)/claw/components/SettingsTab.tsx
+++ b/apps/web/src/app/(app)/claw/components/SettingsTab.tsx
@@ -21,7 +21,6 @@ import { OpenclawImportCard } from './OpenclawImportCard';
 
 import { usePostHog } from 'posthog-js/react';
 import Link from 'next/link';
-import { usePathname, useRouter, useSearchParams } from 'next/navigation';
 import { toast } from 'sonner';
 import { useModelSelectorList } from '@/app/api/openrouter/hooks';
 import { ModelCombobox, type ModelOption } from '@/components/shared/ModelCombobox';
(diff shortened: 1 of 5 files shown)


apps/web/src/app/(app)/claw/components/SettingsTab.tsx
```diff
@@ -1931,6 +2125,40 @@ export function SettingsTab({
     previousStatusRef.current = status.status;
   }, [status.status]);
 
+  // Surface the one-shot success/error param the Google OAuth routes append
+  // when they redirect back to settings, then strip it so it doesn't re-fire.
+  useEffect(() => {
```
jeanduplessis: **WARNING:** Missing-instance OAuth feedback never mounts when instance is absent.

Trace:
1. This effect is responsible for reading `?error=missing_instance` and showing the new toast.
2. `ClawSettingsWithStatus` redirects to `/claw/new` or organization `/claw/new` whenever status is absent or `status.status === null`, and it does not mount `SettingsTab` in that state.
3. Google connect and callback routes redirect missing-instance failures to settings with `?error=missing_instance`.
4. In the no-instance race, settings redirects before this effect can run.

Impact: OAuth connect or callback race involving removed/not-yet-active instance silently bounces user into onboarding. Intended explanation never appears in personal or organization flow.

Please surface redirect feedback above the active-instance gate or preserve it when forwarding to onboarding.

apps/web/src/app/(app)/claw/components/SettingsTab.tsx
```diff
@@ -389,86 +581,84 @@ function GoogleAccountCard({
...
+                    {
+                      " integration listed above. This integration uses KiloCode's official OAuth application and is fully supported going forward."
+                    }
+                  </p>
+                  <p className="text-muted-foreground">
+                    {"For email functionality, use your bot's automatically provisioned "}
+                    <span className="text-foreground font-medium">Inbound Email</span>
+                    {' address below:'}
+                  </p>
+                  {inboundEmailAddress ? (
```
jeanduplessis: **WARNING:** Legacy Gog card can recommend a non-working inbound email alias.

Trace:
1. This branch renders migration guidance based only on `inboundEmailAddress`.
2. Status producers return `inboundEmailAddress` and `inboundEmailEnabled` independently. Existing Inbound Email card already checks both and renders `Unavailable` when disabled.
3. Backend inbound delivery rejects disabled instances with `410`, even when an active alias exists.
4. `GoogleAccountCard` receives no `inboundEmailEnabled` prop, so reachable admin-disabled state still shows exact alias and tells user to use it. Null fallback also says card is below, while actual card renders above and may show unavailable.

Impact: Existing Gog user can disconnect legacy integration expecting replacement email path, then receive failed inbound deliveries because displayed alias is disabled or unavailable.

Please pass `inboundEmailEnabled` into this card and render guidance consistent with actual delivery availability.

jeanduplessis (approved):
Found two edge-state gaps in new settings migration flow:

1. `missing_instance` OAuth redirect feedback is handled inside `SettingsTab`, but that component never mounts when instance is absent because settings redirects to onboarding first.
2. Legacy Gog retirement card recommends Inbound Email based only on alias presence, even though delivery can be disabled independently and backend rejects disabled aliases.

Also noted PR-body cleanup items: remove leftover Visual Changes template comment and keep automated checks out of manual-only Verification section.

apps/web/src/app/(app)/claw/components/SettingsTab.tsx
St0rmz1: Fixed in 5801c6e42.

Moved the OAuth redirect-param handling out of `SettingsTab` and up into `ClawSettingsWithStatus` (the settings page wrapper in `ClawSettingsPage.tsx`), which mounts on the settings route regardless of instance state — so it runs above the no-instance redirect gate. The toast now fires even in the no-instance race, and only strips the param when we're staying on settings (the onboarding redirect drops the query string anyway). The effect waits for status to load so `shouldRedirect` is meaningful before deciding whether to strip. The message map was extracted into a shared `google-oauth-feedback.ts` helper.

apps/web/src/app/(app)/claw/components/SettingsTab.tsx
St0rmz1: Fixed in 5801c6e42.

`GoogleAccountCard` now takes an `inboundEmailEnabled` prop (passed from `status.inboundEmailEnabled`) and only presents the alias as usable when `inboundEmailAddress && inboundEmailEnabled`. When the alias exists but delivery is disabled it tells the user to enable it first rather than showing a dead path; when there's no alias it points to the Inbound Email section without the incorrect "below" direction. This keeps the guidance consistent with the existing Inbound Email card's availability logic and the backend's 410 on disabled aliases.

St0rmz1:
Thanks for the review @jeanduplessis. Addressed both edge-state findings in 5801c6e42:

1. **Missing-instance feedback** — moved the OAuth redirect-param toast handling out of `SettingsTab` into the settings page wrapper (`ClawSettingsPage`), above the no-instance redirect gate, so `?error=missing_instance` still surfaces even when settings bounces to onboarding. Extracted the message map into a shared `google-oauth-feedback.ts`.
2. **Legacy inbound alias** — the legacy Gog card now takes `inboundEmailEnabled` and only presents the alias when delivery is actually enabled (matching the backend's 410 on disabled aliases); otherwise it directs the user to enable/set up Inbound Email without the incorrect directional copy.

Also did the PR-body cleanup you flagged (removed the leftover Visual Changes template comment and dropped automated checks from the manual Verification section). `format:check`, `lint`, and `typecheck` all pass.

PR #3717, opened 2026-06-03
feat(mcp-gateway): implement OAuth runtime gateway

pandemicsyn:
## Summary

- Implement PR2 of the MCP Gateway as a stacked change on `feat/mcp-gateway-scaffold`, adding the app-owned OAuth/control plane, shared gateway domain package, gateway Postgres schema/migration, and Worker-owned runtime proxy.
- Add dynamic client registration, authorization-code plus refresh-token flows, provider authorization and grant persistence, derived connect tokens, route/assignment lifecycle handling, encrypted secrets, per-instance DO refresh coordination, and strict runtime rechecks for scoped connect routes.
- Harden the runtime boundary with route-key rotation, owner/context checks, Origin validation, strict header filtering, non-public upstream rejection, no redirect credential forwarding, explicit bearer provider tokens, consent-gated authorization, and a single consolidated Kilo v1 gateway spec.
- Add local key-generation tooling, copy-ready Worker dev vars, fixture-driven app flow coverage, Worker route/proxy tests, soft-delete cleanup coverage, and IPv4/IPv6 public-address policy regressions.

## Verification

- [x] Started the local `mcp-gateway` Worker and confirmed it booted on `http://0.0.0.0:8806`.
- [x] Requested `GET /health` locally and confirmed a `200` response with `{"status":"ok","service":"mcp-gateway"}`.
- [x] Requested `GET /.well-known/oauth-protected-resource` locally and confirmed protected-resource metadata is returned.
- [x] Requested an unauthenticated scoped `/mcp-connect/...` route locally and confirmed a `401` OAuth challenge with `WWW-Authenticate`.

## Visual Changes

N/A

## Reviewer Notes

- This PR is intentionally stacked on `feat/mcp-gateway-scaffold`; the base should remain that branch until PR1 merges.
- The app owns first-level OAuth and provider callbacks, while the Worker owns protected-resource metadata, runtime proxying, fresh Postgres rechecks, credential injection, and per-instance refresh coordination.
- The consolidated `.specs/mcp-gateway-auth.md` is now the single in-repo contract for Kilo v1; the earlier baseline/profile split was removed.
- The Worker does best-effort dual-stack DoH validation for arbitrary third-party upstreams and explicitly does not follow redirects with injected credentials.


services/mcp-gateway/src/durable-objects/MCPGatewayInstance.do.ts
```diff
@@ -1,8 +1,58 @@
...
+
+  async refreshProviderGrant(input: unknown) {
+    const parsed = RefreshProviderGrantInputSchema.safeParse(input);
+    if (!parsed.success) {
+      return await runRefreshProviderGrant({ env: this.env, sqlite: this.sqlite, input });
+    }
+    const attemptKey = `${parsed.data.instanceId}:${parsed.data.grantId}:${parsed.data.expectedGrantVersion}`;
+    const existing = this.refreshInFlight.get(attemptKey);
+    if (existing) {
+      return await existing;
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in 23ae9b2a8.

apps/web/src/lib/mcp-gateway/lifecycle-service.ts
```diff
@@ -0,0 +1,173 @@
...
+  );
+  await database
+    .delete(mcp_gateway_pending_provider_authorizations)
+    .where(
+      and(
+        eq(mcp_gateway_pending_provider_authorizations.kilo_user_id, userId),
+        inArray(mcp_gateway_pending_provider_authorizations.config_id, configIds)
+      )
+    );
+}
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in 23ae9b2a8.

services/mcp-gateway/src/durable-objects/MCPGatewayInstance.do.ts
```diff
@@ -1,8 +1,58 @@
...
+
   constructor(state: DurableObjectState, env: Env) {
     super(state, env);
+    this.sqlite = drizzle(state.storage, { schema: { mcpGatewayInstanceState } });
+    void state.blockConcurrencyWhile(async () => {
+      const schemaVersion = await state.storage.get<number>('schema_version');
+      if (schemaVersion === undefined) {
+        state.storage.sql.exec(initialMigration);
+        await state.storage.put('schema_version', 1);
+        return;
```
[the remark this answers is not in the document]
pandemicsyn: The void state.blockConcurrencyWhile(...) feedback was not warranted. Cloudflare’s Durable Object API blocks event delivery as soon as blockConcurrencyWhile() is invoked in the constructor; void only marks the returned promise as intentionally unawaited in a synchronous constructor. This is also the existing repository pattern across other Durable Objects.

services/mcp-gateway/src/lib/upstream-proxy.ts
```diff
@@ -0,0 +1,108 @@
...
+  const init: RequestInit = {
+    method: params.request.method,
+    headers,
+    body:
+      params.request.method === 'GET' || params.request.method === 'HEAD'
+        ? undefined
+        : params.request.body,
+    redirect: 'manual',
+  };
+  const response = await fetchImpl(upstream.toString(), init);
```
St0rmz1, line 102: **Low, Security:** The upstream destination is validated and then fetched by hostname, so the validated IPs are never pinned. `validateRedirectTarget` (via `validateResolvedPublicUrl`) resolves the host over DoH and confirms every A/AAAA answer is public, but it returns the original URL and this `fetch` re-resolves the hostname independently. A host whose DNS flips from a public address at validation time to a private one (loopback, link-local, RFC1918) at fetch time bypasses the check. The spec already calls this out as best-effort for v1, and the blast radius is limited because the upstream URL comes from an authenticated config owner and Workers cannot reach cloud metadata endpoints. Flagging so it stays a tracked, conscious limitation. If the threat model later includes untrusted upstreams, pin the validated IP by connecting to the resolved address with the host preserved via SNI/Host header.

apps/web/src/lib/mcp-gateway/token-service.ts
```diff
@@ -0,0 +1,593 @@
...
+    }
+    const key = params.config.jwtKeyset.keys.find(candidate => candidate.keyId === kid);
+    if (!key) {
+      throw createGatewayError(GatewayErrorCode.InvalidGrant, 'Token key is unknown', 401);
+    }
+    const publicKeyPem = createPublicKey({ key: key.publicJwk, format: 'jwk' }).export({
+      format: 'pem',
+      type: 'spki',
+    });
+    const payload = jwt.verify(token, publicKeyPem, {
```
St0rmz1, line 130: **Low, Security:** `verifyAccessToken` checks signature and issuer but not the `aud` claim. Today it is only called by the `userinfo` endpoint, which returns the token subject's own profile and is gated on the `profile` scope, so a gateway token minted for a different connect resource and replayed here just returns the same user's profile. The risk is that this helper looks like a general-purpose verifier and could be reused on a route where audience actually matters, silently accepting tokens scoped to a different resource. Either pin the expected audience here or rename/document it as deliberately audience-agnostic so a future caller does not assume it enforces `aud`.

services/mcp-gateway/src/durable-objects/mcp-gateway-instance/refresh.ts
```diff
@@ -0,0 +1,293 @@
...
+}
+
+async function markRefreshFailure(
+  params: { env: MCPGatewayEnv['Bindings']; sqlite: MCPGatewayInstanceSQLite },
+  instanceId: string,
+  instanceKey: string,
+  grantId: string,
+  expectedGrantVersion: number
+) {
+  const db = getWorkerDb(params.env.HYPERDRIVE.connectionString, { statement_timeout: 5_000 });
```
St0rmz1: **Low, Performance:** `markRefreshFailure` opens a second `getWorkerDb` pool for the same logical operation that `refreshProviderGrant` already opened one for. On the refresh-failure path that is two separate Hyperdrive connections per request. The spec's per-request fresh-wrapper rule forbids caching a pool across requests, but it does not require a new pool within a single refresh. Pass the `db` instance from `refreshProviderGrant` into `markRefreshFailure` (as `recordRefreshAudit` already receives it) to reuse the connection.

apps/web/src/lib/mcp-gateway/token-service.ts
```diff
@@ -0,0 +1,593 @@
...
+    }
+    const updatedAt = new Date(user.updated_at);
+    return {
+      sub: user.id,
+      name: user.google_user_name,
+      preferred_username: user.google_user_email,
+      picture: user.google_user_image_url,
+      updated_at: Number.isNaN(updatedAt.getTime()) ? undefined : updatedAt.toISOString(),
+      email: user.google_user_email,
+      email_verified: true,
```
St0rmz1: **Nit, Security:** `email_verified` is emitted as a constant `true` rather than reflecting the user's actual verification state. A relying party that trusts this claim would treat every returned address as verified. Confirm that is intentional for the `profile`-scoped userinfo response; if email verification is not actually guaranteed for all users, either derive this from real state or omit the claim.

St0rmz1 (approved):

services/mcp-gateway/src/durable-objects/mcp-gateway-instance/refresh.ts
pandemicsyn: Fixed in fb224d5fa.

apps/web/src/lib/mcp-gateway/token-service.ts
pandemicsyn: Fixed in fb224d5fa.

PR #3722, opened 2026-06-04
feat(mcp-gateway): add management dashboard

pandemicsyn:
## Summary

- Add an admin-gated MCP Gateway management dashboard for personal and organization scopes, including connection discovery, setup, connect URL management, assignments, credentials, and provider sign-in controls.
- Introduce an app-owned dashboard tRPC/control-plane surface while preserving the two-plane architecture: `apps/web` manages configuration and OAuth workflows, while `services/mcp-gateway` remains responsible for runtime token verification and credential-injecting proxy behavior.
- Harden dashboard lifecycle boundaries: scope-bind management mutations, persist initial static provider credentials atomically, normalize DB timestamps at the tRPC boundary, reject OAuth client authentication-method changes without a secret lifecycle, require confirmations for destructive actions, and keep the dashboard rollout gated to admins for personal connections.
- Move resource-specific OAuth registration under `/api/mcp-gateway/oauth/register/resource/...` to avoid Next.js dynamic route conflicts while retaining the required registration capability.
- Separate Kilo gateway scopes from upstream provider scopes. This avoids incorrectly sending Kilo's `profile` scope to remote providers, preserves remote protected-resource `resource` values, stores upstream scope provenance on the config, and lets both dashboard sign-in and Codex-triggered sign-in use the same resolved provider OAuth configuration.
- Preserve discovered provider scope provenance by only treating user-edited scope values as overrides, and polish the dashboard's accessibility and interaction details around auth hints, checkbox contrast, internal navigation, and responsive status display.

## Verification

- Manual dashboard demo and OAuth flow verification were shared in Slack.

## Visual Changes

| Before | After |
|---|---|
| MCP Gateway dashboard unavailable (screenshot pending) | Personal connection list and setup flow (screenshot pending) |
| Organization MCP Gateway management unavailable (screenshot pending) | Organization detail, assignment, and credential management surface (screenshot pending) |

## Reviewer Notes

- This PR now targets `main`; the underlying gateway implementation and production hostname work have already landed separately.
- The upstream OAuth scope change adds `provider_scopes`, `provider_scope_source`, and `provider_resource` to the config model because remote provider scopes are not the same thing as Kilo gateway scopes.
- Focus areas: OAuth/provider sign-in setup behavior, tenant-scoped management mutations, destructive action lifecycle semantics, and provider scope/resource resolution.


apps/web/src/app/(app)/cloud/mcp-gateway/page.tsx
```diff
@@ -0,0 +1,14 @@
+import { getUserFromAuth } from '@/lib/user/server';
+import { notFound } from 'next/navigation';
+import { PageContainer } from '@/components/layouts/PageContainer';
+import { McpGatewayListContent } from './McpGatewayListContent';
+
+export default async function McpGatewayPage() {
+  const { user } = await getUserFromAuth({ adminOnly: false });
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in 5e0479f95.

apps/web/src/routers/mcp-gateway-router.ts
```diff
@@ -0,0 +1,647 @@
...
+      organizationId: params.organizationId,
+      userId: params.userId,
+      isGlobalAdmin: params.isGlobalAdmin,
+    });
+  }
+  return resolved;
+}
+
+export const mcpGatewayRouter = createTRPCRouter({
+  discover: mcpGatewayProcedure
```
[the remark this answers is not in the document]
pandemicsyn: gonna intentionally punt on this, because if this is actually useful wanna start caching endpoint responses in kv anyway.

apps/web/src/lib/mcp-gateway/provider-oauth-service.ts
```diff
@@ -249,20 +281,54 @@ export function createProviderOAuthService(params: {
       credentials.metadata.authorization_endpoint
     );
     const tokenEndpoint = await validatePublicHttpsDestination(credentials.metadata.token_endpoint);
+    let providerScopes = resolved.config.provider_scopes;
+    let providerResource = resolved.config.provider_resource
+      ? await validatePublicHttpsDestination(resolved.config.provider_resource)
+      : null;
+    if (providerScopes === null || providerResource === null) {
```
jeanduplessis: **Warning:** `null` provider scopes/resource trigger mutable live rediscovery on every sign-in.

**Trace:** Persisted `null` is also valid final state when provider advertises no scope/resource or admin clears override. This branch nevertheless performs fresh discovery and uses returned values without persisting them, advancing config version, cancelling pending state, revoking grants, or recording config change. Governing spec classifies provider scope/resource changes as material.

**Impact:** Temporary MCP endpoint failure blocks otherwise valid OAuth sign-in. Remote metadata changes can silently alter requested scope/resource without lifecycle protections or auditability.

Distinguish unknown legacy value requiring hydration from authoritative absence, then persist rediscovered material changes through versioned config-update path.

apps/web/src/lib/mcp-gateway/config-service.ts
```diff
@@ -272,6 +453,59 @@ export function createConfigService(params: {
...
+          GatewayErrorCode.InvalidRequest,
+          'Provider scopes require OAuth provider sign-in',
+          400
+        );
+      }
+      const currentScopes = config.provider_scopes ?? null;
+      const sameScopes =
+        currentScopes === null && input.providerScopes === null
+          ? true
+          : currentScopes?.length === input.providerScopes?.length &&
```
jeanduplessis: **Warning:** Scope equality is order-sensitive even though OAuth scope sets are not.

**Trace:** Dashboard accepts whitespace-separated scopes, but `sameScopes` compares each item at same array index. Saving same scopes in different order therefore calls `revokeConfigGrants`, consumes pending sign-ins, and increments config version.

**Impact:** Harmless reordering revokes active provider grants and forces affected users through reauthentication.

Normalize/dedupe scope values before persistence and compare canonical sets rather than input order.

apps/web/src/routers/mcp-gateway-router.ts
```diff
@@ -0,0 +1,717 @@
...
+    .mutation(async ({ input, ctx }) =>
+      withGatewayErrorMapping(async () => {
+        await requireManagedConfig({
+          configId: input.configId,
+          organizationId: input.organizationId,
+          userId: ctx.user.id,
+          isGlobalAdmin: ctx.user.is_admin,
+        });
+        const services = createGatewayServices();
+        const assignment = await services.configService.revokeAssignment({
```
jeanduplessis, line 708: **Warning:** Assignment-removal audit receives no current actor.

**Trace:** Mutation has `ctx.user.id` but passes only config and target user IDs. Service consequently records historical `assignment.assigned_by_kilo_user_id` as removal actor. When another owner/admin revokes assignment, audit identifies original assigner instead of actual caller.

**Impact:** Durable audit trail becomes incorrect, weakening investigation and accountability.

Pass `actorUserId: ctx.user.id` into `revokeAssignment` and use it for `assignment_removed` audit event.

packages/mcp-gateway/src/headers.ts
```diff
@@ -45,6 +47,15 @@ export function isCredentialLikeHeader(name: string): boolean {
   );
 }
 
+function isTransportIdentityHeader(name: string): boolean {
+  const normalized = name.toLowerCase();
+  return (
+    transportIdentityHeaders.has(normalized) ||
+    normalized.startsWith('x-forwarded-') ||
+    normalized.startsWith('cf-')
```
jeanduplessis: **Warning:** Blanket `cf-*` classification rejects valid Cloudflare Access static credentials.

**Trace:** Prefix catches transport metadata such as `CF-Connecting-IP`, but also catches `CF-Access-Client-Id` and `CF-Access-Client-Secret`. Static-header mode intentionally supports configured credential headers, and `parseStaticHeaders` now rejects these before creation/loading.

**Impact:** MCP servers protected by Cloudflare Access cannot use supported static-header authentication.

Block explicit transport-derived Cloudflare headers instead of entire `cf-*` namespace, while retaining credential-confusion protections for auxiliary headers.

apps/web/src/app/(app)/cloud/mcp-gateway/McpGatewayDetailContent.tsx
```diff
@@ -0,0 +1,706 @@
...
+                            </div>
+                            <div className="text-muted-foreground truncate text-xs">
+                              {member?.name ? member.email : assignment.userId}
+                            </div>
+                          </div>
+                          <Button
+                            variant="ghost"
+                            size="sm"
+                            className="text-muted-foreground hover:text-foreground sm:-mr-2"
+                            onClick={() =>
```
jeanduplessis: **Warning:** Assignment revocation executes immediately without confirmation.

**Trace:** Click directly invokes mutation. Service revokes assignment, removes active/needs-reauth instances, and revokes active grants. Route rotation, disable, and deletion on same page all use confirmation dialogs.

**Impact:** Accidental click immediately removes member access and provider state; recovery requires reassignment and provider reauthentication.

Wrap action in destructive confirmation describing instance/grant revocation.

apps/web/src/app/(app)/components/OrganizationAppSidebar.tsx
```diff
@@ -232,6 +232,15 @@ export default function OrganizationAppSidebar({
           },
         ]
       : []),
+    ...(user?.is_admin
```
jeanduplessis, line 235: **Suggestion:** Organization-owner navigation does not match route authorization.

**Trace:** Sidebar entry requires global `user.is_admin`, while organization pages and tRPC procedures permit non-global organization owners. Owners can manage feature through direct URL but receive no navigation entry.

**Impact:** Intended organization managers cannot discover or routinely navigate to dashboard.

Show entry for effective organization owner role as well as global admins.

jeanduplessis (approved):
Reviewed final combined diff, PR body, MCP Gateway auth spec, current caller/callee context, and unresolved GitHub review state.

Submitted 5 new warnings and 1 suggestion inline. Existing discovery authorization/rate-limiting comment already covers the remaining warning and was not duplicated.

Main issues: null provider metadata causes unversioned live rediscovery; reordered scope sets revoke grants; assignment-removal audit records the wrong actor; blanket `cf-*` rejection blocks Cloudflare Access credentials; assignment revocation lacks confirmation; organization owners lack dashboard navigation.

No changed PostHog events were lost, duplicated, or incorrectly gated. Tests cover major authorization, timestamp, OAuth, and credential paths, but not these edge cases. PR body still contains pending verification and screenshot placeholders.

apps/web/src/lib/mcp-gateway/provider-oauth-service.ts
pandemicsyn: Fixed in 4b5f95e78.

apps/web/src/lib/mcp-gateway/config-service.ts
pandemicsyn: Fixed in 4b5f95e78.

apps/web/src/routers/mcp-gateway-router.ts
pandemicsyn: Fixed in 4b5f95e78.

packages/mcp-gateway/src/headers.ts
pandemicsyn: Fixed in 4b5f95e78.

apps/web/src/app/(app)/cloud/mcp-gateway/McpGatewayDetailContent.tsx
pandemicsyn: Fixed in 4b5f95e78.

apps/web/src/app/(app)/cloud/mcp-gateway/ConnectToKiloDialog.tsx
```diff
@@ -0,0 +1,171 @@
...
+                    {connectUrl}
+                  </code>
+                  <CopyButton
+                    value={connectUrl}
+                    ariaLabel="Copy connect URL"
+                    toastLabel="Connect URL copied"
+                  />
+                </div>
+              </Step>
+              <Step number={5}>
```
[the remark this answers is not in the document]
pandemicsyn: authMode is the upstream provider credential mode, not the Kilo client-to-gateway auth mode.
Every MCP Gateway connect URL is still protected by Kilo OAuth, regardless of whether the upstream MCP server uses:
- none
- static_headers
- oauth_dynamic
- oauth_static
So for every Connect URL, Kilo should answer:
- OAuth required: Yes
- Pre-registered client ID: No
- oauth: {} in kilo.json

PR #3737, opened 2026-06-04
fix(kiloclaw): retry Fly secret sync failures

RSO:
## Summary
- Retry transient Fly app secret writes so `setAppSecret` 520/5xx/408/429 responses do not immediately fail KiloClaw provisioning.
- Keep AppDO env-key setup marked pending until Fly confirms `KILOCLAW_ENV_KEY` was synced, allowing the retry alarm to self-heal failed secret propagation.
- Add regression coverage for Fly secret retry and AppDO retry-alarm state after secret sync failure.

## Verification
N/A - backend Worker provisioning change; no manual UI verification performed.

## Visual Changes
N/A

## Reviewer Notes
- `setAppSecret` is retried only for transient statuses and remains single-attempt for non-retryable 4xx responses.
- Existing unrelated worktree changes were stashed separately before creating this PR branch.

### DIFF
--- a/services/kiloclaw/src/fly/secrets.ts
+++ b/services/kiloclaw/src/fly/secrets.ts
@@ -21,6 +21,8 @@ type AppSecret = {
   updated_at?: string;
 };
 
+const SET_SECRET_RETRY_DELAYS_MS = [250, 1000];
+
 async function secretsFetch(
   config: FlySecretsConfig,
   path: string,
@@ -44,6 +46,14 @@ async function assertOk(resp: Response, context: string): Promise<void> {
   }
 }
 
+function isRetryableSecretStatus(status: number): boolean {
+  return status === 408 || status === 429 || status >= 500;
+}
+
+function sleep(ms: number): Promise<void> {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+
 /**
  * Set a single app secret. Returns the secrets version, which should be
  * passed as `min_secrets_version` to createMachine/updateMachine to ensure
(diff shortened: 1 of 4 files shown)


services/kiloclaw/src/fly/secrets.ts
```diff
@@ -56,13 +66,29 @@ export async function setAppSecret(
...
-  const resp = await secretsFetch(config, `/secrets/${encodeURIComponent(name)}`, {
-    method: 'POST',
-    body: JSON.stringify({ value }),
-  });
-  await assertOk(resp, 'setAppSecret');
-  const data: { version?: number } = await resp.json();
-  return { version: data.version ?? 0 };
+  const path = `/secrets/${encodeURIComponent(name)}`;
+
+  for (let attempt = 0; attempt <= SET_SECRET_RETRY_DELAYS_MS.length; attempt++) {
```
St0rmz1, line 71: **Low, Functional:** The retry loop only handles returned responses with a retryable status. If `fetch` itself rejects (connection reset, DNS failure, TLS error during a Fly outage), the rejection escapes on the first attempt with no retry. Those connection-level blips are arguably the most transient failure of all, and the targeted 5xx/520 cases are the ones least likely to recover instantly.

The outer retry alarm still self-heals these, so impact is limited. If you want symmetric coverage, wrap the `secretsFetch` call in a try/catch inside the loop and treat a thrown network error like a retryable status: sleep and continue, throw on the last attempt.

services/kiloclaw/src/fly/secrets.ts
```diff
@@ -44,6 +46,14 @@ async function assertOk(resp: Response, context: string): Promise<void> {
   }
 }
 
+function isRetryableSecretStatus(status: number): boolean {
+  return status === 408 || status === 429 || status >= 500;
```
St0rmz1: **Nit, Design:** `status >= 500` treats every 5xx as retryable, including non-transient codes like 501 and 505. Negligible in practice (Fly does not return these here, and the cost is at most two wasted retries), but if you want precision, an explicit allow list reads clearer: `status === 408 || status === 429 || [500, 502, 503, 504].includes(status)`.

RSO:
@kilocode-bot please address the review comments on this PR. If you want to make a change, make a separate commit for each comment you want to address. Once you've pushed the commits, address/resolve the comments

St0rmz1 (approved):

PR #3739, opened 2026-06-04
feat(kiloclaw): lazy-load file tree directories

RSO:
## Summary
- Add path-scoped, shallow KiloClaw controller file-tree listing and advertise the `files.tree.path` capability.
- Thread optional file-tree paths through Worker, platform, tRPC, and personal/org/admin file editors so folders lazy-load on expansion.
- Fix the KiloClaw dev image push script to pass the monorepo `workspace` build context required by plugin builder stages.

## Verification
- Local debug confirmed stale docker-local instances surfaced the recursive-tree guard before the local image was rebuilt.
- Rebuilt `kiloclaw:local` for docker-local follow-up validation; the running instance was not recreated before handoff.

## Visual Changes

https://github.com/user-attachments/assets/c8fed5a4-1570-4f61-b105-02708dccebbc

## Reviewer Notes
- Existing KiloClaw instances need a restart or recreation with the updated controller image before path-scoped tree loading works.
- The UI strips nested directory children defensively so stale recursive-controller responses produce an actionable error instead of suppressing lazy loads.

### DIFF
--- a/apps/web/src/app/(app)/claw/components/FileEditorShell.tsx
+++ b/apps/web/src/app/(app)/claw/components/FileEditorShell.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useState, useCallback, useRef, type ReactNode } from 'react';
+import { useState, useCallback, useEffect, useRef, type ReactNode } from 'react';
 import { Loader2, RefreshCw } from 'lucide-react';
 import { Button } from '@/components/ui/button';
 import { Alert, AlertDescription } from '@/components/ui/alert';
@@ -18,11 +18,46 @@ import { useResizableSidebar } from '@/hooks/useResizableSidebar';
 import type { FileNode } from '@/lib/kiloclaw/kiloclaw-internal-client';
 import { FileTree } from './FileTree';
 
+function replaceDirectoryChildren(
+  nodes: FileNode[],
+  path: string,
+  children: FileNode[]
+): FileNode[] {
+  return nodes.map(node => {
+    if (node.path === path && node.type === 'directory') {
+      return { ...node, children };
+    }
+    if (node.children) {
+      return { ...node, children: replaceDirectoryChildren(node.children, path, children) };
+    }
+    return node;
+  });
+}
+
+function stripNestedChildren(nodes: FileNode[]): FileNode[] {
+  return nodes.map(node => {
+    if (node.type === 'directory') {
+      return { name: node.name, path: node.path, type: node.type };
+    }
+    return node;
+  });
+}
+
+function validatePathScopedChildren(path: string, children: FileNode[]): void {
+  const prefix = `${path}/`;
+  if (children.every(child => child.path.startsWith(prefix))) return;
+
+  throw new Error(
+    'Controller returned a recursive file tree. Restart this instance with the latest image.'
+  );
+}
+
 export function FileEditorShell({
   tree,
   isLoading,
   error,
   refetch,
+  loadChildren,
   renderPane,
   onClose,
   height,
@@ -31,6 +66,7 @@ export function FileEditorShell({
   isLoading: boolean;
   error: { message: string } | null;
   refetch: () => void;
+  loadChildren?: (path: string) => Promise<FileNode[]>;
   renderPane: (selectedPath: string, onDirtyChange: (dirty: boolean) => void) => ReactNode;
   onClose?: () => void;
   height?: string;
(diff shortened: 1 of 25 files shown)


apps/web/src/app/(app)/claw/components/FileEditorShell.tsx
```diff
@@ -39,9 +75,20 @@ export function FileEditorShell({
...
     { type: 'switch'; path: string } | { type: 'close' } | null
   >(null);
+  const [mergedTree, setMergedTree] = useState<FileNode[] | undefined>(tree);
+  const [loadedPaths, setLoadedPaths] = useState<Set<string>>(new Set());
+  const [loadingPaths, setLoadingPaths] = useState<Set<string>>(new Set());
+  const [loadErrors, setLoadErrors] = useState<Map<string, string>>(new Map());
   const hasUnsavedChangesRef = useRef(false);
   const { width: sidebarWidth, startDrag } = useResizableSidebar();
 
+  useEffect(() => {
```
St0rmz1, line 109: **Moderate, Functional:** This effect clears `loadedPaths` and rebuilds `mergedTree` shallow whenever the `tree` prop changes, but `FileTree`'s `expanded` set is internal state that survives across the change. After any refetch (the manual refresh button, and the admin write mutation which invalidates the root `fileTree` query on save), a directory the user had expanded is still marked expanded but now has `children === undefined` and is no longer in `loadedPaths`. Since loads only fire on a user toggle, the folder renders empty and never reloads until the user collapses and re-expands it.

With the previous recursive tree this did not happen, because a refetch repopulated `node.children` for the whole tree. So this is a regression: the common edit-then-save flow silently empties every open folder in the sidebar.

Suggestion: preserve loaded subtrees across `tree` changes whose paths still exist (merge the new shallow root into the existing `mergedTree` rather than discarding loaded children), or lift `expanded` into this component alongside `loadedPaths` and re-trigger loads for still-expanded paths after a refetch.

apps/web/src/routers/kiloclaw-router.ts
```diff
@@ -4179,16 +4179,21 @@ export const kiloclawRouter = createTRPCRouter({
...
-      const instance = await getActiveInstance(ctx.user.id);
-      const client = new KiloClawInternalClient();
-      const result = await client.getFileTree(ctx.user.id, workerInstanceId(instance));
-      return result.tree;
-    } catch (err) {
-      handleFileOperationError(err, 'fetch file tree');
-    }
-  }),
+  fileTree: clawAccessProcedure
+    .input(z.object({ path: z.string().min(1).optional() }).optional())
```
St0rmz1: **Low, Security:** The new `path` input is unbounded (`z.string().min(1).optional()`). The controller's `resolveSafePath` rejects unsafe values, so this is not a traversal vector, but the string is unbounded before it reaches that layer and gets URL-encoded across two hops (web client to platform to gateway to controller). Per the repo's input-validation convention, every user-supplied zod `.string()` should carry a `.max(N)`. The same applies to the matching inputs in `organization-kiloclaw-router.ts` and `admin-kiloclaw-instances-router.ts`. A shared constant would also let you bound the existing `readFile` path, which is currently unbounded too.

apps/web/src/app/admin/components/KiloclawInstances/AdminFileEditor.tsx
```diff
@@ -112,7 +112,7 @@ export function AdminFileEditor({
       onSuccess: async result => {
         if ('outcome' in result) return;
         await queryClient.invalidateQueries({
-          queryKey: trpc.admin.kiloclawInstances.fileTree.queryKey({ userId, instanceId }),
+          queryKey: trpc.admin.kiloclawInstances.fileTree.queryKey(),
```
St0rmz1: **Low, Design:** Changing the invalidation key to `fileTree.queryKey()` with no args now matches every cached `fileTree` query across all users and instances, not just the one being edited. The widening was needed so path-scoped subqueries also invalidate, but it is broader than intended, and the asymmetry with the `readFile` invalidation just below (still scoped to `{ userId, instanceId }`) is a readability snag. If react-query partial matching allows it, scope to `{ userId, instanceId }` so only the current instance's root and path-scoped trees invalidate.

apps/web/src/routers/organizations/organization-kiloclaw-router.ts
```diff
@@ -1461,16 +1461,21 @@ export const organizationKiloclawRouter = createTRPCRouter({
...
-      const instance = await getActiveOrgInstance(ctx.user.id, input.organizationId);
-      const client = new KiloClawInternalClient();
-      const result = await client.getFileTree(ctx.user.id, workerInstanceId(instance));
-      return result.tree;
-    } catch (err) {
-      handleFileOperationError(err, 'fetch file tree');
-    }
-  }),
+  fileTree: organizationMemberProcedure
+    .input(z.object({ organizationId: z.uuid(), path: z.string().min(1).optional() }))
```
St0rmz1: **Low, Testing:** The controller route, capability advertisement, and gateway encoding all got new tests, which is good. The three tRPC procedures that now accept and forward `path` (personal, org, admin) have none. They are thin pass-throughs, but they are the auth boundary and the point where `path` is wired into `getFileTree({ instanceId, path })`. A regression that drops or mis-routes `path` here would not be caught. Suggestion: add a caller-level test per namespace asserting `path` is forwarded to the internal client, mirroring the existing fileTree procedure tests.

chrarnoldus (approved):

PR #3746, opened 2026-06-04
Make discounted MiniMax M3 a separate model id

chrarnoldus:
Overwriting the original model id causes the model to be unavailable for BYOK: https://kilo-code.slack.com/archives/C0AN54T2P71/p1780602381397329


evanjacobson (approved):

chrarnoldus:
https://app.kilo.ai/byok shows which models are supported for BYOK, M3 should show up here after everything is merged/deployed/synced.

CleanShot 2026-06-04 at 22 01 27@2x

PR #3747, opened 2026-06-04
fix(session-ingest): slim oversized item RPC payloads

johnnyeric:
## Summary

After #3744 batched multiple session-ingest items into one Durable Object RPC, the post-deploy log report showed a concrete retry category for oversized RPC payloads: **RPC >32MiB: 593 retries**. Example error:

`Serialized RPC arguments or return values are limited to 32MiB`

The queue consumer already offloads item data above the DO SQLite row limit to R2 and passes an `r2References` map into `SessionIngestDO.ingest()`. However, it still sent the full parsed item object in the RPC payload. For very large message/part payloads, that meant we paid both costs: the full data was written to R2, but it was also serialized into the DO RPC, where Cloudflare enforces the 32MiB RPC argument/return limit.

This changes the queue consumer so oversized items are still written to R2 unchanged, but the item sent over RPC is slimmed to only the fields the DO needs to identify and route the item. The DO already persists referenced items as an empty inline blob plus `item_data_r2_key`, so the full `item.data` is redundant in the RPC once an R2 reference exists. The chunk byte counter also tracks the slim RPC payload size for offloaded items, rather than the full pre-slim data size, so multiple oversized/R2-backed items do not force unnecessarily tiny chunks.

Specifically:

- `message` keeps only `data.id`
- `part` keeps only `data.id` and `data.messageID`
- `session` keeps metadata fields used by extractors (`title`, `parentID`)
- `session_diff` / `model` keep valid empty payload shapes because full data is available through R2
- metadata/lifecycle-only item types are left unchanged

The existing R2 key generation, DO item identity, queue retry behavior, and R2-backed export/materialization path are unchanged.

Test coverage updated in `services/session-ingest/src/queue-consumer.test.ts`:

- full oversized item data is still written to R2
- the DO RPC receives only the slim item payload
- the DO RPC still receives the correct R2 reference

## Verification

Local automated checks run before updating the PR:

- `pnpm exec oxfmt services/session-ingest/src/queue-consumer.ts services/session-ingest/src/queue-consumer.test.ts`
- `pnpm --filter cloudflare-session-ingest test -- src/queue-consumer.test.ts` passed; Vitest executed the full session-ingest test suite (15 files, 278 tests)

## Visual Changes

N/A

## Reviewer Notes

N/A

### DIFF
--- a/services/session-ingest/src/queue-consumer.test.ts
+++ b/services/session-ingest/src/queue-consumer.test.ts
@@ -184,7 +184,7 @@ describe('queue', () => {
     expect(retry).toHaveBeenCalledWith({ delaySeconds: QUEUE_RETRY_DELAY_SECONDS });
   });
 
-  it('passes full parsed oversized message data and its R2 reference into ingest', async () => {
+  it('passes a slim oversized message and its R2 reference into ingest', async () => {
     const ingest = vi.fn(async () => ({ changes: [] }));
     vi.mocked(getSessionIngestDO).mockReturnValue({ ingest } as never);
     const limit = vi.fn(async () => [{ session_id: 'ses_compacted' }]);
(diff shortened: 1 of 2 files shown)


iscekic (approved):
Approved after correctness review. No externally visible API changes found.

PR #3749, opened 2026-06-04
fix(session-ingest): preserve lifecycle marker order

johnnyeric:
## Summary

This PR addresses the review-report warning: **Same-chunk lifecycle markers no longer preserve order**.

```
WARNING services/session-ingest/src/queue-consumer.ts:304 - Same-chunk lifecycle markers no longer preserve order
Evidence:
+    const items = chunk.splice(0);
...
+      stub => stub.ingest(items, kiloUserId, sessionId, ingestVersion, ingestedAt, r2References),
Trace:
1. Changed code sends multiple ordered items through one SessionIngestDO.ingest call instead of one RPC per item.
2. SessionIngestDO.ingest was checked: it records hasSessionOpen and closeReason while iterating, then always applies open side effects before close side effects.
3. For valid input [session_close, session_open], prior one-item calls ended with cleared close reason and inactivity alarm; batched call restores close reason and replaces alarm with five-second post-close alarm.
Impact: New turn can be finalized as closed and emit metrics prematurely with stale termination reason.
```

After #3744, multiple ordered session-ingest items can be sent through one `SessionIngestDO.ingest()` call. The DO previously collapsed lifecycle markers while iterating: it recorded whether any `session_open` was present and the latest `session_close` reason, then applied open side effects before close side effects. That was equivalent enough when every item arrived in its own RPC, but it is wrong when both markers appear in one batched payload.

For example, with `[session_close, session_open]` in the same payload:

- prior one-item RPC behavior ended with the session opened again
- batched behavior could apply open first and close second, ending with a close reason and post-close alarm

This fix records lifecycle events in payload order and applies the side effects in that same order. That preserves the old per-item ordering semantics while keeping the batched RPC path.

Test coverage added in `services/session-ingest/src/dos/SessionIngestDO.test.ts`:

- a same-batch `[session_close, session_open]` payload applies close side effects first and open side effects second
- the final side effect order matches the prior one-item-RPC semantics

## Verification

Local automated checks run before opening the PR:

- `pnpm exec oxfmt services/session-ingest/src/dos/SessionIngestDO.ts services/session-ingest/src/dos/SessionIngestDO.test.ts`
- `pnpm --filter cloudflare-session-ingest test -- src/dos/SessionIngestDO.test.ts` passed; Vitest executed the full session-ingest test suite (15 files, 279 tests)

## Visual Changes

N/A

## Reviewer Notes

N/A

### DIFF
--- a/services/session-ingest/src/dos/SessionIngestDO.test.ts
+++ b/services/session-ingest/src/dos/SessionIngestDO.test.ts
@@ -1,16 +1,92 @@
 import { describe, expect, it, vi } from 'vitest';
 
+const drizzleMocks = vi.hoisted(() => ({
+  db: undefined as unknown,
+  migrate: vi.fn(),
+}));
+
 vi.mock('cloudflare:workers', () => ({
   DurableObject: class DurableObject {
-    constructor(_state: unknown, _env: unknown) {}
+    ctx: unknown;
+    env: unknown;
+    constructor(state: unknown, env: unknown) {
+      this.ctx = state;
+      this.env = env;
+    }
   },
 }));
 
-import { ingestOrderCursor } from './SessionIngestDO';
+vi.mock('drizzle-orm/durable-sqlite', () => ({
+  drizzle: vi.fn(() => drizzleMocks.db),
+}));
+
+vi.mock('drizzle-orm/durable-sqlite/migrator', () => ({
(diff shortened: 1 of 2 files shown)


iscekic (approved):
Reviewed for correctness. No externally visible API changes found.

PR #3751, opened 2026-06-04
fix(session-ingest): delay metadata changes until side effects finish

johnnyeric:
## Summary

Fixes a metadata consistency race introduced by the queue batching work in #3744, which changed session-ingest from one DO RPC per item to batched `SessionIngestDO.ingest()` calls.

A batched ingest can persist item rows and then perform awaited side effects, like lifecycle alarm scheduling or orphaned R2 blob cleanup. If extractable metadata (`title`, `parentId`, `platform`, `orgId`, `gitUrl`, `gitBranch`, `status`) is written before those side effects finish, a failed RPC can leave the DO metadata updated without returning `changes` to the queue consumer. On retry, the DO sees the same metadata value and does not re-emit it, so Postgres/API consumers can stay stale.

This PR keeps metadata writes aligned with successful side effects:

- lifecycle alarm side effects run before extractable metadata is written, so alarm failure does not mark metadata as already emitted
- `closeReason` stays part of lifecycle handling and is still written before the post-close alarm
- orphaned R2 blob cleanup no longer blocks metadata persistence; it runs in `ctx.waitUntil()` after metadata changes are written/returned
- R2 cleanup failures are logged instead of failing the ingest RPC

The R2 cleanup ordering matters because R2 is external I/O. Awaiting it between timestamp-guarded item writes and metadata persistence can let another request for the same DO interleave, write newer metadata, then get overwritten by stale metadata from the resumed request. Moving cleanup to `waitUntil()` removes that interleaving window.

Test coverage added in `services/session-ingest/src/dos/SessionIngestDO.test.ts`:

- metadata-bearing payload with a failing lifecycle side effect rejects without writing extractable metadata
- `closeReason` is committed before alarm failure, while `title` is not
- orphaned R2 cleanup is scheduled after metadata is written
- stale ingest does not overwrite a simulated newer metadata write during R2 cleanup

## Verification

Local automated checks run after rebasing on `origin/main`:

- `pnpm exec oxfmt services/session-ingest/src/dos/SessionIngestDO.ts services/session-ingest/src/dos/SessionIngestDO.test.ts`
- `pnpm --filter cloudflare-session-ingest test -- src/dos/SessionIngestDO.test.ts` passed; Vitest executed the full session-ingest test suite (15 files, 283 tests)

## Visual Changes

N/A

## Reviewer Notes

N/A

### DIFF
--- a/services/session-ingest/src/dos/SessionIngestDO.ts
+++ b/services/session-ingest/src/dos/SessionIngestDO.ts
@@ -246,25 +246,6 @@ export class SessionIngestDO extends DurableObject<Env> {
       }
     }
 
-    const changes: Changes = [];
-
-    for (const key of Object.keys(incomingByKey) as ExtractableMetaKey[]) {
-      const incoming = incomingByKey[key];
-      if (incoming === undefined) continue;
-      const meta = writeIngestMetaIfChanged(this.db, {
-        key,
-        incomingValue: incoming,
-      });
-      if (meta.changed) {
-        changes.push({ name: key, value: meta.value });
-      }
-    }
-
-    // Clean up orphaned R2 blobs (e.g. replaced or stale oversized items)
-    if (orphanedR2Keys.length > 0) {
-      await this.env.SESSION_INGEST_R2.delete(orphanedR2Keys);
-    }
-
     if (ingestVersion >= 1) {
       // v1 clients send explicit open/close pairs. Only those events drive alarms.
       for (const event of lifecycleEvents) {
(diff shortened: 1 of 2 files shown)


services/session-ingest/src/dos/SessionIngestDO.ts
```diff
@@ -238,20 +238,6 @@ export class SessionIngestDO extends DurableObject<Env> {
...
-      const meta = writeIngestMetaIfChanged(this.db, {
-        key,
-        incomingValue: incoming,
-      });
-      if (meta.changed) {
-        changes.push({ name: key, value: meta.value });
-      }
-    }
-
     // Clean up orphaned R2 blobs (e.g. replaced or stale oversized items)
```
iscekic: This now leaves an awaited R2 delete between the timestamp-guarded item writes and the metadata write/return below. R2 is non-storage I/O, so another request for the same Durable Object can interleave while this await is pending. A stale ingest can pass the timestamp guard, pause here, a newer ingest can update the same session metadata, and then the stale ingest resumes and writes/returns its older title/status/git metadata. The queue consumer applies returned changes to `cli_sessions_v2`, so this becomes visible to API consumers. Please either re-read/recompute the metadata from current persisted item state after the await, or otherwise keep non-storage cleanup out of the window between timestamp validation and metadata persistence.

iscekic (changes requested):
One correctness issue that can leak stale metadata to API consumers.

services/session-ingest/src/dos/SessionIngestDO.ts
johnnyeric: Addressed this issue by using the second option suggested. Deferred the R2 clean up to after the metadata update. Added a new test that failed first and passed after current change.

iscekic (approved):
The previous correctness issue is addressed. The R2 cleanup now runs after metadata persistence via DO waitUntil, so it no longer sits between timestamp-guarded item writes and metadata updates. I re-reviewed the updated head and found no remaining correctness issues in the changed files.

PR #3752, opened 2026-06-04
feat(kiloclaw) multi agent plumbing

St0rmz1:
## Summary

Adds the cloud plumbing to manage OpenClaw agent configuration from the cloud
side. The machine controller already exposes agent config CRUD over
`/_kilo/config/agents*`; this PR wires that contract up through the Durable
Object, the Worker platform routes, and tRPC, in both the personal and org
namespaces. There is no UI in this PR; it is server-side only.

Changes, bottom to top:

- **Durable Object methods** (`kiloclaw-instance/index.ts`, `gateway.ts`):
  `listAgents`, `getAgent`, `updateAgent`, `updateAgentDefaults`, `createAgent`,
  `deleteAgent`. Each is gated on the matching `config.agents.*` capability and
  fails closed (typed `capability_unavailable`) on controllers that do not
  advertise it. `callGatewayController` is widened to PATCH and DELETE.
- **RPC-safe error handling** (`gateway-controller-types.ts`, `gateway.ts`):
  `GatewayControllerError` loses its `.status`/`.code` crossing the DO RPC
  boundary, so the gateway wrappers return an `AgentConfigErrorEnvelope`
  instead of throwing, and the Worker route reconstructs the HTTP response from
  it. Mirrors the existing `kilo-cli-run` / `doctor-run` pattern.
- **Worker routes** (`routes/platform.ts`): `GET/POST/PATCH/DELETE
  /api/platform/agents*` and `PATCH /api/platform/agent-defaults`, mirroring the
  existing `openclaw-config` proxy (internal-key auth, DO-key resolution, opaque
  payload forwarding). Adds `AGENT_CONFIG_ERROR_CODES` plus error reconstruction.
  Non-idempotent mutations opt out of automatic DO retries; mutating calls use a
  longer (180s) request timeout because the controller serializes all agent
  mutations through one per-config queue.
- **Internal client + web types** (`kiloclaw-internal-client.ts`, `types.ts`):
  six client methods and the agent response types, re-declared web-side since
  `apps/web` cannot import from `services/kiloclaw`.
- **tRPC procedures**, personal (`kiloclaw-router.ts`) and org
  (`organization-kiloclaw-router.ts`): six procedures per namespace, with a
  shared `agent-schemas.ts` for deep Zod input validation. Adds 422 and 501
  branches to the error mapping (`422 -> BAD_REQUEST`, `501
  capability_unavailable -> NOT_FOUND`). Org procedures require `organizationId`
  and resolve the instance via `requireOrgInstance`.

No controller, database schema, or migration changes. The capability strings
match the controller's `endpoint-capabilities.ts`.

## Verification

- [ ] No manual testing performed. This PR is server-side plumbing with no UI
  consumer yet, so there is no user-facing path to exercise by hand, and the
  surface is inert (capability-gated, no caller) until the follow-up UI work
  wires it. Behavior is covered by automated tests at each layer: DO/gateway
  (envelope conversion, capability fail-closed, mutation timeouts), Worker
  routes (payload forwarding, error mapping, retry policy), Zod input schemas,
  and tRPC procedures for both namespaces against Postgres.

## Visual Changes

N/A

## Reviewer Notes

- The error path is the most load-bearing part: typed agent errors are
  returned as an envelope and reconstructed in the Worker route, not thrown,
  because custom error properties do not survive the DO RPC boundary. Route stub
  calls use `.then(r => r)` to collapse the RPC union wrapper.
- The surface is capability-gated end to end. The Worker/tRPC layer is the
  enforcement boundary (returns a typed error when the controller does not
  advertise the capability), not just cosmetic UI gating.
- Bindings are out of scope here. Creating an agent without a channel binding
  produces an agent that receives no traffic; editable binding support needs a
  new controller capability and lands in a later PR.
- Follow-ups: the settings UI that consumes these procedures, and the
  controller binding-edit capability.


pandemicsyn (commented):
Posting the verified non-blocking finding from local hyper-review. No must-fix issues found.

apps/web/src/lib/kiloclaw/agent-schemas.ts
```diff
@@ -0,0 +1,129 @@
...
+  'max',
+]);
+const VerboseDefaultSchema = z.enum(['off', 'on', 'full']);
+const ReasoningDefaultSchema = z.enum(['on', 'off', 'stream']);
+
+// A model value to write: primary and/or fallbacks (at least one).
+const ModelInputSchema = z
+  .object({
+    primary: z.string().trim().min(1).max(256).optional(),
+    fallbacks: z.array(z.string().trim().min(1).max(256)).max(20).optional(),
```
pandemicsyn, line 34: Suggested/nit: `fallbacks` accepts an empty array here, so `{ set: { model: { fallbacks: [] } } }` passes even though the model contract says primary and/or fallbacks must include at least one value. Add `.min(1)` to the fallback array or refine on `primary || fallbacks.length > 0`, plus a regression test for empty fallback arrays.

pandemicsyn (approved):

PR #3762, opened 2026-06-05
feat(session-ingest): export session diffs

iscekic:
## Summary

Export the latest `session_diff` payload from session ingest as a top-level `sessionDiff` field so clients can restore filesystem state without scraping message summaries. Update the cloud-agent restore wrapper to prefer that top-level diff payload, apply patch-backed diffs with `git apply --3way`, and retain the existing message-summary fallback for older exports.

## Verification

- [x] No UI path to manually verify; this is backend export/restore behavior. The end-to-end CLI import path was verified against a local fake ingest export in the companion Kilo PR.

## Visual Changes

N/A

## Reviewer Notes

This keeps the ingest export format backward compatible: sessions without a stored diff return `sessionDiff: []`, and restore still falls back to legacy `messages[].info.summary.diffs` data.

### DIFF
--- a/services/cloud-agent-next/wrapper/src/restore-session.test.ts
+++ b/services/cloud-agent-next/wrapper/src/restore-session.test.ts
@@ -22,6 +22,30 @@ function makeSnapshot(diffs: Array<{ file: string; after: string; status: string
   });
 }
 
+function makeSessionDiffSnapshot(patch: string): string {
+  return JSON.stringify({
+    info: snapshotInfo(),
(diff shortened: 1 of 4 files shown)


iscekic:
Fixed in debe6c7f9. `enqueueItemData` now accepts a fallback payload, the `sessionDiff` export path passes `[]` for missing R2-backed diffs, and the integration test covers that missing-blob case.

eshurakov (approved):

PR #3765, opened 2026-06-05
fix(kiloclaw): invalidate destruction deadlines on renewal

jeanduplessis:
## Summary
Prevent successful KiloClaw credit renewals from retaining stale instance-destruction schedules.

### Why this change is needed
A suspended credit-funded subscription could renew to active while keeping its old destruction deadline when async instance startup was accepted but readiness never completed. This left misleading warning state and allowed a stale destruction candidate to race with renewal.

### How this is addressed
- Clear any previous destruction deadline inside every successful active credit-renewal transaction, including duplicate/idempotent reconciliation.
- Retain suspension and auto-resume retry state until actual readiness so interrupted startup continues retrying.
- Revalidate personal warning and destruction candidates against current database state before sending warnings or requesting platform destruction.
- Preserve suspension-cycle email deduplication until readiness completes, then reset it for future suspension lifecycles.
- Document the renewal invariant and final destruction revalidation requirements.

## Human Verification
- Billing lifecycle regression suite passed: 165 tests.
- Instance lifecycle readiness suite passed: 11 tests.
- Independent security, logic, type, data, resource, style, and React review passes found no issues.

### Human Reviewer Flags
- Renewal now invalidates the destruction deadline without clearing `suspended_at`; readiness remains the completion signal for suspension and email-dedupe reset.
- Final database revalidation closes the stale in-memory candidate race, but a small gap remains between revalidation and the external destroy request.
- KiloClaw billing and lifecycle specifications are updated with the new invariants.

### Code Reviewer Agent
<details>

- Normal renewal and duplicate reconciliation share post-advance suspended-recovery behavior.
- Change-log snapshots include the cleared destruction deadline.
- Personal destruction revalidation checks current subscription identity, status, suspension, elapsed deadline, instance identity, personal ownership, and live-instance state.

</details>

### DIFF
--- a/services/kiloclaw-billing/src/lifecycle.ts
+++ b/services/kiloclaw-billing/src/lifecycle.ts
@@ -235,13 +235,33 @@ type OrganizationRecoveryRow = OrganizationEntitlementLifecycleFields & {
   instance_id: string | null;
 };
 
-type OrganizationDestructionRow = OrganizationRecoveryRow & {
+type InstanceDestructionRow = {
+  id: string;
+  user_id: string;
+  instance_id: string | null;
   sandbox_id: string | null;
   instance_name: string | null;
   instance_destroyed_at: string | null;
+  organization_id: string | null;
+  organization_name: string | null;
   plan: KiloClawPlan;
   status: KiloClawSubscriptionStatus;
   email: string;
+  credit_renewal_at: string | null;
+};
+
+type OrganizationDestructionRow = OrganizationEntitlementLifecycleFields & InstanceDestructionRow;
+
+type PersonalDestructionWarningRow = {
+  id: string;
+  user_id: string;
+  email: string;
+  destruction_deadline: string | null;
+  instance_id: string;
+  instance_name: string | null;
+  instance_destroyed_at: string | null;
+  organization_id: string | null;
+  plan: KiloClawPlan;
   credit_renewal_at: string | null;
 };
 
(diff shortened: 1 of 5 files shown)


services/kiloclaw-billing/src/lifecycle.ts
```diff
@@ -1970,6 +2022,14 @@ async function processCreditRenewalRow(
   }
 
   if (outcome.kind === 'duplicate') {
+    await handleCreditRenewalRecoveryAfterAdvance(
```
pandemicsyn, line 2196: Flagging this one just in case. I asked the robot to compare this pr against the spec, this is the one bit that it flagged:

> When the duplicate insert path’s guarded subscription update returns no row, the code still returns kind: 'duplicate' and later runs duplicate side effects such as paid-conversion handling and terminal-failure resolution. The updated spec says duplicate reconciliation should apply side effects only when the same renewal boundary remains current; otherwise it should skip further processing.
  
> Suggested fix direction: split the outcome into reconciled duplicate versus stale/skipped duplicate, return skipped when updatedSubscription is absent, and add a regression test for duplicate insert conflict plus no guarded update row."

But i think the fetch and user-row lock make that path way more narrow than the robot claims? I think this actually fine as-is ?

pandemicsyn (approved):

services/kiloclaw-billing/src/lifecycle.ts
jeanduplessis: Good flag. The user-row lock narrows the race but does not serialize arbitrary subscription-row mutations, and under READ COMMITTED the guarded update can still return no row after the initial read. I changed that case to return a skipped outcome before duplicate recovery, paid-conversion, or terminal-failure-resolution side effects, and added a regression covering duplicate conflict plus zero-row guarded reconciliation. Fixed in c47d83a7c.

PR #3768, opened 2026-06-05
feat(kilo-pass): enforce duplicate-card purchase window

jeanduplessis:
## Summary
Blocks rapid cross-account Kilo Pass card reuse during the 24 hours after its first permanent fingerprint claim, while allowing same-user reuse and later cross-account purchases.

### Why this change is needed
The previous duplicate-card gate could prevent card reuse for months while another account retained an active subscription and could run again on renewals. A rolling accepted-purchase implementation would narrow that behavior but introduce another persistent anti-abuse model alongside existing fingerprint claims and issuance history.

The intended threat model is rapid cross-account reuse shortly after a card first appears, not continuous rolling enforcement after every later purchase.

### How this is addressed
- Reuses permanent welcome-promo fingerprint claims as the atomic first-card-claim record and cooldown timestamp.
- Resolves first claimant through the claim's source invoice, issuance, and subscription.
- Blocks different-user reuse only while the first claim is strictly less than 24 hours old.
- Allows same-user reuse and cross-account reuse at exactly 24 hours or later without refreshing the first claim.
- Limits cancellation enforcement to positive initial Stripe monthly invoices; renewals, yearly purchases, and zero-dollar starts skip the gate.
- Shares exact settled-payment resolution with welcome-promo decisions and never falls back to attached, default, or locally stored cards.
- Uses successful duplicate-block audits for blocked replay authority and matching committed issuances for allowed or fail-open replay authority.
- Removes the branch-only accepted-purchase table and migration.

## Human Verification
- Ran targeted duplicate-card, invoice-paid, checkout, user-retention, settlement-resolution, and email suites: 305 tests passed.
- Verified migration history bootstraps an empty database.
- Verified schema synchronization; schema suite passes 12 of 13 tests, with one unrelated pre-existing enum-registry mismatch.
- Completed independent logic, security, data, type-safety, resource, and maintainability reviews with no findings.

### Human Reviewer Flags
- The cooldown is intentionally anchored to the permanent first fingerprint claim and is not refreshed by later allowed purchases.
- Existing fingerprint claims participate immediately. Missing first-claimant attribution fails open.
- A card allowed after 24 hours remains ineligible for welcome promo and Kilo Pass referral conversion because its permanent claim remains unchanged.
- Successful duplicate-block audits are authoritative blocked-replay evidence; matching committed issuances are authoritative allowed or fail-open replay evidence.
- Stripe cancellation and refund remain best effort inside the database transaction. Current cancellation-email copy states that the charge will be refunded.

### Code Reviewer Agent
<details>

- First-claim uniqueness atomically selects the winner for concurrent card use; no new digest lock or accepted-purchase table remains.
- The exact settled card is resolved once per invoice attempt and shared between duplicate-card and welcome-promo decisions.
- Duplicate audit and operational context contain a fingerprint digest and first-claim identifiers, never the raw fingerprint.
- Blocked handling occurs before scheduled-change credits, issuance, referrals, and affiliate sales.

</details>


apps/web/src/lib/kilo-pass/card-fingerprint-gate.ts
```diff
@@ -1,224 +1,430 @@
...
+      eq(kilo_pass_subscriptions.id, kilo_pass_issuances.kilo_pass_subscription_id)
+    )
+    .where(
+      or(
+        eq(kilo_pass_issuances.stripe_invoice_id, params.stripeInvoiceId),
+        eq(kilo_pass_subscriptions.stripe_subscription_id, params.stripeSubscriptionId)
+      )
+    );
+
+  for (const record of [...blockAudits, ...issuanceRecords]) {
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 40381c37e. Replay authority now resolves and validates an exact invoice issuance first, so later-cycle issuances on the same subscription no longer cause false attribution conflicts. If no exact invoice exists, a separate subscription-level lookup still rejects conflicting payment identity as required by Rule 54. Added regression coverage for replaying the initial invoice after a renewal issuance; focused tests pass (17/17).

.specs/kilo-pass.md
```diff
@@ -239,19 +253,37 @@ with welcome-promo overrides. Yearly subscriptions use a flat 50% monthly bonus.
...
-    exactly-one-winner guarantee across concurrent new subscriptions.
+Rules 51-55 protect against rapid cross-account reuse of a recently first-used card. Allowed purchases do not restart the
+24-hour restriction.
+
+51. A first paid monthly or yearly Stripe subscription purchase MUST be checked after payment and before the purchaser
+    receives Kilo Pass credits or other purchase benefits. Renewals, zero-value starts, later payments for the same
+    subscription, non-card payments, and store-provider purchases MUST NOT be blocked. A later positive monthly payment
+    remains subject to Rule 21 and MAY make its payment instrument count as previously used in future monthly
+    welcome-promo and referral eligibility decisions.
+52. Duplicate-card and monthly welcome-promo or referral decisions for the same purchase MUST be based on the same paid
```
RSO, line 264: > Duplicate-card and monthly welcome-promo or referral decisions for the same purchase

What does "for the same purchase" mean in this case? A purchase can only be a single thing, right?

.specs/kilo-pass.md
```diff
@@ -239,19 +253,37 @@ with welcome-promo overrides. Yearly subscriptions use a flat 50% monthly bonus.
...
+    card settlement. The purchase MUST be allowed when that settlement does not identify exactly one paid card. A provider
```
RSO, line 265: > The purchase MUST be allowed when that settlement does not identify exactly one paid card.

Can you describe this scenario? I'm not sure I follow how a purchase can not identify exactly one paid card.

.specs/kilo-pass.md
```diff
@@ -239,19 +253,37 @@ with welcome-promo overrides. Yearly subscriptions use a flat 50% monthly bonus.
...
+    lookup failure MUST produce `settlement_unresolved` when a monthly welcome-promo decision is required. Unexpected
```
RSO, line 266: What is a "provider lookup failure"?

.specs/kilo-pass.md
```diff
@@ -239,19 +253,37 @@ with welcome-promo overrides. Yearly subscriptions use a flat 50% monthly bonus.
...
+    evidence or attribution failures MUST be reported.
+53. The first eligible purchase made with a card MUST establish that purchaser as the first claimant and MUST be allowed.
+    A different user MUST be blocked only when the first use was less than 24 hours earlier. The first claimant and first-use
+    time MUST NOT change. Same-user purchases and different-user purchases at or after 24 hours MUST be allowed, regardless
```
RSO, line 270: Since you're explicitly saying that same-user purchases must be allowed after 24 hours, does that mean that you want to block them within the 24 hour window?

.specs/kilo-pass.md
```diff
@@ -239,19 +253,37 @@ with welcome-promo overrides. Yearly subscriptions use a flat 50% monthly bonus.
...
+    of subscription status, cancellation, card attachment, or later allowed purchases. A card first used for a yearly
```
RSO, line 271: What does "regardless of card attachment" mean?
Also, "Regardless of later allowed purchases" doesn't make any sense, right, because you can't know at the time whether there'll be any later allowed purchases?

.specs/kilo-pass.md
```diff
@@ -239,19 +253,37 @@ with welcome-promo overrides. Yearly subscriptions use a flat 50% monthly bonus.
...
+    purchase MUST count as previously used in later monthly welcome-promo and referral decisions, while the yearly
+    purchase itself remains ineligible for those benefits. If the first claimant cannot be determined reliably, the
```
RSO, line 273: > while the yearly purchase itself remains ineligible for those benefits.

I don't understand what this means 🤔 the purchase itself is what makes the user eligible for the benefits, right?

.specs/kilo-pass.md
```diff
@@ -239,19 +253,37 @@ with welcome-promo overrides. Yearly subscriptions use a flat 50% monthly bonus.
...
+    purchase MUST be allowed and the failure MUST be reported.
+54. Concurrent eligible purchases using the same card MUST produce one first claimant. Retried processing MUST preserve the
+    original blocked or allowed outcome, including after the 24-hour restriction expires or card evidence becomes
+    unavailable. A blocked outcome MUST take precedence over an allowed outcome. If no prior outcome can be established,
+    the purchase MUST receive the standard duplicate-card decision. Conflicting purchaser, subscription, or payment
+    identity MUST abort processing rather than trigger a new decision.
```
RSO, line 279: What scenario are you covering here? I don't understand what "concurrent eligible purchases" are

.specs/kilo-pass.md
```diff
@@ -239,19 +253,37 @@ with welcome-promo overrides. Yearly subscriptions use a flat 50% monthly bonus.
...
+55. A blocked purchase or retry MUST create no Kilo Pass issuance, credits, referral conversion, referral reward, or
+    affiliate sale. The Kilo Pass subscription MUST remain canceled, and the purchaser MUST receive the permanent
+    `kilo_pass_duplicate_card` block unless another block reason already exists. The system MUST attempt to cancel the
+    Stripe subscription, refund the settled payment, and send a cancellation email that reveals neither the 24-hour
+    restriction nor the matched claimant. Missing refund evidence or provider failure MUST be reported but MUST NOT
```
RSO, line 284: > Missing refund evidence

What is refund evidence?

.specs/kilo-pass.md
jeanduplessis: One paid invoice can drive several decisions: duplicate-card enforcement, welcome-promo eligibility, and referral eligibility. The intent is that they all use the same settled payment method. I'll could rewrite this around the invoice and settlement rather than 'same purchase' if it makes it clearer.

.specs/kilo-pass.md
jeanduplessis: It's the Stripe evidence that can be ambiguous, not the purchase itself. We can see zero or multiple paid invoice-payment records, or one settlement that is non-card or missing card details. We only block when exactly one paid settlement resolves to a card fingerprint.

.specs/kilo-pass.md
jeanduplessis: Provider simply refers to the Stripe API here and the lookup call needed to resolve the settled payment for example when loading the invoice payments.

.specs/kilo-pass.md
jeanduplessis: No. Same-user purchases are always allowed. Only a different user's purchase is blocked during the first 24 hours.

.specs/kilo-pass.md
jeanduplessis: Card attachment means whether the card remains attached to either Stripe customer after the first claim. Detaching it doesn't remove the permanent claim on our side. Later allowed purchases means those purchases don't replace the claimant or refresh the first-use timestamp. I can state this as two separate invariants to make it clearer.

.specs/kilo-pass.md
jeanduplessis: The benefits here are the monthly welcome promo or Kilo Pass referral reward and annual purchases don't qualify for those. They still claim the payment fingerprint, so a later monthly purchase using the same card is treated as previously used regardless of whether the benefit was claimed.

.specs/kilo-pass.md
jeanduplessis: This is two "users" starting paid subscriptions with the same card close enough that both invoice webhooks are processed concurrently. The database claim must select which one is first. This covers the scenario where a user or bot does rapid-fire purchases but the async webhooks are processed in order. It's probably a rare scenario, but the model loves defense in depth.

.specs/kilo-pass.md
jeanduplessis: It's the PaymentIntent or Charge ID needed to issue the refund. The account stays blocked, but the failed refund attempt is reported.

RSO (approved):

PR #3790, opened 2026-06-08
feat(kiloclaw): retire Commit subscriptions

jeanduplessis:
## Summary
Retire new and recurring KiloClaw Commit subscriptions while honoring one qualified final term and requiring explicit Standard continuation.

### Why this change is needed
Commit sales ended at the exclusive `2026-06-06T00:00:00Z` cutoff, but existing customers and confirmed pre-cutoff obligations still need their paid terms honored. Retirement must prevent another six-month charge without silently converting customers, removing paid access, or letting ambiguous Stripe and local state trigger destructive lifecycle actions.

### How this is addressed
- Enforce the cutoff across checkout, credits, Kilo Pass, plan switching, reactivation, Stripe settlement, and support or admin mutation paths.
- Preserve one verified final Commit term, then cancel at its authorized boundary unless the customer explicitly schedules lineage-priced Standard.
- Add durable qualification, final-boundary, continuation, provider-guard, and manual-review state with an operator review queue.
- Guard Stripe-backed terms before renewal and stop pure-credit Commit renewal without deduction, auto top-up, or period advancement.
- Preserve retirement evidence through settlement, referral extensions, and reprovisioning while containing uncertain provider outcomes.
- Remove Commit from selectable customer offers and expose final-term, continuation, undo, and support-review states across billing surfaces.
- Add cutoff capture, checkout expiration, and retirement audit scripts plus source-of-truth specs and regression coverage.

## Human Verification
Local development run through to ensure plan is removed from UI options.

kiloclaw-commit-plan-retired-final

### Human Reviewer Flags
- Review the new additive retirement state model and manual-review containment rules against the updated billing, lifecycle, data-model, Kilo Pass, referral, and Subscription Center specs.
- Validate rollout ordering for migration, pre-cutoff qualification capture, checkout expiration, runtime deployment, and post-deploy audit.
- Review the external-first Stripe mutation tradeoff: provider renewal is contained before guarded local persistence, and uncertain outcomes remain open for operator review rather than unsafe rollback.

### Code Reviewer Agent
<details>

- Cutoff eligibility is exclusive at `2026-06-06T00:00:00Z`; exact-cutoff requests do not qualify.
- Commit retirement remains additive to payment and access status so paid access can continue while renewal is blocked.
- Stripe guard processing uses bounded concurrency, wall-clock limits, and stable keyset continuation.
- Admin review pagination uses `(created_at, id)` cursors and limits concurrent provider snapshots.
- Provider-only dispositions remain durable so later webhook or settlement ingestion cannot recreate forbidden Commit state.

</details>


apps/web/src/lib/kiloclaw/commit-retirement.ts
```diff
@@ -0,0 +1,595 @@
...
+    currentPeriodEndsAt: subscription.current_period_end,
+    providerPeriodEndsAt: getStripeSubscriptionPeriodEnd(live),
+  });
+  if (
+    boundary.kind !== 'verified' ||
+    !timestampsEqual(boundary.finalEndsAt, params.expectedFinalBoundary)
+  ) {
+    reportKiloClawCommitRetirementAnomaly({
+      reason: 'boundary_mismatch',
+      summary: 'Retirement guard boundary could not be verified; provider non-renewal forced.',
```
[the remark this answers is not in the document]
jeanduplessis: Fixed by distinguishing a verified-but-stale boundary from an unverified boundary in the anomaly summary. Provider non-renewal remains fail-closed in both cases.

apps/web/src/lib/kiloclaw/commit-retirement.ts
```diff
@@ -0,0 +1,595 @@
...
+      .where(eq(kiloclaw_subscriptions.id, subscription.id))
+      .for('update')
+      .limit(1);
+    if (
+      !before ||
+      (before.scheduled_plan === 'standard' && before.scheduled_by === 'user') ||
+      before.plan !== 'commit' ||
+      before.status !== 'active' ||
+      before.stripe_subscription_id !== subscription.stripe_subscription_id ||
+      !timestampsEqual(before.current_period_end, subscription.current_period_end)
```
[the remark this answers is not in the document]
jeanduplessis: Fixed by explicitly requiring the locked row's commit_ends_at to match expectedFinalBoundary before persisting guard state. This prevents a stale sweep cursor from overwriting an independently changed final boundary.

apps/web/src/routers/kilo-pass-router.ts
```diff
@@ -1123,12 +1154,115 @@ export const kiloPassRouter = createTRPCRouter({
...
+          code: 'BAD_REQUEST',
+          message: 'Kilo Pass credits are still processing.',
+        });
+      }
+
+      await enrollWithCredits({
+        userId: ctx.user.id,
+        instanceId: instance.id,
+        plan: hostingPlan,
+        hadPaidSubscription: false,
```
pandemicsyn: Must-fix: This hard-codes KiloClaw paid-history eligibility to false for Kilo Pass hosting activation. A user with prior paid KiloClaw history can come through this path and get intro/first-time pricing that normal KiloClaw enrollment would deny. Recompute the prior-paid-subscription check at activation time, or persist and verify a server-authored eligibility value before passing it to enrollWithCredits.

apps/web/src/app/payments/kilo-pass/awarding/KiloPassAwardingCreditsClient.tsx
```diff
@@ -104,40 +109,24 @@ export function KiloPassAwardingCreditsClient() {
   });
 
   const isReady = query.data?.creditsAwarded === true;
+  const isClawAutoActivation =
```
pandemicsyn, line 112: Must-fix: The client only runs the KiloClaw activation mutation when the polling response has a non-none hostingIntent, but the server maps post-cutoff Commit intents to none. That means expired Commit hosting checkouts are treated as ordinary Kilo Pass purchases and redirect to /profile instead of surfacing the expired_commit recovery path. Preserve an expired Commit hosting signal in getCheckoutReturnState, or include expired_commit in this state and route it through activateCheckoutHosting.

apps/web/src/lib/impact/kiloclaw-referrals.ts
```diff
@@ -371,20 +373,36 @@ async function hasDeletedUserEmailTombstone(params: {
...
+function hasUserScheduledStandardContinuation(subscription: KiloClawSubscription): boolean {
+  return subscription.scheduled_plan === 'standard' && subscription.scheduled_by === 'user';
+}
+
+function isEligibleFinalCommitCancellation(subscription: KiloClawSubscription): boolean {
+  return (
+    subscription.plan === 'commit' &&
+    subscription.commit_ends_at !== null &&
+    subscription.current_period_end === subscription.commit_ends_at &&
+    subscription.cancel_at_period_end &&
```
pandemicsyn: Must-fix: This treats any final Commit row with cancel_at_period_end as reward-eligible, but that same column shape can come from an ordinary user cancellation. The referral spec only keeps final Commit terms eligible when non-renewal is solely the retirement guard. Derive this from canonical change-log evidence, or explicitly exclude rows whose latest cancellation mutation was user initiated, and add a regression for user-canceled final Commit terms where commit_ends_at equals current_period_end.

apps/web/src/lib/impact/kiloclaw-referrals.ts
```diff
@@ -936,16 +1043,57 @@ async function applyReferralRewardById(
...
+    const phases = buildExtendedRetirementSchedulePhases({
+      schedule,
+      previousBoundary: result.previousBoundary,
+      newBoundary: result.newBoundary,
+    });
+    if (!phases) {
+      throw new Error('Ambiguous retirement schedule phases');
     }
-  );
+    await stripe.subscriptionSchedules.update(result.scheduleId, { phases });
```
pandemicsyn: Must-fix: The Stripe schedule is updated before the local reward application is persisted, but retries do not recognize an already-extended schedule. If Stripe succeeds and the follow-up local transaction fails, the next run can classify the already-applied provider state as ambiguous and leave the reward in review_required. Make the schedule path idempotent by accepting the already-applied phase shape and then completing the local mutation, or persist/reconcile an operation record around the external update.

services/kiloclaw-billing/src/lifecycle.ts
```diff
@@ -1758,14 +1859,43 @@ async function processCreditRenewalRow(
...
+    const pendingSwitchQualification =
+      current.plan === 'standard' && current.scheduled_plan === 'commit'
+        ? await findLatestPreCutoffUserCommitSwitchQualification(tx, current.id)
+        : null;
+    const retirementDecision = decideCommitRetirementCreditRenewal(
+      current,
+      pendingSwitchQualification !== null
+    );
+    if (retirementDecision.kind === 'ambiguous') {
+      log('error', 'Skipping ambiguous Commit credit renewal', {
```
pandemicsyn, line 1882: Must-fix: Ambiguous Commit renewal evidence fails closed here, but it is only logged by the Worker. The Commit retirement specs require these anomalies to be reported through the operational anomaly path/Sentry as well, so production cases are visible and retryable. Route this branch through a non-sensitive Worker-compatible anomaly reporter while preserving the skipped fail-closed behavior.

pandemicsyn (changes requested):

pandemicsyn (approved):

apps/web/src/lib/impact/kiloclaw-referrals.ts
jeanduplessis: Fixed by requiring canonical retirement-guard change-log evidence for canceling final Commit rows without Standard continuation. Added regression coverage where user cancellation has commit_ends_at equal to current_period_end and remains pending.

PR #3800, opened 2026-06-08
feat(model-experiments): edit model metadata

chrarnoldus:


apps/web/src/app/admin/model-experiments/ModelExperimentsContent.tsx
```diff
@@ -43,15 +44,51 @@ import {
...
+import * as z from 'zod';
+
+const StrictCustomLlmMetadataSchema = deepStrict(CustomLlmMetadataSchema);
 
 const INITIAL_UPSTREAM: ExperimentUpstream = {
   internal_id: '',
   base_url: '',
 };
 
+function parseMetadataJson(
```
marius-kilocode: Not sure if this responsibility should sit here. Shall we split parsing logic and UI at least file wise?

marius-kilocode (approved):

PR #3814, opened 2026-06-08
fix(kilo-pass): reconcile subscription webhooks with Stripe

jeanduplessis:
## Summary
Prevent stale or concurrent Stripe webhook deliveries from resurrecting canceled Kilo Pass subscriptions, while hardening the duplicate-card path that produced the incident.

### Why this change is needed
Out-of-order subscription events allowed an older `customer.subscription.created` payload to overwrite a newer canceled state. This left Stripe canceled and refunded while local checkout logic still treated the subscription as active.

The webhook reconciliation change fixes that root cause: local state previously followed delivery order instead of Stripe's current state. The duplicate-card changes harden the incident-producing path: cancellation and refund outcomes are now accurate, observable, and safely retryable. Duplicate-card enforcement is not the ordering fix itself, but it prevents partial provider failures from leaving misleading audit state or unsafe retry behavior.

### How this is addressed
- Fix the root cause by reconciling each subscription event against current Stripe state under a per-subscription advisory lock.
- Treat Stripe `resource_missing` as canceled and propagate temporary retrieval failures for webhook retry.
- Record stale, duplicate, applied, and failed reconciliation outcomes in existing audit facilities.
- Harden the incident-producing duplicate-card path by committing the block before provider calls, moving cancellation and refund outside the database transaction, and recording their outcomes separately.
- Preserve retry safety through webhook failure propagation and deterministic Stripe idempotency keys without granting credits or reversing the block.
- Keep existing schema unchanged; no webhook ledger, workflow table, or migration is introduced.

## Human Verification
- Focused Kilo Pass subscription, invoice, and duplicate-card webhook tests passed.
- Stripe dispatcher tests passed.
- Seven-focus code review completed; React Code Quality analyzer passed with 100/100 and no applicable React files.

### Human Reviewer Flags
- Current Stripe retrieval is the primary ordering mechanism; subscription-row event metadata was intentionally not added.
- Stripe retrieval occurs while holding the transaction-scoped advisory lock so concurrent handlers cannot persist state out of order.
- Duplicate-card hardening addresses the incident-producing path, while current-state reconciliation is the root-cause fix.
- Kilo Pass spec now requires duplicate-card cancellation failures to trigger webhook retry while refund failures remain observable and nonfatal.

### Code Reviewer Agent
<details>

- `resource_missing` maps to terminal canceled state; other Stripe retrieval failures roll back local mutation and propagate.
- Duplicate-card provider calls run after the database transaction commits.
- Existing audit action values are reused with structured outcome payloads, avoiding a schema migration.
- Store-managed subscriptions are unaffected because routing changes apply only to Stripe Kilo Pass subscription events.

</details>

### DIFF
--- a/apps/web/src/lib/kilo-pass/card-fingerprint-gate.ts
+++ b/apps/web/src/lib/kilo-pass/card-fingerprint-gate.ts
@@ -443,15 +443,9 @@ export async function attemptDuplicateCardProviderEnforcement(params: {
       );
     } catch (error) {
       if (mayIndicateAlreadyCanceledStripeSubscriptionError(error)) {
+        let subscription: Stripe.Subscription;
         try {
-          const subscription = await params.stripe.subscriptions.retrieve(
-            params.stripeSubscriptionId
-          );
-          if (isStripeSubscriptionEnded(subscription.status)) {
-            canceledSubscription = { id: subscription.id };
-          } else {
-            throw error;
-          }
+          subscription = await params.stripe.subscriptions.retrieve(params.stripeSubscriptionId);
         } catch (confirmationError) {
           captureException(confirmationError, {
             tags: {
@@ -460,6 +454,11 @@ export async function attemptDuplicateCardProviderEnforcement(params: {
             },
             extra: operationalContext,
           });
+          throw error;
+        }
+        if (isStripeSubscriptionEnded(subscription.status)) {
+          canceledSubscription = { id: subscription.id };
+        } else {
           throw error;
         }
       } else {
--- a/apps/web/src/lib/kilo-pass/card-fingerprint-gate.ts
+++ b/apps/web/src/lib/kilo-pass/card-fingerprint-gate.ts
@@ -30,6 +30,16 @@ import { createHash } from 'node:crypto';
 
 const KILO_PASS_DUPLICATE_CARD_EMAIL_TYPE = 'kilo_pass_duplicate_card_canceled';
 
+function isAlreadyCanceledStripeSubscriptionError(error: unknown): boolean {
+  const message = error instanceof Error ? error.message : String(error);
+  const normalizedMessage = message.toLowerCase();
+  return (
+    (normalizedMessage.includes('already') && normalizedMessage.includes('cancel')) ||
+    normalizedMessage.includes('has been canceled') ||
+    normalizedMessage.includes('is canceled')
+  );
+}
+
 export type PaymentFingerprintClaimResult = {
   welcomePromoReason: KiloPassWelcomePromoEligibilityReason;
   cardClaim: {
@@ -418,7 +428,7 @@ export async function attemptDuplicateCardProviderEnforcement(params: {
     matchedKiloUserId: params.gateResult.matchedKiloUserId,
   };
 
-  let canceledSubscription: Stripe.Subscription;
+  let canceledSubscription: { id: string };
   try {
     canceledSubscription = await params.stripe.subscriptions.cancel(
       params.stripeSubscriptionId,
@@ -426,11 +436,15 @@ export async function attemptDuplicateCardProviderEnforcement(params: {
       { idempotencyKey: `kilo-pass-duplicate-card-cancel:${params.stripeInvoiceId}` }
     );
   } catch (error) {
-    captureException(error, {
-      tags: { source: 'kilo_pass_duplicate_card_gate', stage: 'subscription_cancel' },
-      extra: operationalContext,
-    });
-    throw error;
+    if (isAlreadyCanceledStripeSubscriptionError(error)) {
+      canceledSubscription = { id: params.stripeSubscriptionId };
+    } else {
+      captureException(error, {
+        tags: { source: 'kilo_pass_duplicate_card_gate', stage: 'subscription_cancel' },
+        extra: operationalContext,
+      });
+      throw error;
+    }
   }
 
   const refundableTarget = params.gateResult.refundableTarget;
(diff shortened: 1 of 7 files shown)


apps/web/src/lib/kilo-pass/stripe-handlers-subscription-events.ts
```diff
@@ -21,130 +22,178 @@ import {
...
+          eventStatus: eventSubscription.status,
+          appliedStatus: stripeStatus,
+          resourceMissing,
+          staleDelivery: eventSubscription.status !== stripeStatus,
+        },
+      });
+    });
+  } catch (error) {
+    captureException(error, {
+      tags: { source: 'kilo_pass_subscription_reconciliation', stage: 'stripe_retrieve' },
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 50a0ffd70. The outer transaction catch now uses stage=subscription_reconciliation, so database, lock, pause, and audit failures are no longer mislabeled as Stripe retrieval failures.

apps/web/src/lib/kilo-pass/stripe-handlers-subscription-events.ts
```diff
@@ -21,130 +22,189 @@ import {
...
+  if (wasDuplicateDelivery) return;
 
   void reportEvents({
     events: [
       {
         type: 'billing.kilo_pass_changed',
         data: {
-          kilo_user_id: kiloUserId,
-          tier,
+          kilo_user_id: eventMetadata.kiloUserId,
```
pandemicsyn: Must-fix: This report still uses `eventMetadata` from the webhook payload even though the transaction may have reconciled and persisted `metadata` from the current Stripe subscription. A stale/out-of-order event can now write the current tier/status to DB but report the old user/tier to `billing.kilo_pass_changed`. Carry the reconciled metadata out of the transaction and report those final values instead.

apps/web/src/lib/kilo-pass/stripe-handlers-invoice-paid.ts
```diff
@@ -69,6 +70,108 @@ import {
...
+  const existingCancellationSuccess = await db.query.kilo_pass_audit_log.findFirst({
+    columns: { id: true },
+    where: and(
+      eq(kilo_pass_audit_log.action, KiloPassAuditLogAction.DuplicateCardSubscriptionCanceled),
+      eq(kilo_pass_audit_log.result, KiloPassAuditLogResult.Success),
+      eq(kilo_pass_audit_log.stripe_invoice_id, enforcement.stripeInvoiceId),
+      sql`${kilo_pass_audit_log.payload_json}->>'outcome' = 'subscription_cancellation'`
+    ),
+  });
+  if (existingCancellationSuccess) return;
```
pandemicsyn: Must-fix: This early return treats a prior successful cancellation audit as complete enforcement even when no refund outcome was recorded. If the refund audit write failed after the cancellation audit committed, a webhook retry would return here and never record/retry the refund outcome, despite the spec requiring separate cancellation and refund audit payloads. Gate cancellation and refund idempotency independently, and only skip once a terminal refund audit exists.

apps/web/src/lib/kilo-pass/card-fingerprint-gate.ts
```diff
@@ -30,6 +30,16 @@ import { createHash } from 'node:crypto';
 
 const KILO_PASS_DUPLICATE_CARD_EMAIL_TYPE = 'kilo_pass_duplicate_card_canceled';
 
+function isAlreadyCanceledStripeSubscriptionError(error: unknown): boolean {
+  const message = error instanceof Error ? error.message : String(error);
+  const normalizedMessage = message.toLowerCase();
+  return (
+    (normalizedMessage.includes('already') && normalizedMessage.includes('cancel')) ||
```
pandemicsyn, line 38: Suggested/nit: This treats broad error-message substrings as proof that the target subscription is already canceled. Because this is a provider enforcement decision, prefer a structured Stripe error code/type or retrieve/confirm the subscription status before suppressing the cancellation error and proceeding to refund.

pandemicsyn (changes requested):
Two must-fix findings and one suggested nit are included inline.

pandemicsyn (approved):

apps/web/src/lib/kilo-pass/stripe-handlers-subscription-events.ts
jeanduplessis: Fixed in b18ecda33: subscription reconciliation now carries the reconciled Stripe metadata out of the transaction and reports final user/tier/status values to billing.kilo_pass_changed.

apps/web/src/lib/kilo-pass/card-fingerprint-gate.ts
```diff
@@ -401,17 +431,52 @@ export async function attemptDuplicateCardProviderEnforcement(params: {
...
+        try {
+          const subscription = await params.stripe.subscriptions.retrieve(
+            params.stripeSubscriptionId
+          );
+          if (isStripeSubscriptionEnded(subscription.status)) {
+            canceledSubscription = { id: subscription.id };
+          } else {
+            throw error;
+          }
+        } catch (confirmationError) {
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 174fad5bd: the confirmation retrieve is now isolated from the active-subscription branch, so Sentry only captures actual retrieve/confirmation failures and active status rethrows the original cancel error without a misleading confirmation alert.

jeanduplessis (commented):
Submitting review replies after addressing and resolving all threads.

PR #3818, opened 2026-06-08
fix(ai-gateway): preserve valid Terminal Bench stats

lambertjosh:
## Summary

- Validate only the Terminal Bench subtree when enriching the model catalog.
- Preserve valid Terminal Bench results when unrelated benchmark providers contain nullable or otherwise incompatible fields.
- Cover the production-shaped `artificialAnalysis.liveCodeBench: null` regression.

## Verification

- Verified locally that GPT-5.5 appears in the Kilo Gateway catalog with its Terminal Bench score and attempt cost while retaining no Artificial Analysis data.

## Visual Changes

N/A

## Reviewer Notes

This keeps Terminal Bench publication isolated from schema drift in unrelated benchmark providers. Existing eligibility requirements remain unchanged.

### DIFF
--- a/apps/web/src/lib/model-stats/terminal-bench.test.ts
+++ b/apps/web/src/lib/model-stats/terminal-bench.test.ts
@@ -46,9 +46,14 @@ function row(overrides: Partial<Parameters<typeof summarizeTerminalBench>[0][num
 describe('summarizeTerminalBench', () => {
   it('publishes only eligible non-internal summaries', () => {
     const stealth = { ...row({ openrouterId: 'stealth/model' }), isStealth: true };
+    const nullable = {
+      ...benchmarks(),
+      artificialAnalysis: { liveCodeBench: null },
+    };
     const summaries = summarizeTerminalBench([
       row(),
       stealth,
+      row({ openrouterId: 'nullable/model', benchmarks: nullable }),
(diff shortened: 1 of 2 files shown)


apps/web/src/lib/model-stats/terminal-bench.test.ts
```diff
@@ -46,9 +46,14 @@ function row(overrides: Partial<Parameters<typeof summarizeTerminalBench>[0][num
...
   it('publishes only eligible non-internal summaries', () => {
     const stealth = { ...row({ openrouterId: 'stealth/model' }), isStealth: true };
+    const nullable = {
+      ...benchmarks(),
+      artificialAnalysis: { liveCodeBench: null },
+    };
     const summaries = summarizeTerminalBench([
       row(),
       stealth,
+      row({ openrouterId: 'nullable/model', benchmarks: nullable }),
```
marius-kilocode, line 56: Why is `nullable/model` eligible? This smells like a bug...

lambertjosh: It's validating that a model with terminal bench data (benchmarks())) but _null_ artificialanalysis data still contains terminal bench data. Previously this would get filtered out. It's not testing any validation of the model or provider name itself, just test data.... if I understood your question correctly.

marius-kilocode (approved):

PR #3819, opened 2026-06-08
feat(kiloclaw) WIP multi agents UI

St0rmz1:
## Summary

Adds an admin-only Agents management page to the KiloClaw web app, for both
personal and organization instances. It consumes the already-merged agent
backend (controller CRUD + bindings via DO/Worker/tRPC) to list the agents
running on a machine and create, edit, and delete them. The surface is gated
behind admin status and per-operation controller capabilities. It is a work in
progress shipped for limited testing.

## Changes

Page and navigation:
- New routes: `/claw/agents` (personal) and `/organizations/[id]/claw/agents`
  (org), reusing the `ClawSettingsPage` shell (context provider, status gate,
  billing wrapper).
- "Agents" item added to the personal and org KiloClaw sidebars, shown only to
  admins.
- A work-in-progress / admin-only note at the top of the page.

Gating:
- Admin-only via `useUser().is_admin`, fail-closed: the nav item is hidden and
  the page redirects non-admins (and unknown/errored admin status) to settings.
- Per-capability gating: the view requires `config.agents.read`; New, edit, and
  delete are each gated on `config.agents.create.basic.cli`,
  `config.agents.update`, and `config.agents.delete.cli`. Older controllers that
  don't advertise the read capability get an "update your machine" state, and a
  controller-version fetch failure shows a distinct error rather than an upgrade
  prompt.

Read view (`AgentsSection`):
- Lists each agent's effective model (including fallback-only models), routed
  channels, and behavioral settings, plus the inherited fleet defaults.

Create / edit / delete:
- Create dialog (name + optional model; workspace path derived from the name).
- Edit dialog (model plus thinking/verbose/reasoning/fast-mode), with an
  "inherit default" option per field, diffed into a set/unset patch and guarded
  by the config etag captured at open time.
- Delete via a confirmation dialog with honest "workspace files may remain"
  copy; the reserved `main` agent cannot be deleted.
- `useClawAgentMutations` dispatcher (personal + org) that injects the org id
  and invalidates the agent list on success.

Model selection:
- Create and edit use the shared `ModelCombobox` backed by a new
  `useClawModelOptions` hook, so the agent dialogs offer the same
  version-filtered model list as the Settings Model Configuration picker, and
  surface catalog/version load failures.

Supporting:
- Shared `THINKING`/`VERBOSE`/`REASONING` option tuples exported from
  `agent-schemas.ts` so the Zod enums and the editor selects share one source.
- `useKiloClawStatus` gained an optional `enabled` flag so the personal status
  query doesn't poll on org routes.

## Verification

Exercised in local dev (personal context):
- [x] Agents nav item and page are hidden/redirected for non-admins, shown for admins
- [x] Read-only fleet view renders the agent list and inherited defaults
- [x] Create / edit / delete dialogs open and submit; the list refreshes after each

Full end-to-end validation (including model selection) is being done via the
live limited-testing rollout.

## Visual Changes

New UI: a dedicated Agents page, an admin-only "Agents" sidebar entry, and
create/edit/delete dialogs. Screenshots to be added.

| Before | After |
|---|---|
|  |  |

## Reviewer Notes

- This is UI only; it is a pure consumer of the already-merged agent backend.
- Gating is client-side (nav visibility, page redirect, per-capability
  controls). The agent tRPC procedures are not admin-gated server-side; whether
  to add that is an open decision.
- The create dialog derives the workspace path by mirroring the controller's
  agent-id normalization. A shared module isn't possible across the
  apps/web ↔ services boundary, so it's re-declared (same pattern as the Zod
  input schemas).

### DIFF
--- a/apps/web/src/app/(app)/claw/components/AgentEditDialog.tsx
+++ b/apps/web/src/app/(app)/claw/components/AgentEditDialog.tsx
@@ -4,6 +4,7 @@ import { useMemo, useState } from 'react';
 import { toast } from 'sonner';
 
 import { AnimatedDots } from './AnimatedDots';
+import { ModelCombobox } from '@/components/shared/ModelCombobox';
 import { Button } from '@/components/ui/button';
 import { Checkbox } from '@/components/ui/checkbox';
 import {
@@ -14,7 +15,6 @@ import {
   DialogHeader,
   DialogTitle,
 } from '@/components/ui/dialog';
-import { Input } from '@/components/ui/input';
 import { Label } from '@/components/ui/label';
 import {
   Select,
@@ -26,6 +26,7 @@ import {
 import type { AgentUpdateInput } from '@/lib/kiloclaw/agent-schemas';
 import type { AgentSummary } from '@/lib/kiloclaw/types';
 import { useClawAgentMutations } from '../hooks/useClawHooks';
+import { useClawModelOptions } from '../hooks/useClawModelOptions';
 
 const INHERIT = 'inherit';
 const THINKING = ['off', 'minimal', 'low', 'medium', 'high', 'xhigh', 'adaptive', 'max'] as const;
@@ -90,6 +91,7 @@ export function AgentEditDialog({
   etag: string;
 }) {
   const { updateAgent } = useClawAgentMutations();
+  const { modelOptions, isLoading: isLoadingModels } = useClawModelOptions();
 
   const initial = useMemo(() => ownModel(agent), [agent]);
   // An agent owns a model when rawModel is set — including a fallback-only model
@@ -220,12 +222,15 @@ export function AgentEditDialog({
             </div>
             {!inheritModel && (
               <>
-                <Input
-                  id="agent-edit-model"
+                <ModelCombobox
+                  label=""
+                  models={modelOptions}
                   value={primary}
-                  maxLength={256}
-                  placeholder="Model id (e.g. kilocode/kilo-auto/balanced)"
-                  onChange={e => setPrimary(e.target.value)}
+                  onValueChange={setPrimary}
+                  isLoading={isLoadingModels}
+                  placeholder="Select a model"
+                  modal
+                  className="w-full"
                 />
                 {initial.fallbacks.length > 0 && (
                   <p className="text-muted-foreground text-xs">
--- a/apps/web/src/app/(app)/claw/components/AgentEditDialog.tsx
+++ b/apps/web/src/app/(app)/claw/components/AgentEditDialog.tsx
@@ -5,6 +5,7 @@ import { toast } from 'sonner';
 
 import { AnimatedDots } from './AnimatedDots';
 import { Button } from '@/components/ui/button';
+import { Checkbox } from '@/components/ui/checkbox';
 import {
   Dialog,
   DialogContent,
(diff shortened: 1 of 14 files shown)


apps/web/src/app/(app)/claw/hooks/useClawModelOptions.ts
```diff
@@ -0,0 +1,79 @@
...
+  const isRunning = status.status === 'running';
+  const { trackedVersion, runningVersion, isLoadingControllerVersion, isControllerVersionError } =
+    useClawUpdateAvailable(status);
+
+  const versionError = isRunning && isControllerVersionError;
+  const modelOptions = useMemo<ModelOption[]>(
+    () =>
+      getSettingsModelOptions({
+        models: (modelsData?.data || []).map(model => ({
+          id: model.id,
```
jeanduplessis, line 51: **WARNING:** Prefix selected catalog IDs before submitting them as KiloClaw agent model refs.

Trace: this hook exposes public catalog IDs such as `anthropic/claude-sonnet-4.5`; `ModelCombobox` returns that exact ID; create and edit submit it unchanged; tRPC, Worker, and controller paths forward/write it unchanged. Existing Settings flow explicitly converts selected IDs to `kilocode/${selectedModel}`, and shared model utilities define `kilocode/` as KiloClaw model namespace.

Impact: creating or editing an agent with a selected model can persist an invalid/unresolved provider reference. Existing correctly prefixed agent models also fail to appear selected in combobox.

apps/web/src/app/(app)/claw/components/AgentEditDialog.tsx
```diff
@@ -0,0 +1,316 @@
...
+type ThinkingOpt = typeof INHERIT | (typeof THINKING_OPTIONS)[number];
+type VerboseOpt = typeof INHERIT | (typeof VERBOSE_OPTIONS)[number];
+type ReasoningOpt = typeof INHERIT | (typeof REASONING_OPTIONS)[number];
+type FastModeOpt = typeof INHERIT | 'on' | 'off';
+
+// Map a controller-reported setting value (string | null) to a select option,
+// matching against the known options rather than casting; anything unrecognized
+// (including null) falls back to INHERIT.
+function toOption<T extends string>(raw: string | null, options: readonly T[]): typeof INHERIT | T {
+  return options.find(opt => opt === raw) ?? INHERIT;
```
jeanduplessis, line 48: **WARNING:** Preserve unknown non-null setting values instead of collapsing them to inheritance.

Trace: response contracts intentionally allow arbitrary `string | null` values for forward compatibility. `toOption()` maps any newer unknown value to `INHERIT`; when user saves an unrelated model or fast-mode edit, patch builder interprets that state as an explicit unset and deletes original setting.

Impact: during newer-controller/older-web version skew, saving unrelated agent edits silently removes newer thinking, verbose, or reasoning settings and changes agent behavior.

jeanduplessis (approved):
Found two concrete model-editing issues:

- Selected catalog model IDs are forwarded without required KiloClaw `kilocode/` prefix.
- Unknown forward-compatible behavior-setting values are converted into destructive unsets when saving unrelated edits.

Other reviewed paths—including capability gating, status transitions, etag handling, delete behavior, PostHog paths, and PR-body claims—look consistent. No unresolved review threads remain.

apps/web/src/app/(app)/claw/hooks/useClawModelOptions.ts
St0rmz1: Fixed: added `addKilocodeModelPrefix`/`stripKilocodeModelPrefix` in `modelSupport.ts` (mirrors the Settings flow). Create now prefixes the selected catalog id with `kilocode/` on submit; edit strips it to seed the combobox (so existing `kilocode/…` models show selected) and re-adds it on save. Fallbacks are preserved verbatim.

apps/web/src/app/(app)/claw/components/AgentEditDialog.tsx
St0rmz1: Fixed: the edit dialog now diffs each setting against its initial select value rather than the raw response. An untouched value (including an unknown forward-compat one, shown as Inherit) compares equal and emits no patch, so newer settings are no longer collapsed to INHERIT and silently unset on an unrelated save.

PR #3825, opened 2026-06-09
feat(cloud-agent): classify retryable runtime failures

eshurakov:
## Summary

- Add a shared `{ code, message, retryable }` client-error contract across Cloud Agent tRPC responses, terminal callbacks, and message-result polling.
- Treat agent, provider, wrapper, network, and timeout failures as retryable regardless of lifecycle stage, while keeping deterministic model, payment, input, metadata, and user-interruption failures non-retryable.
- Preserve typed `payment_required` and `model_missing` failures through wrapper ingest and expose optional `failureStage` so callback clients can choose their recovery strategy.
- Keep Code Review, Security Auto Analysis, Auto Fix, bot, and webhook consumer policies unchanged; the new callback fields remain additive and can be adopted separately.

## Verification

- Manual verification not performed; this is a backend-only contract and classification change.

## Visual Changes

N/A

## Reviewer Notes

- `retryable: true` permits a new logical attempt with a new `messageId`; it does not automatically replay work or make a terminal message ID reusable.
- Failures after `agent_activity` may have partial side effects. The callback's `failureStage` lets clients choose between replaying, continuing, inspecting state, or requesting user input.
- Database telemetry keeps its existing stage/code vocabulary: typed payment/model failures are normalized only for persistence reports, while callbacks and message results retain the exact classification.

### DIFF
--- a/packages/db/src/schema.ts
+++ b/packages/db/src/schema.ts
@@ -4361,6 +4361,7 @@ export type CloudAgentSessionRunFailureCode =
   | 'assistant_error'
   | 'wrapper_error_after_activity'
   | 'missing_assistant_reply'
+  | 'payment_required'
   | 'user_interrupt'
   | 'container_shutdown'
   | 'system_interrupt'
--- a/services/cloud-agent-next/src/session/wrapper-supervisor.ts
+++ b/services/cloud-agent-next/src/session/wrapper-supervisor.ts
@@ -18,6 +18,7 @@ import {
   type SessionMessageState,
   type SessionMessageStorage,
 } from './session-message-state.js';
+import type { WrapperTerminalFailureCode } from '../shared/protocol.js';
 import type { LatestAssistantMessage } from './types.js';
 import {
   clearCurrentWrapperRuntimeFailureState,
@@ -90,6 +91,7 @@ export type WrapperTerminalEvent = {
   error?: string;
   errorSource?: 'assistant';
   interruptionSource?: 'container_shutdown';
+  failureCode?: WrapperTerminalFailureCode;
   gateResult?: 'pass' | 'fail';
   messageIds?: string[];
 };
@@ -926,8 +928,16 @@ export function createWrapperSupervisor(
   }
 
   async function onTerminalEvent(params: WrapperTerminalEvent): Promise<void> {
-    const { wrapperRunId, status, error, errorSource, interruptionSource, gateResult, messageIds } =
-      params;
+    const {
+      wrapperRunId,
+      status,
+      error,
+      errorSource,
+      interruptionSource,
+      failureCode: terminalFailureCode,
+      gateResult,
+      messageIds,
+    } = params;
     const sessionId = getSessionIdForLogs();
     const state = await getWrapperRuntimeState(storage);
     if (
@@ -971,7 +981,7 @@ export function createWrapperSupervisor(
               error: error ?? 'Assistant request failed',
               completionSource: 'wrapper_failure',
               failureStage: 'agent_activity',
-              failureCode: 'assistant_error',
+              failureCode: terminalFailureCode ?? 'assistant_error',
               safeFailureMessage: classifyAssistantFailureMessage(error),
             });
             continue;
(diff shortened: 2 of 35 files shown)


services/cloud-agent-next/src/trpc-error.ts
```diff
@@ -1,33 +1,109 @@
...
+    !PublicErrorCodeSchema.safeParse(cause.error).success ||
+    typeof cause.retryable !== 'boolean'
+  ) {
+    return undefined;
+  }
+  if ('message' in cause && cause.message !== undefined && typeof cause.message !== 'string') {
+    return undefined;
+  }
+  const error = cause.error;
+  if (typeof error !== 'string') return undefined;
```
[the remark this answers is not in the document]
eshurakov: Fixed in 680204aad. The parser now retains the successful Zod parse result and returns `parsedError.data`, removing the redundant guard without adding a type cast.

apps/web/src/app/api/internal/code-review-status/[reviewId]/route.ts
```diff
@@ -189,6 +192,7 @@ function normalizePayload(raw: StatusUpdatePayload): {
     sessionId,
     cliSessionId,
     errorMessage: raw.errorMessage,
+    clientError: 'clientError' in raw ? raw.clientError : undefined,
```
Comment (alex-alecu) [inline apps/web/src/app/api/internal/code-review-status/[reviewId]/route.ts]: Should `clientError.retryable === false` feed the retry decision here? Otherwise Cloud Agent can send a non-retryable failure that the legacy terminalReason/errorMessage heuristics do not recognize, and the code-review callback will still schedule a fresh retry.

services/security-auto-analysis/src/callbacks.ts
```diff
@@ -133,6 +139,7 @@ export function mapAnalysisCallbackFailure(params: {
 
   const failureCode =
     classifyStructuredAnalysisFailure(params.failure) ??
+    classifyAnalysisFailureText(params.clientError?.message) ??
```
alex-alecu: Should the normal failed-callback path pass `clientError` into this mapper too? Right now this branch only has data for already-terminal callbacks, so first-time failures that rely on `clientError.message` still fall back to `START_CALL_AMBIGUOUS`.

services/cloud-agent-next/src/session/wrapper-supervisor.ts
```diff
@@ -971,7 +981,7 @@ export function createWrapperSupervisor(
               error: error ?? 'Assistant request failed',
               completionSource: 'wrapper_failure',
               failureStage: 'agent_activity',
-              failureCode: 'assistant_error',
+              failureCode: terminalFailureCode ?? 'assistant_error',
```
alex-alecu, line 984: Can this emit `payment_required` or `model_missing` before the constraint-dropping migration is applied? The current production check only allows `assistant_error`/`wrapper_error_after_activity` at `agent_activity`, so code-before-migration rollout would fail terminalization.

eshurakov: The rollout risk is valid because Worker and migration deployments are independent. I removed the constraint-dropping migration and retained the existing database constraint. Cloud Agent now maps payment/model classifications to existing stage-compatible codes only at the telemetry persistence boundary; callbacks and message results retain the exact typed code. The new mapping is covered for both post-dispatch and agent-activity stages.

apps/web/src/app/api/internal/code-review-status/[reviewId]/route.ts
Comment (eshurakov) [reply apps/web/src/app/api/internal/code-review-status/[reviewId]/route.ts]: Not applied as a Code Review policy change. This PR is Cloud Agent Next-only, so I removed the `clientError` plumbing from the code-review callback. Code Review continues using its existing `terminalReason`/`errorMessage` retry policy and ignores the additive callback fields.

services/security-auto-analysis/src/callbacks.ts
eshurakov: The normal and already-terminal paths should remain unchanged in this PR. I removed the partial `clientError.message` classification and downstream consumer plumbing instead of extending it, so Security Auto Analysis continues using its existing `failure`/`errorMessage` mapping.

packages/db/src/schema.ts
```diff
@@ -4357,6 +4357,7 @@ export type CloudAgentSessionRunFailureCode =
   | 'assistant_error'
   | 'wrapper_error_after_activity'
   | 'missing_assistant_reply'
+  | 'payment_required'
```
alex-alecu, line 4364: @kilocode-bot Why was this removed? What replaced this?

alex-alecu (approved):

PR #3863, opened 2026-06-09
Disallow experiment models for enterprise organizations

chrarnoldus:
I don't think the current 'allowed when provider list is uninitialized' logic makes sense. There's no way in the UI to observe this state or to go back to it after changing it. I think we would need a new UI if we want to allow experiment models for enterprise orgs.

### DIFF
--- a/apps/web/src/lib/organizations/organization-models.ts
+++ b/apps/web/src/lib/organizations/organization-models.ts
@@ -26,7 +26,7 @@ export async function getAvailableModelsForOrganization(
   }
 
   const responseData = await getEnhancedOpenRouterModels();
-  const restrictionCandidates = responseData.data;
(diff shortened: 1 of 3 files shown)


apps/web/src/lib/organizations/organization-models.ts
```diff
@@ -26,8 +26,7 @@ export async function getAvailableModelsForOrganization(
   }
 
   const responseData = await getEnhancedOpenRouterModels();
-  const experimentModels = await listAvailableExperimentModels();
-  const restrictionCandidates = responseData.data.concat(experimentModels);
+  const restrictionCandidates = responseData.data;
```
chrarnoldus:
```suggestion
  const restrictionCandidates = [...responseData.data];
```

johnnyeric (approved):

PR #3866, opened 2026-06-09
chore(ai-gateway): log client disconnects

chrarnoldus:
## Summary

- Log when an AI gateway client disconnects and aborts its request.
- Include bounded path, elapsed-time, client-request, and session correlation fields without logging request content or credentials.

## Verification

- Manual verification not run because reproducing a client disconnect requires a running gateway stack.

## Visual Changes

N/A

## Reviewer Notes

- The existing request abort signal continues to cancel the upstream provider request; this change only adds an informational log listener.


chrarnoldus:
This is probably not enough, we'll also need the following in vercel.json:

```
{
  "functions": {
    "api/openrouter/responses": {
      "supportsCancellation": true
    }
  }
}
```

https://vercel.com/docs/functions/functions-api-reference

but we can do that in a follow-up PR for make reverts easier.

marius-kilocode (approved):

PR #3868, opened 2026-06-09
feat(security-agent): add auto remediation

jeanduplessis:
## Summary
Adds Security Agent Auto Remediation, including product spec, user controls, remediation attempt tracking, Cloud Agent orchestration, and PR-facing UI.

### Why this change is needed
Security Agent could triage and analyze vulnerable dependencies, but users had no durable way to turn eligible findings into remediation PRs, track attempts, prevent duplicate work, or understand why fix actions were unavailable.

Incoming Security Agent UI changes also refactored finding rows, dialogs, and settings sections, so remediation controls need to match current Kilo design-system structure.

### How this is addressed
- Defines Auto Remediation rules in `.specs/security-agent.md`, ADR/context docs, and planning docs.
- Adds database schema, generated migration, repositories, and lifecycle state for remediation attempts.
- Adds manual and automatic remediation admission, duplicate suppression, cancellation, retry, and Worker processing.
- Connects Cloud Agent remediation execution to Security Agent callbacks and PR-opened/no-changes/failure outcomes.
- Adds manual remediation controls, live polling, PR links, unavailable-state copy, and attempt history in Security Agent findings UI.
- Adds Auto Remediation settings in refactored settings layout.

## Human Verification
- Extensive end-to-end local testing of manual and automatic remediation flows. 
- Example of Pr opened: https://github.com/Kilo-Org/security-agent-testbed/pull/2

<details>
<summary>Browser smoke</summary>

Opened `http://localhost:3000/security-agent/findings` with `agent-browser`; route loaded and displayed the GitHub installation gate for the current local account.

</details>

### Human Reviewer Flags
- Manual remediation is allowed after human review for manual-review or unknown-exploitability findings when concrete fix path exists; automatic remediation still requires stricter eligibility.
- Persistent remediation-attempt tables and generated migration need schema, retention, ownership, and GDPR soft-delete review.
- Remediation UI follows incoming Kilo design-system refactor: row actions outside detail button, extracted modal panels, shared settings sections.
- Existing PR remains draft; title remains unchanged by workflow even though scope now exceeds docs.

### Code Reviewer Agent
<details>

- Remediation admission policy lives in `@kilocode/worker-utils/security-remediation-policy`.
- Manual remediation commands flow through Security Agent routers into `security-auto-analysis`.
- Callback handling updates remediation attempts separately from analysis lifecycle.
- Finding detail polling includes active remediation attempts so open modal updates after PR creation.
- Row-level `View PR` stops row propagation and uses incoming row action slot.
- Rebase regenerated branch migration as `0162_sloppy_luke_cage.sql` after latest `origin/main` added `0161_smiling_wallop.sql`.

</details>


packages/worker-utils/src/security-remediation-policy.ts
```diff
@@ -0,0 +1,313 @@
...
+
+function isRepoInScope(params: {
+  finding: SecurityRemediationFinding;
+  config: SecurityRemediationConfig;
+  repoFullNamesInScope: string[];
+}): boolean {
+  if (params.config.repository_selection_mode === 'all') {
+    return params.repoFullNamesInScope.includes(params.finding.repo_full_name);
+  }
+  return params.repoFullNamesInScope.includes(params.finding.repo_full_name);
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 4adf8949c: simplified isRepoInScope to the single repoFullNamesInScope check and removed the unused config field from the helper signature.

.gitignore
```diff
@@ -134,3 +134,4 @@ run-milvus-test.sh
 !.env.*.example
 !.env.test
 !.envrc
+.env*
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 4adf8949c: removed the duplicate .env* ignore entry.

packages/db/src/schema.ts
```diff
@@ -4971,8 +4971,238 @@ export const security_analysis_owner_state = pgTable(
...
+      table.remediation_id,
+      table.attempt_number
+    ),
+    uniqueIndex('UQ_security_remediation_attempts_active_finding')
+      .on(table.finding_id)
+      .where(sql`${table.status} IN ('queued', 'launching', 'running')`),
+    uniqueIndex('UQ_security_remediation_attempts_active_remediation')
+      .on(table.remediation_id)
+      .where(sql`${table.status} IN ('queued', 'launching', 'running')`),
+    uniqueIndex('UQ_security_remediation_attempts_finding_fingerprint_terminal')
```
St0rmz1, line 5144: **Moderate, Functional:** This partial unique index covers `blocked` and `no_changes_needed`, but those statuses are also in `RETRYABLE_TERMINAL_STATUSES` and `canRetry` surfaces them to the UI (`apps/web/src/lib/security-agent/db/security-remediation.ts:304`).

When a user retries a `blocked` or `no_changes_needed` attempt without re-analyzing, `computeSecurityRemediationAnalysisFingerprint` returns the same fingerprint, so `admitRemediationAttempt` inserts a new `queued` attempt that collides with the still-present terminal row on `(finding_id, analysis_fingerprint)`. The transaction throws and the user gets a generic "Failed to queue remediation" error. Retry of `failed`/`cancelled` works because those are excluded from the predicate, so the feature is half-broken in a way that is easy to miss. This contradicts the spec ("Users MAY retry failed, blocked, no-changes-needed, or cancelled attempts").

Pick one source of truth: either drop `blocked` and `no_changes_needed` from this index predicate (keep the active states plus `pr_opened`), or have the retry path transition the prior terminal attempt out of the predicate before inserting, or narrow the retryable set to `failed`/`cancelled`. A test that retries a `blocked` attempt with an unchanged fingerprint would lock this down.

apps/web/src/routers/organizations/organization-security-agent-router.ts
```diff
@@ -60,6 +60,15 @@ export const organizationSecurityAgentRouter = createTRPCRouter({
   startAnalysis: organizationMemberMutationProcedure
     .input(OrganizationIdInputSchema.merge(handlers.startAnalysis.inputSchema))
     .mutation(handlers.startAnalysis.handler),
+  startRemediation: organizationBillingMutationProcedure
```
St0rmz1: **Moderate, Design:** `startRemediation`, `retryRemediation`, and `cancelRemediation` are wired to `organizationBillingMutationProcedure`, but both the plan ("Manual remediation actions ... org: `organizationMemberMutationProcedure`") and the Permissions section of `.specs/security-agent.md` ("Organization manual remediation actions MAY be available to organization members") call for member-level access. It is also inconsistent with the sibling `startAnalysis` and `triggerSync`, which are `organizationMemberMutationProcedure`.

This is more restrictive than designed, so it is not a security issue, but it silently locks regular org members out of a feature the spec promises them, and the "Organization Member Starts Remediation" scenario would not hold. If members are meant to start/retry/cancel, switch these three to `organizationMemberMutationProcedure`. If billing-gating is deliberate (remediation opens PRs, so gate it like settings/`dismissFinding`), please update the spec and plan so the source of truth matches the code.

services/security-auto-analysis/src/remediation.ts
```diff
@@ -0,0 +1,1679 @@
...
+  const ownerCondition =
+    owner.type === 'org'
+      ? eq(security_findings.owned_by_organization_id, owner.id)
+      : eq(security_findings.owned_by_user_id, owner.id);
+  const findings = await db
+    .select({ id: security_findings.id })
+    .from(security_findings)
+    .where(and(ownerCondition, eq(security_findings.status, 'open')))
+    .orderBy(asc(security_findings.created_at))
+    .limit(200);
```
St0rmz1: **Low, Performance:** The include-existing scan is capped at 200 open findings with no pagination, re-enqueue, or user-visible signal when the cap is hit, so an owner with more than 200 eligible findings gets partial coverage that reads as "all existing findings queued." Separately, each finding in this loop calls `admitRemediationAttempt`, which calls `getRuntimeConfig` again, re-fetching `agent_configs` and `platform_integrations` once per finding (an N+1 against config that does not change within the command).

Surface the cap in `result_metadata` (and log it) so truncation is visible, and hoist the single `getRuntimeConfig(owner)` out of the per-finding loop by passing it into a lower-level admit path.

services/security-auto-analysis/src/remediation.ts
```diff
@@ -0,0 +1,1679 @@
...
+  }
+  await db
+    .update(security_remediation_attempts)
+    .set({
+      cancellation_requested_at: sql`now()`,
+      cancellation_requested_by_user_id: params.request.actorUserId,
+      updated_at: sql`now()`,
+    })
+    .where(eq(security_remediation_attempts.id, attempt.id));
+  if (attempt.cloud_agent_session_id) {
```
St0rmz1, line 1667: **Low, Functional:** Cancellation only sends `interruptSession` when `cloud_agent_session_id` is set. An attempt claimed into `launching` but not yet past `prepareSession` has no session id, so this branch is skipped, only `cancellation_requested_at` is set. Since `launchAttempt` never re-checks `cancellation_requested_at`, the session then launches and runs to completion, and the `interrupted` to `cancelled` callback mapping never fires because nothing interrupted it. The user is told cancellation was requested while the agent proceeds to open a PR.

Narrow race, but worth closing: re-check `cancellation_requested_at` inside `launchAttempt` after `prepareSession` (and again before `initiate`) and short-circuit to cancelled, or issue the interrupt as soon as the session id is known.

services/security-auto-analysis/src/dispatcher.ts
```diff
@@ -57,15 +61,37 @@ export async function dispatchDueOwners(env: CloudflareEnv): Promise<{
     await env.OWNER_QUEUE.sendBatch(messages.slice(i, i + QUEUE_SEND_BATCH_LIMIT));
   }
 
+  const remediationAttemptIds = await discoverQueuedRemediationAttempts(
```
St0rmz1, line 64: **Low, Design:** This dispatches attempts that already exist in `queued` state and relies on `next_retry_at` for launch retries, but there is no reconciler that scans eligible findings to recover missed admissions. If `maybeAdmitAutoRemediationForCompletedAnalysis` throws before the attempt row is created (or marks it `QUEUE_ADMISSION_FAILED`), the callback retry hits `already-terminal` and remediation is never re-attempted automatically. The plan called for a reconciler that recovers missed `auto_policy` admissions using the shared eligibility function, which is not present here.

Either add the scan-based admission reconciler, or note in the plan/spec that auto-admission recovery is deferred for v1 and manual start is the fallback so the gap is intentional and documented.

St0rmz1 (approved):

packages/db/src/schema.ts
jeanduplessis: Fixed in 657aacedf: regenerated the branch migration with the fingerprint unique index scoped to queued/launching/running/pr_opened only, so blocked and no_changes_needed attempts can be retried with the same analysis fingerprint. Added a policy regression for manual retry of retryable terminal outcomes.

services/security-auto-analysis/src/remediation.ts
jeanduplessis: Fixed in 657aacedf: include-existing now fetches runtime config once per command, passes it into admission, scans limit + 1, logs when the cap is hit, and writes scanLimit/truncated/candidateCount into command result metadata.

apps/web/src/routers/organizations/organization-security-agent-router.ts
jeanduplessis: Fixed in 657aacedf: startRemediation, retryRemediation, and cancelRemediation now use organizationMemberMutationProcedure to match the spec and sibling Security Agent actions.

services/security-auto-analysis/src/remediation.ts
jeanduplessis: Fixed in 657aacedf: launch now re-checks cancellation after prepareSession and again before initiate. If cancellation was requested, it finalizes the attempt as cancelled and interrupts the prepared session when a Cloud Agent session id exists.

services/security-auto-analysis/src/dispatcher.ts
jeanduplessis: Addressed in 657aacedf: documented the scan-based auto-admission reconciler as deferred from v1, with manual remediation as the fallback for missed automatic admissions.

PR #3874, opened 2026-06-09
feat(auto-routing): add auto routing classifier

iscekic:
## Summary

- Adds a new `services/auto-routing` Cloudflare Worker with authenticated `/decide` and `/health` endpoints for mirrored OpenRouter gateway traffic.
- Mirrors Next.js OpenRouter gateway requests in the background, including redacted headers, session/task IDs, request path, timestamp, and original body for `/chat/completions`, `/responses`, and `/messages`.
- Implements the first classifier pass with the OpenRouter SDK, a Gemma default classifier model (`google/gemma-4-31b-it`), strict taxonomy validation, KV-configurable `classifier_model`, and a stable `decision: null` response while routing decisions remain a follow-up.
- Adds Analytics Engine metrics for classifier status, model, session ID, request shape, confidence, cost, duration, and mirrored body size, plus Worker infra/config for KV, Secrets Store, local dev, workspace wiring, and async backend auth.

## Verification

E2E tested locally.

## Visual Changes

N/A

## Reviewer Notes

- The Worker classifies mirrored traffic and records analytics, but does not change the selected model yet.
- Please review the mirrored-payload data handling, redaction, classifier taxonomy, and Analytics Engine schema carefully.
- Deployment requires `AUTO_ROUTING_CONFIG`, `AUTO_ROUTING_CLASSIFIER_METRICS`, `OPENROUTER_API_KEY`, and `INTERNAL_API_SECRET_PROD` bindings to be present.


apps/web/src/lib/ai-gateway/auto-routing-mirror.ts
```diff
@@ -0,0 +1,69 @@
...
+  const response = await fetch(`${workerUrl}/decide`, {
+    method: 'POST',
+    headers: new Headers({
+      authorization: `Bearer ${authToken}`,
+      'content-type': 'application/json',
+    }),
+    body: JSON.stringify({
+      path,
+      receivedAt: new Date().toISOString(),
+      headers: serializeHeaders(request.headers),
```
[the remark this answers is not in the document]
iscekic: Fixed in c190057ec: mirrored request headers are now passed through `redactSensitiveHeaders` before being sent to `/decide`, and the mirror test asserts the user `Authorization` value is `[REDACTED]`.

services/auto-routing/src/model-classifier.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  return {
+    cost: result.usage?.cost ?? null,
+    classification: parseClassifierOutput(extractClassifierText(result)),
+  };
+}
+
+function extractClassifierText(result: ChatResult) {
+  const content = result.choices[0]?.message.content as unknown;
```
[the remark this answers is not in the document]
iscekic: Fixed in c190057ec: removed the `as unknown` cast. The SDK content value is now assigned to an explicitly `unknown` local before the existing runtime string narrowing, which keeps the lint boundary explicit without a cast.

apps/web/src/lib/ai-gateway/auto-routing-admin-client.ts
```diff
@@ -0,0 +1,61 @@
...
+      status: 500,
+      body: { error: 'Auto routing worker is not configured' },
+    };
+  }
+
+  const response = await fetch(`${AUTO_ROUTING_WORKER_URL}${path}`, {
+    ...init,
+    headers: {
+      authorization: `Bearer ${INTERNAL_API_SECRET}`,
+      ...(init.headers as Record<string, string> | undefined),
```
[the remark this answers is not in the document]
iscekic: Fixed in 9d53cfddd: narrowed the admin worker client init type so callers can only pass plain header objects, which removes the unsafe RequestInit.headers cast/spread.

apps/web/src/lib/ai-gateway/auto-routing-admin-client.ts
```diff
@@ -0,0 +1,61 @@
...
+    },
+  });
+
+  return {
+    status: response.status,
+    body: (await response.json()) as T | ErrorBody,
```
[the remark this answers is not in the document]
iscekic: Fixed in 9d53cfddd: added Zod schemas for the auto-routing admin worker responses and parse the worker JSON before returning it, so this no longer casts unvalidated response bodies.

services/auto-routing/src/admin-classifier-analytics.ts
```diff
@@ -0,0 +1,254 @@
...
+      avgConfidence: numberValue(row.avg_confidence),
+    })),
+    classifierModelBreakdown: modelRows.map(row => ({
+      classifierModel: row.classifier_model,
+      requests: numberValue(row.requests),
+    })),
+  });
+};
+
+export const classifierAnalyticsPeriods = Object.keys(PERIODS) as AnalyticsPeriod[];
```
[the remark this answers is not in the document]
iscekic: Fixed in 9d53cfddd: removed the unused classifierAnalyticsPeriods export.

apps/web/src/app/admin/auto-routing/AutoRoutingAdminContent.tsx
```diff
@@ -0,0 +1,449 @@
...
+};
+
+async function parseAdminResponse<T extends object>(response: Response): Promise<T> {
+  const body = (await response.json()) as T | AdminApiError;
+  if (!response.ok) {
+    throw new Error(
+      'error' in body && body.error ? body.error : `Request failed: ${response.status}`
+    );
+  }
+  return body as T;
```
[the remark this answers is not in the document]
iscekic: Fixed in 9d53cfddd: the admin UI response helper now parses JSON as unknown, validates error bodies separately, and validates successful responses with the relevant Zod schema instead of casting.

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -166,6 +167,8 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
 
   // Parse body first to check model before auth (needed for anonymous access)
   const requestBodyText = await request.text();
+  const authPromise = getUserFromAuth({ adminOnly: false });
+  scheduleAutoRoutingMirror({ request, path, bodyText: requestBodyText, authContext: authPromise });
```
Comment (pandemicsyn) [inline apps/web/src/app/api/openrouter/[...path]/route.ts]: Must-fix: this mirror is scheduled before the request has passed auth, paid-model access, rate limits, org/model/provider restrictions, and data-collection checks. Because the mirror payload includes the raw body and the worker sends prompt-derived summaries to the classifier, requests that are later rejected can still leak prompt prefixes into classifier processing. Move mirroring until after the gateway has accepted the request, or pass only a normalized/sanitized classifier summary and explicitly skip every denial path.

apps/web/src/app/admin/auto-routing/AutoRoutingAdminContent.tsx
```diff
@@ -0,0 +1,475 @@
...
+            onValueChange={setSelectedModel}
+            isLoading={openRouterModelsQuery.isLoading || classifierModelQuery.isLoading}
+            error={openRouterModelsQuery.error?.message ?? undefined}
+            placeholder={classifierModelQuery.data?.defaultModel ?? 'Select classifier model'}
+            className="w-full"
+          />
+          <Button
+            type="button"
+            onClick={() => saveMutation.mutate(selectedModel)}
+            disabled={!hasModelChange || saveMutation.isPending}
```
pandemicsyn, line 395: Must-fix: if `classifierModelQuery` fails while the OpenRouter model list succeeds, `currentModel` falls back to `''`, the config-load error is hidden, and this save button becomes enabled after any selection. That lets an admin overwrite classifier configuration without knowing the current value failed to load. Surface `classifierModelQuery.error` and disable saving until the classifier model query has successfully loaded.

apps/web/src/app/admin/auto-routing/AutoRoutingAdminContent.tsx
```diff
@@ -0,0 +1,475 @@
...
+
+      {analyticsQuery.error ? (
+        <div className="border-destructive/40 bg-destructive/10 text-destructive rounded-md border px-3 py-2 text-sm">
+          {analyticsQuery.error instanceof Error
+            ? analyticsQuery.error.message
+            : 'Failed to load classifier analytics'}
+        </div>
+      ) : null}
+
+      <div className="grid gap-4 md:grid-cols-2 xl:grid-cols-4">
```
pandemicsyn: Suggested/nit: when the initial analytics query fails, the banner shows an error but the metric cards still render derived `0`/`0%` values and the tables render `No data`. That can make a load failure look like real zero telemetry. Render an error state instead when `analyticsQuery.isError && !analyticsQuery.data`, or keep prior data only when it actually exists.

apps/web/src/app/admin/auto-routing/AutoRoutingAdminContent.tsx
```diff
@@ -0,0 +1,475 @@
...
+            OpenRouter credits are reported as credits.
+          </p>
+        </div>
+        <div className="flex w-fit rounded-md border p-1">
+          {periods.map(option => (
+            <Button
+              key={option.value}
+              type="button"
+              size="sm"
+              variant={period === option.value ? 'secondary' : 'ghost'}
```
pandemicsyn, line 421: Suggested/nit: this period picker is mutually exclusive, but the selected state is only visual through the button variant. Screen readers get four unrelated buttons. Use a tabs/toggle-group control or add `aria-pressed={period === option.value}` with an accessible group label.

apps/web/src/app/admin/auto-routing/AutoRoutingAdminContent.tsx
```diff
@@ -0,0 +1,475 @@
...
+          <p className="text-muted-foreground text-sm">
+            Classifier configuration and routing telemetry.
+          </p>
+        </div>
+        <Button
+          type="button"
+          variant="outline"
+          onClick={() => {
+            void classifierModelQuery.refetch();
+            void analyticsQuery.refetch();
```
pandemicsyn, line 366: Suggested/nit: the page-level Refresh action refetches classifier config and analytics, but not `openRouterModelsQuery`, so a transient model-list failure cannot be recovered from this visible refresh button. Also refetch the model list and include its fetching state in the button disabled/spinner state.

packages/auto-routing-contracts/src/index.ts
```diff
@@ -0,0 +1,172 @@
+import * as z from 'zod';
+
+export const MirrorPathSchema = z.enum(['/chat/completions', '/responses', '/messages']);
+export type MirrorPath = z.infer<typeof MirrorPathSchema>;
+
+export const MirrorPayloadSchema = z.object({
+  path: MirrorPathSchema,
+  receivedAt: z.string().min(1),
```
pandemicsyn: Suggested/nit: `receivedAt` accepts any non-empty string even though the sender emits an ISO timestamp. Tightening this cross-service contract would catch malformed payloads earlier. Use `z.string().datetime()` or an equivalent strict timestamp schema.

services/auto-routing/src/admin-classifier-analytics.ts
```diff
@@ -0,0 +1,255 @@
...
+      body: sql,
+    }
+  );
+
+  if (!response.ok) {
+    const errorText = await response.text();
+    throw new Error(`Analytics Engine query failed (${response.status}): ${errorText}`);
+  }
+
+  const result = (await response.json()) as AnalyticsEngineResponse<T>;
```
pandemicsyn: Suggested/nit: this trusts the external Analytics Engine JSON shape via a cast and returns `result.data` without runtime validation. If the API shape changes or returns malformed JSON, later mapping code fails in less obvious ways. Validate `{ data: z.array(rowSchema) }` per query, or at least guard that `result.data` is an array before returning it.

pandemicsyn (changes requested):
Couple inline from the robot reviewing it locally, but other than that lgtm.

pandemicsyn (approved):

apps/web/src/app/admin/auto-routing/AutoRoutingAdminContent.tsx
iscekic: Fixed in 7c7c85723: initial analytics load failures now render an error state instead of zero-valued metric cards and empty tables. Prior data still remains visible on later refetch errors.

apps/web/src/app/admin/auto-routing/AutoRoutingAdminContent.tsx
iscekic: Fixed in 7c7c85723: the period control now has an accessible group label and exposes the selected state with aria-pressed.

apps/web/src/app/admin/auto-routing/AutoRoutingAdminContent.tsx
iscekic: Fixed in 7c7c85723: the page Refresh action now also refetches OpenRouter models and reflects that query in the button disabled/spinner state.

services/auto-routing/src/admin-classifier-analytics.ts
iscekic: Fixed in 7c7c85723: Analytics Engine responses are validated with Zod row schemas at the fetch boundary, and malformed responses now fail explicitly.

apps/web/src/app/api/openrouter/[...path]/route.ts
Comment (iscekic) [reply apps/web/src/app/api/openrouter/[...path]/route.ts]: Fixed in 7c7c85723: moved the auto-routing mirror scheduling until after local gateway validation, auth, rate limit, provider/model restrictions, data-collection checks, and abuse decision gates have accepted the request.

apps/web/src/app/admin/auto-routing/AutoRoutingAdminContent.tsx
iscekic: Fixed in 7c7c85723: classifier model load errors now surface in the model control, and saving stays disabled until the current classifier model query has successfully loaded.

packages/auto-routing-contracts/src/index.ts
iscekic: Fixed in 7c7c85723: receivedAt now uses z.string().datetime(), with a regression assertion for malformed timestamps.

services/auto-routing/src/model-classifier.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  return {
+    cost: result.usage?.cost ?? null,
+    classifierModel,
+    classification: parseClassifierOutput(extractClassifierText(result)),
+  };
+}
+
+function extractClassifierText(result: ChatResult) {
+  const content: unknown = result.choices[0]?.message.content;
```
jeanduplessis, line 89: **WARNING:** Paid classifier responses lose cost and model attribution when output validation fails.

**Trace:** OpenRouter can return billed `usage.cost` and response content that is empty, invalid JSON, or taxonomy-invalid. `parseClassifierOutput` then throws before `cost` and `classifierModel` reach `decideHandler`; catch path records model as `unknown` and cost as zero.

**Impact:** Actual classifier spend is underreported and failures cannot be attributed to configured model. Preserve upstream cost/model before parsing, then include them in classifier-error metrics.

apps/web/src/app/admin/auto-routing/AutoRoutingAdminContent.tsx
```diff
@@ -0,0 +1,499 @@
...
+  return new Intl.NumberFormat('en-US', { maximumFractionDigits }).format(value);
+}
+
+function formatPercent(value: number) {
+  return `${formatDecimal(value * 100, 1)}%`;
+}
+
+function formatCredits(value: number) {
+  return new Intl.NumberFormat('en-US', {
+    minimumFractionDigits: value > 0 && value < 1 ? 4 : 2,
```
jeanduplessis, line 114: **WARNING:** Four-decimal formatting displays normal low-volume classifier spend as zero.

**Trace:** Classifier fixtures use costs such as `0.00000123`, and analytics fixture totals `0.0000123`. Limiting every value below one to four decimal places renders that total as `0.0000`.

**Impact:** Dashboard hides nonzero spend until totals cross rounding threshold. Use adaptive precision or significant digits for sub-credit values.

services/auto-routing/src/openrouter.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    const client = await createOpenRouterClient({
+      OPENROUTER_API_KEY: {
+        get: async () => 'sk-or-test',
+      },
+    } satisfies Pick<Env, 'OPENROUTER_API_KEY'>);
+
+    expect(client).toHaveProperty('chat');
+
+    const options = (client as OpenRouterClientOptions)._options;
+    expect(options.apiKey).toBe('sk-or-test');
```
jeanduplessis: **SUGGESTION:** Test asserts private OpenRouter SDK state through `_options`.

**Trace:** `_options` is not part of public `OpenRouter` API; local cast invents its shape rather than verifying observable request behavior.

**Impact:** Compatible SDK upgrades can break test solely because private storage changes. Prefer exercising request configuration through supported client behavior.

jeanduplessis (approved):
Reviewed final combined diff at `7c7c857`. Latest commit fixes pre-acceptance mirroring through an existing review thread, so that finding is deduped. Remaining follow-ups: classifier parse failures underreport billed cost/model, low-volume costs render as zero, and one SDK test depends on private state. Submitting approval per requested disposition.

services/auto-routing/src/openrouter.test.ts
iscekic: Fixed in 72c7f70bb: the OpenRouter client test now mocks the SDK constructor and asserts the supported constructor options instead of reading the client private `_options` field.

apps/web/src/app/admin/auto-routing/AutoRoutingAdminContent.tsx
iscekic: Fixed in 72c7f70bb: tiny positive credit values now use significant-digit formatting so low-volume classifier spend no longer renders as `0.0000`.

services/auto-routing/src/model-classifier.ts
iscekic: Fixed in 72c7f70bb: `classifyWithOpenRouter` now captures the OpenRouter `usage.cost` and classifier model before output parsing, wraps validation failures in `ClassifierRunError`, and the decide handler records that metadata for `classifier_error` analytics.

services/auto-routing/src/decide.ts
```diff
@@ -0,0 +1,100 @@
...
+
+function getClassifierFailureMetadata(error: unknown): {
+  cost?: number | null;
+  classifierModel?: string;
+} {
+  if (!error || typeof error !== 'object') {
+    return {};
+  }
+
+  const maybeMetadata = error as { cost?: unknown; classifierModel?: unknown };
```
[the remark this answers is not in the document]
iscekic: Fixed in f0da9df02: `getClassifierFailureMetadata` now narrows with `instanceof ClassifierRunError` and returns attribution only for that error type. The worker test mock preserves the real exported class so the existing classifier-error attribution regression still exercises the same path.

PR #3902, opened 2026-06-10
fix(models): serve discounted Qwen models through OpenRouter

chrarnoldus:
## Summary

- Serve Qwen3.7 Max and Qwen3.7 Plus through their standard OpenRouter model entries; the direct Alibaba integration is no longer used for these models.
- Define Kilo's authoritative, tier-aware prices for both models in `custom-pricing.ts`, including cache pricing and the Qwen3.7 Plus long-context tier.
- Show the custom prices and human-readable discount labels in model lists, and calculate billed usage from token counts while retaining OpenRouter's reported cost in `market_cost` metadata.

## Verification

- Verified focused custom-pricing tests for both Qwen3.7 Max and Qwen3.7 Plus, including model-list prices, discount labels, cache pricing, and usage pricing tiers.

## Visual Changes

N/A

## Reviewer Notes

The `percentage` field is display-only. Billing uses the explicit pricing tables and does not derive charges from OpenRouter's reported market cost.


apps/web/src/lib/ai-gateway/custom-pricing.ts
```diff
@@ -0,0 +1,118 @@
...
+
+const TOKENS_256K = 256 * 1024;
+
+export type CustomPricing = {
+  pricing: PricingTiers;
+  /** Human-readable discount shown in the model name; never used in calculations. */
+  percentage?: number;
+};
+
+export const customPricingByModelId: Record<string, CustomPricing> = {
```
marius-kilocode, line 22: 2 shouts of my bot here:

- the long-context tier starts at 256 * 1024, and the selector uses <=, so exactly 262,144 tokens gets the 3× price. The description says the surcharge applies when input exceeds 256K, implying the tier should start at 262,145. The deleted test covered this boundary, but the replacement does not.

- applyCustomPricingToModel() always exposes tier 0 pricing. Qwen Plus is therefore advertised at $0.32/$1.28 per million while billing switches to $0.96/$3.84 for long contexts. Any client-side estimate will understate long-context cost by 3×.

apps/web/src/lib/ai-gateway/custom-pricing.ts
```diff
@@ -0,0 +1,118 @@
...
+    customPricing.percentage === undefined ? '' : ` (${customPricing.percentage}% off)`;
+
+  return {
+    ...model,
+    name: model.name + discountSuffix,
+    pricing: applyPricing(model.pricing, customPricing.pricing[0].pricing),
+  };
+}
+
+export function calculateCustomCost_mUsd(
```
marius-kilocode, line 101: Bot: silently clamps negative uncached tokens to zero, whereas the old path emitted a Sentry warning. That could conceal malformed usage and undercharge.

marius-kilocode (approved):

apps/web/src/lib/ai-gateway/custom-pricing.ts
chrarnoldus: Confirmed that the long-context tier intentionally starts at exactly 262,144 input tokens. Added an explicit code comment and regression assertions for 262,143 (base tier) versus 262,144 (expensive tier). The model-list path continues to expose tier 0 pricing; that behavior predates this change and is intentionally left unchanged here.

apps/web/src/lib/ai-gateway/custom-pricing.ts
chrarnoldus: Added a Sentry error event when cache counters would produce a negative uncached-token count. The existing clamp-to-zero billing behavior remains unchanged, and the regression test now verifies both the event and resulting cache-only charge.

PR #3905, opened 2026-06-10
fix(web): include Vercel-only provider models

chrarnoldus:
## Summary

- Include public Kilo models in provider snapshots when they are available only through Vercel AI Gateway.
- Resolve each public model through the existing Vercel model ID mapping, then map Vercel endpoint provider tags back to OpenRouter provider slugs.
- Preserve explicitly configured inference-provider associations while deduplicating providers.

## Verification

- Not manually tested; this sync depends on live OpenRouter and Vercel AI Gateway metadata endpoints.

## Visual Changes

N/A

## Reviewer Notes

- Automated validation was intentionally deferred until after the initial push. A focused validation attempt triggered dependency installation and timed out during the repository prepare hook under Node 22; CI should run with the required Node 24 runtime.


chrarnoldus:
Superseded by a corrected implementation that augments OpenRouter models with providers discovered through Vercel AI Gateway metadata.

PR #3918, opened 2026-06-10
fix(code-review): log repo size

alex-alecu:
## Summary
This adds a safe size check before a cloud code review starts. When the review session is ready, the logs now show the review id, both session ids, and the repo size if the provider gave one. If the size check fails, the review still starts and logs `null`. Retry attempts keep the same size value so later logs can be matched.

## Verification
N/A - this is a backend log-only change with no screen to click through.

## Visual Changes
N/A

## Reviewer Notes
The size is the value GitHub or GitLab reports for stored repo data. It is not the exact amount cloned over the network.


apps/web/src/lib/integrations/platforms/github/adapter.ts
```diff
@@ -711,6 +711,22 @@ export async function fetchGitHubRepositoryDefaultBranch(params: {
...
+}): Promise<string | null> {
+  const { token, owner, repo } = params;
+  const octokit = new Octokit({ auth: token });
+  const { data } = await octokit.repos.get({ owner, repo });
+
+  if (typeof data.size !== 'number') {
+    return null;
+  }
+
+  return `${Math.round(data.size / 1024)} MB`;
```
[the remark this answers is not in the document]
alex-alecu: Fixed in 15c6274.

apps/web/src/lib/integrations/platforms/gitlab/adapter.ts
```diff
@@ -1850,6 +1850,53 @@ export async function getGitLabProject(
...
+    });
+    throw new Error(`GitLab project repository size fetch failed: ${response.status}`);
+  }
+
+  const repositorySizeBytes = readGitLabRepositorySizeBytes(await response.json());
+  if (repositorySizeBytes === null) {
+    return null;
+  }
+
+  return `${Math.round(repositorySizeBytes / (1024 * 1024))} MB`;
```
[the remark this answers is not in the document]
alex-alecu: Fixed in b0e95d9.

apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
```diff
@@ -175,6 +180,27 @@ export async function prepareReviewPayload(
         githubToken = installationToken;
         const [repoOwner, repoName] = review.repo_full_name.split('/');
 
+        if (repoOwner && repoName) {
```
RSO: When is this false?

apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
```diff
@@ -548,6 +583,30 @@ export async function prepareReviewPayload(
...
+    return repositorySize;
+  } catch (error) {
+    warnExceptInTest('[prepareReviewPayload] Repository size lookup failed; continuing', {
+      platform: params.platform,
+      repoFullName: params.repoFullName,
+      error: getReviewInstructionsFetchErrorMetadata(error),
+    });
+    return null;
+  }
+}
```
RSO: This abstraction feels a bit pointless to me, it seems weird to create a method just to wrap another method with a log/try-catch. Do we even need the try-catch? The GitLab method feels pretty safe already

RSO (approved):

apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
alex-alecu: Done in 66623fd.

apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
alex-alecu: Done in 713679f. I kept the safety catch inline so a size lookup cannot stop a review.

PR #3922, opened 2026-06-10
fix(organizations): enforce invite email on acceptance

RSO:
## Summary
- Require the signed-in user's normalized email to match the email stored on an organization invitation before accepting it.
- Leave mismatched invitations pending and avoid creating memberships when an invite code is used by the wrong account.
- Add regression coverage for mismatched users and normalized Gmail/Googlemail invite matching.

## Verification
Ran local verification by inviting `remon@kilocode.ai` to an organization and making sure a user with another email wasn't allowed to accept the invite, but `remon@kilocode.ai` was.

## Reviewer Notes
- The acceptance check uses `normalized_email` when available and falls back to normalizing `google_user_email` for legacy users.

### DIFF
--- a/apps/web/src/lib/organizations/organizations.test.ts
+++ b/apps/web/src/lib/organizations/organizations.test.ts
@@ -1478,6 +1478,63 @@ describe('Organizations', () => {
         }
       });
 
+      test('should reject invitation when accepting user email does not match invite email', async () => {
+        const owner = await insertTestUser();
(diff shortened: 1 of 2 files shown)


pandemicsyn (approved):

PradyumnTiwareNexus:
Thank you for the quick fix and the detailed explanation. The changes look good from my review. I appreciate the prompt response.

PR #3935, opened 2026-06-10
fix(security-agent): complete async command lifecycle

jeanduplessis:
## Summary
Security Agent async commands now settle through command-specific UI polling and configurable command maintenance, so queue admission is no longer treated as completed work.

### Why this change is needed
Security Agent actions can return after durable queue admission while Worker consumers finish later. The remaining UI hardening still had broad terminal invalidation, partly hard-coded queued copy, missing provider-helper coverage, and fixed command retention. A late `.env*` ignore rule also made env template exceptions ineffective.

### How this is addressed
- Track recovered and newly accepted commands through helpers that distinguish active commands, terminal commands, and one-time success callbacks.
- Scope terminal invalidation by command type so sync refreshes permission and repository freshness while dismissal and analysis refresh only finding-derived views.
- Consolidate Security Agent queued/admission copy across sync, dismissal, analysis, initial sync, and backlog warnings.
- Keep dismissal UI in queued state until command terminal success or no-op, then close via terminal success callback.
- Remove dormant repository refresh props from Security Agent config while keeping targeted GitHub permission refresh in layout.
- Make Security Agent command accepted/running timeouts and retention configurable through Worker vars while preserving current defaults.
- Move `.env*` ignore before explicit env-file exceptions so examples, `.env.test`, and `.envrc` remain visible to git.

## Human Verification
- `pnpm --filter web test -- src/components/security-agent/SecurityAgentContext.test.ts src/components/security-agent/security-agent-command-copy.test.ts src/components/security-agent/security-agent-command-invalidation.test.ts src/components/security-agent/manual-analysis-admission-copy.test.ts`
- `pnpm --filter cloudflare-security-auto-analysis test -- src/command-lifecycle-config.test.ts`
- `npx -y react-doctor@latest . --verbose --diff` from `apps/web` reported no issues.
- `git check-ignore -v .env.example .env.test .env.foo.example .env.foo` confirmed env examples are unignored and stray `.env`-prefixed files remain ignored.

### Human Reviewer Flags
- Terminal invalidation is intentionally scoped by command type; sync remains the only terminal command that refreshes GitHub permission and repository freshness.
- Command lifecycle retention is now configurable in Worker vars, with the current 30-day default retained.
- `.plans/security-agent-async-command-lifecycle.md` is ignored by git in this worktree; local status was updated for working context but is not part of this PR.

### Code Reviewer Agent
<details>

- Provider helper tests cover reload recovery, active-action disabling, terminal settlement filtering, and dismissal success callback timing.
- Shared copy tests guard queued/admission terminology and dismissal terminal copy separately.
- Worker config tests cover defaults, positive integer overrides, and invalid override fallback.

</details>


apps/web/src/components/security-agent/security-agent-command-invalidation.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const dismissalScopes = [
+  'findings',
+  'findingDetails',
+  'stats',
+  'dashboardStats',
+  'autoDismissEligible',
+] as const satisfies readonly SecurityAgentInvalidationScope[];
+
+const analysisScopes = [
```
St0rmz1, line 36: **Low, Functional:** `analysisScopes` omits `autoDismissEligible`, but a completed manual analysis is exactly what produces the triage confidence that drives auto-dismiss eligibility (`getAutoDismissEligible` -> `countEligibleForAutoDismiss`, returning `byConfidence`). Both `syncScopes` and `dismissalScopes` invalidate `autoDismissEligible`; only analysis drops it. After a `start_analysis` command settles with auto-dismiss enabled, the eligible count/panel can stay stale until the next sync or a full reload, so a user may not see findings that just became eligible. Suggest adding `'autoDismissEligible'` to `analysisScopes`, or a short comment confirming analysis output never changes eligibility.

apps/web/src/components/security-agent/SecurityAgentContext.tsx
```diff
@@ -573,7 +749,13 @@ function useSecurityAgentProviderValue(
         toast.success('Findings deleted', {
           description: `${data.deletedCount} findings were permanently deleted`,
         });
-        void queryClient.invalidateQueries();
+        invalidateSecurityAgentQueryScopes([
+          'findings',
+          'stats',
+          'dashboardStats',
+          'orphanedRepositories',
```
St0rmz1: **Low, Functional:** Both `deleteFindings` `onSuccess` handlers invalidate `findings, stats, dashboardStats, orphanedRepositories, autoDismissEligible` but drop `findingDetails`, which the previous blanket `queryClient.invalidateQueries()` covered. If a finding detail view is open or cached for a finding that gets bulk-deleted, its `getFinding` data stays stale until refetched on its own. Low impact since the next fetch would 404, but it is a behavior change from the prior broad invalidation. Suggest adding `'findingDetails'` for parity unless the detail view is guaranteed closed during bulk delete. (Same applies to the personal handler below.)

St0rmz1 (approved):

apps/web/src/components/security-agent/security-agent-command-invalidation.ts
jeanduplessis: Valid. Manual analysis can change triage confidence and therefore auto-dismiss eligibility. Added autoDismissEligible to start_analysis invalidation scopes and covered it in the scope test. Commit: afc166846.

apps/web/src/components/security-agent/SecurityAgentContext.tsx
jeanduplessis: Valid. Bulk delete should invalidate cached finding details too. Added a shared deletedSecurityAgentFindingsScopes list with findingDetails and reused it in org and personal delete handlers, with test coverage. Commit: afc166846.

PR #3937, opened 2026-06-10
fix(cloud-agent): harden workspace bootstrap

eshurakov:
### Why

Large repositories can remain healthy while cloning or checking out for longer than the wrapper's previous two-minute wall-clock limit. When those operations were interrupted, the remaining `.git` directory could also make a later attempt reuse an incomplete workspace, turning one timeout into repeated setup failures.

### What was done

- Replace fixed long-command timeouts with a two-minute output-inactivity watchdog and a five-minute hard limit for Git and setup commands.
- Bound the complete workspace phase to eight minutes and propagate cancellation through Git, setup, snapshot restore, and patch application, preserving headroom inside the existing ten-minute wrapper readiness budget.
- Mark bootstrap attempts as pending and write a Git completion marker only after branch preparation, session restore, and setup commands finish; incomplete new workspaces are removed and cloned again.
- Report sanitized Git percentages and generic setup activity without exposing credentials or setup output, then advance the status to `Starting Kilo...` while the runtime starts.
- Bound captured subprocess output and failed-workspace cleanup so noisy or interrupted commands cannot grow memory or cleanup time without limit.

### High-level architecture

```mermaid
sequenceDiagram
  participant Orchestrator
  participant Wrapper
  participant Workspace
  participant Kilo
  Orchestrator->>Wrapper: POST /session/ready (10-minute outer budget)
  Wrapper->>Workspace: Prepare repository and session (8-minute shared budget)
  alt Workspace is complete
    Workspace-->>Wrapper: Reuse and refresh credentials
  else Workspace is incomplete or cold
    Wrapper->>Workspace: Mark pending, remove stale state, clone, restore, and run setup
    Wrapper->>Workspace: Write completion marker
  end
  Wrapper->>Kilo: Start runtime
  Wrapper-->>Orchestrator: Session ready
```

### Architecture decision

**Decision:** Keep bootstrap lifecycle policy in the wrapper and combine activity-aware command watchdogs with a shared workspace deadline and explicit completion markers.

**Context:** A single elapsed timeout could not distinguish a slow, active checkout from a stalled process, while `.git` existence alone could not distinguish a usable workspace from an interrupted clone.

**Rationale:** The wrapper owns the subprocesses and persisted workspace state, so it can observe output, cancel child processes, and write completion state at the point where bootstrap actually succeeds. Layered two-minute inactivity, five-minute command, eight-minute workspace, and ten-minute readiness limits keep each boundary finite without colliding with startup cleanup.

**Alternatives considered:**

- **Only increase the old Git timeout.** This would allow larger repositories but would still wait too long on silent hangs and reuse interrupted workspaces.
- **Continue treating any `.git` directory as warm.** This avoids recloning but preserves the failure mode where partial checkouts are mistaken for complete workspaces.

**Consequences:** Active long-running operations receive more time and interrupted new workspaces recover deterministically. Silent commands can still time out after two minutes, markerless legacy workspaces use a compatibility heuristic, and cleanup remains best-effort so the original setup failure is preserved.

## Verification

- [ ] Verified locally

## Visual Changes

Screenshot 2026-06-10 at 15 02 36

## Reviewer Notes

- Review the nested timeout policy and cancellation propagation through repository preparation and snapshot restore.
- Markerless legacy workspaces are migrated when Kilo auth exists and Git validates the worktree. A narrowly interrupted pre-marker workspace can still satisfy that heuristic; this is accepted because sessions are predominantly new.
- The separate App Builder `Unauthorized: Invalid token` clone failure is intentionally not addressed by this PR.

### DIFF
--- a/services/cloud-agent-next/wrapper/src/session-bootstrap.ts
+++ b/services/cloud-agent-next/wrapper/src/session-bootstrap.ts
@@ -10,6 +10,7 @@ import { buildCloudAgentRules } from '../../src/shared/cloud-agent-rules.js';
 import {
   createSafeProcessDiagnostic,
   git,
+  isTimeoutTermination,
   logToFile,
   runProcess,
   type ExecResult,
@@ -21,6 +22,10 @@ import { WrapperBootstrapError, workspaceBootstrapError } from './bootstrap-erro
 
 const LONG_COMMAND_INACTIVITY_TIMEOUT_MS = 120_000;
 const LONG_COMMAND_HARD_TIMEOUT_MS = 300_000;
+// Setup commands may legitimately stay silent for minutes (piped tools often
+// buffer), unlike git commands which run with --progress, so they get a more
+// lenient silence watchdog.
+const SETUP_COMMAND_INACTIVITY_TIMEOUT_MS = 4 * 60_000;
 const WORKSPACE_PREPARATION_TIMEOUT_MS = 8 * 60_000;
 const WORKSPACE_CLEANUP_TIMEOUT_MS = 60_000;
 const SHORT_GIT_COMMAND_TIMEOUT_MS = 120_000;
@@ -77,11 +82,7 @@ const GIT_FAILURE_PATTERNS = [
 ] as const;
 
 function classifyGitFailure(result: ExecResult, operation: 'clone' | 'checkout') {
-  if (
-    result.terminationReason === 'timeout' ||
-    result.terminationReason === 'inactivity_timeout' ||
-    result.terminationReason === 'hard_timeout'
-  ) {
+  if (isTimeoutTermination(result)) {
     return operation === 'clone' ? 'git_clone_timeout' : 'git_checkout_timeout';
   }
   const output = `${result.stderr}\n${result.stdout}`;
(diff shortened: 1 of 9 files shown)


services/cloud-agent-next/wrapper/src/session-bootstrap.ts
```diff
@@ -71,21 +131,68 @@ async function exists(filePath: string): Promise<boolean> {
...
+
+async function isCompleteGitWorkspace(
+  request: WrapperSessionReadyRequest,
+  runGit: GitRunner
+): Promise<boolean> {
+  const { workspacePath, sessionHome } = request.workspace;
+  if (!(await exists(path.join(workspacePath, '.git')))) return false;
+  if (await exists(gitBootstrapMarkerPath(workspacePath))) return true;
+  if (await exists(bootstrapPendingMarkerPath(sessionHome))) return false;
+  if (!(await exists(sessionAuthFilePath(sessionHome)))) return false;
```
alex-alecu: Can this require evidence written after restore/setup rather than only `auth.json`? The old flow writes auth before both potentially long steps, so a legacy attempt interrupted during restore or setup is migrated as complete and will skip recovery.

eshurakov: Fixed in 9c0f319be. Markerless workspaces are now always treated as incomplete and rebuilt; only `kilo-bootstrap-complete`, written after restore and setup commands finish, qualifies a workspace for the warm path.

alex-alecu (approved):

PR #3939, opened 2026-06-10
Cloud Agent - Preserve Structured Failure Context

eshurakov:
## Summary

- Preserve structured Cloud Agent failure codes, workspace subtypes, retryability, and safe messages across wrapper, queue, callback, result, and reporting boundaries.
- Correct provider, shutdown, workspace bootstrap, and retry-exhaustion classification while preventing raw process output or credentials from crossing service boundaries.
- Update callback consumers to prefer validated structured failures with legacy fallback during rolling deployments.

## Verification

N/A

## Visual Changes

N/A

## Reviewer Notes

The shared failure contract is strict for producers and tolerant for callback consumers so mixed-version deployments retain legacy `errorMessage` compatibility.


services/cloud-agent-next/src/session/agent-runtime.ts
```diff
@@ -38,7 +38,7 @@ import {
   type WrapperLease,
 } from './wrapper-runtime-state.js';
 
-export const WRAPPER_NO_OUTPUT_TIMEOUT_MS = 5 * 60 * 1000;
+export const WRAPPER_NO_OUTPUT_TIMEOUT_MS = 330_000;
```
RSO, line 42: This feels arbitrary? Why the extra 30k ms?

RSO (approved):
This was a chonker to review. Seems like a lot of mangling of error types, but overall the right side of the diff makes more sense than the left side :D

services/cloud-agent-next/src/session/agent-runtime.ts
eshurakov: @RSO I added 30 sec, because our gateway has 5 min response timeout for upstream provider. And cloud agent has 5 min of inactivity timeout.
They were racing and it is not often clear why cloud agent stalled. By changing timeout I'm hoping that we will get a better error when it is actually an issue with an upstream provider.

PR #3943, opened 2026-06-10
fix(model-selector): show accurate free and training indicators

chrarnoldus:
## Summary

Model selectors now present Free and prompt-training indicators as independent model properties across the web and mobile apps.

- A model displays the Free badge only when its `isFree` metadata is `true`.
- A model displays the book indicator only when `mayTrainOnYourPrompts` is `true`, including paid models that may train on prompts.
- Web and mobile selector option mappings retain both fields so every affected model picker applies the same rules.
- Existing indicator styling, sizing, accessibility behavior, and upstream model metadata construction remain unchanged.

## Verification

- [ ] Manual UI verification was not performed because local app services were not started.

## Visual Changes

No screenshots captured. The existing visuals are unchanged; the indicators now appear for the correct models.

## Reviewer Notes

The expected combinations are:

| `isFree` | `mayTrainOnYourPrompts` | Indicators |
|---|---|---|
| `true` | `true` | Free badge and book |
| `true` | `false` or absent | Free badge only |
| `false` or absent | `true` | Book only |
| `false` or absent | `false` or absent | Neither |

Formatting, web and mobile typechecks, changed-file lint, and the mobile test suite passed. Targeted web Jest tests could not run because the local Postgres test database was unavailable.


chrarnoldus:
These "Stealth" models now also show the training indicator rather than just free models.

jeanduplessis (approved):

PR #3957, opened 2026-06-11
refactor(cloud-agent): remove legacy runtime

eshurakov:
## Summary

- Remove the legacy Cloud Agent worker, web routers, clients, UI surfaces, workspace entries, CI jobs, and deployment configuration.
- Make Code Review orchestration Cloud Agent Next-only while preserving cancellation, continuation, callback completion, and fresh-session recovery.
- Preserve historical V1 session and review access through PostgreSQL/R2, including local deletion and read-only historical logs without runtime polling.

## Verification

- Not manually verified in the browser because no local development services were running.

## Visual Changes

- No screenshots. Nonterminal historical V1 code reviews now render persisted session logs with a `Historical` status instead of waiting for live events.

## Reviewer Notes

- This PR removes repository deployment inputs but does not delete deployed legacy Cloudflare resources; those must be decommissioned separately after rollout.
- Historical PostgreSQL/R2 data and `agent_version` metadata are intentionally retained and never select a runtime backend.


apps/web/src/lib/code-reviews/dispatch/dispatch-pending-reviews.ts
```diff
@@ -540,7 +538,7 @@ async function dispatchReservedReview(reservation: ReservedReview, owner: Owner)
   try {
     await db
       .update(cloud_agent_code_reviews)
-      .set({ agent_version: agentVersion })
+      .set({ agent_version: 'v2' })
```
alex-alecu, line 569: Could this mark a review as V2 even when an older code-review worker handles the preceding request? During a web-first rollout, omitting agentVersion makes that worker select V1, so streaming and cancellation can target the wrong backend.

services/code-review-infra/src/code-review-orchestrator.ts
```diff
@@ -1153,47 +1023,12 @@ export class CodeReviewOrchestrator extends DurableObject<Env> {
...
-    await this.loadState();
-
-    if (!this.state) {
-      return { events: [] };
-    }
-
-    return {
-      events: this.state.events || [],
-    };
+    await this.getCloudAgentNextClient().interruptSession(
```
alex-alecu, line 1040: What happens if an already-running V1 Durable Object is cancelled during this rollout? Its persisted session ID belongs to the legacy worker, but this now always calls Cloud Agent Next, so the legacy run can keep posting after the review is marked cancelled.

pnpm-lock.yaml
```diff
@@ -1463,70 +1463,6 @@ importers:
...
-        specifier: 'catalog:'
-        version: 5.9.3
-      vitest:
-        specifier: 'catalog:'
-        version: 4.1.6(@opentelemetry/api@1.9.1)(@types/node@24.12.4)(@vitest/coverage-v8@4.1.6)(@vitest/ui@4.1.6)(esbuild@0.27.4)(jiti@2.7.0)(terser@5.46.0)(tsx@4.21.0)(yaml@2.8.4)
-      wrangler:
-        specifier: 'catalog:'
-        version: 4.98.0(@cloudflare/workers-types@4.20260605.1)(bufferutil@4.1.0)(utf-8-validate@6.0.6)
-
   services/cloud-agent-next:
```
alex-alecu, line 1668: Can we also remove event-source-polyfill from apps/web/package.json and regenerate the lockfile? Its only application import was in the deleted cloud-agent-client.ts, and knip now reports the dependency as unused.

apps/web/src/lib/code-reviews/dispatch/dispatch-pending-reviews.ts
eshurakov: Resolved by current branch state. New code reviews are created with `agent_version: v2` in `createCodeReview`, and the worker now ignores obsolete `agentVersion` request input, so there is no web-first V1 selection path left.

services/code-review-infra/src/code-review-orchestrator.ts
eshurakov: No app-layer compatibility path is needed now: there are no live V1 Cloud Agent sessions or runs left in Cloudflare, and the remaining old Cloudflare resources are being removed through infra cleanup. This cancellation path can stay cloud-agent-next only.

pnpm-lock.yaml
eshurakov: Fixed in `b9faf5fee`: removed `event-source-polyfill` from `apps/web/package.json` and removed its importer/package/snapshot entries from `pnpm-lock.yaml`. `eventsource-parser` remains because it is still used.

apps/web/src/routers/cli-sessions-router.ts
```diff
@@ -478,13 +476,7 @@ export const cliSessionsRouter = createTRPCRouter({
...
     const { session_id } = input;
 
-    const session = await getSessionWithExistsCheck(session_id, ctx.user.id);
-
-    if (session.cloud_agent_session_id) {
-      const authToken = generateApiToken(ctx.user);
-      const cloudAgentClient = createCloudAgentClient(authToken);
-      await cloudAgentClient.deleteSession(session.cloud_agent_session_id);
-    }
+    await getSessionWithExistsCheck(session_id, ctx.user.id);
```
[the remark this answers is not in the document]
eshurakov: Closing this without reintroducing the legacy client path. There are no live V1 Cloud Agent sessions or runs left in Cloudflare, and the remaining old Cloudflare resources are being removed through infra cleanup, so this delete mutation does not need to tear down legacy runtime state.

alex-alecu (approved):
LGTM ✅

Minor duplication: 

`checkEligibility` / repo-listing procedures were migrated into the two cloud-agent-next routers, duplicating identical `MIN_BALANCE_DOLLARS = 1 + eligibility` logic in personal and org routers => `organization-cloud-agent-next-router.ts / cloud-agent-next-router.ts | ~393 / ~568`

PR #3964, opened 2026-06-11
perf(auto-routing): slim classifier prompt input

iscekic:
## Summary
Slims the worker-to-classifier LLM request while keeping the Next.js mirror payload unchanged. The worker still receives the full client request body, but the classifier prompt now receives a smaller request summary: API kind, short system hint, cleaned initial user prompt, cleaned latest differing user prompt, message count, and tools presence.

Adds worker parser coverage for long sessions so later user redirects are preserved without sending provider hints, requested model, stream flags, tool definitions, assistant history, tool outputs, or full history to the classifier LLM. The classifier prompt now names initial versus latest user turns explicitly, uses even 800-character caps for both, and tells the model to prefer the latest turn when they conflict.

Updates the auto-routing admin analytics panel to show average mirrored body size in KB rounded to two decimals, and moves Task Types and Task Subtypes into their own breakdown row below Status and Classifier Models.

## Verification
Live classifier eval over 24 representative request cases with 3 repeats each: 72/72 valid classifier outputs, 95.8% task-type accuracy, 91.7% subtype accuracy, average classifier-message size 9170 bytes, p50 latency 509ms, p95 latency 3656ms, total OpenRouter cost 0.00063063 credits.

Broad slimming sweep across 12 prompt-summary variants, 20 short/long cases, 2 repeats each: 480 live classifier calls. First-only and latest-only were weak on long sessions; all-user and assistant-history variants lowered subtype accuracy. The strongest balanced no-assistant variant was cleaned first+latest user prompts with 800/800 caps.

Confirmation sweep across the top 4 variants, 20 short/long cases, 8 repeats each: 640 live classifier calls. The implemented cleaned first+latest user variant produced 159/160 valid outputs, 99.4% task accuracy, 94.4% subtype accuracy, 98.8% short task/subtype accuracy, 100% long task accuracy, 90% long subtype accuracy, average classifier-message size 9234 bytes, and p95 latency 954ms. The last-assistant variant reached 100% task accuracy but dropped to 88.1% overall subtype and 76.3% long subtype, so this PR keeps assistant messages out of the classifier prompt.

## Visual Changes
Auto Routing admin analytics now shows Body Size in KB, and Task Types / Task Subtypes are grouped into a separate row below Status / Classifier Models.

## Reviewer Notes
Next.js still mirrors the full raw client request body to the worker. This PR only changes what the worker places inside the classifier LLM prompt.

`latestUserPromptPrefix` is optional in the shared response contract for compatibility, but the worker populates it as `null` unless a later user turn differs from the initial user prompt. The low-confidence fallback classifier now also prefers the latest user prompt when present.

Latency was noisy in live OpenRouter sampling. The latest confirmation was under the aspirational 1000ms p95 target, and the accepted 2000ms p95 production target should still be verified after deploy with Analytics Engine/Axiom traffic, as in the previous classifier PR.


services/auto-routing/src/classifier-input.ts
```diff
@@ -170,8 +178,19 @@ function firstPromptPrefix(messages: Message[], role: string) {
   return textPrefix(message?.content);
 }
 
+function latestPromptPrefix(messages: Message[], role: string) {
+  const first = messages.find(item => item.role === role);
+  const latest = messages.findLast(item => item.role === role);
```
jeanduplessis: **Warning:** trailing Anthropic tool results can discard the latest meaningful user redirect.

**Trace:**
1. Gateway mirrors the original `/messages` body to this worker.
2. A normal tool-using sequence can be: initial user instruction -> later user redirect -> assistant `tool_use` -> user-role `tool_result`.
3. `latestPromptPrefix(messages, 'user')` calls `findLast` before cleaning, so it selects the trailing `tool_result` message rather than the preceding redirect.
4. `textPrefix` calls `textFromValue`, which intentionally converts `type: "tool_result"` content to an empty string through `REDUNDANT_CONTENT_TYPES`.
5. `latestPrefix` becomes `null`, and this function returns `null` without scanning backward for the previous cleaned, non-empty user message.
6. `buildClassifierPromptSummary` therefore sends no latest prompt, while `fallbackClassifierOutput` falls back to `userPromptPrefix`; both classify the stale initial intent despite latest-turn precedence.

**Impact:** after a user redirects the task and a tool runs, the next Anthropic request can be classified from the original instruction instead of the current one.

Please derive first/latest prompts from cleaned, non-empty user messages and add a `/messages` regression case for the sequence above.

jeanduplessis (approved):

services/auto-routing/src/classifier-input.ts
iscekic: Fixed in fa230e877. First/latest prompt selection now derives from cleaned, non-empty messages, so trailing user-role Anthropic tool_result messages are ignored instead of erasing the latest meaningful redirect. Added a /messages regression for initial user -> redirect -> assistant tool_use -> user tool_result.

PR #3968, opened 2026-06-11
perf(auto-routing): cache classifier decisions per conversation and fix classifier fallbacks

iscekic:
## Summary

Cuts auto-routing `/decide` latency and classifier cost by caching classification decisions per conversation in a Durable Object, and fixes the two root causes of classifier fallbacks (truncated output and prompt injection). All changes are model-agnostic and were validated in prod via direct wrangler deploys, measured through Axiom (`cloudflare-logpush`, filtered by `ScriptVersion.ID`).

## Why a cache works

Agent sessions mirror every API call, so identical classifier inputs repeat heavily: in a 10-minute baseline window, 14.5k requests contained only ~4.6k unique classifier inputs (~68% duplicates; ~75% ignoring conversation depth). Every duplicate was a redundant ~800ms OpenRouter call.

## Changes

**Latency / cost**
- `AutoRoutingDecisionCacheDO`: SQLite-backed DO keyed by conversation (`sessionId`, or `content:<fingerprint>` when the client sends no session header — the first user prompt + system prompt fingerprint is stable within a conversation). Entries are keyed by `<classifierModel>:<contentHash>` (bucketed message count), 30 min TTL, whole object wiped by alarm after 2h idle. DO gives read-after-write consistency for same-session bursts (KV would not).
- Successful classifications are cached; fallbacks are never cached. Cache hits skip the OpenRouter call entirely and report `cost: 0`.
- Isolate-local memoization of the KV-configured classifier model (60s) and the OpenRouter client/API key (5 min) — removes a KV read and a secrets-store read per request.
- OpenRouter `sessionId` sticky routing (conversation key) to maximize provider-side prompt cache hits.

**Accuracy (fewer confidence-0 keyword fallbacks)**
- `maxTokens` 160 → 256: 77% of baseline fallbacks were output truncated at 160 tokens (`finishReason: length`).
- Prompt hardening: the request summary is wrapped in explicit untrusted-data markers; the classifier was previously following instructions embedded in mirrored user prompts (e.g. answering the user's "return tags as JSON" instead of classifying).
- Retry-once on invalid/empty classifier output: the residual failures are transient provider glitches (responses cut off after a few tokens with `finishReason: stop`); one retry recovers them.

**Observability**
- New structured `auto_routing_decision` log line per decision with duration, cache hit, content hashes, session id, colo, and failure diagnostics (`finishReason`, token counts, output head/tail). Successes are sampled; failures always log.
- Sample rate is the KV key `decision_log_sample_rate` (default 0.01, 60s isolate cache) so logging can be turned up/down without a redeploy. Currently set to `0.01`.
- Analytics Engine `double7` = cache hit; admin analytics summary and admin UI now show cached requests / cache hit rate.

## Results (Axiom, per deployment version)

| Metric | Baseline | Final | |
|---|---|---|---|
| Decision latency p50 | 710 ms | ~19 ms | cache hits (~75%) are ~19 ms; only novel content pays the model call |
| Decision latency avg | 811 ms | ~390 ms blended | |
| Cache hit rate | — | 75.5% | near the 75.5% theoretical duplicate ceiling |
| Fallback rate | 3.3% of decisions | ~0% | 0 fallbacks in the final verification window; retries recovered all residual glitches |
| Paid classifier calls | 100% | ~25% | cache hits cost 0 credits |

Verification queries filter `['cloudflare-logpush']` by `ScriptName == "auto-routing"` and the exact `ScriptVersion.ID`; baseline was instrumented with the same decision log before any behavior change.

## Notes

- The worker currently running in prod is this branch's final state (deployed via wrangler during experimentation). Merging will trigger the regular CI deploy of the same code.
- Admin "Classifier Latency" (AE `double1`) intentionally keeps measuring only model-call duration; cache hits write 0 so the existing latency analytics stay comparable.
- Per-model-call p50 in the cached world reads higher than baseline because only novel content reaches the model now (the cheap repeats that used to pull the percentile down are served from cache).
- Future: the per-conversation DO is a natural place to keep prior decisions per session once `/decide` responses start driving real routing.


services/auto-routing/src/openrouter.ts
```diff
@@ -5,10 +5,30 @@ type OpenRouterEnv = Pick<Env, 'OPENROUTER_API_KEY'>;
...
-export async function createOpenRouterClient(env: OpenRouterEnv): Promise<OpenRouter> {
-  return new OpenRouter({
-    apiKey: await env.OPENROUTER_API_KEY.get(),
-    httpReferer: OPENROUTER_HTTP_REFERER,
-    appTitle: OPENROUTER_APP_TITLE,
+// Isolate-local client cache so each classification does not re-read the
+// API key from the secrets store. The TTL keeps key rotations effective.
+const CLIENT_CACHE_TTL_MS = 300_000;
+
+let cachedClient: { clientPromise: Promise<OpenRouter>; expiresAt: number } | null = null;
```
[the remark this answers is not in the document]
iscekic: Valid per the AGENTS.md rule. Fixed in 0f495835f by caching only the API key string (a plain value) and constructing the OpenRouter client per request — client construction is just object setup around global fetch, so nothing transport-owning is held at module scope anymore. The 5-minute TTL on the key read is the accepted rotation lag and is documented at the cache site.

services/auto-routing/src/model-classifier.ts
```diff
@@ -53,22 +63,57 @@ export class ClassifierRunError extends Error {
...
+  // retry recovers most of those classifications.
+  const firstAttempt = await runClassifierAttempt(client, input, classifierModel, options);
+  if (!firstAttempt.fallback) {
+    return firstAttempt;
+  }
+
+  const retryAttempt = await runClassifierAttempt(client, input, classifierModel, options);
+  const cost = sumCosts(firstAttempt.cost, retryAttempt.cost);
+  if (!retryAttempt.fallback) {
+    return { ...retryAttempt, cost, retried: true };
```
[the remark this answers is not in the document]
iscekic: Good catch. Fixed in 0f495835f: ClassifierRunResult now carries firstAttemptFailure (reason, failureStage, finishReason) when a retry happened, it's included in the auto_routing_decision log line, and retried decisions bypass success sampling so they're always visible in Axiom.

services/auto-routing/src/decide.ts
```diff
@@ -157,10 +284,109 @@ export const decideHandler: Handler<HonoEnv> = async c => {
...
   const startedAt = performance.now();
+  const [hashes, classifierModel, successSampleRate] = await Promise.all([
+    computeContentHashes(classifierInput.data),
+    getClassifierModel(c.env),
+    getDecisionLogSampleRate(c.env),
+  ]);
+  // Stable conversation identity even when the client sends no session id:
+  // the first user prompt and system prompt do not change within a
+  // conversation, so their fingerprint identifies it.
+  const conversationKey = parsed.data.sessionId ?? `content:${hashes.loose}`;
```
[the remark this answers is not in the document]
iscekic: The entry-key collision is harmless as you note (entries are keyed by the exact hash), and sticky-session semantics favor keeping the loose hash as the conversation key. But the alarm-starvation/unbounded-growth point was real: fixed in 0f495835f by replacing the idle deleteAll alarm with a fixed-period sweep — the alarm is only scheduled when none is pending, expired entries are deleted in 128-key batches, and the alarm reschedules only while live entries remain. Storage per DO is now bounded to one TTL window regardless of how many conversations share the object. Covered by new decision-cache.test.ts.

services/auto-routing/src/decide.ts
```diff
@@ -158,63 +268,61 @@ export const decideHandler: Handler<HonoEnv> = async c => {
   }
 
   const startedAt = performance.now();
+  const [hashes, classifierModel, successSampleRate] = await Promise.all([
```
jeanduplessis, line 271: **[WARNING - MUST-FIX] Config lookup failures escape classifier fallback path**

**Trace:**
1. `computeContentHashes`, `getClassifierModel`, and `getDecisionLogSampleRate` now run before the classification `try`.
2. In the base code, `getClassifierModel()` ran inside `classifyNormalizedInput()`, whose invocation was inside that `try`.
3. A rejection from either KV lookup now reaches the global Hono error handler instead of producing the established empty-decision response and classifier-error telemetry.

**Impact:** Transient KV/config failures turn best-effort background classification into an HTTP 500 from `/decide`. The gateway mirror records request failure rather than receiving the normal degraded `200` response.

**Fix direction:** Include setup and cache lookup inside the guarded decision path, or make config reads fail closed to documented defaults while preserving error telemetry.

services/auto-routing/src/conversation-identity.ts
```diff
@@ -0,0 +1,52 @@
...
+// the first user prompt and system prompt do not change within a
+// conversation, so their fingerprint identifies it.
+export function deriveConversationKey(sessionId: string | null, hashes: ContentHashes): string {
+  return sessionId ?? `content:${hashes.loose}`;
+}
+
+export async function computeContentHashes(
+  input: NormalizedClassifierInput
+): Promise<ContentHashes> {
+  const base = [
```
jeanduplessis: **[WARNING - MUST-FIX] Raw delimiter serialization lets distinct classifier inputs share a cache key**

**Trace:**
1. Prompt fields accept arbitrary text, including literal `|` characters.
2. The classifier prompt preserves fields separately through JSON, but cache hashing joins them without escaping or length-prefixing.
3. For example, `system="a|b", user="c"` and `system="a", user="b|c"` produce the same serialized base when remaining fields match. Both `loose` and `exact` hashes therefore collide before SHA-256 is applied.

**Impact:** Different requests can receive the same cache entry and reuse an incorrect classification for up to 30 minutes, especially for no-session requests sharing the content-derived Durable Object.

**Fix direction:** Hash an unambiguous canonical encoding such as `JSON.stringify([apiKind, hasTools, system, user, latest, messageBucket])`, with regression coverage for delimiter-bearing prompts.

services/auto-routing/src/decide.ts
```diff
@@ -52,81 +75,168 @@ function classifierErrorStatus(error: unknown): `classifier_error:${string}` {
...
+              ...callDetails,
+              fallbackReason: fallback.reason,
+              ...(fallback.failureStage ? { classifierFailureStage: fallback.failureStage } : {}),
+              ...(fallback.schemaIssueSummary?.length
+                ? { classifierSchemaIssueSummary: fallback.schemaIssueSummary }
+                : {}),
+              ...(fallback.topLevelKeys?.length
+                ? { classifierOutputTopLevelKeys: fallback.topLevelKeys }
+                : {}),
+              ...(meta ? { textHead: meta.textHead, textTail: meta.textTail } : {}),
```
jeanduplessis: **[WARNING - MUST-FIX] Fallback telemetry logs unredacted raw model output derived from user prompts**

**Trace:**
1. Mirrored user and system prompt prefixes are sent to the classifier as explicitly untrusted data.
2. Raw external model text is captured before schema validation.
3. Every fallback bypasses sampling and logs `textHead` and `textTail`. Fallback is precisely the path where prompt injection, malformed output, or prompt echo occurred.

**Impact:** A classifier response can echo prompt content containing credentials, tokens, PII, or customer data into persistent Worker/Axiom logs, violating repository sensitive-logging rules.

**Fix direction:** Remove raw text fragments from telemetry. Retain non-content diagnostics such as output length, hash, finish reason, token counts, parse stage, and top-level keys.

services/auto-routing/src/model-classifier.ts
```diff
@@ -51,24 +66,61 @@ export class ClassifierRunError extends Error {
...
+): Promise<ClassifierRunResult> {
+  // Invalid output is usually a transient provider glitch (responses cut
+  // off after a handful of tokens with a "stop" finish reason), so one
+  // retry recovers most of those classifications.
+  const firstAttempt = await runClassifierAttempt(client, input, classifierModel, options);
+  if (!firstAttempt.fallback) {
+    return firstAttempt;
+  }
+
+  const retryAttempt = await runClassifierAttempt(client, input, classifierModel, options);
```
jeanduplessis: **[WARNING - MUST-FIX] Retry exceptions discard already-incurred first-attempt cost and diagnostics**

**Trace:**
1. The first attempt can return a fallback with known billed cost and model-call metadata.
2. If the second attempt throws, execution never reaches cost aggregation or `firstAttemptFailure` construction.
3. `decide.ts` catches only the raw retry error. Ordinary transport errors carry no first-attempt cost, so analytics records zero/null and omits the first failure details.

**Impact:** Requests with an invalid first response followed by a failed retry underreport OpenRouter spend and lose diagnostics, making admin cost totals and failure analysis inaccurate.

**Fix direction:** Catch retry exceptions inside `classifyWithOpenRouter()` and return or throw a structured result preserving first-attempt cost, model, and failure metadata.

jeanduplessis (commented):
Automated review: 4 verified warnings.

jeanduplessis (approved):

services/auto-routing/src/decide.ts
iscekic: Valid — `getClassifierModel`/`getDecisionLogSampleRate` ran before the `try`, so a KV rejection escaped to the global handler as a 500. (`getCachedClassification` already swallows its own errors, so these two config reads were the only unguarded source.) Fixed in f5acadc08: both getters now fail closed to their documented defaults (`DEFAULT_CLASSIFIER_MODEL` / `0.01`) and emit an `auto_routing_config_read_failed` warn. The rejected load is not cached (ttlCached evicts it), so the next request retries KV.

services/auto-routing/src/conversation-identity.ts
iscekic: Valid. Fixed in f5acadc08: `computeContentHashes` now hashes `JSON.stringify` of the field array instead of a `|`-join, so `system="a|b", user="c"` and `system="a", user="b|c"` no longer collide. Regression coverage added in `conversation-identity.test.ts`.

services/auto-routing/src/decide.ts
iscekic: Valid. Fixed in f5acadc08: `textHead`/`textTail` are removed from `ClassifierModelCallMeta` and the decision log, replaced with a non-content `textLength`. `finishReason`, token counts, parse stage, and top-level keys remain — enough to separate truncation from prompt echo without persisting user-derived text.

services/auto-routing/src/model-classifier.ts
iscekic: Valid. Fixed in f5acadc08: the retry attempt in `classifyWithOpenRouter` is now wrapped — if it throws, we throw a `ClassifierRunError` carrying the first attempt's cost, model, and failure metadata (`failureStage`/`schemaIssueSummary`/`topLevelKeys`) with the retry error message. `decide.ts`'s error path reads those off `ClassifierRunError`, so spend and diagnostics stay accurate. Covered in `model-classifier.test.ts`.

PR #3975, opened 2026-06-11
perf(auto-routing): mirror normalized classifier input instead of full request body

iscekic:
## What

Follow-up to #3968. The gateway now normalizes the request body into the compact classifier input *before* mirroring, and the payload carries caller identity so the worker can scope its decision cache per user.

### Gateway (`apps/web`)
- `scheduleAutoRoutingMirror` builds the payload from the already-parsed body using `normalizeClassifierInput` (moved into `@kilocode/auto-routing-contracts`); normalization runs inside `after()`, off the request path.
- `requestedModel` and provider hints are captured before auto-model resolution / provider transforms mutate the parsed body.
- Requests that fail normalization (the worker's old `invalid_body` cases) skip the mirror call entirely.
- The full request body is no longer sent: mirrored bodies currently average **331KB** (p50 231KB, p90 771KB, p99 1.7MB); the normalized payload is ~1-3KB — roughly a 99% cut in mirror egress and the worker no longer JSON-parses those bodies.

### Worker (`services/auto-routing`)
- New `MirrorPayloadSchema`: `{ input, userId, sessionId, machineId, clientRequestId, mode, userAgent, bodyBytes }` (breaking, no dual-shape support — we are pre-routing, shadow-only).
- Conversation key is now always user-scoped: `user:<id>:task:<sessionId>` → `user:<id>:machine:<machineId>` → `user:<id>:content:<hash>`. Distinct callers can no longer share a decision-cache DO, and the machine id provides a stable per-device identity when clients send no session header (replacing most content-fingerprint fallbacks).
- Decision logs now include `userId` and the `clientRequestId` value (joinable to PostHog `Feedback Submitted.parentMessageID`).
- `bodyBytes` is kept as an analytics dimension (now reported by the gateway).

## Deploy ordering

The worker only accepts the new payload shape. CI redeploys the worker on merge and Vercel ships the gateway from the same merge; during the small skew window mirrored requests get 400s, which only affects shadow analytics (verified in prod: the new worker served live traffic for ~4 minutes with clean 400s and zero exceptions before being rolled back).

## Measurement

Real before/after numbers (payload size, worker CPU, sessionId coverage) can only be gathered once the gateway side is deployed, since all current traffic is old-shape. Plan after merge: compare `bodyBytes` vs actual payload sizes, cache hit rate (expect ≥ current 73%, plus extra hits from machine-scoped identity), and `invalid_envelope` rate (~0 expected after rollout).


packages/auto-routing-contracts/src/index.ts
```diff
@@ -1,52 +1,33 @@
...
+  NormalizedClassifierInputSchema,
+  type JsonValue,
+  type NormalizedClassifierInput,
+} from './input';
 
+// What the gateway mirrors to the auto-routing worker per request: the
+// already-normalized classifier input plus caller identity. The gateway
+// normalizes before sending so the multi-hundred-KB request body never
+// leaves it, and skips the mirror entirely when normalization fails.
 export const MirrorPayloadSchema = z.object({
```
pandemicsyn, line 14: Must-fix: This replaces the deployed mirror envelope with a normalized-only schema, but the gateway and auto-routing worker deploy independently. During rollout, either old gateway -> new worker or new gateway -> old worker can reject every mirror as `invalid_envelope` and disable classification. Accept both envelope versions in the worker for one deploy, normalizing legacy full-body payloads there, then remove legacy support later.

packages/auto-routing-contracts/src/normalize.ts
```diff
@@ -0,0 +1,314 @@
...
+  if (typeof value === 'string' || typeof value === 'boolean') {
+    return value;
+  }
+
+  if (typeof value === 'number') {
+    return Number.isFinite(value) ? value : null;
+  }
+
+  if (Array.isArray(value)) {
+    return value.map(item => toJsonValue(item, budget));
```
pandemicsyn: Must-fix: The node budget does not actually bound client-controlled work here: `map` still visits every array element after the budget is exhausted, and the object branch uses `Object.entries`, which materializes all keys. Since `redactProviderHints` runs on the gateway path, a huge `provider` or `providerOptions` payload can still impose O(body) CPU/memory. Use explicit bounded loops that break on budget exhaustion and append a sentinel.

services/auto-routing/src/decide.ts
```diff
@@ -218,24 +220,25 @@ function recordDecision(
...
+      messageCount: ctx.payload.input.messageCount,
+      bodyBytes: ctx.payload.bodyBytes,
       taskType: summary.classification?.taskType ?? null,
       subtaskType: summary.classification?.subtaskType ?? null,
       confidence: summary.classification?.confidence ?? null,
-      hasMachineId: 'x-kilocode-machineid' in ctx.headers,
-      hasClientRequestId: 'x-kilo-request' in ctx.headers,
-      mode: ctx.headers['x-kilocode-mode'] ?? null,
-      uaPrefix: ctx.headers['user-agent']?.slice(0, 40) ?? null,
+      userId: ctx.payload.userId,
```
pandemicsyn: Must-fix: This logs raw `userId`; for anonymous traffic the gateway sends `anon:<ip>`, so sampled successes and all failure logs can persist client IPs. Keep raw identity for cache scoping only; remove it from telemetry or replace it with a one-way hash / non-PII identity class.

pandemicsyn (changes requested):
couple of minor things

pandemicsyn (approved):
lgtm, robot flagged a few things but i think they’re area more `nits` than must fixes.

packages/auto-routing-contracts/src/index.ts
iscekic: This is fine, we don't care for backwards compatibility.

packages/auto-routing-contracts/src/normalize.ts
iscekic: Fixed in f8841edb3: `toJsonValue` now walks arrays and objects with bounded loops that break the moment the budget is exhausted and append a `[TRUNCATED]` sentinel. Objects iterate via `for-in` + `Object.hasOwn` instead of `Object.entries`, so a key-dense value never materializes all entries, and the budget is consumed per property as well as per node (so redacted-key properties can't bypass it). Added tests covering both the 10k-element array and 10k-key object cases.

services/auto-routing/src/decide.ts
iscekic: Fixed in f8841edb3: the decision log now carries a one-way `userIdHash` (sha256-derived, same mechanism as the outbound OpenRouter session id) plus an `isAnonymousUser` flag instead of the raw id, so client IPs from `anon:<ip>` ids never reach persisted logs. The raw user id stays confined to Durable Object cache scoping. Test asserts the log line does not contain the raw id.

PR #3981, opened 2026-06-11
feat(organizations): add gated organization mode defaults

pandemicsyn:
## Summary

- Add optional `defaultModel` support to organization mode config so organizations can choose a model per mode before CLI and VSCode clients begin consuming that field.
- This enables organizations to steer different workflows toward different models, such as using a stronger planning model for `architect` while keeping `code` on a faster coding model, without removing user-level model overrides.
- The follow-up CLI and VSCode work will consume these mode defaults below user-driven selection, so cloud admins can establish sensible organizational defaults while users retain the ability to override them locally.
- This is the cloud-side first step of a staged rollout: it establishes the additive storage contract, validation, audit behavior, and UI controls while keeping existing clients unchanged.
- Add server-side validation for concrete model IDs, organization allow/deny policy checks, enterprise-only setting behavior, audit diffs, and release-flag-gated writes through `org-default-model-config`.
- Add the cloud UI picker and mode-card display behind the same flag, including built-in override clear behavior, downgraded-org clear-only behavior, and development-mode visibility.
- Keep the picker sourced from the existing org-scoped model list so mode defaults are selected only from models currently allowed for that organization.
- Make partial mode config updates atomic in JSONB so concurrent description/default-model edits do not overwrite each other.

## Verification

API response sample

```
❯ http GET http://localhost:3000/api/organizations/9c3780e5-69ce-4d18-a0a8-b1ec0998db83/modes \
  Authorization:"Bearer [redacted]" \
  X-KiloCode-OrganizationId:"9c3780e5-69ce-4d18-a0a8-b1ec0998db83"
HTTP/1.1 200 OK
<headers trimmed>

{
    "modes": [
        {
            "config": {
                "customInstructions": "",
                "defaultModel": "kilo-auto/balanced",
                "description": "Write, modify, and refactor code",
                "groups": [
                    "browser",
                    "command",
                    "edit",
                    "mcp",
                    "read"
                ],
                "roleDefinition": "You are Kilo Code, a highly skilled software engineer with extensive knowledge in many programming languages, frameworks, design patterns, and best practices.",
                "whenToUse": "Use this mode when you need to write, modify, or refactor code. Ideal for implementing features, fixing bugs, creating new files, or making code improvements across any programming language or framework."
            },
            "created_at": "2026-06-11 21:54:37.803677+00",
            "created_by": "[redacted]",
            "id": "3b325243-bd36-48aa-b9ee-3cb4e8a83cc1",
            "name": "Code",
            "organization_id": "9c3780e5-69ce-4d18-a0a8-b1ec0998db83",
            "slug": "code",
            "updated_at": "2026-06-12 14:14:30.317325+00"
        }
    ]
}
```

## Visual Changes

| Before | After |
|---|---|
| Mode drawer without a mode default model control | Mode drawer with gated `Mode Default Model` picker |
| Mode cards without model metadata | Mode cards showing gated default-model metadata |
| Built-in mode overrides only supported config edits | Built-in mode overrides can clear the last default-model override and revert |

SCR-20260612-ipys

Screenshot 2026-06-12 at 9 18 04 AM

Screenshot 2026-06-12 at 9 18 16 AM

## Reviewer Notes

- This is intentionally additive and safe for existing mode rows because `defaultModel` is optional in the JSONB config.
- The flag is off for all users except `*kilocode.ai` emails, so existing cloud behavior remains unchanged until the CLI PR lands and is vetted.
- Setting a mode default is enterprise-only; clearing an existing default remains allowed after downgrade.
- Stored defaults that later become disallowed do not block unrelated mode edits.
- The main risk area is the distinction between clearing a mode-specific default and deleting the last built-in override row.

### DIFF
--- a/apps/web/src/components/organizations/custom-modes/CustomModesLayout.tsx
+++ b/apps/web/src/components/organizations/custom-modes/CustomModesLayout.tsx
@@ -27,6 +27,7 @@ import { ModeDrawer } from './ModeDrawer';
 import { NewModeForm } from './NewModeForm';
 import { EditModeForm } from './EditModeForm';
 import { useOrganizationReadOnly } from '@/lib/organizations/use-organization-read-only';
+import { useFeatureFlagEnabled } from 'posthog-js/react';
 
 type CustomModesLayoutProps = {
   organizationId: string;
@@ -43,9 +44,16 @@ type ModesListProps = {
   readonly: boolean;
   onDeleteClick: (mode: DisplayMode) => void;
   onEditClick: (mode: DisplayMode) => void;
+  isDefaultModelConfigEnabled: boolean;
 };
 
-function ModesList({ modes, readonly, onDeleteClick, onEditClick }: ModesListProps) {
+function ModesList({
+  modes,
+  readonly,
+  onDeleteClick,
+  onEditClick,
+  isDefaultModelConfigEnabled,
+}: ModesListProps) {
   return (
     <div className="grid grid-cols-1 gap-4 md:grid-cols-2 lg:grid-cols-3">
       {modes.map((mode, index) => (
@@ -99,7 +107,15 @@ function ModesList({ modes, readonly, onDeleteClick, onEditClick }: ModesListPro
               </div>
             </CardHeader>
             <CardContent className="flex-1">
-              <div className="flex h-full flex-col">
+              <div className="flex h-full flex-col gap-4">
+                {isDefaultModelConfigEnabled && mode.config.defaultModel && (
+                  <div className="flex flex-wrap items-center gap-2">
+                    <span className="text-muted-foreground text-xs font-medium">Default model</span>
+                    <Badge variant="secondary" className="max-w-full font-mono text-xs">
+                      <span className="break-all">{mode.config.defaultModel}</span>
+                    </Badge>
+                  </div>
+                )}
                 {mode.config?.groups && mode.config.groups.length > 0 && (
                   <div className="flex-1">
                     <h4 className="mb-2 text-sm font-medium">Available Tools</h4>
@@ -153,7 +169,9 @@ export function CustomModesLayout({ organizationId }: CustomModesLayoutProps) {
   const [drawerMode, setDrawerMode] = useState<'create' | 'edit'>('create');
   const [editingMode, setEditingMode] = useState<DisplayMode | null>(null);
   const isReadOnly = useOrganizationReadOnly(organizationId);
-
+  const isDefaultModelFeatureEnabled = useFeatureFlagEnabled('org-default-model-config');
+  const isDevelopment = process.env.NODE_ENV === 'development';
+  const isDefaultModelConfigEnabled = isDevelopment || isDefaultModelFeatureEnabled === true;
   const readonly = isReadOnly;
 
   // Separate built-in modes and custom modes
@@ -297,6 +315,7 @@ export function CustomModesLayout({ organizationId }: CustomModesLayoutProps) {
               readonly={readonly}
               onDeleteClick={openDeleteDialog}
               onEditClick={handleEditMode}
+              isDefaultModelConfigEnabled={isDefaultModelConfigEnabled}
             />
           </div>
 
@@ -310,6 +329,7 @@ export function CustomModesLayout({ organizationId }: CustomModesLayoutProps) {
                 readonly={readonly}
                 onDeleteClick={openDeleteDialog}
                 onEditClick={handleEditMode}
+                isDefaultModelConfigEnabled={isDefaultModelConfigEnabled}
               />
             </div>
           )}
@@ -383,6 +403,7 @@ export function CustomModesLayout({ organizationId }: CustomModesLayoutProps) {
               defaultModeSlug={
                 editingMode?.isDefault && !editingMode?.isOverridden ? editingMode.slug : undefined
               }
+              isDefaultModelConfigEnabled={isDefaultModelConfigEnabled}
               onSuccess={handleDrawerClose}
               onCancel={handleDrawerClose}
             />
@@ -390,6 +411,8 @@ export function CustomModesLayout({ organizationId }: CustomModesLayoutProps) {
             <EditModeForm
               organizationId={organizationId}
               modeId={editingMode.id}
+              defaultModeSlug={editingMode.isDefault ? editingMode.slug : undefined}
+              isDefaultModelConfigEnabled={isDefaultModelConfigEnabled}
               onSuccess={handleDrawerClose}
               onCancel={handleDrawerClose}
             />
--- a/apps/web/src/components/organizations/custom-modes/CustomModesLayout.tsx
+++ b/apps/web/src/components/organizations/custom-modes/CustomModesLayout.tsx
@@ -2,7 +2,11 @@
 
 import { useState, useMemo } from 'react';
 import { motion } from 'motion/react';
-import { useOrganizationModes, useDeleteOrganizationMode } from '@/app/api/organizations/hooks';
+import {
+  useOrganizationModes,
+  useDeleteOrganizationMode,
+  useOrganizationWithMembers,
+} from '@/app/api/organizations/hooks';
 import { Button } from '@/components/ui/button';
 import { LockableContainer } from '../LockableContainer';
 import { Badge } from '@/components/ui/badge';
@@ -162,6 +166,7 @@ function ModesList({
 
 export function CustomModesLayout({ organizationId }: CustomModesLayoutProps) {
   const { data, isLoading, error } = useOrganizationModes(organizationId);
+  const { data: organizationData } = useOrganizationWithMembers(organizationId);
   const deleteMutation = useDeleteOrganizationMode();
   const [deleteDialogOpen, setDeleteDialogOpen] = useState(false);
   const [modeToDelete, setModeToDelete] = useState<DisplayMode | null>(null);
@@ -172,6 +177,7 @@ export function CustomModesLayout({ organizationId }: CustomModesLayoutProps) {
   const isDefaultModelFeatureEnabled = useFeatureFlagEnabled('org-default-model-config');
   const isDevelopment = process.env.NODE_ENV === 'development';
   const isDefaultModelConfigEnabled = isDevelopment || isDefaultModelFeatureEnabled === true;
+  const canSetDefaultModel = organizationData?.plan === 'enterprise';
   const readonly = isReadOnly;
 
   // Separate built-in modes and custom modes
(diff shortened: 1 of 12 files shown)


apps/web/src/components/organizations/custom-modes/EditModeForm.tsx
```diff
@@ -4,26 +4,79 @@ import {
...
+    return false;
+  }
+
+  return (
+    formData.name === defaultMode.name &&
+    formData.slug === defaultMode.slug &&
+    formData.roleDefinition === defaultMode.config.roleDefinition &&
+    normalizeOptionalValue(formData.description) === defaultMode.config.description &&
+    normalizeOptionalValue(formData.whenToUse) === defaultMode.config.whenToUse &&
+    JSON.stringify(formData.groups) === JSON.stringify(defaultMode.config.groups) &&
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in 166e95acc.

apps/web/src/routers/organizations/organization-modes-router.ts
```diff
@@ -142,34 +284,60 @@ export const organizationModesRouter = createTRPCRouter({
...
+        ) {
           const oldValue = existingMode.config.customInstructions || '(empty)';
           const newValue = updates.config.customInstructions || '(empty)';
           configChanges.push(
             `customInstructions: "${oldValue.substring(0, 50)}${oldValue.length > 50 ? '...' : ''}" → "${newValue.substring(0, 50)}${newValue.length > 50 ? '...' : ''}"`
           );
         }
+        if (
+          'defaultModel' in updates.config &&
+          updates.config.defaultModel !== existingMode.config.defaultModel
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in 166e95acc.

apps/web/src/routers/organizations/organization-modes-router.ts
```diff
@@ -51,6 +71,98 @@ const ModeIdInputSchema = OrganizationIdInputSchema.extend({
   modeId: z.uuid(),
 });
 
+function hasDefaultModelUpdate(config: DefaultModelConfig | undefined): boolean {
+  return !!config && Object.prototype.hasOwnProperty.call(config, 'defaultModel');
```
jrf0110: Some of these validator/guard functions are little confusing to follow and have sort of confusing names (e.g. `hasDefaultModelUpdate` is weird because what is the "Update" part about)

Perhaps a simple zod schema and safe parse could replace many of these `has*` and `ensure*` functions

pandemicsyn: Refactored in 1277c378e. Kept the existing Zod input boundary, but replaced the ambiguous has*/ensure* guards with an explicit `DefaultModelChange` discriminated union (`none` / `clear` / `set`) so absent/null/string semantics are visible at call sites.

jeanduplessis (commented):
## Judgement Review

### Problem statement
- `non-blocking`: PR explains staged rollout well. Could add one sentence naming expected client precedence order in final shipped behavior: user local override > org mode default > existing fallback.

### Solution
- Pragmatic and clear solution, no feedback.

### Risk
- No feedback.

### Architecture
- `blocking`: Mode default model writes diverge from existing org model-setting ownership. Current org default model and access-list mutations use organizationBillingMutationProcedure (owner or billing manager), and UI copy says only owners can edit model access. This PR adds per-mode model defaults through organizationMemberMutationProcedure. That creates two architecture concepts for who owns org model policy. Align defaultModel writes with existing org model-setting ownership, or document why mode defaults intentionally follow broader custom-mode permissions.

### Operability
- `blocking`: Rollback and stale-default behavior are unclear. Disabling org-default-model-config stops UI/write access, but /api/organizations/[id]/modes still returns persisted config.defaultModel. Also, if org policy later denies stored model, test coverage keeps stale value in mode config. Future clients may consume invalid defaults unless endpoint filters them or clients must validate. Add explicit contract plus tests: either server filters/omits invalid mode defaults, settings changes clear invalid mode defaults, or clients must validate against allowed model list before use.
- `non-blocking`: Add short rollback note: Something like disable flag, clear affected config.defaultModel values if clients already shipped, verify mode API no longer returns active defaults.

### Verification
- `blocking`: Verification section lacks confirmation of manual verification for main user journeys: flag on/off behavior, setting and clearing a mode default in UI, built-in override clear/revert behavior, Teams org rejection, denied model rejection, and API payload after set/clear. 
- `non-blocking`: Add verification steps for reviewer to confirm these flows locally.

### Agent Entropy
- `non-blocking`: Add short comment or doc near OrganizationModeConfigSchema naming relationship to organization.settings.default_model: org default is global fallback; mode default is per-mode override; user local selection still wins. This will prevent future agent confusion once client work lands.

apps/web/src/lib/organizations/organization-modes.ts
```diff
@@ -79,14 +80,36 @@ export async function updateOrganizationMode(
...
+          eq(orgnaization_modes.id, modeId),
+          eq(orgnaization_modes.organization_id, organizationId)
+        )
+      );
+
+    if (!existingMode) {
+      return null;
+    }
+
+    updateData.config = mergeToSatisfy({
```
jeanduplessis: **WARNING — MUST-FIX:** Concurrent partial config updates can silently overwrite each other.

**Trace:**
1. This changed path reads the current JSON config, merges the partial input in application memory, then writes the complete object in a separate query.
2. The router now treats partial config updates as supported behavior.
3. Two concurrent requests can read the same snapshot, merge different fields, and let the later full-object write erase the earlier update.

**Impact:** Concurrent edits, such as a description change and a default-model change, can both report success while one change is silently lost.

**Fix direction:** Merge the JSONB patch atomically in the update statement, or protect the read/write sequence with row locking or optimistic version checking.

apps/web/src/components/organizations/custom-modes/CustomModesLayout.tsx
```diff
@@ -153,7 +169,9 @@ export function CustomModesLayout({ organizationId }: CustomModesLayoutProps) {
   const [drawerMode, setDrawerMode] = useState<'create' | 'edit'>('create');
   const [editingMode, setEditingMode] = useState<DisplayMode | null>(null);
   const isReadOnly = useOrganizationReadOnly(organizationId);
-
+  const isDefaultModelFeatureEnabled = useFeatureFlagEnabled('org-default-model-config');
+  const isDevelopment = process.env.NODE_ENV === 'development';
+  const isDefaultModelConfigEnabled = isDevelopment || isDefaultModelFeatureEnabled === true;
```
jeanduplessis, line 179: **WARNING — MUST-FIX:** Feature-flag state alone exposes the model picker to organizations whose writes the server rejects.

**Trace:**
1. UI capability depends only on the user feature flag or development mode.
2. The organization model endpoint still returns selectable models for non-Enterprise organizations.
3. Create and update procedures reject setting `defaultModel` unless the organization plan is `enterprise`.

**Impact:** Flag-enabled Teams users see a functional picker, can select a model, then receive `FORBIDDEN` on every save.

**Fix direction:** Derive visibility from a server-authoritative capability that combines release flag and Enterprise entitlement; otherwise hide or disable the control with explanatory copy.

apps/web/src/components/organizations/custom-modes/ModeForm.tsx
```diff
@@ -237,6 +261,10 @@ export function ModeForm({
       newErrors.slug = `A mode with the slug "${formData.slug}" already exists`;
     }
 
+    if (isDefaultModelConfigEnabled && hasUnavailableDefaultModel) {
```
jeanduplessis: **WARNING — MUST-FIX:** An unchanged stale default blocks unrelated mode edits.

**Trace:**
1. Organization policy can later deny a previously stored mode default.
2. The allowed-model query then omits the current value, making `hasUnavailableDefaultModel` true.
3. This validation rejects every submit, including description, instruction, and tool-only changes, even when `defaultModel` was not changed.
4. The backend explicitly permits unrelated edits while preserving the stale value, so UI and server contracts diverge.

**Impact:** A policy change can make an existing mode uneditable unless the user also clears or replaces its default model.

**Fix direction:** Require an allowed value only when `defaultModel` changed. Keep an unchanged stale value visible as unavailable while preserving clear and replace controls.

apps/web/src/components/organizations/custom-modes/ModeForm.tsx
```diff
@@ -418,6 +447,74 @@ export function ModeForm({
...
+                    : modelOptions.length === 0
+                      ? 'No organization-allowed models are available.'
+                      : 'Members can still override this locally in Kilo Code.'}
+              </p>
+              {hasUnavailableDefaultModel && (
+                <p className="text-sm text-amber-600">
+                  Choose an allowed model or clear this value before saving.
+                </p>
+              )}
+              {errors.defaultModel && <p className="text-sm text-red-600">{errors.defaultModel}</p>}
```
jeanduplessis: **SUGGESTION — SUGGESTED:** Associate the default-model validation error with the select control.

**Trace:**
1. `SelectTrigger` references help text only.
2. The validation error has no ID, live region, or alert role.
3. The trigger does not expose `aria-invalid`.

**Impact:** Screen-reader users may receive no programmatic indication that submission failed or which control needs correction.

**Fix direction:** Give the error a stable ID, include it in `aria-describedby` when present, set `aria-invalid`, and announce it with `role="alert"` or a live region.

apps/web/src/components/organizations/custom-modes/EditModeForm.tsx
```diff
@@ -4,26 +4,88 @@ import {
...
+  onCancel,
+}: EditModeFormProps) {
   const { data, isLoading, error } = useOrganizationModeById(organizationId, modeId);
   const { data: modesData } = useOrganizationModes(organizationId);
   const updateMutation = useUpdateOrganizationMode();
+  const deleteMutation = useDeleteOrganizationMode();
 
   const handleSubmit = async (formData: ModeFormData) => {
     try {
+      if (
```
jeanduplessis, line 71: **SUGGESTION — SUGGESTED:** Add behavioral coverage for the destructive built-in revert decision.

**Trace:**
1. This new branch decides whether saving clears one field or deletes the entire built-in override row.
2. The decision depends on normalized optional values and group equality.
3. Router and persistence tests cannot exercise this client-side branch, and no custom-mode component tests cover it.

**Impact:** A regression could delete remaining customization or leave an unnecessary override row without current tests detecting it.

**Fix direction:** Add focused tests for matching and non-matching built-in state, reordered groups, clearing the last override, and clearing the model while other customized fields remain.

jeanduplessis (approved):

PR #3982, opened 2026-06-11
feat(auto-routing): benchmark-driven decision engine and kilo-auto/efficient

iscekic:
# Benchmark-driven decision engine and `kilo-auto/efficient`

## Summary

Adds a benchmark-driven model-routing pipeline behind a new hidden virtual model, `kilo-auto/efficient`: route each request to the cheapest model that is proven (by our own benchmarks) to be accurate enough for the request's difficulty.

> Durable docs: **`docs/adr/0002-auto-routing-efficient.md`** (design, invariants, billing/sticky-session rules, alternatives, rollout/rollback) and **`services/auto-routing-benchmark/README.md`** (local run, env, DLQ/container debugging).

**Why now.** `balanced` routes every request to one fixed model; we have no mechanism to send easy requests to a cheaper model that our own measurements show is good enough. The goal is a measured cost reduction versus `balanced` at non-inferior accuracy. The model ships **hidden** specifically so this is validated on Kilo team traffic before any real-user exposure — graduation from hidden is a judgement call on that traffic (target signals: cost-per-request down vs. balanced, accuracy not worse, fallback rate not regressed), not a code switch. See the ADR for the graduation criteria.

Three moving parts:

- **`services/auto-routing-benchmark`** (new Cloudflare Worker): runs two deterministic benchmarks — classifier prompt replay via OpenRouter, and decider golden tasks through the real `kilo` CLI in a Cloudflare Container — writes normalized results to D1, and publishes a routing table (per-difficulty-tier ranked candidates) plus a classifier winner.
- **`services/auto-routing`** (existing worker): `/decide` classifies the request, derives a difficulty tier, and picks the cheapest above-threshold model from the routing table, with session-sticky decisions held in a Durable Object.
- **`apps/web`** (gateway): exposes `kilo-auto/efficient`, blocks on `/decide` with a 2s timeout, falls back to the balanced Qwen default, bills the classifier LLM cost to the requesting user, and adds an admin panel for the whole pipeline.

Shared classifier code (prompt, parsing, fallback, taxonomy, tier derivation, routing-table schema) moves into the new `packages/auto-routing-contracts` package, so the benchmark replays exactly the code the production worker executes.

## Architecture

```
client request (model = kilo-auto/efficient)
        │
        ▼
┌────────────────────────────┐  POST /decide (2s timeout;      ┌─────────────────────────────────┐
│ apps/web gateway           │  null/timeout → balanced Qwen)  │ services/auto-routing           │
│ · resolves kilo-auto/*     │ ───────────────────────────────▶│ · classify request (LLM)        │
│ · applies pinned           │                                 │ · derive difficulty tier        │
│   reasoningEffort          │                                 │ · cheapest above-threshold pick │
│ · bills classifier cost    │                                 │ · sticky decision per           │
│ · admin panel              │                                 │   conversation (DO)             │
└──────────┬─────────────────┘                                 └──────────────┬──────────────────┘
           │ admin proxy +                                     routing table & classifier winner:
           │ 6h token mint                                     isolate cache 60s → KV 1h →
           │ (internal secret)                                 service binding origin
           ▼                                                                  ▼
┌───────────────────────────────────────────────────────────────────────────────────────────────┐
│ services/auto-routing-benchmark (new worker)                                                   │
│ · /admin/config /admin/runs /admin/routing-table /admin/classifier-winner /admin/debug-cli     │
│ · Queue (+DLQ) fans out per-model jobs                                                         │
│ · classifier bench: 72-case prompt replay via OpenRouter                                       │
│ · decider bench: 76 golden tasks via `kilo` CLI in a Cloudflare Container                      │
│ · D1 (drizzle, normalized): runs, case results, summaries, published routing tables            │
└───────────────────────────────────────────────────────────────────────────────────────────────┘
```

### Classifier benchmark

Replays 72 normalized classifier inputs through OpenRouter using the exact production classifier code (`@kilocode/auto-routing-contracts/classifier`). Each output is graded per-field against a hand-labeled expectation via `CLASSIFIER_FIELD_WEIGHTS` (`src/grading.ts`): taskType 0.25, reasoningComplexity 0.20, contextComplexity 0.15, executionMode 0.15, subtaskType 0.10, requiresTools 0.10, riskLevel 0.05. Heuristic-fallback outputs score 0.

The winner (`src/winner.ts`) is the cheapest model meeting the run's accuracy threshold (most accurate one if none do). It feeds the worker's classifier-model resolution chain (below).

### Decider benchmark

Runs 76 golden tasks per candidate model through the real `kilo` CLI (`@kilocode/cli`) inside a Cloudflare Container (`container/Dockerfile` + `container/server.mjs`, node:22-slim, standard-2). Grading is purely mechanical — `exact` / `contains_all` / `regex` / `json_equal` checks, no LLM judges; golden answers were hand-derived and mechanically re-verified where executable. Cases include genuinely agentic tasks performed with file/terminal tools inside the container (deterministic: no repo, no network).

Execution details:

- One queue message per (model, 10-case chunk); each chunk gets its own container instance (`runId:model:chunk`) so models/chunks never share state. CLI runs are serialized per instance (the CLI's sqlite state is not safe under concurrent first runs); a `/warmup` endpoint absorbs the one-time sqlite migration before the case loop.
- Each candidate's pinned `reasoningEffort` is forwarded as the CLI's `--variant`, so the benchmark measures the model exactly as it will be served.
- The CLI authenticates as a real Kilo user: the worker mints a short-lived token once per queue message via `apps/web`'s internal endpoint (token only ever lives in a child-process env var, never logged or written to disk).
- Empty-output sessions (exit 0, no assistant text) are retried once, mirroring the production classifier's retry policy; costs of both attempts are summed.

### Datasets

Both datasets cover all 18 (taskType, subtaskType) taxonomy pairs with at least 4 cases per pair — enforced by tests (`src/datasets/*.test.ts`). Decider cases each carry exactly one difficulty tier with at least 4 distinct task types per tier.

### D1 schema (`src/db-schema.ts`)

Fully normalized, zero JSON blob columns, composite-PK-only access:

- `benchmark_config` + `config_classifier_models` + `config_decider_models` — admin config (incl. per-decider-model `reasoning_effort`).
- `benchmark_runs` — carries a **config snapshot** (`min_accuracy`, `switch_cost_factor`, `max_concurrency`, `benchmark_user_id`) taken at `startRun` time, so mid-run admin edits can't skew results. All job processing and publishing reads the snapshot, never live config.
- `run_models` — which models were enqueued vs. skipped, with the pinned `reasoning_effort` snapshot.
- `case_results` — per (run, model, case) score/latency/cost plus diagnostics (classifier fallback reason, CLI exit code/output prefix/event tail).
- `model_summaries` — per (run, model, tier) aggregates. **Carried summaries**: models with prior results are skipped on new runs (their latest summaries are copied in with `carried=true`), so re-runs only spend on new candidates; the admin can force a full re-run.
- `routing_tables` + `routing_table_candidates` — published tables, queryable history.

Single squashed baseline migration (`migrations/0000_absent_wallow.sql`), applied by a `predeploy` script (`wrangler d1 migrations apply --remote`) which the CI deploy workflow now runs for any worker that defines one (`.github/workflows/deploy-workers.yml`).

### Publishing

On run completion the worker builds the routing table from the run's own snapshot (`src/routing-table-builder.ts`): per tier, candidates are ranked best-bang-for-buck (above-threshold cheapest-first, below-threshold by accuracy). Models with zero graded cases or no cost signal in a tier are excluded; if any tier ends up empty the publish is skipped and the previous table stays live (schema enforces `.min(1)` per tier). Publishing only **deletes** the KV cache keys so the auto-routing worker repopulates from D1 on the next read.

## Decision engine (`services/auto-routing`)

`/decide` (existing endpoint, now decision-capable):

1. Classifies the request (per-conversation classification cache in a Durable Object, keyed by classifier model + content hash).
2. Derives a difficulty tier (`deriveDifficultyTier` in contracts: reasoning complexity dominates at 2x weight; context, execution mode, and risk nudge borderline cases).
3. Picks from the routing table (`src/decision-engine.ts`): cheapest above-threshold candidate for the tier — **unless the session has an incumbent**.

**Session stickiness**: the conversation's Durable Object remembers the last served model. The incumbent is kept while it still meets the tier's accuracy threshold, unless the fresh pick is cheaper by more than the table's `switchCostFactor`. Rationale (commented in code): a model switch discards the provider's prompt cache, and rebuilding it costs full-price input tokens (4–10x cache-read rates) on a context that dominates agent-session spend — switching only pays off when recurring per-turn savings clearly exceed that one-time penalty. Sticky state trusts only real classifier output: heuristic fallbacks never re-anchor the session's model.

**Routing table access**: read-through chain — isolate-local 60s TTL cache → KV (1h TTL, shared `AUTO_ROUTING_CONFIG` namespace) → service binding to the benchmark worker's D1-backed `/admin/routing-table`. Corrupt KV values are treated as misses; origin failures degrade to null (no decision) rather than erroring the request.

**Classifier model resolution** (`src/classifier-config.ts`): admin KV override → benchmark winner (same KV read-through, derived on read) → built-in default `google/gemini-2.5-flash-lite`. A benchmark-origin failure never discards a healthy override.

## Gateway (`apps/web`)

- **`kilo-auto/efficient`** (`src/lib/ai-gateway/auto-model/index.ts`): hidden virtual model (excluded from `/models`, usable by id) with the same catalog properties as balanced — intended to eventually replace it, hidden while validated on Kilo team traffic.
- **Resolution** (`auto-model/resolution.ts` + `auto-routing-decision.ts`): blocks on `/decide` with a 2s timeout; on a decision, serves the decided model and applies its pinned `reasoningEffort` so it runs under the same conditions the benchmark measured. On null/timeout/error, serves `BALANCED_QWEN_MODEL` — an efficient request never degrades below balanced.
- **Billing**: the classifier LLM cost returned by `/decide` is billed to the requesting user as a separate microdollar usage row (`requested_model: kilo-auto/efficient`), so routing overhead is visible and attributed rather than absorbed.
- **Admin panel** (`admin/auto-routing/BenchmarksSection.tsx`, proxied through admin API routes with the internal secret): config editor (classifier/decider model lists, per-decider `reasoningEffort`, `minAccuracy`, `switchCostFactor`, `maxConcurrency`, `benchmarkUserId`), run triggers with a force-rerun toggle, run history, and the live published routing table.
- **Config-save validation** (`admin/api/auto-routing/benchmark-config/route.ts`): every decider model must be servable on **all** gateway chat API kinds (`chat_completions`, `responses`, `messages`) by the provider the gateway would route it to — the routing table deliberately carries no per-protocol metadata, so this invariant is enforced at write time.
- **Token mint** (`api/internal/auto-routing-benchmark/token/route.ts`): POST gated by `INTERNAL_API_SECRET`; mints a 6h full user API token (`tokenSource: auto-routing-benchmark`) for the decider CLI's identity/billing.

## Design properties

- **No fabricated data anywhere.** There is no default routing table: `/decide` returns null decisions until a benchmark publishes one, and the gateway serves balanced fallbacks. There is no default benchmark config: runs refuse to start until an admin saves one (and decider runs additionally fail fast without a `benchmarkUserId`).
- **Deterministic, reproducible grading.** Mechanical checks only; run-level config snapshots; routing tables built from the run's snapshot, not live config.
- **Cheap iteration.** Carried summaries mean adding one candidate model re-benchmarks only that model; config-only changes (model removed, threshold tweaked) republish instantly with zero spend.
- **Graceful degradation at every layer.** Corrupt KV → miss; origin failure → previous behavior; classifier failure → no decision → balanced fallback; publish failure → previous table stays live.

## Infrastructure

- D1 `auto-routing-benchmark` in region EEUR, primary in Frankfurt (colo FRA — next to the backend; verified via `wrangler d1 info`).
- Queue `auto-routing-benchmark-jobs` (max_concurrency 4, max_retries 2) + DLQ `auto-routing-benchmark-dlq`.
- Container app `auto-routing-benchmark-runner` (standard-2, max 40 instances), image built and pushed by `wrangler deploy`.
- Service binding `auto-routing` → `auto-routing-benchmark`; shared KV namespace `AUTO_ROUTING_CONFIG`.
- Both workers already run this branch's code; the D1 database is empty pending admin setup.

## Post-merge deploy / cutover checklist

1. Merge → Vercel ships the gateway side (`kilo-auto/efficient`, admin panel, token mint).
2. First post-merge worker deploy runs the D1 migration via the new CI predeploy hook — CI's `CLOUDFLARE_API_TOKEN` needs D1 edit permission (the deploy will surface it if missing).
3. Admin saves a benchmark config: `benchmarkUserId` is required for decider runs (consider a dedicated service account — its account is billed for CLI usage); suggested `switchCostFactor` starting value: 3.
4. Trigger a classifier run and a decider run from the admin panel.
5. Clear the leftover `classifier_model` KV override (currently set to flash-lite) if the benchmark winner should drive classifier selection.

## Reviewer notes

- The `exact` decider check also accepts the **last non-empty output line** (`src/grading.ts`): agent harnesses sometimes prepend preamble despite instructions; wrong answers fail either way.
- `@kilocode/cli@latest` is resolved at image **build** time, i.e. each deploy pins whatever was `latest` then; re-deploy to pick up a newer CLI.
- The token-mint endpoint is gated by `INTERNAL_API_SECRET` and can mint for **any** user id; scoping it to the configured benchmark user is a reasonable follow-up.
- The decider benchmark exercises models through `chat_completions` only (the CLI's path). Config-save validation guarantees candidates are *servable* on all three chat API kinds, but accuracy is only *measured* on one.

## Alternatives considered

Full rationale in the ADR; in brief:

- **Reuse model-experiment tooling** — experiments are explicit, user-selected preview ids that (per `.specs/model-experiments.md`) must never enter automatic `kilo-auto` candidate sets, and carry no per-difficulty accuracy/cost signal. They can't drive automatic routing.
- **Offline benchmarks + hand-maintained routing tables** — a hand-maintained table is fabricated data that drifts, has no reproducible provenance, and can't be re-derived after a model/prompt change. Making the benchmark the source of truth is the point.
- **A narrower first PR** (classifier-only, or routing without benchmarks) — the pieces aren't independently useful: a routing engine with no published table has nothing to route from; a benchmark with no consumer publishes into the void. The smallest *shippable* unit is the full loop behind a hidden model, which is why it ships hidden rather than as smaller live increments.
- **LLM-judged grading** — rejected for reproducibility; mechanical checks are deterministic.

## Rollback

`kilo-auto/efficient` is hidden and additive, so rollback is containment, not revert (detail in the ADR):

- **Disable the model** — it's hidden, so no `/models` consumer depends on it; the gateway serves balanced on any null decision, so reverting the gateway deploy (or forcing null decisions) degrades cleanly to balanced.
- **Clear published artifacts** — delete the routing-table + classifier-winner keys in the `AUTO_ROUTING_CONFIG` KV namespace; `/decide` returns null until a benchmark republishes (balanced everywhere).
- **Stop benchmark activity** — stop triggering runs from the admin panel; in-flight queue jobs drain or fail into the DLQ.
- **Worker rollback** — redeploy the previous `auto-routing` / `auto-routing-benchmark` versions. The D1 schema is additive; a failed predeploy migration fails the deploy before serving, leaving the prior version live.

## DLQ runbook

Failed queue messages land in `auto-routing-benchmark-dlq` after `max_retries` (2). A message is one (model, chunk) job, so a DLQ'd message means that chunk produced no results and its model's tier summaries are incomplete. Inspect via the Cloudflare dashboard (Workers → Queues → `auto-routing-benchmark-dlq`); the body is the JSON job (`runId`, `model`, `chunk`, case ids). Replay by re-running the affected model with the admin **force** toggle once the cause is fixed (carried summaries re-benchmark only that model). A wedged `running` run is swept to `failed` on the next `GET /admin/runs`. Full version in `services/auto-routing-benchmark/README.md`.

## Verification

Local stack via `pnpm dev:start auto-routing` (gateway :3000, auto-routing :8810, benchmark worker :8814), seeded users `auto-routing-cli-local` (credits) and `auto-routing-admin-local` (admin).

Automated — all green:

- `pnpm --filter auto-routing-benchmark test` (worker, incl. C1 batch slicing, C4 carry-identity, C5 409 + stale-sweep, C6 duplicate-id)
- `pnpm --filter @kilocode/auto-routing-contracts test` (C6 duplicate-id validation)
- `apps/web` jest: openrouter `route.test.ts` (C2 anon-skip + exit-safe billing, C7 BYOK-still-bills, C8 no first-usage events), `benchmark-config/route.test.ts` (C3 experiment-reserved 400), `BenchmarksSection` suite (C9–C13)
- `tsgo --noEmit` clean for `apps/web` and the worker

Manual end-to-end (against the local stack):

- **Gateway routing + billing** — a real `kilo-auto/efficient` chat completion (`POST /api/openrouter/chat/completions`) routed to `google/gemini-2.5-flash-lite` and returned 200. The DB then showed **two** usage rows for the user: the final-inference row *and* a separate `auto-routing/classifier` overhead row (`requested_model: kilo-auto/efficient`) — confirming the classifier cost is billed independently (C2/C7), even though the final inference reported `is_byok: true`.
- **C6 duplicate ids** — `PUT /admin/config` with a duplicated decider id, and again with a duplicated classifier id, each returned `400` with a field-specific `Duplicate model id` issue (path `deciderModels[1]` / `classifierModels[1]`).
- **C3 experiment-reserved** — seeded a `model_experiment` row, then `PUT /admin/api/auto-routing/benchmark-config` (admin session) with that id as a decider model returned `400` ("must not be model-experiment public ids"); removed the row afterward.
- **C5 stale sweep** — `GET /admin/runs` executes `sweepStaleRuns` against live D1 and returns cleanly. (The 409 admission path and partial unique index are covered by unit tests; triggering a live conflict requires a full ~860-call classifier run, not worth the spend for a unit-tested path.)

### Validate locally

```bash
pnpm dev:start auto-routing
# mint a gateway token for the seeded benchmark user
SECRET=$(grep '^INTERNAL_API_SECRET=' .env.local | cut -d= -f2- | tr -d '"')
TOKEN=$(curl -s -X POST http://localhost:3000/api/internal/auto-routing-benchmark/token \
  -H "Authorization: Bearer $SECRET" -H 'Content-Type: application/json' \
  -d '{"userId":"auto-routing-cli-local"}' | python3 -c 'import sys,json;print(json.load(sys.stdin)["token"])')
# route a request through kilo-auto/efficient
curl -s http://localhost:3000/api/openrouter/chat/completions \
  -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
  -d '{"model":"kilo-auto/efficient","messages":[{"role":"user","content":"Write a haiku about routing."}]}'
# worker admin (benchmark config / runs / routing table)
curl -s http://localhost:8814/admin/config -H "Authorization: Bearer $SECRET"
```

The benchmark worker is already wired into the dev runner (`pnpm dev:start auto-routing`).


services/auto-routing-benchmark/container/server.mjs
```diff
@@ -0,0 +1,157 @@
...
+          stdoutLines,
+          stderrTail: redactedStderrTail,
+        });
+      };
+
+      child.on('error', err => {
+        stderrTail = (stderrTail + `\nspawn error: ${err.message}`).slice(-STDERR_CAP_BYTES);
+        void finish(-1);
+      });
+      child.on('close', code => {
```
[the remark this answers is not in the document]
iscekic: Fixed in ba3b3be8b: `finish()` is now guarded by a `finished` flag, so the first of error/close wins and the spawn-error append can't race the resolve.

services/auto-routing-benchmark/src/index.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export default {
+  fetch: app.fetch,
+  async scheduled(controller: ScheduledController, env: Env, ctx: ExecutionContext): Promise<void> {
+    const kind = controller.cron === DECIDER_CRON ? 'decider' : 'classifier';
+    ctx.waitUntil(startRun(env, kind));
+  },
+  async queue(batch: MessageBatch<BenchmarkJobMessage>, env: Env): Promise<void> {
+    for (const message of batch.messages) {
+      await processJob(env, message.body);
```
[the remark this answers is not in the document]
iscekic: Documented in ba3b3be8b: the queue handler now carries a comment spelling out the throw-to-retry contract (no ack on throw → retry → DLQ after max_retries) and why adding a catch there would silently drop chunks.

services/auto-routing-benchmark/src/run.ts
```diff
@@ -0,0 +1,527 @@
...
+  const workers = Array.from({ length: Math.min(concurrency, queue.length) }, async () => {
+    for (let item = queue.shift(); item !== undefined; item = queue.shift()) {
+      await fn(item);
+    }
+  });
+  await Promise.all(workers);
+}
+
+async function finalizeRunIfComplete(env: Env, runId: string, kind: BenchmarkKind): Promise<void> {
+  const state = await getRunState(env, runId);
```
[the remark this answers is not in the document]
iscekic: Fixed in 82aef0be2 — `finalizeRunIfComplete` now takes the `RunState` its callers already hold (`processJob` passes its loaded state; `startRun`'s zero-enqueued path builds it from the rows it just wrote), removing the extra round-trip. Same commit also removes the live `getBenchmarkConfig` read in the decider publish path: the routing table is now built from the run's own model snapshot (`run_models` + the run's `min_accuracy`), restoring the snapshot-isolation the columns exist for.

apps/web/src/lib/ai-gateway/auto-routing-mirror.ts
```diff
@@ -4,11 +4,11 @@ import { after } from 'next/server';
 import { AUTO_ROUTING_WORKER_URL, INTERNAL_API_SECRET } from '@/lib/config.server';
 import { warnExceptInTest } from '@/lib/utils.server';
 
-type ScheduleAutoRoutingMirrorParams = {
+// Shared base params for both the mirror (fire-and-forget) and the
+// efficient-decision (blocking) call sites. Keep in sync with
+// EfficientDecisionParams in auto-routing-decision.ts.
```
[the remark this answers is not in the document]
iscekic: Fixed in a449c26eb — comment updated; `EfficientDecisionParams` is an alias of `DecideBaseParams` now, so there's nothing to keep in sync.

services/auto-routing-benchmark/src/admin.ts
```diff
@@ -0,0 +1,76 @@
...
+    const runs: BenchmarkRun[] = await listRuns(c.env.BENCH_DB, limit);
+    return c.json({ runs });
+  });
+
+  app.post(
+    '/admin/runs',
+    zodJsonValidator(StartBenchmarkRunRequestSchema, { errorMessage: 'Invalid run request' }),
+    async c => {
+      const { kind, force } = c.req.valid('json');
+      return c.json(await startRun(c.env, kind, { force }));
```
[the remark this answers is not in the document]
iscekic: Good catch — fixed in 165240b37. Lifted the `getBenchmarkConfig` null-check into the `/admin/runs` handler (mirroring the `/admin/debug-cli` guard) so a missing config now returns 400 instead of letting `startRun`'s precondition throw surface as a 500. Updated the test to assert 400 + the error body.

jeanduplessis (commented):
## Problem Statement
- `non-blocking`: Problem statement is technically clear, but "why now" could be sharper. Adding expected cost target/validation cohort/success bar for deciding when kilo-auto/efficient graduates from hidden/team validation to broader use would be helpful.

## Chosen Solution
- `blocking`: This PR adds new worker, D1 schema, queue, container runner, admin UI, gateway routing, and billing path in one merge. PR body explains the design, but not why this is simplest pragmatic solution versus smaller staged options. Adding a short "Alternatives considered" section covering why existing model-experiment tooling, offline benchmarks/manual routing tables, or a narrower first PR would not meet current goal would be helpful in making a judgement call on whether this solution is the right one right now.

## Risk
- `non-blocking`: Add rollback path info. Current checklist explains cutover, not rollback. For example adding a short note like: How to disable/hide kilo-auto/efficient, clear routing/classifier KV, stop benchmark runs, roll back worker if D1/predeploy fails.

## Architecture
- `blocking`: This adds new auto-routing product/domain rules, but there is no durable architecture doc. PR body explains the design well, but future agents and reviewers need repo-local source of truth after merge. Adding a  docs/adrs/xxxx-auto-routing.md or similar file covering the hidden kilo-auto/efficient contract, benchmark/routing-table ownership, fallback behavior, billing policy, sticky-session rule, and rollout/cutover invariants seems like a good idea.

## Operability
- `non-blocking`: Add rollback notes to the post-merge checklist: hide/stop using kilo-auto/efficient, clear classifier/routing KV overrides if needed, pause benchmark runs, and redeploy previous worker/gateway version.
- `non-blocking`: Add short DLQ runbook. The worker has auto-routing-benchmark-dlq, but reviewer cannot tell how failed chunks are inspected, replayed, or declared failed from the PR body.

## Verification
- `blocking`: Add a Verification section to the PR body with commands actually run and manual checks performed. For this PR size, as a reviewer I want to see explicit evidence that manual verification was exercised. There's also UI changes which could be documented with some screenshots. 
- `non-blocking`: Since this is a new feature, it would be good to have the steps to validate the feature in local development environment in the PR description for the reviewer.

## Agent Entropy
- `blocking`: Add durable agent-facing docs for auto-routing. Future agents need one source of truth for the benchmark worker, routing table contract, hidden model behavior, fallback guarantees, token boundary, and "what not to change" invariants. The PR body is not enough for a new domain this large.
- `non-blocking`: Add a short README in services/auto-routing-benchmark/ with local run commands, required env, and how to debug DLQ/container failures. 
- `non-blocking`: Consider adding the benchmark service as part of the dev environment runner script.

services/auto-routing-benchmark/src/run.ts
```diff
@@ -0,0 +1,605 @@
...
+      }))
+    );
+    return { runId, enqueuedModels: enqueuedModelIds.length, skippedModels };
+  }
+
+  // Decider: one message per (model, rep, chunk) so each queue invocation stays
+  // bounded. finalizeRunIfComplete expects enqueuedModels × DECIDER_CASES × repetitions rows.
+  const chunks = chunkArray(DECIDER_CASES, DECIDER_CHUNK_SIZE);
+  const messages = buildDeciderMessages(runId, kind, enqueuedModelIds, repetitions, chunks);
+  await env.BENCH_QUEUE.sendBatch(messages);
```
jeanduplessis: **[WARNING — MUST-FIX]** Decider fan-out can exceed Cloudflare Queue's 100-message `sendBatch` limit.

**Evidence:** This call sends the complete `models × repetitions × ceil(76 / 5)` fan-out in one batch.

**Trace:**
1. Seventy-six cases produce 16 messages per model per repetition.
2. Config permits up to five repetitions and does not cap decider-model count.
3. Two models with four repetitions produce 128 messages; seven models with one repetition produce 112.
4. Other queue producers in this repository explicitly split `sendBatch` calls at 100.
5. The run row is inserted before enqueue, so batch rejection leaves a `running` run without a complete result set.

**Impact:** Valid benchmark configurations fail to start and remain stuck until stale-run handling eventually executes.

**Fix direction:** Send messages in slices of at most 100. On partial enqueue failure, mark the run failed or make dispatch resumable.

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -261,8 +270,30 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
...
 
   let autoModel: string | null = null;
+  let classifierCostUsd = 0;
   if (isKiloAutoModel(requestedModelLowerCased)) {
     autoModel = requestedModelLowerCased;
+    const efficientDecision =
+      requestedModelLowerCased === KILO_AUTO_EFFICIENT_MODEL.id
+        ? async () => {
+            const user = (await authPromise).user;
+            const result = await fetchEfficientAutoDecision({
```
Comment (jeanduplessis) [inline apps/web/src/app/api/openrouter/[...path]/route.ts:287]: **[WARNING — MUST-FIX]** Paid classification runs before request access checks, leaving rejected requests unbilled.

**Evidence:** `fetchEfficientAutoDecision` can incur classifier cost here, while billing is not scheduled until after upstream inference.

**Trace:**
1. Efficient resolution calls `/decide` before IP, paid-auth, provider, API-kind, balance, organization, and abuse rejection paths.
2. `/decide` can perform a paid classifier call and return positive cost.
3. Classifier usage is scheduled only later in the successful upstream path.
4. Any intervening return drops the known classifier cost.
5. Tests cover successful upstream billing but not paid-classification early exits.

**Impact:** Rejected or unauthenticated requests can repeatedly consume Kilo-funded classifier inference without usage attribution.

**Fix direction:** Complete authentication and invariant access checks before paid classification, or schedule positive classifier cost through an exit-safe billing lifecycle immediately after the decision returns.

apps/web/src/app/admin/api/auto-routing/benchmark-config/route.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const parsed = BenchmarkConfigSchema.safeParse(rawBody);
+  if (!parsed.success) {
+    return NextResponse.json({ error: 'Invalid benchmark config' }, { status: 400 });
+  }
+
+  // Routing-table candidates carry no per-protocol metadata, so every decider
+  // model must be servable on ALL gateway chat API kinds by the provider the
+  // gateway would route it to.
+  const unsupported = parsed.data.deciderModels
```
jeanduplessis, line 58: **[WARNING — MUST-FIX]** Validation permits model-experiment public IDs in automatic routing.

**Evidence:** Candidate validation checks only static gateway protocol support.

**Trace:**
1. Unknown IDs, including experiment public IDs, are treated as OpenRouter-backed by `gatewayChatApisForModel`.
2. Actual provider resolution checks experiment membership first and serves active experiments through a direct provider supporting only `chat_completions`.
3. Saved IDs enter benchmark routing tables and can be automatically selected for `kilo-auto/efficient`.
4. `.specs/model-experiments.md` explicitly prohibits experiment IDs from automatic `kilo-auto` candidate sets.
5. Route tests cover protocol support but not experiment ownership.

**Impact:** Users selecting `kilo-auto/efficient` can be silently routed into preview experiment traffic without explicit opt-in; `responses` and `messages` can also fail despite config validation accepting the model.

**Fix direction:** Reject any decider candidate reserved by a model experiment. Add route-level tests covering active, paused, draft, and completed experiment IDs according to spec ownership rules.

services/auto-routing-benchmark/src/run.ts
```diff
@@ -0,0 +1,605 @@
...
+    kind === 'classifier' ? config.classifierRepetitions : config.deciderRepetitions;
+  const models =
+    kind === 'classifier' ? config.classifierModels : config.deciderModels.map(m => m.id);
+
+  // Models with prior results are skipped (their latest summaries are carried
+  // into this run's aggregate) unless the admin forces a full re-run.
+  const priorSummaries = options.force
+    ? new Map<string, BenchmarkModelSummary[]>()
+    : await getLatestSummariesByModel(env.BENCH_DB, kind);
+  const enqueuedModelIds = models.filter(m => !priorSummaries.has(m));
```
jeanduplessis: **[WARNING — MUST-FIX]** Carried summaries ignore changed benchmark identity and serving configuration.

**Evidence:** Prior results are considered reusable solely when the same model ID exists in the latest completed summaries.

**Trace:**
1. Current runs can change reasoning effort, repetitions, datasets, classifier prompt, CLI/image version, or execution settings.
2. A matching model ID is skipped even when those benchmark inputs changed.
3. Publishing then combines old measurements with the current `run_models.reasoning_effort`.
4. Production can therefore serve a configuration that was never measured by the carried summaries.
5. Tests do not cover carry invalidation after benchmark-identity changes.

**Impact:** Routing tables can select models using accuracy and cost measured under different serving conditions, causing inaccurate or unexpectedly expensive routing.

**Fix direction:** Persist a benchmark identity/version containing model, reasoning effort, dataset revision, repetitions, CLI/image version, and relevant execution inputs. Carry summaries only on an exact identity match.

services/auto-routing-benchmark/src/admin.ts
```diff
@@ -0,0 +1,83 @@
...
+    async c => {
+      const { kind, force } = c.req.valid('json');
+      const config = await getBenchmarkConfig(c.env.BENCH_DB);
+      if (!config) {
+        return c.json(
+          { error: 'benchmark config not set: save it in the admin panel before starting a run' },
+          400
+        );
+      }
+      return c.json(await startRun(c.env, kind, { force }));
```
jeanduplessis: **[WARNING — MUST-FIX]** Run admission lacks a coherent server-side state machine.

**Evidence:** This endpoint starts a run without atomically rejecting or superseding an existing `running` run of the same kind.

**Trace:**
1. The server sweeps stale runs only inside `startRun`.
2. The UI disables run buttons whenever a listed run is `running`, so a dead-lettered run prevents the only action that performs that sweep.
3. Concurrent direct requests or multiple tabs bypass client gating and can create overlapping runs.
4. Routing tables are selected by publication time.
5. An older run completing later can therefore replace a newer run's published table.

**Impact:** Failed runs can wedge the admin UI indefinitely, while concurrent starts can publish stale configuration over newer benchmark results.

**Fix direction:** Atomically sweep and claim one active run per kind on the server, expose stale recovery independently of starting a run, and prevent publication when a newer run exists.

packages/auto-routing-contracts/src/benchmark.ts
```diff
@@ -0,0 +1,119 @@
...
+  id: z.string().trim().min(1),
+  // Passed to the kilo CLI as --variant during the benchmark and carried into
+  // the routing table so serving uses the same effort the model was graded
+  // with. Null for models without (or not using) configurable reasoning.
+  reasoningEffort: ReasoningEffortSchema.nullable().default(null),
+});
+export type BenchmarkDeciderModel = z.infer<typeof BenchmarkDeciderModelSchema>;
+
+export const BenchmarkConfigSchema = z.object({
+  classifierModels: z.array(z.string().trim().min(1)).min(1),
```
jeanduplessis: **[WARNING — MUST-FIX]** Duplicate model IDs pass validation but violate D1 primary keys.

**Evidence:** Both model arrays validate entries and minimum length but not uniqueness.

**Trace:**
1. The UI permits duplicate classifier lines and duplicate decider rows.
2. `config_classifier_models.model` and `config_decider_models.model` are primary keys.
3. `replaceConfig` inserts arrays directly without deduplication or conflict handling.
4. Duplicate input therefore reaches D1 as a constraint violation.
5. Existing config tests do not cover duplicates.

**Impact:** A valid-looking admin save returns a server error instead of an actionable validation response.

**Fix direction:** Enforce unique normalized IDs in `BenchmarkConfigSchema` with field-specific issues, then add route and persistence regression tests.

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -839,6 +873,71 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
 
   accountForMicrodollarUsage(clonedReponse, usageContext, openrouterRequestSpan);
 
+  if (classifierCostUsd > 0 && !usageContext.user_byok) {
```
Comment (jeanduplessis) [inline apps/web/src/app/api/openrouter/[...path]/route.ts]: **[WARNING — MUST-FIX]** BYOK final inference suppresses billing for Kilo-funded classification.

**Evidence:** Positive classifier cost is discarded whenever the resolved inference provider uses user BYOK.

**Trace:**
1. Auto-routing classification uses the worker-owned OpenRouter credential.
2. Final provider BYOK status does not change classifier cost ownership.
3. This condition skips the classifier usage row for every BYOK final request.
4. The added test explicitly codifies no classifier billing for BYOK.
5. The PR contract says classifier LLM cost is billed to the requesting user without this exception.

**Impact:** BYOK users receive repeated Kilo-funded classification without usage attribution, creating systematic platform spend and a broken stated billing contract.

**Fix direction:** Base classifier billing on classifier credential ownership rather than final-provider BYOK status, or explicitly change the product contract and funding policy.

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -839,6 +873,71 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
...
+            promptInfo: {
+              system_prompt_prefix: '',
+              system_prompt_length: 0,
+              user_prompt_prefix: '',
+            },
+            max_tokens: null,
+            has_middle_out_transform: null,
+            isStreaming: false,
+            prior_microdollar_usage: usageContext.prior_microdollar_usage,
+            posthog_distinct_id: usageContext.posthog_distinct_id,
```
Comment (jeanduplessis) [inline apps/web/src/app/api/openrouter/[...path]/route.ts]: **[WARNING — MUST-FIX]** Classifier overhead participates in generic first-usage PostHog tracking.

**Evidence:** The overhead row reuses the primary request's `prior_microdollar_usage` and `posthog_distinct_id` before calling generic `logMicrodollarUsage`.

**Trace:**
1. Primary usage and classifier usage run as separate asynchronous `after()` tasks.
2. Both call `logMicrodollarUsage` with the same prior usage and PostHog identity.
3. `logMicrodollarUsage` checks `isFirstUsage` before insertion and emits generic `first_usage` and `first_microdollar_usage`.
4. Concurrent checks can both observe no usage, and classifier usage can insert before a streaming primary response completes.
5. Route tests mock `logMicrodollarUsage`, so event ordering is untested.

**Impact:** First-usage analytics can duplicate or identify `auto-routing/classifier` as the user's first model usage instead of the requested inference.

**Fix direction:** Suppress generic lifecycle PostHog events for internal classifier-overhead rows while retaining database billing. Add coverage proving one efficient request emits at most one primary first-usage event.

apps/web/src/app/admin/auto-routing/BenchmarksSection.tsx
```diff
@@ -0,0 +1,911 @@
...
+  config,
+  onSaved,
+}: {
+  config: BenchmarkConfig | null;
+  onSaved: (next: { config: BenchmarkConfig | null }) => void;
+}) {
+  const [form, setForm] = useState(() => configToFormState(config));
+
+  // Sync when config changes from outside (initial load / after save)
+  useEffect(() => {
```
jeanduplessis, line 215: **[WARNING — MUST-FIX]** Background config refresh can erase unsaved edits.

**Evidence:** Every `config` object change replaces the entire independent form state.

**Trace:**
1. Query-backed config is copied into local editable state.
2. This effect resets the whole form whenever query data changes.
3. Global QueryClient marks data stale after one minute and retains default focus refetch behavior.
4. External saves or focus refetches can return a new config while the local form is dirty.
5. No dirty-state or conflict check protects local edits.

**Impact:** Admins can lose partially entered model lists, benchmark user ID, and tuning values without warning.

**Fix direction:** Track dirty state and reset only after initial load or successful local save. Surface an explicit reload/conflict action for remote updates while dirty.

apps/web/src/app/admin/auto-routing/BenchmarksSection.tsx
```diff
@@ -0,0 +1,911 @@
...
+    onSuccess: (data, variables) => {
+      const kindLabel = variables.kind === 'classifier' ? 'Classifier' : 'Decider';
+      if (data.enqueuedModels === 0) {
+        toast.success(`All models already have results — republished from existing data`);
+      } else {
+        toast.success(
+          `${kindLabel} benchmark started — ${data.enqueuedModels} models enqueued, ${data.skippedModels.length} skipped`
+        );
+      }
+      void queryClient.invalidateQueries({ queryKey: ['auto-routing', 'benchmark-runs'] });
```
jeanduplessis, line 839: **[WARNING — MUST-FIX]** Completed runs do not refresh published routing output.

**Evidence:** Run start and polling refresh only the benchmark-runs query.

**Trace:**
1. Run finalization publishes a routing table or classifier winner.
2. Routing table data uses independent `['auto-routing', 'benchmark-routing-table']` state.
3. No transition handler invalidates derived output when polling observes completion.
4. The query therefore retains prior published data until focus refetch, manual reload, or unrelated invalidation.
5. Existing UI tests do not cover completion-driven cache invalidation.

**Impact:** Admins can see a completed benchmark while the published routing table still shows previous data or `No routing table published yet`.

**Fix direction:** Detect running-to-terminal transitions and invalidate the routing table for decider runs plus classifier-model/winner data for classifier runs.

apps/web/src/app/admin/auto-routing/BenchmarksSection.tsx
```diff
@@ -0,0 +1,911 @@
...
+    );
+  }
+
+  return (
+    <>
+      {runs.map(run => {
+        const expanded = expandedIds.has(run.id);
+        return (
+          <React.Fragment key={run.id}>
+            <TableRow
```
jeanduplessis: **[WARNING — MUST-FIX]** Expandable run rows are pointer-only.

**Evidence:** Click behavior and `aria-expanded` are attached to a native `<tr>`.

**Trace:**
1. `<tr>` is not focusable and does not activate with Enter or Space.
2. `aria-expanded` does not add button semantics or an accessible name.
3. The chevron cell contains no interactive control.
4. No alternate keyboard path expands summaries.

**Impact:** Keyboard and switch-device users cannot inspect run summaries; screen-reader users receive ambiguous expansion state.

**Fix direction:** Place a labeled button in the first cell with `aria-expanded` and `aria-controls`. Keep the row non-interactive or mirror row clicks to the same control.

apps/web/src/app/admin/auto-routing/BenchmarksSection.tsx
```diff
@@ -0,0 +1,911 @@
...
+        </TableCell>
+      </TableRow>
+    );
+  }
+
+  return (
+    <>
+      <TableRow className="bg-muted/30">
+        <TableCell colSpan={6} className="px-4 py-2">
+          <Table>
```
jeanduplessis: **[SUGGESTION — SUGGESTED]** Wide benchmark tables lack horizontal-overflow containment.

**Evidence:** The nested table contains up to ten non-wrapping headers plus a model identifier, while the shared `Table` primitive provides no overflow wrapper.

**Trace:**
1. Shared table headers use `whitespace-nowrap`.
2. This parent cell and card provide no local horizontal scrolling.
3. Narrow viewports therefore expand the entire admin surface rather than scrolling the data region.

**Impact:** Tablet, mobile, and narrow-laptop users may lose access to columns and surrounding controls.

**Fix direction:** Wrap wide tables in `overflow-x-auto` containers and give the tables appropriate minimum widths.

apps/web/src/app/admin/auto-routing/BenchmarksSection.tsx
```diff
@@ -0,0 +1,911 @@
...
+              </TableCell>
+              <TableCell className="py-2 capitalize text-sm">{run.kind}</TableCell>
+              <TableCell className="py-2">
+                <Badge variant={statusBadgeVariant(run.status)} className="capitalize">
+                  {run.status}
+                </Badge>
+              </TableCell>
+              <TableCell className="py-2 text-xs tabular-nums">{run.startedAt}</TableCell>
+              <TableCell className="py-2 text-xs tabular-nums">{run.completedAt ?? '—'}</TableCell>
+              <TableCell className="py-2 text-xs text-destructive max-w-48 truncate">
```
jeanduplessis: **[SUGGESTION — SUGGESTED]** Run errors are permanently truncated.

**Evidence:** The API provides the full error, but this cell applies a fixed maximum width and `truncate`.

**Trace:**
1. Expanded rows show model summaries only.
2. No tooltip, details panel, title, copy action, or accessible description exposes the omitted text.
3. Longer upstream and configuration errors therefore cannot be recovered from this UI.

**Impact:** Admins cannot read diagnostic details needed to investigate failed benchmark runs.

**Fix direction:** Show the full error in expanded details or provide an accessible tooltip/popover and copy action.

jeanduplessis (changes requested):
See inline findings.

services/auto-routing-benchmark/src/run.ts
iscekic: Fixed in 0844c4812. The decider fan-out (and the classifier fan-out, which is also unbounded by model count) now goes through `enqueueRunMessages`, which slices into `QUEUE_SEND_BATCH_LIMIT` (100) sized `sendBatch` calls — same pattern as `services/security-sync`. A mid-dispatch failure marks the run failed (visible in the admin panel) before rethrowing, rather than leaving a partially-enqueued run stuck in `running` until the 6h stale sweep. Added a regression test asserting a 112-message decider fan-out splits into [100, 12].

apps/web/src/app/api/openrouter/[...path]/route.ts
Comment (iscekic) [reply apps/web/src/app/api/openrouter/[...path]/route.ts]: Fixed in 0d9d6d2ea. The classifier-overhead row now passes `posthog_distinct_id: undefined`, so it can't emit the generic `first_usage` / `first_microdollar_usage` events (both are gated on `posthog_distinct_id` in `processUsage`) or race the primary usage row into mis-attributing `auto-routing/classifier` as the user's first model. DB billing is unaffected — it keys on `kiloUserId`. Added an assertion that the classifier context carries no distinct id.

apps/web/src/app/api/openrouter/[...path]/route.ts
Comment (iscekic) [reply apps/web/src/app/api/openrouter/[...path]/route.ts]: Fixed in f00b6196b. Classifier billing no longer depends on the final provider's BYOK status — the classifier always runs on Kilo's own OpenRouter credential, so the overhead is owed regardless. The guard is now `classifierCostUsd > 0` alone; the row remains `is_byok: false` / `user_byok: false`. Flipped the former "does not bill when user is BYOK" test to assert billing does occur for BYOK final inference.

apps/web/src/app/api/openrouter/[...path]/route.ts
Comment (iscekic) [reply apps/web/src/app/api/openrouter/[...path]/route.ts:287]: Fixed in a8d0cd7b8, addressing both halves of the concern:

1. **Unauthenticated requests no longer trigger paid classification.** The `/decide` call in the `efficientDecision` closure now bails out (→ balanced fallback) when there's no authenticated user. Since `kilo-auto/efficient` resolves to a paid model, unauthenticated callers are rejected downstream anyway, so this just stops Kilo-funded classifier spend on traffic there's no one to bill.

2. **Authenticated requests bill exit-safely.** The classifier-usage `after()` is now registered immediately after auth resolves — before the abuse block, provider/api-kind rejection, balance/org checks, and upstream-error paths — rather than at the tail of the success path. Any early return after that point still persists the row (`after()` keeps the instance alive).

Combined with the BYOK fix (f00b6196b), classifier cost is now attributed on `classifierCostUsd > 0` alone for any authenticated request. Added tests for the unauthenticated-skip path and a downstream-rejection (abuse-block) path that still bills.

packages/auto-routing-contracts/src/benchmark.ts
iscekic: Fixed in 188dfe70e. `BenchmarkConfigSchema` now has a `superRefine` that flags duplicate (trim-normalized) ids in both `classifierModels` and `deciderModels` with field-specific issues (`path: [list, index]`), so a duplicate save returns a 400 with an actionable message instead of hitting the D1 PK constraint as a 500. Added contract-level tests (duplicate classifier, trim-normalized duplicate decider, distinct-ids accepted) and a route-level test asserting PUT /admin/config returns 400 + does not call replaceConfig.

apps/web/src/app/admin/api/auto-routing/benchmark-config/route.ts
iscekic: Fixed in 4c0a18d1a. Benchmark-config save now rejects any decider candidate whose public id is owned by a model experiment, returning a 400. The new `findExperimentReservedModelIds` helper queries `model_experiment` with **no status filter**, so it covers ownership across draft/active/paused/completed — not just the active|paused Redis membership used on the hot path (`isPublicIdExperimented`). This enforces the `.specs/model-experiments.md` rule that experimented public ids must never enter kilo-auto candidate sets. Added a route test asserting the rejection and that the check runs against the decider model ids.

services/auto-routing-benchmark/src/run.ts
iscekic: Fixed in 75c762ab1. Carries are now identity-gated rather than model-id-only:

- Added `benchmark_runs.engine_identity` (migration 0001) = `v{ENGINE_VERSION}:{fnv1a(datasetSignature)}`, where the dataset signature is the case ids + grading checks/expectations, and `BENCHMARK_ENGINE_VERSION` is a manual lever for execution changes not visible in the dataset (grading logic, CLI invocation/variant handling, the image's pinned `@kilocode/cli`).
- `getLatestSummariesByModel` now also returns the prior run's `engineIdentity`, `repetitions`, and the model's `reasoning_effort`.
- `startRun` carries a model's prior summaries only when **all three** match the current run (engine identity, repetitions, reasoning effort); otherwise it re-benchmarks. This closes the specific hole where publishing combined old measurements with the current `run_models.reasoning_effort`.

On the CLI/image-version point: the container pins `@kilocode/cli@latest` at build time, which the worker can't read at runtime, so that dimension is covered by bumping `BENCHMARK_ENGINE_VERSION` on a CLI-affecting redeploy (legacy rows with an empty identity also never match, so they re-benchmark). Added a `computeEngineIdentity` unit test plus carry-match / effort-mismatch integration tests.

services/auto-routing-benchmark/src/admin.ts
iscekic: Fixed in 3166effa2 with a coherent server-side admission state machine:

1. **Atomic one-active-run-per-kind.** Added a partial unique index (`UQ_benchmark_runs_one_running_per_kind` on `kind` WHERE status='running') as the atomic backstop. `startRun` pre-checks via `getRunningRun` and throws `RunAlreadyActiveError` → the admin route returns **409** (not a 500). Concurrent POSTs/tabs that slip past the pre-check lose the insert on the index and are also translated to 409. This prevents overlapping runs.
2. **Stale recovery independent of starting.** `sweepStaleRuns` now runs on `GET /admin/runs` too, so a dead/dead-lettered run surfaces as `failed` (freeing the slot) without needing a new run — fixes the deadlock where the UI disables Start while a run shows `running`.
3. **No stale publication.** `finalizeRunIfComplete` skips publishing the routing table / classifier winner when `existsNewerCompletedRun` is true, so a slow older run can't overwrite a newer run's published results (publication is selected by publish time). The run is still marked `completed`; only its publication is suppressed.

Added tests for the 409 conflict and the sweep-on-list. (Also squashed the branch's D1 migrations to a single baseline.)

apps/web/src/app/admin/auto-routing/BenchmarksSection.tsx
iscekic: Fixed in b19c57a11. The config editor now tracks a `dirty` flag: every user edit goes through an `updateForm` wrapper that sets it, and the sync-from-server effect only resets the form while pristine — so a background refetch (the runs poll / focus refetch) can't discard in-progress edits. After a successful save the form re-syncs and dirty clears. Added an explicit "Discard & reload" action (with an "Unsaved changes" indicator) for the conflict case.

apps/web/src/app/admin/auto-routing/BenchmarksSection.tsx
iscekic: Fixed in b19c57a11. Added an effect that watches the running→terminal edge (via a `prevHasRunningRun` ref) and, when the last running run finishes, invalidates the `['auto-routing','benchmark-routing-table']` and `['auto-routing','benchmark-config']` queries. Run completion publishes the routing table / classifier winner, so the published-output view now refreshes instead of showing stale data or "No routing table published yet" until a focus refetch.

apps/web/src/app/admin/auto-routing/BenchmarksSection.tsx
iscekic: Fixed in b19c57a11. The first cell now holds a real `<button>` with `aria-expanded`, `aria-controls` (pointing at the expanded summaries region's id), and an accessible label ("Expand/Collapse {kind} run details"). It's keyboard-focusable and toggles on Enter/Space. The row `onClick` is kept purely as a mouse convenience (the button stops propagation so it doesn't double-toggle), and `aria-expanded` was removed from the `<tr>` so the button is the single source of expansion state.

apps/web/src/app/admin/auto-routing/BenchmarksSection.tsx
iscekic: Fixed in b19c57a11. Wrapped the wide nested summary table and the routing-table tier tables in `overflow-x-auto` containers and gave those tables `min-w-max` so they scroll horizontally within their card instead of expanding the whole admin surface on narrow viewports.

apps/web/src/app/admin/auto-routing/BenchmarksSection.tsx
iscekic: Fixed in b19c57a11. The expanded run row now renders the full error text (a `whitespace-pre-wrap break-words` block) above the summaries, so failed runs are fully diagnosable from the UI. Also added a `title` tooltip on the truncated Error cell in the collapsed row for a quick hover read.

iscekic:
Addressed the points from this review (commit 637d695f1 + PR body update):

**Architecture / Agent Entropy (blocking)** — Added `docs/adr/0002-auto-routing-efficient.md`: the hidden `kilo-auto/efficient` contract, benchmark/routing-table ownership (the benchmark worker is sole writer), fallback guarantees, billing policy, sticky-session rule, token boundary, rollout/cutover, and an explicit "invariants / what not to change" section. Added `services/auto-routing-benchmark/README.md` with local run commands, required env/secrets, D1 inspection, and container/DLQ debugging.

**Chosen Solution (blocking)** — Added an "Alternatives considered" section to the PR body (and the ADR): why model-experiment tooling, offline/hand-maintained routing tables, a narrower first PR, and LLM-judged grading were each rejected, and why the smallest *shippable* unit is the full loop behind a hidden model.

**Verification (blocking)** — Added a "Verification" section to the PR body: the automated suites run (with which comment each covers) and the manual end-to-end checks performed against the local stack — including the real `kilo-auto/efficient` request that produced both a final-inference row and a separate `auto-routing/classifier` billing row, and the C3/C6/C5 admin-path checks. Also added a copy-pasteable "Validate locally" block.

**Problem Statement — "why now" (non-blocking)** — Added a "Why now" paragraph: the cost-reduction goal, the team-traffic validation cohort, and the graduation signals (cost down vs. balanced, accuracy not worse, fallback rate not regressed). Graduation criteria live in the ADR, not in code.

**Risk / Operability — rollback (non-blocking)** — Added a "Rollback" section (PR body + ADR): disable the hidden model (degrades to balanced), clear the routing-table/classifier-winner KV keys, stop benchmark runs, redeploy prior worker/gateway versions; the additive D1 schema means a failed predeploy migration fails the deploy before serving.

**Operability — DLQ runbook (non-blocking)** — Added a "DLQ runbook" section (PR body + README): a DLQ message is one (model, chunk) job; how to inspect it, how to replay via the admin force toggle, and how a wedged `running` run is swept to `failed`.

**Operability — dev runner (non-blocking)** — The benchmark service is already part of the dev runner (`pnpm dev:start auto-routing` brings up the gateway, auto-routing worker, and this worker); documented in the README.

UI screenshots: the admin panel is exercised in the local verification above; happy to attach screenshots if you'd like them in the body.

jeanduplessis (approved):
I ran this in local dev and managed to get the benchmarks running. Since this feature is still locked behind the kilo org, the risk to roll this out seems low.

Tech debt note: The admin UI is not really easy to understand and a user with no prior knowledge of the feature will likely be lost. There's a lot of knobs and inputs that you can tweak that could have real impact. We might want to do a UX pass over it at some point once the feature is more settled.

PR #3990, opened 2026-06-12
fix(code-review): add model diagnostics

alex-alecu:
## Summary
- Adds private runtime diagnostics when code reviews fail because a model is unavailable, so support can see the requested model, live model count, and close matches without showing those details publicly.
- Keeps GitHub, GitLab, and public Cloud Agent messages generic while the signed-in dashboard gets a clearer private error.
- Does the live model lookup in the wrapper and gives it a short timeout so a slow lookup cannot block failure reporting.
- Expands `isModelNotFoundMessage` regex to also match `unknown model` and `invalid model` patterns, keeping it in sync with `classifyAssistantFailureMessage` in `safe-failure-projection.ts` so those error variants no longer skip diagnostics.

## Verification
Manual staging checks:
1. Start a code review with a model that is not available.
2. Confirm the signed-in dashboard shows the missing model, model count, and up to five alternatives.
3. Confirm GitHub or GitLab only shows the generic model-not-found message.
4. Search logs and Sentry by the review id and confirm they include the full runtime model list.

## Visual Changes
N/A

## Reviewer Notes
- The full model list is private diagnostic data and should not appear in public review output.
- The wrapper drops diagnostics if the live model lookup is slow, so terminal failure reporting keeps moving.
- `isModelNotFoundMessage` now mirrors the regex pattern in `classifyAssistantFailureMessage`: `/\b(model (?:was )?not found|unknown model|invalid model)\b/i`.

Built for alex-alecu by [Kilo](https://kilo.ai)


services/cloud-agent-next/wrapper/package.json
```diff
@@ -8,7 +8,8 @@
     "typecheck": "tsgo --noEmit"
   },
   "dependencies": {
-    "@kilocode/sdk": "7.3.21"
+    "@kilocode/sdk": "7.3.21",
+    "fuzzysort": "3.1.0"
```
alex-alecu, line 12: @eshurakov how do you feel about adding a new dependency? When a model is not found, I do a fuzzy search on the available runtime (CLI) models against the one set in web and return just a few.

services/cloud-agent-next/wrapper/src/connection.ts
```diff
@@ -803,6 +807,82 @@ export function createConnectionManager(
     return `Insufficient credits: ${eventType}`;
   }
 
+  function isModelNotFoundMessage(message: string): boolean {
+    return /\bmodel\s+(?:was\s+)?not\s+found\b/i.test(message);
```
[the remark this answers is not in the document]
alex-alecu: @kilocode-bot fix it

services/cloud-agent-next/wrapper/src/connection.ts
```diff
@@ -762,15 +766,15 @@ export function createConnectionManager(
       return undefined;
     }
 
+    const terminalErrorText = getTerminalErrorText(eventType, properties);
     let code: WrapperTerminalFailureCode | undefined;
     if (eventType === 'payment_required' || eventType === 'insufficient_funds') {
       code = 'payment_required';
     } else if (eventType !== 'usage_limit_exceeded') {
```
alex-alecu, line 773: Could this keep inspecting the full error object for billing markers? `getTerminalErrorText` selects one message, so a session error like `{ name: "PaymentRequiredError", data: { message: "Request failed" } }` was terminal before this change but is now ignored.

alex-alecu: Fixed in 957f429.

pandemicsyn (approved):

PR #4000, opened 2026-06-12
Deprecate Codestral-only keys

chrarnoldus:
## Summary

- Allow standard Mistral BYOK keys to route Codestral models through Mistral's FIM endpoint.
- Preserve existing legacy Codestral-only keys, including their dedicated endpoint and priority when both key types are configured.
- Stop offering legacy Codestral-only keys in the add-key dialog while keeping existing keys visible and editable.
- Include Codestral models in the supported-model metadata shown for Mistral BYOK.

## Verification

- Not manually tested; the changes were verified with targeted linting and web type checking during implementation.

## Visual Changes

N/A — the legacy Codestral-only provider option is removed from the add-key dropdown; existing key management remains unchanged.

## Reviewer Notes

Existing Codestral-only keys remain backward compatible and take priority over standard Mistral keys. New configurations use the standard Mistral provider.

### DIFF
--- a/apps/web/src/lib/ai-gateway/byok/index.ts
+++ b/apps/web/src/lib/ai-gateway/byok/index.ts
@@ -15,21 +15,21 @@ import type { BYOKResult } from '@/lib/ai-gateway/providers/types';
 import { getVercelModelsMetadata } from '@/lib/ai-gateway/providers/gateway-models-cache';
 
 export async function getModelUserByokProviders(modelId: string): Promise<UserByokProviderId[]> {
-  if (isCodestralModel(modelId)) {
-    return ['codestral'];
-  }
   const vercelModelMetadata = await getVercelModelsMetadata();
   if (Object.keys(vercelModelMetadata).length === 0) {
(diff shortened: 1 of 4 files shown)


apps/web/src/lib/ai-gateway/byok/index.ts
```diff
@@ -15,22 +15,22 @@ import type { BYOKResult } from '@/lib/ai-gateway/providers/types';
...
+  const providers: UserByokProviderId[] =
     vercelModelMetadata[mapModelIdToVercel(modelId, false)]?.endpoints
       .map(ep => VercelUserByokInferenceProviderIdSchema.safeParse(ep.provider_name ?? ep.tag).data)
       .filter(providerId => providerId !== undefined) ?? [];
   if (providers.length === 0) {
     console.debug(`[getModelUserByokProviders] no user byok providers for ${modelId}`);
     return [];
   }
+  if (isCodestralModel(modelId)) {
+    providers.splice(0, 0, 'codestral');
```
chrarnoldus:
```suggestion
    providers.unshift('codestral');
```

markijbema (approved):

PR #4034, opened 2026-06-15
Feat/kiloclaw agents bindings and reconcile

St0rmz1:
## Summary

Adds agent management to the KiloClaw agents page: route channels to agents and
set each agent's model inline on the list, edit the inherited defaults, and edit
per-agent advanced behavior. Agent config edits do not hot-reload the running
gateway, so the page tracks unapplied changes and offers a single restart to
apply them. Agent mutations are also made resilient to edge-gateway timeouts that
may have already applied server-side, and the create path is trimmed to avoid a
redundant controller subprocess on its latency-sensitive round trip.

## Changes

Agents page UI
- Move model and channel routing onto each agent row (`AgentInlineControls`,
  `AgentsSection`): a model picker and channel toggle chips, instead of a
  separate bindings dialog.
- Reduce the per-agent edit dialog (`AgentEditDialog`) to advanced behavior only
  (thinking, reasoning, verbose, fast mode), each with inline helper text wired
  via `aria-describedby`.
- Add an inherited-defaults editor (`AgentDefaultsDialog`) for the default model,
  thinking, and verbose that agents inherit unless they set their own.
- Simplify the create dialog (`AgentCreateDialog`).
- `ConfirmActionDialog` gains a semantic `confirmVariant` (e.g. destructive) and
  a configurable `cancelLabel` so the delete confirm can read "Keep agent".

Apply-on-restart
- Agent config changes persist immediately but only take effect after a gateway
  restart. The page counts unapplied changes per instance (persisted in
  localStorage, keyed by the Postgres instance id) and shows a "Restart OpenClaw"
  banner. Inline saves show loading toasts because the round trips can be slow.

Resilient mutations
- Create, update, delete, bindings, and defaults mutations reconcile ambiguous
  failures: on an ambiguous error they refetch the agent list and check whether
  the intended end state holds, so a gateway timeout that already applied
  server-side is not reported as a failure.
- `isAmbiguousAgentMutationError` distinguishes a genuine timeout (which may have
  applied) from an explicit controller failure. The tRPC routers propagate the
  controller's error code as `data.upstreamCode`, and the classifier reconciles
  `INTERNAL_SERVER_ERROR` unless the upstream code is a definite failure
  (`openclaw_cli_failed`, `agent_binding_rollback_failed`).
- Every mutation invalidates the list on success and reconciles on its own error
  path, so the list is fetched once per outcome (no duplicate refetch).
- Gate the OpenRouter model-catalog fetch behind `enabled` (`useClawModelOptions`,
  `useModelSelectorList`) so it does not fire on a stopped machine or for
  read-only users.

Controller
- On create, skip the extra `openclaw agents bindings` subprocess when the create
  requested no `--bind`; an agent created without bindings has none to report, and
  the client refetches the authoritative list regardless. Updates controller spec
  rule 17 to match.
- Document `normalizeAgentId` as the authoritative implementation and its web-side
  mirror.

Agent-id parity
- The web app re-declares `normalizeAgentId` because the architecture wall blocks
  importing controller code. Both test suites now load one shared corpus
  (`agent-id-corpus.json`) so the two implementations cannot silently drift.

Tests
- Classifier and reconcile tests (`useClawHooks.test.ts`), shared agent-id parity
  corpus on both sides, and controller create/bindings tests
  (`config-agents.test.ts`, `openclaw-agent-config.test.ts`).

## Verification

- [x] Routed Discord and Telegram between agents from the list, restarted, and
      confirmed each agent answered on its assigned channel.
- [x] Confirmed a model change took effect only after restarting OpenClaw.
- [x] Confirmed the "restart required" banner counts unapplied edits and clears
      after a restart.

## Visual Changes

| Before | After |
|---|---|
|  |  |

## Reviewer Notes

- The web `normalizeAgentId` is a deliberate mirror of the controller's; the
  shared corpus is the drift guard. Changing the rule means updating the shared
  file plus both implementations.
- Reconcile correctness hinges on the upstream-code deny-list: a definite
  controller failure must never be reconciled into a false success.


apps/web/src/app/(app)/claw/components/AgentDefaultsDialog.tsx
```diff
@@ -0,0 +1,318 @@
...
+
+  const hasChanges = Object.keys(patch.set).length > 0 || patch.unset.length > 0;
+  const canSubmit = hasChanges && !modelInvalid && !updateDefaults.isPending;
+
+  // Does a refetched defaults snapshot reflect the patch we tried to write?
+  // (Primary + thinking/verbose + unsets are the signal; deep fallback compare is
+  // skipped — a primary match is enough to confirm the write landed.)
+  const defaultsApplied = (d: AgentDefaultsSummary): boolean => {
+    const { set, unset } = patch;
+    if (set.model?.primary !== undefined && d.model?.primary !== set.model.primary) return false;
```
pandemicsyn: Must-fix: this reconcile check can turn a fallback-only defaults update into a false success. When `set.model` only contains `fallbacks`, no model field is checked, so any refetched defaults snapshot passes even if the fallback state did not land. Please compare the whole requested model shape, including expected primary and exact fallback array equality.

apps/web/src/app/(app)/claw/components/AgentInlineControls.tsx
```diff
@@ -0,0 +1,186 @@
...
+  return (
+    <div className="flex flex-wrap items-center gap-1.5">
+      {catalog.map(channel => {
+        const isSelected = selected.has(channel.id);
+        const owner = channelOwner.get(channel.id);
+        // Owned by another agent (not this one) → can't route here until moved.
+        const ownedElsewhere = owner !== undefined && !isSelected;
+        const blocked = saving || ownedElsewhere;
+        return (
+          <button
```
pandemicsyn: Must-fix: disabled chips rely on `title` for the reason they cannot be selected, but disabled buttons are not keyboard-focusable and `title` is hover-only/unreliable on touch. Please expose the reason via visible text or an accessible description, or use a focusable `aria-disabled` control with `aria-describedby`.

apps/web/src/app/(app)/claw/components/AgentsSection.tsx
```diff
@@ -250,6 +602,19 @@ export function AgentsSection({
...
+        />
+      )}
+
+      {defaultsTarget && data && (
+        <AgentDefaultsDialog
+          open
+          onOpenChange={open => {
+            if (!open) setDefaultsTarget(null);
+          }}
+          defaults={data.defaults}
```
pandemicsyn: Suggested/nit: this freezes only the `etag`; `defaults` still comes from live query data while the dialog is open. A background refetch can change the baseline under the mounted form and make a stale patch look valid. Consider storing `{ defaults, etag }` together in `defaultsTarget`, or rebasing/resetting the form when the target changes.

apps/web/src/app/(app)/claw/components/AgentEditDialog.tsx
```diff
@@ -227,76 +234,38 @@ export function AgentEditDialog({
     <Dialog open={open} onOpenChange={updateAgent.isPending ? undefined : onOpenChange}>
       <DialogContent className="max-w-md">
```
pandemicsyn: Suggested/nit: this dialog now has several always-visible helper paragraphs but no viewport height cap or scroll region. On short mobile viewports or larger text settings, the footer can be pushed off-screen. Please add a mobile-safe max height and scroll behavior for the dialog/body.

apps/web/src/app/(app)/claw/components/AgentsSection.tsx
```diff
@@ -146,63 +241,294 @@ function DefaultsRow({ defaults }: { defaults: AgentDefaultsSummary }) {
     : 'none';
 
   return (
-    <div className="text-muted-foreground bg-muted/30 px-4 py-3 text-xs">
-      <span className="font-medium">Inherited defaults</span> · Model: {modelLabel}
-      {settings.length > 0 && ` · ${settings.join(' · ')}`}
+    <div className="bg-muted/30 flex items-center justify-between gap-2 px-4 py-3">
```
pandemicsyn, line 244: Suggested/nit: long model identifiers can overflow this single non-wrapping row and push the edit button off-screen on narrow viewports. Consider adding `min-w-0` plus wrapping/truncation, or stacking this row on small screens.

pandemicsyn (approved):

apps/web/src/app/(app)/claw/components/AgentDefaultsDialog.tsx
```diff
@@ -0,0 +1,330 @@
...
+  // primary AND exact fallback array for every field the patch actually set.
+  const defaultsApplied = (d: AgentDefaultsSummary): boolean => {
+    const { set, unset } = patch;
+    if (set.model !== undefined) {
+      if (d.model == null) return false;
+      if (set.model.primary !== undefined && d.model.primary !== set.model.primary) return false;
+      if (set.model.fallbacks !== undefined) {
+        const want = set.model.fallbacks;
+        const got = d.model.fallbacks;
+        if (got.length !== want.length) return false;
```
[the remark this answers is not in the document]
St0rmz1: False positive: the optional `{ fallbacks?: string[] }` shape is the input/patch type (`set.model` / `ModelInputSchema`), not the refetched summary. Here `d` is `AgentDefaultsSummary`, whose `model` is `AgentModelSummary | null`:

```ts
export type AgentModelSummary = {
  primary: string | null;
  fallbacks: string[]; // required, not optional
};
```

After the `d.model == null` guard, `d.model.fallbacks` is `string[]`, never `undefined`, so `.length` can't throw. Typecheck passing on `got.length` with no optional chaining confirms it (under strict TS, `string[] | undefined` would be a compile error there). Leaving as-is.

PR #4047, opened 2026-06-16
fix(kiloclaw): prevent checkout settlement quarantine

jeanduplessis:
## Summary
Prevent KiloClaw checkout flows from charging or deducting against local subscription lineage that cannot settle canonically.

### Why this change is needed
Canceled legacy KiloClaw rows exposed two incompatible paths. Kilo Pass checkout selected current pricing, but return-page activation compared that intent with the historical row price and failed after Kilo Pass credits settled. A later user-initiated direct Stripe checkout could create and charge a current-price subscription even though both canonical webhook handlers rejected the local target, leaving Stripe active while local access remained canceled.

### How this is addressed
- Apply one authoritative enrollment price-version policy across checkout previews, Kilo Pass metadata, activation, and credit enrollment.
- Reject canceled legacy bundled-hosting checkout before Kilo Pass payment instead of rewriting immutable historical lineage.
- Preserve credits and the complete canceled predecessor for already-completed legacy sessions, returning a permanent support-led reprovision outcome without a hosting deduction.
- Require settled Kilo Pass base credits and revalidate instance, ownership, lineage, status, and price intent before deducting credits.
- Return typed activation outcomes so only transient settlement or operational failures offer same-session retry; permanent stale, invalid, missing, destroyed, and reprovision-required states receive distinct recovery guidance.
- Reject missing, destroyed, or incompatible canceled legacy targets before direct Stripe Checkout creates a provider session.
- Preserve valid trial, current-price, pure-credit, and intentional direct-Stripe paths, including both webhook delivery orders.
- Distinguish credit-funded hosting recovery from a separate recurring Stripe hosting subscription in billing UI.

## Human Verification
- Reproduced the original failure before implementation: canceled legacy row with current-price Kilo Pass metadata returned `Checkout hosting price version no longer matches instance`.
- Added failing review regressions proving legacy activation rewrote `2026-03-19` to `2026-05-10` and permanent activation failures were not classified for recovery.
- Verified KiloClaw billing, Kilo Pass activation, pricing policy, webhook-order, destroyed-anchor, immutable-lineage, and recovery-copy coverage across 5 focused suites with 340 passing tests.
- Reviewed behavior against KiloClaw billing, billing lifecycle, and data-model specifications.
- Ran React Doctor on changed UI files; score remained 84/100 with no regression.

### Human Reviewer Flags
- Existing one-subscription-per-instance schema makes safe same-instance successor creation non-trivial. This PR chooses containment: new canceled-legacy bundled checkouts fail before payment; already-paid sessions retain awarded credits and route to support-led fresh provisioning.
- No canceled legacy row has its price version, provider ownership, status, transfer pointer, or change log rewritten by recovery.
- Incompatible canceled legacy rows use the same fail-before-provider approach for direct Stripe checkout and direct credit enrollment.
- This change does not cancel, refund, or reconcile provider subscriptions or Checkout Sessions created before deployment. Existing production divergence remains an operational incident task.
- No production systems or production data were changed.

### Code Reviewer Agent
<details>

- Shared price resolution preserves live trial lineage pricing and selects current pricing for fresh enrollment intent after canceled history.
- The credit-enrollment transaction detects duplicate retries first, then rolls back new deductions when canceled legacy history requires a fresh billing anchor.
- Kilo Pass activation requires a matching base issuance item, validates typed checkout metadata, rejects destroyed or organization-owned targets, and returns an exhaustive retryable/action-required outcome union.
- Canceled legacy Kilo Pass checkout is rejected before provider session creation; completed historical sessions return `requires_reprovision` without local billing mutation.
- Direct Stripe preflight runs before provider subscription/session lookups or creation for unsafe local targets.
- `subscription.created processed` emits only when a local mutation occurred.

</details>

### DIFF
--- a/apps/web/src/app/(app)/claw/components/billing/billing-types.ts
+++ b/apps/web/src/app/(app)/claw/components/billing/billing-types.ts
@@ -262,6 +262,63 @@ export function formatKiloClawFirstChargeLabel(params: {
   return `${price}/month`;
 }
 
+export function getKiloClawFundingChoiceCopy(params: {
+  plan: ClawPlan;
+  costMicrodollars: number;
+}): {
+  creditHeading: string;
+  creditDescription: string;
+  creditButtonLabel: string;
+  stripeDividerLabel: string;
+  stripeButtonLabel: string;
+  stripeDescription: string;
+} {
+  const planLabel = params.plan === 'commit' ? 'Commit' : 'Standard';
+  const priceLabel = formatKiloClawFirstChargeLabel(params);
+
+  return {
+    creditHeading: 'Credit-funded hosting',
+    creditDescription: `${planLabel} first charge: ${priceLabel}. Future hosting charges use your credit balance.`,
+    creditButtonLabel: `Activate ${planLabel} with credits`,
+    stripeDividerLabel: 'or start a separate Stripe subscription',
+    stripeButtonLabel: `Subscribe with Stripe, ${priceLabel}`,
+    stripeDescription:
+      params.plan === 'standard'
+        ? 'Creates a separate recurring Stripe charge for hosting.'
+        : 'Creates separate Stripe billing for hosting.',
+  };
+}
+
+export function getKiloPassHostingRecoveryCopy(
+  hostingIntent: 'none' | 'expired_commit' | 'standard' | 'commit'
+): {
+  title: string;
+  description: string;
+  destination: string;
+  destinationLabel: string;
+  canRetry: boolean;
+} {
+  if (hostingIntent === 'expired_commit') {
+    return {
+      title: 'Kilo Pass credits are ready',
+      description:
+        'Commit hosting is no longer available. Choose Standard and fund hosting from your credit balance.',
+      destination: '/claw/subscription',
+      destinationLabel: 'Choose Standard hosting',
+      canRetry: false,
+    };
+  }
+
+  return {
+    title: 'Credit-funded hosting needs attention',
+    description:
+      'Your Kilo Pass credits are ready, but hosting activation did not finish. Retry credit-funded activation or choose a hosting plan.',
+    destination: '/claw/subscription',
+    destinationLabel: 'Choose hosting plan',
+    canRetry: true,
+  };
+}
+
 /** e.g. "Commit ($51/mo)" or "Standard ($55/mo)" */
 export function planLabel(plan: ClawPlan, priceVersion?: string): string {
   if (plan === 'commit') {
--- a/apps/web/src/app/(app)/claw/components/billing/billing-types.ts
+++ b/apps/web/src/app/(app)/claw/components/billing/billing-types.ts
@@ -289,34 +289,128 @@ export function getKiloClawFundingChoiceCopy(params: {
   };
 }
 
-export function getKiloPassHostingRecoveryCopy(
-  hostingIntent: 'none' | 'expired_commit' | 'standard' | 'commit'
-): {
+export type KiloPassHostingRecoveryReason =
+  | 'credits_not_settled'
+  | 'enrollment_failed'
+  | 'requires_reprovision'
+  | 'missing_instance'
+  | 'destroyed_instance'
+  | 'stale_intent'
+  | 'invalid_intent'
+  | 'insufficient_credits'
+  | 'expired_commit'
+  | 'unexpected_error';
+
+export function getKiloPassHostingRecoveryCopy(reason: KiloPassHostingRecoveryReason): {
   title: string;
   description: string;
-  destination: string;
-  destinationLabel: string;
+  destination: string | null;
+  destinationLabel: string | null;
   canRetry: boolean;
+  showSupport: boolean;
 } {
-  if (hostingIntent === 'expired_commit') {
-    return {
-      title: 'Kilo Pass credits are ready',
-      description:
-        'Commit hosting is no longer available. Choose Standard and fund hosting from your credit balance.',
-      destination: '/claw/subscription',
-      destinationLabel: 'Choose Standard hosting',
-      canRetry: false,
-    };
+  switch (reason) {
+    case 'credits_not_settled':
+      return {
+        title: 'Hosting credits are still processing',
+        description:
+          'Your Kilo Pass payment is complete, but hosting credits are not ready yet. Retry credit-funded activation shortly.',
+        destination: null,
+        destinationLabel: null,
+        canRetry: true,
+        showSupport: false,
+      };
+    case 'enrollment_failed':
+      return {
+        title: 'Credit-funded hosting needs attention',
+        description:
+          'Your credits are ready, but hosting activation did not finish. Retry credit-funded activation or choose a hosting plan.',
+        destination: '/claw/subscription',
+        destinationLabel: 'Choose hosting plan',
+        canRetry: true,
+        showSupport: true,
+      };
+    case 'requires_reprovision':
+      return {
+        title: 'Reprovision before activating hosting',
+        description:
+          'Your Kilo Pass credits are ready. Your previous KiloClaw billing history must remain unchanged, so support must help provision a fresh KiloClaw before hosting can use those credits.',
+        destination: null,
+        destinationLabel: null,
+        canRetry: false,
+        showSupport: true,
+      };
+    case 'missing_instance':
+      return {
+        title: 'Provision KiloClaw before activating hosting',
+        description:
+          'Your Kilo Pass credits are ready, but no KiloClaw instance is available for hosting. Contact support to provision one without risking billing history.',
+        destination: null,
+        destinationLabel: null,
+        canRetry: false,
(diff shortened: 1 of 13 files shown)


apps/web/src/lib/kiloclaw/credit-billing.ts
```diff
@@ -1249,9 +1271,19 @@ export async function enrollWithCredits(params: {
   }
 
   const isLiveTrialLineage = existingSub?.status === 'trialing';
-  const kiloclawPriceVersion = isLiveTrialLineage
-    ? existingSub.kiloclaw_price_version
-    : CURRENT_KILOCLAW_PRICE_VERSION;
+  const kiloclawPriceVersion = resolveKiloClawEnrollmentPriceVersion(
```
jeanduplessis, line 1319: **[WARNING - MUST-FIX] Canceled legacy activation rewrites existing lineage immutable price version**

Trace:
1. This resolver returns current pricing for every non-trialing row, including canceled legacy rows.
2. `enrollWithCredits` then uses that version in existing `ON CONFLICT(instance_id) DO UPDATE` behavior, overwriting the same subscription row. Changed test coverage explicitly expects `2026-03-19` to become `2026-05-10`.
3. `.specs/kiloclaw-billing.md` and `.specs/kiloclaw-datamodel.md` require canceled rows to retain historical price version and later enrollment to create a fresh current-price lineage.

Impact: Canonical row loses immutable lineage evidence. Reporting, reconciliation, support, entitlement, and retirement logic can no longer rely on `kiloclaw_price_version` as historical lineage identity.

Fix direction: Preserve canceled predecessor unchanged and create fresh current-price lineage through a new billing anchor or reprovision path. If audited in-place reset is intended product behavior, update authoritative specs and audit every lineage consumer before merging.

apps/web/src/app/(app)/claw/components/billing/billing-types.ts
```diff
@@ -262,6 +262,63 @@ export function formatKiloClawFirstChargeLabel(params: {
...
+    };
+  }
+
+  return {
+    title: 'Credit-funded hosting needs attention',
+    description:
+      'Your Kilo Pass credits are ready, but hosting activation did not finish. Retry credit-funded activation or choose a hosting plan.',
+    destination: '/claw/subscription',
+    destinationLabel: 'Choose hosting plan',
+    canRetry: true,
```
jeanduplessis: **[WARNING - MUST-FIX] Permanent activation failures are presented as retryable**

Trace:
1. Recovery behavior is derived only from original `hostingIntent`; every `standard` or `commit` failure receives `canRetry: true`.
2. Server returns permanent failures for invalid metadata, missing instance, destroyed billing anchor, and stale price intent. Retrying same Checkout Session cannot repair these states.
3. Client clears error and repeats same mutation. For destroyed or missing anchors, secondary `/claw/subscription` path also targets billing flows that reject unusable anchor.

Impact: Customers can enter repeat-failure loop after Kilo Pass payment, with no actionable reprovision or support path despite server knowing failure is permanent.

Fix direction: Return structured failure reason and retryability from activation. Show retry only for transient settlement or enrollment failures; route destroyed or missing anchors to reprovisioning and stale or invalid intents to fresh checkout or support.

jeanduplessis (commented):
Two verified must-fix findings from full PR review.

apps/web/src/lib/kiloclaw/credit-billing.ts
jeanduplessis: Fixed in 9ca067421. Canceled legacy rows are no longer reactivated or repriced in place. New Kilo Pass + hosting attempts now fail before Stripe Checkout creation, direct credit enrollment rolls back before spend or subscription mutation, and already-completed Kilo Pass sessions return a permanent requires_reprovision outcome while preserving awarded credits and the predecessor row byte-for-byte. This uses the smallest safe containment path because a fresh self-service successor requires a new provisioned instance under the one-subscription-per-instance constraint. Regression coverage asserts the legacy price, provider ID, status, transfer pointer, change log, and used-credit counter remain unchanged.

apps/web/src/app/(app)/claw/components/billing/billing-types.ts
jeanduplessis: Fixed in 9ca067421. activateCheckoutHosting now returns a Zod-validated discriminated union: activated, not_requested, retryable_failure, or action_required. Only credits_not_settled and unexpected operational enrollment failures expose retry. Invalid/stale intents route to a fresh plan selection; missing, destroyed, and legacy-reprovision states are non-retryable and show support-led recovery; expired Commit routes to Standard. Client no longer renders raw server errors. Table coverage asserts retryability and destination policy for every reason.

apps/web/src/routers/kilo-pass-router.ts
```diff
@@ -1272,18 +1508,89 @@ export const kiloPassRouter = createTRPCRouter({
...
+            checkout_session_id: input.sessionId,
+            instance_id: instance.id,
+            outcome: 'concurrent_idempotent_noop',
+            intended_price_version: priceVersion,
+            duration_ms: Date.now() - startedAt,
+          });
+          return { outcome: 'activated', hostingIntent: hostingPlan };
+        }
+
+        const actionRequiredReason = errorMessage.includes('Insufficient credit balance')
```
St0rmz1: **Low, Design:** This recovery classification keys off `error.message` substrings, and any message that matches none of these branches falls through to `retryable_failure` / `enrollment_failed`, which renders a retry button. That is the same failure class this PR set out to fix (a permanent failure presented as retryable). The thrown strings live in `credit-billing.ts` with no compile-time or test link to this matcher, so rewording a message there would silently downgrade a permanent condition to a retryable one. The integration tests reach the permanent reasons through the pre-checks that return before `enrollWithCredits`, so these catch-block branches are not directly asserted.

Suggestion: have `enrollWithCredits` throw typed errors (an error subclass or an error carrying a `reason` field) and switch on that instead of message text. At minimum, export the message constants from `credit-billing.ts` and reference them in both the matcher and a unit test, so a reword breaks the build rather than the recovery UI.

packages/db/src/kiloclaw-pricing-catalog.ts
```diff
@@ -63,6 +63,22 @@ export function getKiloClawPricingCatalogEntry(priceVersion: string): KiloClawPr
...
+        kiloclawPriceVersion: string;
+      }
+    | null
+    | undefined
+): KiloClawPriceVersion {
+  if (subscription?.status !== 'trialing') {
+    return CURRENT_KILOCLAW_PRICE_VERSION;
+  }
+
+  return getKiloClawPricingCatalogEntry(subscription.kiloclawPriceVersion).priceVersion;
```
St0rmz1, line 79: **Low, Functional:** For a trialing subscription this returns `getKiloClawPricingCatalogEntry(...).priceVersion`, and `getKiloClawPricingCatalogEntry` throws on an unknown version. `getPersonalBillingStatus` calls this on a read path that previously passed `sub.kiloclaw_price_version` through without validation. So a trialing row carrying a price version not in the catalog now throws and breaks the billing status query for that user, where before it rendered. With only `LEGACY` and `CURRENT` in the catalog today this is a narrow edge, but it is a fail-closed behavior change on a read path.

Suggestion: confirm the fail-closed behavior on the read path is intended. If a legacy or corrupt trialing version is reachable in production, consider tolerating it in the status read while still rejecting it on the write and enrollment paths.

St0rmz1 (approved):

packages/db/src/kiloclaw-pricing-catalog.ts
jeanduplessis: Confirmed fail-closed behavior is intended by `.specs/kiloclaw-billing.md` Pricing Versions rule 2: runtime MUST fail closed when a subscription references an unknown version. This is also not a new billing-status failure: on the base branch, `getPersonalBillingStatus` immediately passed the raw trial version to `getKiloClawPricingCatalogEntry` and both plan-cost helpers, which already throw for an unknown version. The shared resolver makes that existing invariant explicit. I added direct regression coverage for an unknown trial enrollment version.

apps/web/src/routers/kilo-pass-router.ts
jeanduplessis: Valid. I replaced `enrollWithCredits` message matching with a typed `CreditEnrollmentError` reason union and exhaustive classification maps in both router consumers. Typed permanent failures now produce non-retryable recovery, while untyped infrastructure failures remain retryable. Added a settled-but-insufficient-credits activation regression. Changes are verified locally but not committed, so I am leaving this thread unresolved until the branch is updated.

jeanduplessis: Implemented and pushed in 0c7cd8a3e. `enrollWithCredits` now emits typed reasons, both router consumers use exhaustive compile-time mappings, and the catch-path recovery regression is covered.

PR #3796, opened 2026-06-08
feat(review-memory): add review memory

alex-alecu:
## Summary

- Adds opt-in Review Memory for GitHub code reviews so repeated maintainer feedback can become REVIEW.md guidance.
- Lets users analyze last 2 weeks of replies (first inline comment to kilo reviews), review evidence, edit or dismiss proposals, and open a separate REVIEW.md PR when approved.
- Keeps this version small compared with the original #3448 flow: GitHub only, replies only, manual analysis only, and no footer CTA.

## Verification

1. Enable Review Memory from the GitHub Memory tab on the Code Reviewer page.
2. Reply as a repo owner, member, or collaborator to a Kilo inline review comment.
3. Open the Memory tab, pick the repo, and click Analyze now.
4. Review the proposal, edit the guidance, then open the REVIEW.md PR.

## Visual Changes

Adds a Memory tab and proposal review panel to the Code Reviewer settings.

Screenshot 2026-06-08 at 15 52 59

## Reviewer Notes

<details>
<summary>Scope compared to #3448</summary>

Reference PR: https://github.com/Kilo-Org/cloud/pull/3448

This keeps the core idea: learn from review feedback, draft editable REVIEW.md guidance, and open a human-approved change request.

Discarded from #3448:

- GitLab support.
- Broad signal capture from approvals, change requests, reactions, emojis, and review threads.
- Keyword classification during webhook ingest.
- Subject sync, fingerprinting, and aggregation state tables.
- Automatic aggregation cron and cooldown logic.
- Footer CTA/deep links added to Kilo review comments.
- Subscription gating, expanded role matrix, and audit-log plumbing.
- Proposal scopes/types beyond repository-level REVIEW.md guidance.
- Historical backfill and broader dashboard memory controls.

</details>

<details>
<summary>Market research from #3448</summary>

Closest competitor analogs:

| Competitor | Similar feature | How it compares to Kilo review memory |
|---|---|---|
| CodeRabbit | Learnings from `@coderabbitai` chat/review interactions, plus `emit path instructions` to open a PR updating `.coderabbit.yaml`. | Closest to Kilo's `REVIEW.md` PR flow. Learnings are internal, editable/deletable/exportable in the dashboard, and applied automatically. `emit path instructions` is similar to turning recurring feedback into reviewable repo guidance, but targets `.coderabbit.yaml` path instructions. |
[...]

PR #4048, opened 2026-06-16
fix(review-memory): accept contributor feedback

alex-alecu:
## Why

Followup to https://github.com/Kilo-Org/cloud/pull/3796

GitHub can mark a helpful repo user as a contributor instead of a member. Review Memory was skipping those replies, so it missed feedback it should learn from.

## What changed

Review Memory now accepts replies from contributors when all other checks pass. New or unknown commenters are still skipped. A reply must still be to a Kilo review comment, the feature must still be turned on, and only the first reply is saved. The app text and the pull request text now match this rule.

## How to test

1. Turn on Review Memory for a GitHub repo.
2. Reply to a Kilo review comment as a contributor who has already worked on the repo.
3. Check that Review Memory stores the reply.
4. Reply as a first-time contributor and check that Review Memory skips it.


alex-alecu:
Will start another PR

PR #4049, opened 2026-06-16
feat(code-review): resolve fixed threads

alex-alecu:
## Summary

Incremental code reviews can now close an old Kilo review thread only when the new review says the issue is fully fixed. The server checks that the thread still belongs to Kilo, has no replies, is still safe to close, and matches the commit the review saw. If anything is missing, stale, or changed, nothing is closed.

## Verification

1. Run an incremental GitHub code review on a staging PR with a fixed Kilo inline thread that has no replies.
2. Confirm that fixed thread is resolved.
3. Add a reply to another Kilo thread and confirm it stays open after the review.

## Visual Changes

N/A

## Reviewer Notes

GitHub permissions should be checked in staging because closing threads depends on `viewerCanResolve`.


alex-alecu:
Done here in a less programatic way by using direct `gh` cli from main agent https://github.com/Kilo-Org/cloud/pull/4116

PR #4050, opened 2026-06-16
feat(web): add environment management workflow

RSO:
## Summary

- Add one idempotent `pnpm web:env set` command that asks whether a variable is sensitive, then updates both Vercel projects across Development, Staging, and Production.
- Mirror sensitive Production values to the shared 1Password vault while leaving Development exportable for local setup.
- Update tracked dotenv defaults, highlight matching remote values without blocking intentional reuse, and run contributor setup smoke checks when relevant environment templates change.

## Verification

- Confirmed the sensitivity prompt rejects sensitive `NEXT_PUBLIC_*` values before accessing remote providers.
- Confirmed local argument validation, pinned Vercel CLI invocation, and 1Password create/update template behavior.
- Did not run the main command against live Vercel resources during implementation.

## Visual Changes

CleanShot 2026-06-17 at 09 06 50@2x
CleanShot 2026-06-17 at 09 06 32@2x

## Reviewer Notes

- The implementation is intentionally small and idempotent: partial failures are recovered by rerunning the same command rather than maintaining transaction or resume state.
- Multiline values use `--development-file`, `--staging-file`, and `--production-file`; normal values use hidden terminal prompts.
- Matching tracked and remote values produce a highlighted warning but remain allowed.


RSO:
@kilocode-bot please review the changes in this PR

markijbema (approved):

PR #4054, opened 2026-06-16
Feat/kiloclaw image model catalog refresh

St0rmz1:
## Summary

OpenClaw's live model discovery does not populate the agent model registry that
capability resolution reads, so kilocode provider models resolve as text-only and
image understanding is refused even though the Kilo Gateway advertises vision
support. This adds a temporary workaround in the controller: it periodically
fetches the gateway model catalog and writes it into
`models.providers.kilocode.models` in `openclaw.json`, which the capability gate
does read. OpenClaw hot-reloads that change without a restart. Remove this once
OpenClaw feeds discovery into capability resolution.

## Changes

- New `services/kiloclaw/controller/src/model-catalog-refresh.ts`:
  - Fetches `GET /api/gateway/models` using the instance's API key, organization,
    and feature headers.
  - Validates the response with zod per entry (tolerant of null numeric/price
    fields) and skips any entry that does not parse, so one bad entry cannot drop
    the whole catalog.
  - Maps each model to the OpenClaw provider model shape, deriving `input`
    modalities (`text` / `image`) from `architecture.input_modalities`.
  - Writes `models.providers.kilocode.models` with `atomicWrite` (mode `0o600`),
    preserving all other config fields. Skips the write when the catalog is
    unchanged or when the kilocode provider block does not exist yet.
  - Runs once about 30 seconds after the gateway starts, then every 24 hours.
    Follows the existing `gmail-watch-renewal` timer pattern, and is a no-op when
    `KILOCODE_API_KEY` is unset.
- `services/kiloclaw/controller/src/index.ts`: starts the refresh after the
  gateway starts and stops it during shutdown.
- Unit tests covering mapping, header auth, null/unparseable entries, the
  skip-when-unchanged and missing-block guards, and the start/stop scheduling.

## Verification

- [x] Ran the built image locally in Docker. About 30 seconds after gateway start
  the controller logged `[model-catalog-refresh] seeded 332 kilocode models`,
  `openclaw.json` `models.providers.kilocode.models` held 332 entries with
  `anthropic/claude-sonnet-4.6` as `["text","image"]`, and the gateway logged
  `config hot reload applied (models.providers.kilocode.models)` with no restart.
  `openclaw infer image describe` no longer fails the capability gate with
  "model does not support images ... input: text"; it now reaches the model (the
  local image's key returns a 401 sign-in error, unrelated to this change).

## Visual Changes

N/A

## Reviewer Notes

- This is a temporary workaround. The real fix is in OpenClaw (feeding live
  discovery into the model registry / capability resolution); this module should
  be removed when that lands.
- `config-writer` still writes `models: []` at boot, doctor, and restore. The
  refresh re-seeds after those events; the initial run is about 30 seconds after
  gateway start, so there is a short window where the catalog is empty.
- Concurrency: the read, mutate, and write run in one synchronous step (no await
  between read and write), so no in-process writer can lose an update. The only
  residual is the external gateway process writing the file in the brief window
  between read and rename, which skip-when-unchanged keeps rare.
- The seeded `models` array is the full gateway catalog (a few hundred entries)
  written into `openclaw.json` and re-read on hot reload.

### DIFF
--- a/apps/web/src/app/(app)/claw/components/changelog-data.ts
+++ b/apps/web/src/app/(app)/claw/components/changelog-data.ts
@@ -11,6 +11,13 @@ export type ChangelogEntry = {
 // Newest entries first. Developers add new entries to the top of this array.
 export const CHANGELOG_ENTRIES: ChangelogEntry[] = [
   {
+    date: '2026-06-16',
+    description:
+      "Fixed image recognition for KiloClaw agents. Images sent to an agent could not be processed and failed with 'model does not support images', because the models that support images were not being offered to the image pipeline. KiloClaw now refreshes the model list so image understanding works again. Redeploy your instance to pick up the fix.",
+    category: 'bugfix',
+    deployHint: 'redeploy_required',
+  },
+  {
     date: '2026-06-15',
     description: 'OpenClaw 2026.6.5 is now generally available to everyone.',
     category: 'feature',
--- a/services/kiloclaw/controller/src/index.ts
+++ b/services/kiloclaw/controller/src/index.ts
@@ -52,6 +52,7 @@ import { writeGogCredentials } from './gog-credentials';
 import { installGogShim } from './gog-shim';
 import { migrateLegacyGoogleCredentialsToBroker } from './legacy-google-migration';
 import { startWatchRenewal, stopWatchRenewal } from './gmail-watch-renewal';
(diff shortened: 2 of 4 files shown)


pandemicsyn (approved):

St0rmz1:
Just for reference the mentioned bug in OpenClaw is documented at
https://github.com/openclaw/openclaw/issues/93775

PR #4056, opened 2026-06-17
feat(ai-gateway): add organization auto model routing

pandemicsyn:
## Summary

- Adds `kilo-auto/org`, an Enterprise-only Organization Auto model that lets an organization choose which model should handle each Kilo mode.
- `kilo-auto/org` is intentionally used rather than an org-name-derived ID: organization names are org owner mutable and not unique, and identity comes from authenticated organization context. This avoids rename migrations, stale cache keys, privacy leakage, and unbounded telemetry/catalog cardinality. Also, I think this name fits because its kilo’s “org” auto routing model. 
- Organization Auto stores routes in `organizations.settings.org_auto_model`:
  - If a request mode has an explicit route, Organization Auto uses that model.
  - If a request mode has no explicit route, Organization Auto uses the configured fallback model.
  - If a request uses an unknown/custom mode, Organization Auto also uses the fallback model unless that exact mode slug has a route.
  - Exact route slugs win first, so `build` uses `routes.build` before falling back to `routes.code`, and `plan` uses `routes.plan` before falling back to `routes.architect`.
- Replaces the unshipped `org-default-model-config` mode-default rollout with the new `organization-auto-model-routing` PostHog toggle for canonical organization-wide routing.
- Adds Enterprise-only enable/disable/fallback controls, canonical mode route configuration, route-aware mode lifecycle behavior, and exact route audit messages.
- Restricts provider, model, default-model, and Organization Auto routing policy changes to organization owners and admins; billing managers retain billing-focused access.
- Replaces the default-model dialog with a unified configuration flow: Organization Auto and a specific model are staged choices, while the global default remains a quiet explicit reset action.
- Organization Auto is presented as the recommended path with accessible radio semantics, keyboard navigation, and explicit selected-state affordances.
- Reset now explicitly disables Organization Auto and returns the organization to global-default resolution while preserving dormant routes and fallback settings for future re-enable.
- Custom Modes now labels stored routes as inactive when Organization Auto is off instead of implying they are currently applied.
- Removes the unshipped `OrganizationModeConfig.defaultModel` bridge and stores all routing only in `organizations.settings.org_auto_model`.
- Adds organization-owned direct BYOK route support with credential ownership checks, direct-provider routing, zero Kilo hosted cost, and no fallback to OpenRouter/Vercel when credentials disappear.
- Adds client-safe shared Organization Auto helpers so browser components do not import server-only DB/pg modules.

## Verification

- [x] Verified the default behavior chooser in the browser, including selected-state styling, the Recommended badge, and focused-but-unselected options.
- [x] Verified keyboard navigation for the radio chooser with ArrowUp/Down/Left/Right, Home, and End.
- [x] Verified Organization Auto and Specific model dialog states at the compact dialog width.
- [x] Verified Organization Auto API behavior manually with `/tmp/org-auto-api-log.sh` against a local service, including modes/defaults/models and optional gateway requests.

## Visual Changes

Screenshot 2026-06-17 at 5 01 46 PM

Screenshot 2026-06-17 at 5 02 18 PM

Screenshot 2026-06-17 at 5 02 31 PM

Screenshot 2026-06-17 at 5 03 28 PM

Loom:

https://www.loom.com/share/30ded7b1543c4b2aa0d1e5989dc04d68

## Reviewer Notes

- Organization Auto UI/configuration writes are gated by the PostHog release toggle `organization-auto-model-routing`, replacing the old `org-default-model-config` toggle. Runtime routing remains data-driven for already-configured organizations, and disable remains available during rollback.
- Stripe webhook/admin plan transitions intentionally do not clear the stored sentinel; catalog/default/runtime eligibility gates make dormant routes safe while preserving future re-enable behavior. The explicit global reset path now disables active Organization Auto while preserving those dormant route settings.
- Review BYOK fallback protection, and route-management authorization as the highest-risk areas.
- Provider/model/default/routing policy is intentionally owner/admin-only because it controls organization-wide model access, cost exposure, and data egress.


pandemicsyn:
Fixed review feedback in b3107a90b: BYOK ownership validation now accepts and uses the active transaction client inside Organization Auto settings/mode mutations.

apps/web/src/routers/organizations/organization-settings-router.ts
```diff
@@ -167,121 +316,575 @@ export const organizationsSettingsRouter = createTRPCRouter({
...
+          action: 'organization.settings.change',
+          actor_email: ctx.user.google_user_email,
+          actor_id: ctx.user.id,
+          actor_name: ctx.user.google_user_name,
+          message:
+            behavior === 'auto'
+              ? 'Configured Organization Auto default behavior.'
+              : behavior === 'specific'
+                ? `Configured specific organization default model: ${settings.default_model}`
+                : existingOrg.settings.default_model === ORG_AUTO_MODEL.id
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in f753b11d1. Captured the locked pre-mutation default inside the transaction and added audit coverage.

pandemicsyn:
(will fix up the merge conflicts as they pop up - but not actively hacking on this beyond that)

apps/web/src/lib/organizations/organization-auto-model.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  if (normalizedModelId.startsWith('kilo-auto/')) {
+    if (!isOrganizationAutoTargetModel(normalizedModelId)) {
+      return {
+        kind: 'error',
+        message: `Organization Auto route target '${targetModelId}' is not a supported auto tier.`,
+      };
+    }
+
+    return { kind: 'ok', modelId: normalizedModelId };
```
St0rmz1, line 166: **Moderate, Functional:** Auto-tier route and fallback targets skip the organization's model policy here. For a `kilo-auto/*` target this returns `ok` right after the `isOrganizationAutoTargetModel` check, so the `provider_allow_list` / `model_deny_list` predicate (applied below for concrete catalog ids) never runs. At runtime the tier resolves to a concrete model: for example `kilo-auto/balanced` resolves to an Alibaba-only Qwen model. An Enterprise org with a restrictive `provider_allow_list` can save a route or fallback pointing at `kilo-auto/balanced`, get a success response, and then have every request through that route fail downstream with "No allowed providers". The admin gets no signal at configuration time that the target is unusable under their own policy. Consider resolving the auto tier to its concrete target and running it through the same allow predicate so the admin is warned, or disallowing auto-tier targets when the org has an active `provider_allow_list`.

apps/web/src/routers/organizations/organization-settings-router.ts
```diff
@@ -167,121 +316,577 @@ export const organizationsSettingsRouter = createTRPCRouter({
...
+              org_auto_model: {
+                ...orgAutoModel,
+                routes,
+                fallback_model: validation.modelId,
+              },
+            };
+          },
+          tx
+        );
+        await createAuditLog({
```
St0rmz1: **Low, Design:** This audit log fires unconditionally on every call, including no-op saves (global to global, or auto to auto with identical fallback and routes). The sibling endpoints in this router (`setOrganizationAutoRoute`, `setOrganizationAutoFallback`, `enableOrganizationAuto`) guard their audit writes on an actual change. Consider comparing the resulting `default_model` and `org_auto_model` against the previous values and skipping the audit entry when nothing changed, to keep the change history readable and consistent with the other endpoints.

apps/web/src/routers/organizations/organization-modes-router.ts
```diff
@@ -183,46 +246,85 @@ export const organizationModesRouter = createTRPCRouter({
...
-        organizationId,
-        ctx.user.id,
-        name,
-        slug,
-        normalizeModeConfig(config)
-      );
+      let createdMode: OrganizationMode | null | undefined;
+      let routeAuditMessage: string | undefined;
+      await db.transaction(async tx => {
+        await mutateOrganizationSettings(
```
St0rmz1, line 234: **Low, Performance:** Wrapping mode create (and update/delete) in `mutateOrganizationSettings` takes a `FOR UPDATE` lock on the `organizations` row even when `route_model` is undefined and no setting actually changes. That serializes all mode CRUD for an organization behind the org row lock, where previously these were independent inserts/updates. Modes are a low-frequency admin operation so the impact is small, but if you want to avoid the contention you could acquire the settings lock only when a route change is actually requested and run the mode mutation without it otherwise.

apps/web/src/components/organizations/providers-and-models/DefaultModelDialog.tsx
```diff
@@ -21,75 +18,230 @@ import {
...
+  const [selectedFallbackModel, setSelectedFallbackModel] = useState('');
 
-      // Invalidate the defaults query to refresh the display
-      await queryClient.invalidateQueries({
-        queryKey: ['organization-defaults', organizationId],
-      });
+  const availableModels = (openRouterModels?.data ?? []).filter(
+    model => model.id !== ORG_AUTO_MODEL.id
+  );
+  const autoTargetModels = useMemo(
```
St0rmz1, line 181: **Nit, Performance:** `availableModels` above is a fresh `.filter()` result on every render, so this `useMemo` keyed on `[availableModels]` recomputes each render and never actually memoizes. Either wrap `availableModels` in its own `useMemo`, or key this memo on `openRouterModels?.data` and do both filters inside it.

apps/web/src/lib/organizations/organization-auto-model.ts
pandemicsyn: Fixed in 9f9f610ea. Auto-tier route and fallback targets now reject active model policies; admins must choose a concrete allowed model.

apps/web/src/routers/organizations/organization-settings-router.ts
pandemicsyn: Fixed in 9f9f610ea. No-op behavior saves now return locked settings unchanged and skip audit writes.

apps/web/src/components/organizations/providers-and-models/DefaultModelDialog.tsx
pandemicsyn: Fixed in 9f9f610ea. Memoized availableModels from OpenRouter data so autoTargetModels now has a stable dependency.

apps/web/src/routers/organizations/organization-modes-router.ts
pandemicsyn: The observation is accurate, but the suggested optimization is risky because update/delete can still mutate route state without route_model when slugs change or routed modes are deleted. Mode CRUD is low-frequency admin work, and the current lock ordering preserves route consistency.

apps/web/src/lib/ai-gateway/auto-model/index.ts
```diff
@@ -22,6 +22,7 @@ type AutoModel = {
   supports_images: boolean;
   supports_pdf: boolean;
   opencode_settings: OpenCodeSettings | undefined;
+  status?: 'public' | 'hidden';
```
chrarnoldus: Is this new property used in practice or the product of consecutive merge conflicts?

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -353,6 +369,7 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
   const isRateLimitedFreeModelRequest =
     isKiloExclusiveFreeModel(effectiveModelIdLowerCased) ||
     autoModel === KILO_AUTO_FREE_MODEL.id ||
+    routingTarget === KILO_AUTO_FREE_MODEL.id ||
```
Comment (chrarnoldus) [inline apps/web/src/app/api/openrouter/[...path]/route.ts:370]: Is this the only reason we need the extra routingTarget variable?

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -561,6 +578,29 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
...
+  }
+
+  // Request-level data-collection opt-out: a caller can set
+  // `provider.data_collection: 'deny'` or `provider.zdr: true` on any
+  // request to opt that single request out of training/data-retention.
+  // Direct experiment upstreams ignore those OpenRouter/Vercel flags
+  // (we never reach OpenRouter), but we still capture the prompt to R2
+  // for partner evaluation — which violates the caller's stated
+  // intent. Refuse here regardless of org settings, anon/BYOK status,
+  // or the org-level check below.
```
Comment (chrarnoldus) [inline apps/web/src/app/api/openrouter/[...path]/route.ts:601]: looks like a random addition?

chrarnoldus:
There are downsides of not telling the client which model is used:

* Context window and multi-modal support may be off.
* A suboptimal API format may be used.
* If a modified prompt is available for the model family it will not be used.
* Model variants will not be selectable (this is probably the easiest one to fix by making it part of the mode config).

Some of this can be mitigated by setting the `opencode` property if models happen to align, but I'm not sure there's an easy solution in general.

apps/web/src/app/api/openrouter/[...path]/route.ts
Comment (pandemicsyn) [reply apps/web/src/app/api/openrouter/[...path]/route.ts:370]: Clarified in 67a608ccb. routingTarget is needed for nested free-tier rate limiting and direct-BYOK ownership validation after resolution.

apps/web/src/app/api/openrouter/[...path]/route.ts
Comment (pandemicsyn) [reply apps/web/src/app/api/openrouter/[...path]/route.ts:601]: Clarified in 67a608ccb. Added a comment explaining the post-resolution uses of routingTarget.

apps/web/src/lib/ai-gateway/auto-model/index.ts
pandemicsyn: Fixed in 67a608ccb. Removed the unused AutoModel status property and hidden marker from Organization Auto.

apps/web/src/components/organizations/providers-and-models/DefaultModelDialog.tsx
```diff
@@ -21,75 +18,242 @@ import {
...
+
+function BehaviorChooser({
+  value,
+  onChange,
+}: {
+  value: DefaultBehavior;
+  onChange: (value: DefaultBehavior) => void;
+}) {
+  const itemRefs = useRef<Partial<Record<DefaultBehavior, HTMLButtonElement | null>>>({});
+
```
iscekic, line 79: **Standards:** `BehaviorChooser` hand-rolls a ~70-line radiogroup (roles, tabIndex, arrow/Home/End handling) while `components/ui/radio-group.tsx` already exists. AGENTS.md: "Use existing dependencies before implementing custom solutions."

apps/web/src/components/organizations/providers-and-models/DefaultModelDialog.tsx
```diff
@@ -21,75 +18,242 @@ import {
...
+  },
+  {
+    value: 'specific',
+    title: 'Specific model',
+    description: 'Pin a single model as the organization default.',
+  },
+];
+
+const BEHAVIOR_ORDER = BEHAVIOR_OPTIONS.map(option => option.value);
+
```
iscekic, line 70: **Spec (partial):** the PR describes "one unified behavior save flow for Organization Auto, a specific model, or the global default" — but the radiogroup only offers auto/specific; global default lives in a separate Reset link outside the chooser. Server-side flow is unified, UI is partial.

apps/web/src/components/organizations/providers-and-models/DefaultModelDialog.tsx
```diff
@@ -21,75 +18,242 @@ import {
...
+      availableModels.filter(model => {
+        if (model.id.startsWith(CUSTOM_LLM_PREFIX)) return false;
+        if (model.id.startsWith('kilo-auto/')) {
+          return !hasActiveModelPolicy && isOrganizationAutoTargetModel(model.id);
+        }
+        return true;
+      }),
+    [availableModels, hasActiveModelPolicy]
+  );
+  const defaultAutoFallback = hasActiveModelPolicy
```
iscekic: **Questionable default:** with an active model policy and no stored fallback, `defaultAutoFallback` silently falls back to `autoTargetModels[0]` — enabling Auto can persist an arbitrary catalog-first fallback the user never chose. Also: `'kilo-auto/balanced'` is hardcoded here (and at ~208) instead of `KILO_AUTO_BALANCED_MODEL.id`.

apps/web/src/lib/ai-gateway/auto-model/resolution.ts
```diff
@@ -88,16 +100,152 @@ function gatewaySupportsApiKind(
...
+): string | undefined {
+  const routes = settings.org_auto_model?.routes ?? {};
+  const mode = modeHeader?.trim() ?? '';
+  const normalizedMode = mode.toLowerCase();
+
+  if (normalizedMode && hasOrganizationAutoRoute(routes, normalizedMode)) {
+    return routes[normalizedMode];
+  }
+
+  if (normalizedMode === 'code' || normalizedMode === 'build') {
```
iscekic: **Unspecified behavior:** the PR description only promises `build` → `routes.build` before `routes.code` and `plan` → `routes.plan` before `routes.architect`, but this implements *symmetric* fallbacks too — mode `code` with no `routes.code` silently uses `routes.build`, and `architect` uses `routes.plan`. A route saved for a custom "build" mode captures canonical `code` traffic. Intentional?

apps/web/src/lib/ai-gateway/auto-model/resolution.ts
```diff
@@ -88,16 +100,152 @@ function gatewaySupportsApiKind(
...
+  if (normalizedMode === 'plan' || normalizedMode === 'architect') {
+    if (hasOrganizationAutoRoute(routes, 'architect')) return routes.architect;
+    if (hasOrganizationAutoRoute(routes, 'plan')) return routes.plan;
+    return settings.org_auto_model?.fallback_model;
+  }
+
+  return settings.org_auto_model?.fallback_model;
+}
+
+function isSupportedNestedAutoTarget(modelId: string): boolean {
```
iscekic: **Duplication:** `isSupportedNestedAutoTarget` is byte-identical to `isOrganizationAutoTargetModel` in `organization-auto-model-shared.ts`; similarly `hasOrganizationAutoRoute` above re-implements the `hasOwnProperty` guard from `getOrganizationAutoRoute` (a third copy lives in `organization-modes-router.ts:84`).

apps/web/src/lib/ai-gateway/auto-model/resolution.ts
```diff
@@ -88,16 +100,152 @@ function gatewaySupportsApiKind(
...
+  }
+
+  if (isSupportedNestedAutoTarget(validation.modelId)) {
+    const nestedResult = await resolveAutoModel(
+      {
+        ...params,
+        model: validation.modelId,
+      },
+      userPromise,
+      balancePromise
```
iscekic, line 219: **Dead check:** `validation.modelId === ORG_AUTO_MODEL.id` re-checks a self-target case that `validateOrganizationAutoTarget` (and the zod refine in `schema-types.ts`) already rejects.

apps/web/src/routers/organizations/organization-modes-router.ts
```diff
@@ -60,121 +64,180 @@ const UpdateModeInputSchema = OrganizationIdInputSchema.extend({
...
-function assertDefaultModelCanBeSet(
-  organization: NonNullable<Awaited<ReturnType<typeof getOrganizationById>>>,
-  change: DefaultModelChange
-): void {
-  if (change.kind !== 'set') {
-    return;
-  }
-
-  if (organization.plan !== 'enterprise') {
+async function assertOrganizationAutoWriteEnabled(userId: string): Promise<void> {
```
iscekic: **Duplication:** `assertOrganizationAutoWriteEnabled` / `assertOrganizationAutoEligible` are defined verbatim here and in `organization-settings-router.ts:174,186` — changing the rollout gate requires edits in both routers. Worth extracting to a shared module.

packages/db/src/schema-types.ts
```diff
@@ -733,12 +733,44 @@ export const OrganizationPlanSchema = z.enum(['teams', 'enterprise']);
...
+    'Organization Auto route slug must contain only lowercase letters, numbers, and hyphens'
+  );
+
+const OrganizationAutoModelTargetSchema = z
+  .string()
+  .min(1, 'Organization Auto route target is required')
+  .max(200, 'Organization Auto route target must be less than 200 characters')
+  .refine(value => !value.endsWith('/*'), {
+    message: 'Organization Auto route target must be a concrete model identifier',
+  })
```
iscekic, line 754: **Magic literals:** `'kilo-auto/org'` here (vs `ORG_AUTO_MODEL.id`) and `100` at ~759 (vs `MAX_ORGANIZATION_AUTO_ROUTES`). The cross-package import direction explains it, but the constants could live here in `packages/db` and be imported by the app.

apps/web/src/app/api/organizations/hooks.ts
```diff
@@ -180,28 +180,97 @@ export function useUpdateCompanyDomain() {
...
+  const onSuccess = useInvalidateOrganizationDataAndDefaults();
+
+  return useMutation(
+    trpc.organizations.settings.updateDefaultModel.mutationOptions({
+      onSuccess,
+    })
+  );
+}
+
+export function useEnableOrganizationAuto() {
```
iscekic: **Dead API surface:** `useEnableOrganizationAuto` / `useDisableOrganizationAuto` (and `useSetOrganizationAutoFallback` below) have no UI callers — everything routes through `configureOrganizationDefaultBehavior`. Consider dropping the unused procedures + hooks until something needs them.

apps/web/src/app/api/organizations/hooks.ts
```diff
@@ -180,28 +180,97 @@ export function useUpdateCompanyDomain() {
...
+  const trpc = useTRPC();
+  const onSuccess = useInvalidateOrganizationDataAndDefaults();
+  return useMutation(
+    trpc.organizations.settings.configureOrganizationDefaultBehavior.mutationOptions({
+      onSuccess,
+    })
+  );
+}
+
+export function useSetOrganizationAutoFallback() {
```
iscekic: **Duplication:** the three auto-route hooks repeat an identical `onSuccess` invalidation instead of reusing `useInvalidateOrganizationDataAndDefaults` defined just above.

apps/web/src/routers/organizations/organization-settings-router.ts
```diff
@@ -172,126 +362,597 @@ export const organizationsSettingsRouter = createTRPCRouter({
       return result;
     }),
 
-  updateAllowLists: organizationBillingMutationProcedure
+  updateAllowLists: organizationOwnerMutationProcedure
```
iscekic, line 330: **Scope creep / permission change:** `updateAllowLists` (and `updateDefaultModel` at ~433) escalated from `organizationBillingMutationProcedure` to `organizationOwnerMutationProcedure` — a permission-model change not mentioned in the PR description. Worth calling out explicitly.

apps/web/src/routers/organizations/organization-settings-router.ts
```diff
@@ -172,126 +362,597 @@ export const organizationsSettingsRouter = createTRPCRouter({
...
-      // Check if default_model needs to be cleared when access lists change
-      if (
-        (provider_allow_list !== undefined || model_deny_list !== undefined) &&
-        currentSettings.default_model
-      ) {
-        const isAllowed = createAllowPredicateFromRestrictions({
-          providerAllowList: settingsUpdate.provider_allow_list,
-          modelDenyList: settingsUpdate.model_deny_list ?? [],
+      if (existingOrg.settings.default_model === ORG_AUTO_MODEL.id) {
+        throw new TRPCError({
```
iscekic, line 414: **Stale message:** "Disable Organization Auto through its dedicated controls" points at enable/disable controls no UI actually exposes (the dialog goes through `configureOrganizationDefaultBehavior`).

apps/web/src/routers/organizations/organization-settings-router.ts
```diff
@@ -172,126 +362,597 @@ export const organizationsSettingsRouter = createTRPCRouter({
...
+          message: createDefaultModelDiffMessage(previousSettings, settings),
+          organization_id: organizationId,
+          tx,
+        });
+        return settings;
+      });
+      return { settings: updatedSettings };
+    }),
+
+  enableOrganizationAuto: organizationOwnerMutationProcedure
```
iscekic: **Duplication:** the router repeats the same ~60-line shape several times (pre-checks re-run inside `mutateOrganizationSettings`, closure-mutated `previous`/`didChange`, transaction + `createAuditLog`); this and `configureOrganizationDefaultBehavior` additionally share the seeded-routes revalidation loop and `JSON.stringify` change-detection. A shared helper would collapse a lot of this file.

apps/web/src/routers/organizations/organization-settings-router.ts
```diff
@@ -172,126 +362,597 @@ export const organizationsSettingsRouter = createTRPCRouter({
...
+        return settings;
       });
 
-      return {
-        settings: updatedSettings,
-      };
+      return { settings: updatedSettings };
     }),
 
   updateDataCollection: organizationBillingMutationProcedure
```
iscekic, line 713: **Scope creep:** `updateDataCollection` and the following unrelated mutations (projects UI, code indexing, minimum balance) are rewritten onto the new transactional `mutateOrganizationSettings` — settings become schema-validated + row-locked, a behavior-affecting refactor beyond what the PR description covers.

apps/web/src/components/organizations/custom-modes/EditModeForm.tsx
```diff
@@ -31,7 +35,17 @@ function normalizeGroups(groups: unknown): string[] | undefined {
     return undefined;
```
iscekic, line 35: **Scope creep:** the new `edit`-group normalization changes `matchesBuiltInModeState`, altering when an edit is treated as a built-in revert — an unrequested behavior change worth a mention in the PR description.

apps/web/src/components/organizations/custom-modes/NewModeForm.tsx
```diff
@@ -50,8 +71,28 @@ export function NewModeForm({
...
+        model_id: targetModelId,
+      });
+    } else {
+      await clearRouteMutation.mutateAsync({ organizationId, mode_slug: modeSlug });
+    }
+  };
+
   const handleSubmit = async (data: ModeFormData) => {
     try {
+      if (defaultModeSlug && matchesBuiltInModeState(data, defaultModeSlug)) {
```
iscekic, line 88: **Minor:** the built-in route-only path calls `persistRoute` and toasts "route updated successfully" even when the route is unchanged — the server no-ops but the toast still claims success.

iscekic (approved):
LGTM :shipit: — approving. Ran an automated two-axis review (repo standards + PR-description-as-spec); nothing blocking, but here's what the bot turned up, left as inline comments. Core behaviors all verified correct: routing resolution (exact slug → alias → fallback), enterprise gating, BYOK ownership checks + no-fallback guard, reset preserving dormant routes, bridge removal, client-safe helpers.

apps/web/src/components/organizations/providers-and-models/DefaultModelDialog.tsx
pandemicsyn: Fixed in 32b30a937. Restricted organizations now require an explicit concrete fallback, and the balanced default uses the shared model constant.

apps/web/src/lib/ai-gateway/auto-model/resolution.ts
pandemicsyn: Fixed in 32b30a937. Alias fallback is now one-way: build to code and plan to architect, with regression coverage.

apps/web/src/lib/ai-gateway/auto-model/resolution.ts
pandemicsyn: Fixed in 32b30a937. Resolution and mode lifecycle code now reuse shared target and route-presence helpers.

apps/web/src/routers/organizations/organization-modes-router.ts
pandemicsyn: Fixed in 32b30a937. Both routers now use shared rollout and Enterprise eligibility assertions.

apps/web/src/app/api/organizations/hooks.ts
pandemicsyn: Fixed in 32b30a937. Removed the unused dedicated enable, disable, and fallback procedures and hooks.

packages/db/src/schema-types.ts
pandemicsyn: Fixed in 32b30a937. The sentinel ID and route limit now come from the shared DB contract.

apps/web/src/app/api/organizations/hooks.ts
pandemicsyn: Fixed in 32b30a937. Route mutations now reuse the shared organization/default invalidation callback.

apps/web/src/routers/organizations/organization-settings-router.ts
pandemicsyn: Fixed in 32b30a937. The error now directs callers to the unified default-model behavior flow.

apps/web/src/routers/organizations/organization-settings-router.ts
pandemicsyn: Fixed in 3d6adeac7. Restored the existing persistence and audit pattern for unrelated settings mutations.

apps/web/src/components/organizations/providers-and-models/DefaultModelDialog.tsx
pandemicsyn: This presentation is intentional: Organization Auto and Specific model are staged choices, while returning to the global default is a quieter explicit reset action. The server mutation remains unified. I clarified the PR description so it no longer implies all three behaviors are radio options.

apps/web/src/components/organizations/providers-and-models/DefaultModelDialog.tsx
pandemicsyn: The shared RadioGroup provides individual radio buttons but not the group labeling, roving tab index, or Arrow/Home/End behavior used here. Replacing this chooser with it would regress keyboard and screen-reader behavior, so I would prefer to improve the shared primitive separately before migrating this UI.

apps/web/src/lib/ai-gateway/auto-model/resolution.ts
pandemicsyn: This is intentionally retained as a fail-closed guard at the recursive resolution boundary. The validator and schema reject self-targets today, but this prevents accidental recursion if persisted data or validation behavior changes later; the code now comments that intent.

apps/web/src/routers/organizations/organization-settings-router.ts
pandemicsyn: Agreed that this needed to be explicit. The owner/admin-only policy is intentional because these settings control organization-wide model access, cost exposure, and data egress. The PR Summary and Reviewer Notes now call out the permission change.

apps/web/src/routers/organizations/organization-settings-router.ts
pandemicsyn: The transaction blocks have similar structure but different validation, no-op detection, rollout gating, mutation, and audit conditions. A generic wrapper would need enough callbacks and flags that it would obscure those boundaries, so I would defer broader transaction/audit abstraction to a focused refactor.

apps/web/src/components/organizations/custom-modes/EditModeForm.tsx
pandemicsyn: The normalization is intentional: it makes semantically equivalent edit-group configurations compare equal despite object property order or omitted versus empty optional descriptions. Without it, an unchanged built-in mode can be incorrectly treated as a custom override instead of a revert.

apps/web/src/components/organizations/custom-modes/NewModeForm.tsx
pandemicsyn: An unchanged route cannot be submitted through the supported UI because ModeForm disables Save when the form is not dirty. The server no-op is additional protection for direct callers, so this success toast only follows a user-visible change in the normal flow.

PR #4081, opened 2026-06-17
feat(security-agent): add audit report backend

jeanduplessis:
## Summary

Security Agent now has the backend/API foundation for durable Security Finding Activity Events and owner-scoped Audit Reports, split out so it can land before the web UI.

### Why this change is needed

Security owners need period-bounded evidence of material Security Finding actions and outcomes for investigation and compliance work without implying complete legacy reconstruction or aggregate historical SLA compliance. The schema, writers, and API need to land first because deployed services will write audit evidence before the UI consumes it.

### How this is addressed

- Extends the canonical audit ledger with owner-scoped, idempotent, deletion-safe Security Finding Activity Event fields and a regenerated migration.
- Adds shared audit-event helpers used by Security Sync and Security Auto Analysis flows.
- Emits reportable evidence across sync, analysis, dismissal, remediation, and deletion paths.
- Adds server-side report assembly with UTC date ranges, deterministic finding timelines, privacy controls, fail-closed query budgets, and personal/organization tRPC endpoints.
- Updates the Security Agent spec and context contract for Audit Report evidence, access, deletion, privacy, and availability guarantees.

## Verification

- Reviewed the backend split against the original #4069 scope and confirmed it excludes web route pages, React components, shared UI primitive changes, and the UI-only package dependency.
- No browser or manual runtime verification was performed; this PR contains backend/API/database changes only.

## Visual Changes

N/A

### Human Reviewer Flags

- Deploy this PR before #4082. The migration must be applied before deployed writers emit the new Security Finding Activity Event fields and actions.
- Schema-first deployment is required because authoritative finding transitions now write audit evidence in the same transaction.
- Synchronous report assembly intentionally returns no partial data when its event-count, serialized-size, or timeout budgets are exceeded.
- Organization report access is broader than ordinary member access and intentionally remains available after Security Agent, subscription, or integration disablement for authorized viewers.

### Code Reviewer Agent

<details>

- Audit events use owner-partitioned idempotency, compact deletion-safe snapshots, strict payload validation, and event-time actor attribution.
- Report assembly uses owner-scoped keyset pagination, deterministic grouping, safe link projection, and explicit legacy-evidence handling.
- Focus review on route authorization, real-database multi-page report behavior, deleted-finding end-to-end evidence, and migration behavior on populated audit data.
- The UI is intentionally excluded and stacked in #4082.

</details>


packages/db/src/migrations/0167_flawless_true_believers.sql
```diff
@@ -0,0 +1,16 @@
+ALTER TABLE "security_audit_log" DROP CONSTRAINT "security_audit_log_action_check";--> statement-breakpoint
+ALTER TABLE "security_audit_log" ADD COLUMN "actor_type" text;--> statement-breakpoint
+ALTER TABLE "security_audit_log" ADD COLUMN "finding_id" uuid;--> statement-breakpoint
+ALTER TABLE "security_audit_log" ADD COLUMN "occurred_at" timestamp with time zone;--> statement-breakpoint
+ALTER TABLE "security_audit_log" ADD COLUMN "source_occurred_at" timestamp with time zone;--> statement-breakpoint
+ALTER TABLE "security_audit_log" ADD COLUMN "event_key" text;--> statement-breakpoint
+ALTER TABLE "security_audit_log" ADD COLUMN "schema_version" smallint;--> statement-breakpoint
+ALTER TABLE "security_audit_log" ADD COLUMN "finding_snapshot" jsonb;--> statement-breakpoint
+ALTER TABLE "security_audit_log" ADD COLUMN "source_context" text;--> statement-breakpoint
+CREATE UNIQUE INDEX "UQ_security_audit_log_org_event_key" ON "security_audit_log" USING btree ("owned_by_organization_id","event_key") WHERE "security_audit_log"."owned_by_organization_id" IS NOT NULL AND "security_audit_log"."event_key" IS NOT NULL;--> statement-breakpoint
```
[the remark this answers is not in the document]
jeanduplessis: Leaving the generated migration as-is. The table is still small enough that we are accepting the non-concurrent index build/lock cost for this migration.

apps/web/src/lib/security-agent/services/auto-dismiss-service.ts
```diff
@@ -42,6 +54,39 @@ function toOwner(securityOwner: SecurityReviewOwner, userId: string): Owner {
...
+  if ('organizationId' in owner && owner.organizationId) {
+    return eq(security_findings.owned_by_organization_id, owner.organizationId);
+  }
+  if ('userId' in owner && owner.userId) {
+    return eq(security_findings.owned_by_user_id, owner.userId);
+  }
+  throw new Error('Invalid owner: must have either organizationId or userId');
+}
+
+function toAuditFinding(
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in ab275929b by removing the identity wrapper and passing the updated finding directly into the audit event writer.

apps/web/src/lib/security-agent/services/auto-dismiss-service.ts
```diff
@@ -420,17 +540,26 @@ export async function countEligibleForAutoDismiss(
...
+      !triage ||
+      analysis?.sandboxAnalysis !== undefined ||
+      triage.needsSandboxAnalysis !== false ||
+      triage.suggestedAction !== 'dismiss'
+    ) {
+      continue;
     }
+
+    eligible++;
+    byConfidence[triage.confidence]++;
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in ab275929b by validating triage confidence against the supported values before incrementing the count, with a regression test for malformed persisted confidence.

services/security-sync/src/sync.ts
```diff
@@ -1575,6 +1907,14 @@ async function syncRepo(params: {
           slaDueAt,
         });
 
+        await writeSyncFindingAuditEvents(tx, {
```
St0rmz1, line 1910: **Moderate, Functional:** The audit event write shares this transaction with the finding upsert, so a snapshot validation failure rolls back the upsert and the finding never syncs.

`writeSyncFindingAuditEvents` builds the snapshot from the finding (including `title` from the upstream advisory), and `SecurityFindingAuditSnapshotSchema.superRefine(validateSafeAuditJson)` throws if any string value matches `EMAIL_VALUE_PATTERN`, plus strict `dependabot_html_url: z.string().url()` and `cvss_score: z.number().finite()`. When that throws, the whole transaction rolls back and the per-finding `catch` below only increments `result.errors`. The finding is never persisted and every later sync run hits the same failure with no recovery path.

Before this change, any alert `parseDependabotAlert` could parse would sync. Now a parseable alert whose snapshot trips validation (most plausibly a `title` containing an email-like substring, or a malformed `dependabot_html_url`) silently fails to sync forever, so a real vulnerability never appears in the dashboard and the only signal is a `console.error`.

Decide whether audit-write failure should be allowed to fail the authoritative finding mutation. If atomicity is intended, sanitize/coerce the customer-uncontrolled fields (`title`, `dependabot_html_url`, `cvss_score`) instead of throwing, so upstream data can never block the core write. If audit evidence is best-effort relative to sync, isolate the audit write so its failure is logged without rolling back the upsert. Either way, add a test with an advisory title containing an email substring and a malformed `html_url`.

apps/web/src/lib/security-agent/db/security-audit-report.ts
```diff
@@ -0,0 +1,994 @@
...
+    byAction[row.action] = (byAction[row.action] ?? 0) + 1;
+  }
+
+  return {
+    reportVersion: SECURITY_AGENT_AUDIT_REPORT_VERSION,
+    owner: params.owner,
+    period: params.period,
+    generatedAt: params.generatedAt,
+    dataThrough: params.dataThrough,
+    reliableCoverageStart: resolveSecurityAgentAuditReliableCoverageStart(),
```
St0rmz1, line 541: **Low, Functional:** `resolveSecurityAgentAuditReliableCoverageStart()` is called unconditionally here during report assembly and throws a plain `Error` if `SECURITY_AGENT_AUDIT_RELIABLE_COVERAGE_START` is unset. I do not see the var added to the env schema or `.env.example` in this PR, and because it throws a plain `Error` (not `SecurityAgentAuditReportQueryError`), `assembleAuditReportResponse` rethrows it as a 500 rather than the graceful `query_failed` path.

If the var is not configured in every environment running `apps/web` before #4082 ships the UI, `getAuditReport` fails hard for all callers, and it is easy to miss because the rest of the report path degrades gracefully. Add it to the environment schema / `.env.example` and confirm it is set in all deploy targets, or provide a safe default. Consider mapping the missing-config case to `query_failed` (or an explicit config error) instead of an unhandled 500.

apps/web/src/lib/security-agent/db/security-audit-report.ts
```diff
@@ -0,0 +1,994 @@
...
+  }
+
+  try {
+    return await withSecurityAgentAuditReportTimeout(
+      tx
+        .select({
+          id: security_audit_log.id,
+          action: security_audit_log.action,
+          actor_id: security_audit_log.actor_id,
+          actor_email: security_audit_log.actor_email,
```
St0rmz1: **Nit, Design:** `actor_email` is selected here and carried on `AuditReportRow`, but `buildReportActor` only reads `actor_id`, `actor_name`, and `actor_type`, and nothing else consumes it. It is not leaked to the client, but it is an unnecessary column fetch, and given the PII posture of this report it is cleaner not to load actor emails into report memory at all. Consider dropping it from the select.

St0rmz1 (approved):

apps/web/src/lib/security-agent/db/security-audit-report.ts
jeanduplessis: Fixed in 7174b18d8. Missing or invalid SECURITY_AGENT_AUDIT_RELIABLE_COVERAGE_START now raises SecurityAgentAuditReportQueryError so the existing router path returns query_failed instead of a plain 500. Also added the env var to .env.local.example, apps/web/.env.development.local.example, apps/web/.env.test, and ENVIRONMENT.md, with test coverage asserting the error type.

apps/web/src/lib/security-agent/db/security-audit-report.ts
jeanduplessis: Fixed in 7174b18d8. Removed actor_email from the report row type and scan select, and updated the row fixtures so report actor rendering continues to rely only on actor_id, actor_name, and actor_type.

services/security-sync/src/sync.ts
jeanduplessis: Fixed in 7174b18d8. Kept the finding mutation and audit event atomic, but made the snapshot builder sanitize source-boundary evidence before validation: email-like snapshot text is redacted, malformed or non-http(s) Dependabot URLs are omitted, and non-finite CVSS values are dropped. Added a sync regression test that imports a parseable alert with an email-like advisory title and malformed html_url and verifies sync plus audit insertion still succeed.

apps/web/src/lib/security-agent/db/security-audit-report.ts
```diff
@@ -0,0 +1,997 @@
...
+  }
+
+  const fixedAt = stringFromSnapshot(snapshot, 'fixed_at');
+  if (fixedAt) {
+    const fixedAtMs = Date.parse(fixedAt);
+    if (Number.isNaN(fixedAtMs)) {
+      return { status: 'unknown', deadline, reason: 'invalid_terminal_timestamp' };
+    }
+    return {
+      status: fixedAtMs <= deadlineMs ? 'terminal_met' : 'terminal_missed',
```
pandemicsyn: Must-fix: This treats `fixed_at === sla_due_at` as `terminal_met`. The spec defines SLA report evidence as before vs at/after the recorded deadline, and SLA breach semantics make equality breached/missed. Use `fixedAtMs < deadlineMs` for met, otherwise `terminal_missed`, and add an equality regression test.

apps/web/src/lib/security-agent/db/security-audit-report.ts
```diff
@@ -0,0 +1,997 @@
...
+    return { status: 'unknown', deadline, reason: 'missing_open_status_evidence' };
+  }
+
+  const cutoffMs = Date.parse(dataThrough);
+  if (Number.isNaN(cutoffMs)) {
+    return { status: 'unknown', deadline, reason: 'invalid_report_cutoff' };
+  }
+
+  return {
+    status: cutoffMs <= deadlineMs ? 'open_within_deadline' : 'open_past_deadline',
```
pandemicsyn: Must-fix: This reports `dataThrough === sla_due_at` as `open_within_deadline`. At the exact deadline an open finding is already at/after the deadline, so the report should show it as past deadline. Use `cutoffMs < deadlineMs` for within-deadline and cover the equality case in tests.

apps/web/src/lib/security-agent/db/security-audit-report.ts
```diff
@@ -0,0 +1,997 @@
...
+    patchedVersion: stringFromSnapshot(latestSnapshot, 'patched_version'),
+    ghsaId: stringFromSnapshot(latestSnapshot, 'ghsa_id'),
+    cveId: stringFromSnapshot(latestSnapshot, 'cve_id'),
+    cweIds: stringArrayFromSnapshot(latestSnapshot, 'cwe_ids'),
+    cvssScore: cvssFromSnapshot(latestSnapshot),
+    dependabotUrl: safeUrl(stringFromSnapshot(latestSnapshot, 'dependabot_html_url')),
+    firstDetectedAt: stringFromSnapshot(evidenceSnapshot, 'first_detected_at'),
+    canonicalFindingId: stringFromSnapshot(latestSnapshot, 'canonical_finding_id'),
+    deleted,
+    sla: buildSlaEvidence(evidenceSnapshot, reportParams.dataThrough),
```
pandemicsyn: Must-fix: Deleted findings still derive SLA evidence from the latest snapshot. If the final deletion evidence still has `status: "open"`, the report can show `deleted: true` and `sla.status: "open_past_deadline"`, which is contradictory and not trustworthy evidence. Pass deletion evidence into SLA derivation and return `unknown` unless a terminal timestamp is recorded.

apps/web/src/lib/security-agent/db/security-audit-report.ts
```diff
@@ -0,0 +1,997 @@
...
+  reportParams: {
+    dataThrough: string;
+    isRequestingUserKiloAdmin: boolean;
+  }
+): SecurityFindingAuditSection {
+  rows.sort(
+    (left, right) =>
+      left.effective_at.localeCompare(right.effective_at) || left.id.localeCompare(right.id)
+  );
+  const latestSnapshot = latestFindingSnapshot(rows);
```
pandemicsyn, line 836: Must-fix: SLA state is derived only from in-period snapshots, but open/terminal status is evaluated against `dataThrough`. A report can include an in-period open snapshot, miss a later out-of-period fixed event, and incorrectly report the finding as open/past deadline. Load latest state through `dataThrough` for included findings, or return `unknown` when the period rows do not prove cutoff state.

pandemicsyn (changes requested):
Found a few audit report SLA correctness issues that should be fixed before merge.

pandemicsyn (approved):

apps/web/src/lib/security-agent/db/security-audit-report.ts
jeanduplessis: Fixed in fb06cdb0c. Report assembly now loads each included finding's latest owner-scoped audit snapshot and deletion evidence through dataThrough inside the same repeatable-read transaction. Timelines stay period-bounded while status and SLA use cutoff state. Integration coverage verifies later fixed and deleted events affect state without appearing in the selected-period timeline.

services/security-sync/src/sync.ts
jeanduplessis: Follow-up in fb06cdb0c: source text sanitization now redacts every email-like substring, not only the first. Expanded the sync regression to cover two addresses plus malformed html_url, closing the remaining validation-failure path.

apps/web/src/lib/security-agent/db/security-audit-report.ts
jeanduplessis: Fixed in fb06cdb0c. Terminal timestamps now count as met only when strictly before the recorded deadline; exact equality returns terminal_missed. Added equality regression coverage.

apps/web/src/lib/security-agent/db/security-audit-report.ts
jeanduplessis: Fixed in fb06cdb0c. SLA derivation now receives deletion evidence and returns unknown for deleted findings without a recorded terminal timestamp, while preserving trustworthy terminal evidence. Added deleted/open and fixed-then-deleted regressions.

apps/web/src/lib/security-agent/db/security-audit-report.ts
jeanduplessis: Fixed in fb06cdb0c. Open findings now remain within deadline only while dataThrough is strictly before the deadline; exact equality returns open_past_deadline. Added equality regression coverage.

packages/db/src/migrations/0167_flawless_true_believers.sql
```diff
@@ -0,0 +1,16 @@
...
+CREATE UNIQUE INDEX "UQ_security_audit_log_user_event_key" ON "security_audit_log" USING btree ("owned_by_user_id","event_key") WHERE "security_audit_log"."owned_by_user_id" IS NOT NULL AND "security_audit_log"."event_key" IS NOT NULL;--> statement-breakpoint
+CREATE INDEX "IDX_security_audit_log_org_occurred" ON "security_audit_log" USING btree ("owned_by_organization_id","occurred_at","id") WHERE "security_audit_log"."owned_by_organization_id" IS NOT NULL AND "security_audit_log"."occurred_at" IS NOT NULL;--> statement-breakpoint
+CREATE INDEX "IDX_security_audit_log_user_occurred" ON "security_audit_log" USING btree ("owned_by_user_id","occurred_at","id") WHERE "security_audit_log"."owned_by_user_id" IS NOT NULL AND "security_audit_log"."occurred_at" IS NOT NULL;--> statement-breakpoint
+ALTER TABLE "security_audit_log" ADD CONSTRAINT "security_audit_log_actor_type_check" CHECK ("security_audit_log"."actor_type" IN ('customer_user', 'kilo_admin', 'system'));--> statement-breakpoint
```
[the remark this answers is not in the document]
jeanduplessis: This is acceptable as the table is small still.

PR #4082, opened 2026-06-17
feat(security-agent): add audit report UI

jeanduplessis:
## Summary

Security Agent now has the web UI for owner-scoped Audit Reports and a refreshed management experience, stacked on the backend/API foundation in #4081.

### Why this change is needed

Security owners need an interactive way to review recorded Security Finding evidence by period, repository, severity, and state. Splitting the UI keeps the deployable backend migration and API review separate from the larger visual and interaction changes.

### How this is addressed

- Adds personal and organization Audit Report routes with an interactive report page, UTC date-range selection, provenance messaging, evidence filters, and deterministic finding timelines.
- Refreshes Security Agent dashboard, finding list, finding detail, settings, navigation, help, and action surfaces around clearer user actions and outcomes.
- Adds supporting Security Agent UI helpers for dismissal, remediation-unavailable copy, finding-list presentation, and command invalidation.
- Adds shared UI primitive updates plus `react-day-picker` for report date-range selection.

## Verification

- Reviewed the stacked diff against #4081 and confirmed it is limited to web routes, React components, shared UI primitives, and the UI-only package/lockfile changes.
- No browser or manual visual verification was performed in this environment; screenshots were not captured.

## Visual Changes

localhost_3000_security-agent (6)
localhost_3000_security-agent (7)
localhost_3000_security-agent (8)
localhost_3000_security-agent (9)
localhost_3000_security-agent (10)
localhost_3000_security-agent (11)
localhost_3000_security-agent (12)
localhost_3000_security-agent (13)
localhost_3000_security-agent (14)
localhost_3000_security-agent (15)
localhost_3000_security-agent (16)
localhost_3000_security-agent (17)

### Human Reviewer Flags

- Review after #4081. This PR compiles against the audit report tRPC endpoints and report types introduced there.
- This branch changes shared UI primitives and global app styling alongside Security Agent surfaces, so visual regression scope is broader than the report page alone.
- Confirm the final navigation and historical-access behavior matches the Audit Report guarantees introduced in #4081.
- Browser screenshots were not captured during PR preparation.

### Code Reviewer Agent

<details>

- Audit Report UI keeps filtering client-side so matching finding groups retain their complete in-period timelines.
- Date range and filter state are URL-backed, with repository options derived from recorded report evidence rather than current repository selection.
- Focus review on route gating, responsive layout, keyboard/focus behavior for report filters, and finding detail action availability copy.
- Backend/API behavior is intentionally excluded and lives in #4081.

</details>


apps/web/src/components/security-agent/SecurityAuditReportPage.tsx
```diff
@@ -0,0 +1,1647 @@
...
+        ...state,
+        draftFilters: ALL_AUDIT_REPORT_FILTERS,
+        submittedFilters: ALL_AUDIT_REPORT_FILTERS,
+      };
+  }
+}
+
+export function SecurityAuditReportPage() {
+  const trpc = useTRPC();
+  const searchParams = useSearchParams();
```
St0rmz1, line 259: **Low, Functional:** Filter and date-range state is read from the URL on mount but never written back. `useSearchParams()` is read once inside the `useReducer` initializer (here and the `parseAuditReportFilters` call below), and `submit-report`/filter changes only update local reducer state. There is no `router.replace`/`pushState` anywhere in the file.

The PR notes describe this state as "URL-backed," but it is one-way: a pasted deep link is honored on first load, yet once the user changes the period or a filter the address bar goes stale. They cannot bookmark or share the view they are looking at, and back/forward will not move between filter states. For an audit report this sharing case is plausible.

Suggestion: on `submit-report` (and optionally filter changes), push the current range and filters to the URL with `router.replace`. If read-only deep-linking is the intended scope, consider softening the "URL-backed" wording so it does not imply round-tripping.

apps/web/src/components/security-agent/security-finding-list-presentation.ts
```diff
@@ -0,0 +1,186 @@
...
+
+export function getFindingListGridClass(showSla: boolean) {
+  return showSla ? gridWithSla : gridWithoutSla;
+}
+
+function isSupersededFinding(finding: SecurityFinding) {
+  return finding.status === 'ignored' && finding.ignored_reason?.startsWith('superseded:');
+}
+
+export function getAnalysisPresentation(finding: SecurityFinding): FindingStatusPresentation {
```
St0rmz1: **Low, Design:** This `getAnalysisPresentation` and `getAnalysisStatus` in `FindingDetailDialog.tsx` both map `analysis_status` + `sandboxAnalysis.isExploitable` + `triage.suggestedAction` to a label and tone, and they have already drifted. This helper omits the `sandbox.extractionStatus === 'failed'` and `isExploitable === 'unknown'` branches the dialog handles, and labels the not-exploitable case "Not exploitable" where the dialog says "No reachable path."

The result is that the same finding can read differently depending on whether it is shown in the list or the detail dialog. Only this helper is unit-tested.

Suggestion: extract one shared, tested helper for the analysis-status-to-presentation mapping (surfaces can still pick their own icon/tooltip from a common state), or at minimum add a test around the dialog's `getAnalysisStatus` so the two stay aligned.

apps/web/src/components/security-agent/SecurityAuditReportPage.tsx
```diff
@@ -0,0 +1,1647 @@
...
+function formatDateInput(value: Date): string {
+  return value.toISOString().slice(0, 10);
+}
+
+function formatDate(value: string): string {
+  return REPORT_DATE_FORMATTER.format(new Date(value));
+}
+
+function formatReportDateTime(value: string): string {
+  return formatDateTime24Hour(value).replace(/, (\d{2}:\d{2} UTC)$/, ' at $1');
```
St0rmz1: **Nit, Design:** `formatReportDateTime` reshapes the formatted string with `.replace(/, (\d{2}:\d{2} UTC)$/, ' at $1')`, which depends on the exact `Intl.DateTimeFormat` output shape. If an ICU/runtime version formats the parts differently the replace silently no-ops and you get the unmodified string. Degradation is graceful, just brittle. Assembling the date and time from `formatToParts` (or composing two formatters) would avoid relying on the output layout.

apps/web/src/components/security-agent/SecurityAuditReportPage.tsx
```diff
@@ -0,0 +1,1647 @@
...
+    ? normalizeAuditReportRepositoryFilter(draftFilters, repositoryOptions)
+    : draftFilters;
+  const effectiveSubmittedFilters = unfilteredReport
+    ? normalizeAuditReportRepositoryFilter(submittedFilters, repositoryOptions)
+    : submittedFilters;
+  const report = unfilteredReport
+    ? filterSecurityAgentAuditReport(unfilteredReport, effectiveSubmittedFilters)
+    : null;
+
+  function handleGenerateReport(event: FormEvent<HTMLFormElement>) {
```
pandemicsyn, line 334: Must-fix: clicking Generate report with the same submitted date range only dispatches reducer state; it does not refetch the TanStack Query because the query key is unchanged. That makes the "generate the report again" retry path a no-op after transient load/query failures, can leave the displayed data cutoff stale after new Security Finding activity, and skips a fresh backend report access for what the UI presents as a new generation. Capture `refetch` from `useQuery` and call it when the submitted range is unchanged, or add an explicit generation nonce/invalidation so same-range submits force a new request.

pandemicsyn (approved):

apps/web/src/components/security-agent/SecurityAuditReportPage.tsx
jeanduplessis: Fixed in f5443814b. Same-range Generate report submissions now call the TanStack Query refetch function, so transient failures retry and the report data cutoff refreshes. A changed range still fetches through the query-key update.

apps/web/src/components/security-agent/SecurityAuditReportPage.tsx
jeanduplessis: Valid; fixed in f5443814b. Report timestamps now compose independently formatted UTC date and time values instead of reshaping Intl output with a regex, preserving the existing display without depending on ICU punctuation.

apps/web/src/components/security-agent/SecurityAuditReportPage.tsx
jeanduplessis: Valid; fixed in f5443814b. Submitted dates and filters now round-trip through the URL with unrelated parameters preserved, Clear filters updates the URL, and reducer state resynchronizes when search parameters change so browser navigation and shared links reflect the displayed report.

apps/web/src/components/security-agent/security-finding-list-presentation.ts
jeanduplessis: Valid; fixed in f5443814b. List and detail now use a shared getFindingAnalysisState classifier. Extraction failures and unknown exploitability consistently show Needs review, and the not-exploitable label is aligned to No reachable path. Added regression coverage for both missing branches.

PR #4088, opened 2026-06-18
fix(email): prevent organization invite autolinks

RSO:
## Summary
- Render organization names in team invitation emails with HTML escaping plus autolink neutralization so URL-like names cannot become active links in mail clients.
- Preserve existing organization-name validation semantics to avoid breaking production organizations with established punctuation and naming formats.
- Add regression coverage for email rendering hardening and for accepting URL-like organization names at input while neutralizing them at output.

## Verification
Not manually verified; no visual UI changes.

## Visual Changes
N/A

## Reviewer Notes
Prompted by a HackerOne report about URL-like organization names becoming clickable in official team invitation emails. The existing template already escaped HTML, so the fix targets mail-client autolinking rather than raw HTML injection or broad input rejection.

### DIFF
--- a/apps/web/src/lib/email.ts
+++ b/apps/web/src/lib/email.ts
@@ -79,6 +79,20 @@ export class RawHtml {
 
 type TemplateVars = Record<string, string | RawHtml>;
 
+const AUTOLINK_PROTOCOL_PATTERN = /\b(https?):\/\//gi;
+const AUTOLINK_WWW_PATTERN = /\bwww\./gi;
+const AUTOLINK_BARE_DOMAIN_PATTERN =
+  /\b([a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?)(\.)([a-z]{2,})(?=\b)/gi;
+
+export function renderNonAutolinkedText(str: string): RawHtml {
+  return new RawHtml(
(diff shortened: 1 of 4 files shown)


apps/web/src/lib/email.ts
```diff
@@ -79,6 +79,20 @@ export class RawHtml {
...
+const AUTOLINK_WWW_PATTERN = /\bwww\./gi;
+const AUTOLINK_BARE_DOMAIN_PATTERN =
+  /\b([a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?)(\.)([a-z]{2,})(?=\b)/gi;
+
+export function renderNonAutolinkedText(str: string): RawHtml {
+  return new RawHtml(
+    escapeHtml(str)
+      .replace(AUTOLINK_PROTOCOL_PATTERN, '$1:&#8203;//')
+      .replace(AUTOLINK_WWW_PATTERN, 'www&#8203;.')
+      .replace(AUTOLINK_BARE_DOMAIN_PATTERN, '$1&#8203;$2$3')
```
markijbema: this feels really close to just replacing all : / . given we know that there arent that many in a name

markijbema (approved):

PR #4090, opened 2026-06-18
fix(review-memory): enforce structured output

alex-alecu:
## Summary

Review Memory now asks the model for schema-checked JSON instead of trying to recover JSON from plain text. Bad or incomplete replies fail before any GitHub branch or pull request is created, while safe request details make failures easier to trace without storing prompts or model output. Structured-output routing is applied only where the selected provider supports it.

## Verification

1. Approve a Review Memory proposal with a valid model reply and confirm the REVIEW.md pull request opens.
2. Return malformed or cut-off model output and confirm the proposal becomes retryable without creating a GitHub branch.
3. Use a direct provider and confirm OpenRouter-only routing settings are not sent upstream.

## Visual Changes

N/A

## Reviewer Notes

Provider-facing schemas remove unsupported rules, but the complete Zod schemas still validate every model reply inside the app. Failure reports contain only fixed messages, counts, and request IDs; they do not contain prompts, repository content, or raw model output.


apps/web/src/lib/code-reviews/review-memory/change-request.ts
```diff
@@ -229,6 +343,124 @@ function assertGitHubPermissions(integration: PlatformIntegration): void {
...
+  const stackFrames = error.stack
+    .split('\n')
+    .slice(1)
+    .filter(line => /^\s*at\s/.test(line))
+    .slice(0, 20)
+    .join('\n');
+  return stackFrames ? sanitizeReviewMemoryDiagnosticMessage(stackFrames) : null;
+}
+
+function getSafeErrorClass(error: unknown): string {
```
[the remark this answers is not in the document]
alex-alecu: Fixed in 77a4160.

apps/web/src/lib/ai-gateway/providers/apply-provider-specific-logic.ts
```diff
@@ -113,6 +113,24 @@ export function applyGatewayModelsFallback(
...
+    providerId !== 'openrouter' ||
+    requestToMutate.kind !== 'chat_completions' ||
+    requestToMutate.body.response_format?.type !== 'json_schema'
+  ) {
+    return;
+  }
+
+  requestToMutate.body.provider = {
+    ...requestToMutate.body.provider,
+    require_parameters: true,
```
chrarnoldus: in what scenarios does setting this parameter make a difference?

alex-alecu: It helps filter OpenRouter models that don't support `response_format: { type: "json_schema" }`

chrarnoldus: can you set the parameter from the client?

alex-alecu: Moved this out of the gateway. Review Memory now sets `provider.require_parameters = true` from its OpenAI-compatible client via `transformRequestBody`, so the requirement is scoped to these structured-output calls instead of generic gateway behavior. Pushed in 6e7f00a4a.

chrarnoldus (approved):

PR #4091, opened 2026-06-18
fix(review): use kilocode-bot model for fix sessions

alex-alecu:
## Summary
- Resolve review fix-session models from the exact GitHub integration row that created the review, matching KiloCode Bot's `metadata.model_slug` behavior.
- Fall back to `kilo-auto/frontier` when the review has no linked integration, the row was deleted, or the stored bot model is invalid.
- Fail closed when a linked integration row no longer matches the authorized review's platform or user/organization owner, avoiding owner-wide installation lookup.

## Reviewer Notes
Cloud Agent still validates the submitted catalog model in the correct personal or organization context. Historical `cloud_agent_code_reviews.model` usage data remains reporting-only and is not reused for fix sessions.

### DIFF
--- a/apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts
+++ b/apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts
@@ -1,10 +1,7 @@
 import type { NextRequest } from 'next/server';
 import { NextResponse } from 'next/server';
 import { buildFixReviewPrompt } from '@/lib/code-reviews/prompts/fix-review-prompt';
-import {
-  DEFAULT_CODE_REVIEW_MODE,
-  DEFAULT_CODE_REVIEW_MODEL,
-} from '@/lib/code-reviews/core/constants';
+import { DEFAULT_CODE_REVIEW_MODE } from '@/lib/code-reviews/core/constants';
 import { createCallerFactory, createTRPCContext } from '@/lib/trpc/init';
 import { rootRouter } from '@/routers/root-router';
 import { TRPCError } from '@trpc/server';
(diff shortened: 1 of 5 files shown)


apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts
```diff
@@ -72,17 +71,29 @@ export async function GET(request: NextRequest, context: RouteContext) {
...
-
   try {
     const organizationId = review.owned_by_organization_id;
+    const integration = review.platform_integration_id
+      ? await getIntegrationById(review.platform_integration_id)
+      : null;
+
+    if (
+      integration &&
+      (integration.platform !== review.platform ||
```
Comment (RSO) [inline apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts:82]: How would this happen? We fetch the integration based on `review.platform_integration_id`, so this if is checking whether we have reviews that have a `platform_integration_id` that belongs to an integration that doesn't have the same platform as the review?

apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts
```diff
@@ -72,17 +71,29 @@ export async function GET(request: NextRequest, context: RouteContext) {
...
+        integration.owned_by_user_id !== review.owned_by_user_id ||
+        integration.owned_by_organization_id !== review.owned_by_organization_id)
```
Comment (RSO) [inline apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts:84]: Also here, what's the case that we're covering for?

apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts
```diff
@@ -72,17 +71,29 @@ export async function GET(request: NextRequest, context: RouteContext) {
...
+    ) {
+      return redirectToError(url.origin, 'fix_session_failed');
+    }
+
+    const sessionInput = {
+      githubRepo: review.repo_full_name,
+      prompt: buildFixReviewPrompt(review.pr_url),
+      mode: DEFAULT_CODE_REVIEW_MODE,
```
Comment (RSO) [inline apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts:92]: Shouldn't this be `code` mode if it's trying to fix things?

RSO (approved):

apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts
Comment (alex-alecu) [reply apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts:82]: It should not happen in normal data. This fails closed if a saved review points at an integration from the wrong platform, so we do not use the wrong bot settings.

apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts
Comment (alex-alecu) [reply apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts:92]: It is already code mode. DEFAULT_CODE_REVIEW_MODE is 'code'.

apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts
Comment (alex-alecu) [reply apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts:84]: This is the same safety check for owner data. If the saved review points at another user or org's integration, we stop instead of using that bot model.

apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts
Comment (alex-alecu) [reply apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts:82]: E.g. kilocode-bot enabled for Github and code reviewer enabled for Gitlab. Rare, but possible

PR #4111, opened 2026-06-18
fix(mcp-gateway): harden OAuth consent screen

pandemicsyn:
## Summary

- Protect users from OAuth consent phishing where publicly registered MCP clients can supply self-asserted metadata and callback URIs.
- Present the requesting client as unverified without displaying its self-asserted app name, and show the stable client ID, exact callback destination, MCP connection, configured endpoint, owner context, granting Kilo account, and effective broad MCP access before approval.
- Add an explicit denial path and harden browser approval with user/request binding, per-flow cookies, server-enforced expiry, safe POST redirects, and restrictive response headers. The change preserves the existing OAuth protocol and requires no database migration.

## Verification

- [x] Manually verified the local consent screen displays the unverified-client warning, connection and account context, broad-access warning, callback destination, and allow/deny actions.
- [x] Manually denied access to Kilo CLI and confirmed the browser follows the `303` to the localhost callback, displays the authorization failure, and releases the CLI from `Starting OAuth flow` without issuing an authorization code.

## Visual Changes

New consent screen (lots of warnings, way more details):

Screenshot 2026-06-18 at 2 27 58 PM

What kilo shows when you click deny:

Screenshot 2026-06-18 at 2 28 01 PM

## Reviewer Notes

- For Kilo CLI after a user clicks **Deny access**, the browser follows the gateway's `303` to the CLI's localhost callback, shows `Authorization Failed` and the denial description, and the terminal stops `Starting OAuth flow` and reports `Authentication failed`. The consent CSP explicitly allows the validated callback origin so the browser can complete that redirect.
- The approval signature now binds the authenticated user and full OAuth request context; scoped POST errors use `303` so form bodies are not replayed to client callbacks.
- Approval remains signed and stateless with a five-minute server-enforced lifetime, and the change adds no schema or migration work.

### DIFF
--- a/apps/web/src/app/api/mcp-gateway/oauth/authorize/route.ts
+++ b/apps/web/src/app/api/mcp-gateway/oauth/authorize/route.ts
@@ -30,11 +30,12 @@ const consentDecisionValues = ['allow', 'deny'] as const;
 type ConsentDecision = (typeof consentDecisionValues)[number];
 
 function consentSecurityHeaders(redirectUri: string) {
-  const callbackOrigin = new URL(redirectUri).origin;
+  const callback = new URL(redirectUri);
+  const callbackSource = callback.protocol === 'http:' ? ` ${callback.origin}` : '';
   return {
     'Cache-Control': 'no-store',
     Pragma: 'no-cache',
-    'Content-Security-Policy': `default-src 'none'; style-src 'unsafe-inline'; form-action 'self' ${callbackOrigin}; frame-ancestors 'none'; base-uri 'none'; object-src 'none'`,
+    'Content-Security-Policy': `default-src 'none'; style-src 'unsafe-inline'; form-action 'self' https:${callbackSource}; frame-ancestors 'none'; base-uri 'none'; object-src 'none'`,
     'X-Frame-Options': 'DENY',
     'X-Content-Type-Options': 'nosniff',
     'Referrer-Policy': 'no-referrer',
--- a/apps/web/src/app/api/mcp-gateway/oauth/authorize/route.ts
+++ b/apps/web/src/app/api/mcp-gateway/oauth/authorize/route.ts
@@ -154,7 +154,6 @@ function clearConsentCookie(response: NextResponse, request: NextRequest, approv
 function consentDocument(params: {
   action: string;
   approvalState: string;
-  clientName: string;
   clientId: string;
   redirectUri: string;
   connectionName: string;
@@ -270,7 +269,6 @@ function consentDocument(params: {
         line-height: 1.55;
         text-wrap: pretty;
       }
-      .client-name { color: var(--foreground); font-weight: 700; }
       .client-id { color: var(--muted); font-size: 0.76rem; }
       .warning {
         margin-bottom: 1.25rem;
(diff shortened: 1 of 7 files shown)


apps/web/src/app/api/mcp-gateway/oauth/authorize/route.ts
```diff
@@ -25,6 +26,21 @@ const authorizationSingletonParams = [
...
 ] as const;
+const consentDecisionValues = ['allow', 'deny'] as const;
+type ConsentDecision = (typeof consentDecisionValues)[number];
+
+function consentSecurityHeaders(redirectUri: string) {
+  const callbackOrigin = new URL(redirectUri).origin;
+  return {
+    'Cache-Control': 'no-store',
+    Pragma: 'no-cache',
+    'Content-Security-Policy': `default-src 'none'; style-src 'unsafe-inline'; form-action 'self' ${callbackOrigin}; frame-ancestors 'none'; base-uri 'none'; object-src 'none'`,
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in acc50a02a. Consent forms now allow HTTPS provider redirect chains while retaining an exact exception for validated HTTP loopback callbacks.

apps/web/src/app/api/mcp-gateway/oauth/authorize/route.ts
```diff
@@ -25,6 +26,22 @@ const authorizationSingletonParams = [
...
+const consentDecisionValues = ['allow', 'deny'] as const;
+type ConsentDecision = (typeof consentDecisionValues)[number];
+
+function consentSecurityHeaders(redirectUri: string) {
+  const callback = new URL(redirectUri);
+  const callbackSource = callback.protocol === 'http:' ? ` ${callback.origin}` : '';
+  return {
+    'Cache-Control': 'no-store',
+    Pragma: 'no-cache',
+    'Content-Security-Policy': `default-src 'none'; style-src 'unsafe-inline'; form-action 'self' https:${callbackSource}; frame-ancestors 'none'; base-uri 'none'; object-src 'none'`,
```
St0rmz1, line 38: **Low, Security:** `form-action 'self' https:${callbackSource}` permits form-driven navigation to any HTTPS origin (the explicit origin is only added for HTTP loopback). This is the intended fix for provider redirects (the post-approval `303` targets an arbitrary HTTPS provider URL not known at render time), so it cannot be pinned tighter without breaking OAuth-backed connections. The residual risk is low: `default-src 'none'` blocks scripts and every interpolated value is escaped. Suggest a one-line comment here recording why `https:` is intentionally broad, so a future tightening does not silently break provider redirects.

apps/web/src/app/api/mcp-gateway/oauth/authorize/route.ts
```diff
@@ -73,49 +93,111 @@ async function authorizeRequest(
...
   resource: string;
   scopes: string[];
+  oauthState: string | null;
+  codeChallenge: string | null;
+  codeChallengeMethod: string | null;
   executionContext: ReturnType<typeof executionContextFromAuth>;
   secret: string;
 }) {
-  return hmacValue(JSON.stringify(params), params.secret);
+  const { secret, ...payload } = params;
```
St0rmz1, line 119: **Low, Design:** The signature is `hmacValue(JSON.stringify(payload), secret)`, so it depends on object key insertion order. The two call sites that build this payload by hand (the GET issue site around line 440 and the POST verify site around line 532) currently use identical key order, but if a future edit reorders or adds a field in only one of them, `JSON.stringify` produces a different string and every approval silently fails with `access_denied`. The GET to POST roundtrip tests would catch a pure reorder, but consider extracting a single helper that takes the raw inputs and returns the canonical signing payload, called from both sites, so the invariant cannot be broken one-sided.

St0rmz1 (approved):

PR #4117, opened 2026-06-19
ci(cloud-agent-next): add staging worker config

RSO:
## Summary

Adds an explicit `staging` Wrangler environment for `services/cloud-agent-next` so it can participate in the staging Worker deployment matrix without reusing production routes, queues, R2 buckets, or container capacity.

## Verification

Manual verification was limited to local config inspection and Wrangler dry-run behavior; live Cloudflare resource checks were blocked because local Wrangler auth does not have access to the configured Kilo Cloudflare account.

## Visual Changes

N/A

## Reviewer Notes

This PR targets `foamy-drip` so the diff only contains the cloud-agent-next staging configuration layered on top of the staging deployment workflow PR.

The staging config points service bindings at staging-named dependencies (`session-ingest-staging`, `git-token-service-staging`, `notifications-staging`) so staging traffic is not silently routed to production workers. Those dependency deployments must exist before a live deploy can fully succeed.


RSO:
Closing in favor of #4118, which is based directly on main and contains only the cloud-agent-next staging config change.

PR #4137, opened 2026-06-20
fix: redirect unauthenticated users from subscriptions page

### LINKED ISSUE #4136: Unauthenticated users can access /subscriptions and see raw authentication error
### What happened?

When a logged-out user navigates directly to https://app.kilo.ai/subscriptions, the page loads and displays the Subscriptions UI shell.

The page then attempts to fetch data, which returns a 401 Unauthorized response. Instead of redirecting the user to sign in (or showing a proper authentication screen), the UI displays the following error:

User not authenticated - no user to set on context

### Area

Bead Board / Dashboard

### Town ID (if known)

_No response_

### Steps to reproduce

1. Sign out of Kilo.
2. Open https://app.kilo.ai/subscriptions directly.
3. Observe that the page loads instead of redirecting.
4. Open DevTools → Network.
5. Observe an API request returning 401 Unauthorized.
6. The UI displays:
User not authenticated - no user to set on context

### Relevant logs or screenshots

Image

FaisalMisbah23:
## Summary

Fix unauthenticated access handling for the subscriptions page.

Previously, a logged-out user could navigate directly to `/subscriptions`. The page would render and then display an authentication error after a subscription-related API request returned 401.

This change adds an authentication check at the subscriptions layout level using `getUserFromAuthOrRedirect()`, ensuring unauthenticated users are redirected before subscription content is rendered.

Fixes #4136

## Verification

* [x] Logged out and attempted to access `/subscriptions`
* [x] Confirmed unauthenticated users are redirected instead of seeing the authentication error
* [x] Added test coverage for the subscriptions layout authentication requirement

## Visual Changes

N/A

## Reviewer Notes

This change applies the authentication guard at the layout level so all routes under `/subscriptions` require authentication before rendering.


apps/web/src/app/(app)/subscriptions/layout.test.tsx
```diff
@@ -0,0 +1,59 @@
+import type { ReactNode } from 'react';
```
jeanduplessis: Thanks for adding regression coverage here. One repository-specific detail: Jest currently matches only `**/src/**/*.test.ts` in `apps/web/jest.config.ts`, so this `.test.tsx` file is not discovered. Running it through the normal test configuration reports `No tests found`. Please rename this file to `layout.test.ts` (it contains no JSX) so the test runs in CI. This is required because the PR otherwise adds no executable regression test.

apps/web/src/app/(app)/subscriptions/layout.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+    customer_source: null,
+  } as User;
+}
+
+describe('SubscriptionsLayout', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('requires authentication before rendering subscription pages', async () => {
```
jeanduplessis: Could you please add an unauthenticated case instead of covering only successful authentication? Have `mockGetUserFromAuthOrRedirect` reject with a redirect sentinel and assert that the layout propagates the rejection rather than returning `children`. This verifies that auth failure stops rendering and that the guard remains awaited. Please also rename this happy-path case to describe authenticated behavior. This is required because the current assertion does not exercise the bug this PR fixes.

apps/web/src/app/(app)/subscriptions/layout.test.tsx
```diff
@@ -0,0 +1,59 @@
+import type { ReactNode } from 'react';
+import type { User } from '@kilocode/db/schema';
+
+const mockGetUserFromAuthOrRedirect = jest.fn<Promise<User>, []>();
+
+jest.mock('@/lib/user/server', () => ({
+  getUserFromAuthOrRedirect: () => mockGetUserFromAuthOrRedirect(),
+}));
+
+function makeUser(): User {
```
jeanduplessis: Please replace this hand-built `makeUser()` object and the `as User` cast with the shared `defineTestUser()` helper from `@/tests/helpers/user.helper`. The current object already omits required `User` fields, while the cast hides that drift; the shared helper uses `satisfies User` and stays aligned with the schema. This cleanup is required so the new regression test follows the repository fixture pattern and does not introduce a second stale User definition.

jeanduplessis (changes requested):
Thanks for reporting and fixing this. I reproduced the issue and confirmed the layout-level guard is the right approach and matches existing protected-layout patterns in this codebase. I am requesting changes only for regression coverage: the new test is not currently discovered, it does not exercise the unauthenticated path, and its local User fixture hides schema drift. Once those items are addressed, the production change looks good.

FaisalMisbah23:
@jeanduplessis Thanks for your feedback I missed the test discovery issue. I have updated layout.test.ts to use defineTestUser() and added the unauthenticated redirect sentinel case.

jeanduplessis (approved):

PR #4163, opened 2026-06-22
fix(kiloclaw): serialize credit billing mutations

jeanduplessis:
## Summary
Prevent concurrent KiloClaw credit billing operations from overdrawing balances or collapsing distinct Stripe payments.

### Why this change is needed
Credit enrollment checked available balance before entering its transaction, so concurrent enrollments for one user could make decisions from the same stale balance. Stripe-funded settlement also keyed deductions by billing period, which could treat separate payments in one period as duplicates and continue after an unexpected deduction conflict.

### How this is addressed
- Serialize each user's credit enrollment balance decision and mutation with a row lock.
- Revalidate target instance state inside transaction before deducting credits.
- Key Stripe-funded deductions by payment identity and roll back settlement when deduction cannot be recorded.
- Add regression coverage for concurrent enrollments, distinct same-period Stripe payments, and deduction conflicts.

## Human Verification
- Confirmed behavior against KiloClaw billing and billing-lifecycle specs.
- Ran targeted KiloClaw billing tests: 2 suites and 231 tests passed.

### Human Reviewer Flags
- Credit enrollment now holds user row lock while calculating effective balance and applying deduction, intentionally serializing concurrent billing operations for same user.
- Stripe-funded deduction identity changes from period-based to payment-based so distinct provider payments remain independently balance-neutral.

### Code Reviewer Agent
<details>

- Settlement deposit and deduction remain in one database transaction; deduction conflicts now throw so deposit and subscription mutation roll back together.
- Transaction rechecks instance ownership and destruction state after acquiring user lock.
- Existing settlement category parser remains compatible because settlement categories do not encode instance IDs.

</details>

### DIFF
--- a/apps/web/src/lib/kilo-pass/cancel-and-refund.ts
+++ b/apps/web/src/lib/kilo-pass/cancel-and-refund.ts
@@ -178,16 +178,6 @@ export async function cancelAndRefundKiloPassForUser({
   }
 
   const balanceResetAmountUsd = await db.transaction(async tx => {
-    const freshUserRows = await tx
-      .select({
-        microdollars_used: kilocode_users.microdollars_used,
-        total_microdollars_acquired: kilocode_users.total_microdollars_acquired,
-      })
-      .from(kilocode_users)
-      .where(eq(kilocode_users.id, userId))
-      .for('update')
-      .limit(1);
-
     await tx
       .update(kilo_pass_subscriptions)
       .set({
@@ -208,6 +198,16 @@ export async function cancelAndRefundKiloPassForUser({
         })
         .where(eq(kilocode_users.id, userId));
     }
+
+    const freshUserRows = await tx
+      .select({
+        microdollars_used: kilocode_users.microdollars_used,
+        total_microdollars_acquired: kilocode_users.total_microdollars_acquired,
+      })
+      .from(kilocode_users)
+      .where(eq(kilocode_users.id, userId))
+      .for('update')
+      .limit(1);
     const freshUser = freshUserRows[0];
     const currentBalanceMicrodollars = freshUser
       ? freshUser.total_microdollars_acquired - freshUser.microdollars_used
--- a/apps/web/src/lib/kilo-pass/cancel-and-refund.ts
+++ b/apps/web/src/lib/kilo-pass/cancel-and-refund.ts
@@ -178,6 +178,16 @@ export async function cancelAndRefundKiloPassForUser({
   }
 
   const balanceResetAmountUsd = await db.transaction(async tx => {
+    const freshUserRows = await tx
+      .select({
+        microdollars_used: kilocode_users.microdollars_used,
+        total_microdollars_acquired: kilocode_users.total_microdollars_acquired,
+      })
+      .from(kilocode_users)
+      .where(eq(kilocode_users.id, userId))
+      .for('update')
+      .limit(1);
+
     await tx
       .update(kilo_pass_subscriptions)
       .set({
@@ -198,16 +208,6 @@ export async function cancelAndRefundKiloPassForUser({
         })
         .where(eq(kilocode_users.id, userId));
     }
-
-    const freshUserRows = await tx
-      .select({
-        microdollars_used: kilocode_users.microdollars_used,
-        total_microdollars_acquired: kilocode_users.total_microdollars_acquired,
-      })
-      .from(kilocode_users)
-      .where(eq(kilocode_users.id, userId))
-      .for('update')
-      .limit(1);
     const freshUser = freshUserRows[0];
     const currentBalanceMicrodollars = freshUser
       ? freshUser.total_microdollars_acquired - freshUser.microdollars_used
--- a/apps/web/src/lib/kiloclaw/credit-billing.ts
+++ b/apps/web/src/lib/kiloclaw/credit-billing.ts
@@ -1,6 +1,6 @@
 import 'server-only';
 
-import { and, desc, eq, isNotNull, isNull, sql } from 'drizzle-orm';
+import { and, desc, eq, gte, isNotNull, isNull, like, lte, ne, sql } from 'drizzle-orm';
 import { addMonths, format } from 'date-fns';
 
 import { db } from '@/lib/drizzle';
@@ -638,6 +638,7 @@ export async function applyStripeFundedKiloClawPeriod(
     }
     const commitEndsAt = plan === 'commit' ? targetRow.commit_ends_at : null;
 
+    const deductionCategory = `kiloclaw-settlement:${stripeSubscriptionId}:payment:${stripePaymentId}`;
     const deposited = await processTopUp(
       user,
       amountCents,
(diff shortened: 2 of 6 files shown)


apps/web/src/lib/kiloclaw/credit-billing.ts
```diff
@@ -661,7 +659,7 @@ export async function applyStripeFundedKiloClawPeriod(
       return;
     }
 
-    const deductionCategory = `kiloclaw-settlement:${stripeSubscriptionId}:${periodStartDate}`;
+    const deductionCategory = `kiloclaw-settlement:${stripeSubscriptionId}:payment:${stripePaymentId}`;
```
St0rmz1: **Moderate, Functional:** The duplicate deposit branch returns before checking whether this payment has a matching deduction. This preserves historical cases where the old period based key allowed a second payment deposit but skipped its deduction, so replaying that payment still reports success while leaving the balance inflated. Please reconcile duplicate deposits by checking for the payment deduction and safely accounting for legacy period based deductions so an existing deduction is not applied twice. A regression test could seed two payment deposits with one legacy deduction, then replay the unmatched payment.

apps/web/src/lib/kiloclaw/credit-billing.ts
```diff
@@ -1435,7 +1409,32 @@ export async function enrollWithCredits(params: {
       );
     }
 
-    // 5a: Insert negative credit transaction with period-encoded idempotency key
+    const projectedUsage = user.microdollars_used + costMicrodollars;
```
St0rmz1, line 1485: **Moderate, Functional:** Intro pricing eligibility is calculated before the user lock is acquired. Two concurrent enrollments can both observe no prior paid subscription; after serialization, the second request still uses the intro price even though the first has established paid history. Please recompute paid subscription history inside the transaction after acquiring the user lock, then derive the intro decision, cost, SKU, and effective balance from that current state. A concurrency test with two eligible legacy trial lineages should assert that only one enrollment receives intro pricing.

apps/web/src/lib/kiloclaw/credit-billing.ts
```diff
@@ -1435,7 +1409,32 @@ export async function enrollWithCredits(params: {
...
+    const effectiveThreshold = getEffectiveKiloPassThreshold(user.kilo_pass_threshold);
+    const kiloPassSubscription =
+      effectiveThreshold !== null && projectedUsage >= effectiveThreshold
+        ? await getKiloPassStateForUser(tx, userId)
```
St0rmz1: **Moderate, Functional:** The effective balance decision reads Kilo Pass entitlement while holding the user lock, but entitlement cancellation can update the Kilo Pass row before waiting for that same user lock during balance reset. Enrollment can therefore read the old active entitlement and approve a deduction using a projected bonus that will not be issued after cancellation commits, potentially leaving the user with a negative balance. Please establish a consistent lock order for entitlement changes and balance decisions, or lock and revalidate the selected Kilo Pass subscription before committing the deduction. A race test should cover cancellation overlapping enrollment.

apps/web/src/lib/kiloclaw/credit-billing.ts
St0rmz1, line 1485: **Moderate, Functional:** This revalidation reads the subscription without locking it. A lifecycle update can commit after this check but before the later upsert, allowing enrollment to overwrite fresh status or suspension state while also using the stale pre-transaction `wasSuspended` value. That can activate billing without running the required compute resume path. Please lock the subscription row with `FOR UPDATE` and derive suspension state from the locked row before building the mutation and post-commit resume decision. Add a race test where suspension overlaps enrollment.

St0rmz1 (approved):

apps/web/src/lib/kiloclaw/credit-billing.ts
jeanduplessis: Valid. Fixed in 054144eff. Duplicate deposits now reconcile payment-scoped deductions and detect unmatched same-period settlement deposits relative to legacy deductions before advancing subscription state. Added regression coverage for two deposits with one legacy deduction.

apps/web/src/lib/kiloclaw/credit-billing.ts
jeanduplessis: Valid. Fixed in 054144eff. Effective-balance projection now locks Kilo Pass subscription rows while holding the user lock, and admin cancellation now acquires that same user lock before changing entitlement or resetting balance, establishing user -> entitlement lock order.

apps/web/src/lib/kiloclaw/credit-billing.ts
jeanduplessis: Valid. Fixed in 054144eff. Enrollment now locks the current KiloClaw subscription row with FOR UPDATE and derives eligibility, suspension state, trial transition state, and post-commit resume behavior from that locked row.

jeanduplessis: Valid. Fixed in 054144eff. Intro eligibility, price version, cost, SKU, and paid-history lookup now run after acquiring the user lock. Added concurrent legacy-trial coverage proving only one enrollment receives intro pricing.

apps/web/src/lib/kilo-pass/cancel-and-refund.ts
```diff
@@ -178,6 +178,16 @@ export async function cancelAndRefundKiloPassForUser({
   }
 
   const balanceResetAmountUsd = await db.transaction(async tx => {
+    const freshUserRows = await tx
```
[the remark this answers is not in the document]
jeanduplessis: Valid. Fixed in 0c4692b45 by restoring subscription-first mutation before acquiring the user row lock, matching existing Kilo Pass invoice handling and avoiding the inverse lock order.

apps/web/src/lib/kiloclaw/credit-billing.ts
```diff
@@ -652,42 +651,103 @@ export async function applyStripeFundedKiloClawPeriod(
...
+      const periodSettlementDeposits = matchingPaymentDeduction
+        ? []
+        : await tx
+            .select({ id: credit_transactions.id })
+            .from(credit_transactions)
+            .where(
+              and(
+                eq(credit_transactions.kilo_user_id, userId),
+                eq(credit_transactions.amount_microdollars, amountMicrodollars),
+                eq(credit_transactions.description, `KiloClaw ${plan} settlement`),
```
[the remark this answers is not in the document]
jeanduplessis: Valid. Fixed in 0c4692b45. Reconciliation no longer counts user-wide same-amount deposits. It now uses exact payment deduction, subscription-scoped legacy deduction, and exact subscription period-advance evidence. Added a two-subscription regression test.

apps/web/src/lib/kiloclaw/credit-billing.ts
```diff
@@ -652,42 +651,101 @@ export async function applyStripeFundedKiloClawPeriod(
...
+                eq(kiloclaw_subscription_change_log.subscription_id, targetRow.id),
+                eq(kiloclaw_subscription_change_log.action, 'period_advanced'),
+                eq(kiloclaw_subscription_change_log.reason, 'stripe_invoice_settlement'),
+                sql`${kiloclaw_subscription_change_log.after_state}->>'current_period_start' = ${periodStart}`,
+                sql`${kiloclaw_subscription_change_log.after_state}->>'current_period_end' = ${periodEnd}`
+              )
+            )
+            .limit(1);
+      const shouldReconcilePaymentDeduction =
+        !matchingPaymentDeduction && (!legacyPeriodDeduction || !!settledSamePeriod);
```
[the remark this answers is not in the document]
jeanduplessis: Valid. Fixed in a29d2baec. Reconciliation now requires absence of both a legacy period deduction and exact same-period settlement evidence before adding a payment-scoped deduction. Existing legacy-covered replays and ambiguous two-deposit/one-legacy states fail closed without another balance mutation. Added regression coverage.

apps/web/src/lib/kiloclaw/stripe-funded-settlement.test.ts
```diff
@@ -579,4 +580,409 @@ describe('Stripe-funded KiloClaw settlement', () => {
...
+        plan: 'standard',
+        priceVersion: CURRENT_KILOCLAW_PRICE_VERSION,
+        amountMicrodollars: 55_000_000,
+        periodStart,
+        periodEnd,
+      })
+    ).resolves.toBe(true);
+
+    await expect(readUser(user.id)).resolves.toMatchObject({
+      total_microdollars_acquired: 55_000_000,
```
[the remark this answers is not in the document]
jeanduplessis: This warning conflates two cases. A normal legacy replay with one deposit and one legacy deduction remains balance-neutral and is covered by the adjacent `does not double-deduct a replay already covered by a legacy period deduction` test, which asserts `0`. This fixture intentionally has two indistinguishable deposits but only one legacy period deduction. Because legacy rows do not identify which payment they cover, forcing this assertion to `0` would require guessing and can deduct another subscription’s payment. Safe behavior avoids a second irreversible deduction in this ambiguous historical state; changing the test as suggested would reintroduce cross-subscription or double-charge bugs from prior threads.

PR #4190, opened 2026-06-23
fix(cloud-agent-next): fail over shared sandbox timeouts

eshurakov:
### Why

Shared Cloud Agent sessions derive a stable sandbox identity, so repeated `wrapper.discovery.listProcesses` timeouts can send every new session for the same shared identity back to the same unhealthy sandbox. This change gives future sessions a narrow, durable escape route while keeping each existing session pinned to the assignment it was created with.

### What changed

- Resolve each new shared-session assignment from a KV-backed derivation suffix and persist the selected route key, suffix, and sandbox ID as an immutable session snapshot.
- Count only fresh, typed `listProcesses` timeout observations in durable sandbox-recovery state owned by wrapper supervision. A delivery observation starts an evidence episode; a cleanup observation can complete the two-timeout failover threshold.
- Publish `shared-slot-v1` independently of pending-message terminalization and interruption. Publication deadlines are persisted, retried after 2, 4, and 8 seconds, and honored by the shared Durable Object alarm.
- Bound physical wrapper cleanup to five attempts: immediate, then after 5, 10, 10, and 10 seconds. A fifth failure quarantines the session, blocks delivery, and stops further sandbox probes.
- Repair malformed persisted wrapper leases into the same conservative quarantine instead of treating ownership as empty or dead-lettering the session. Session deletion still makes one best-effort per-session sandbox cleanup call before discarding Durable Object state.
- Bound every shared-route KV read and write to five seconds. Allocation intentionally fails closed when routing state cannot be read or is invalid.

### High-level architecture

```mermaid
sequenceDiagram
  participant API as Cloud Agent API
  participant KV as Shared route KV
  participant DO as Session Durable Object
  participant Queue as Pending-message queue
  participant Runtime as AgentRuntime
  participant Supervisor as Wrapper supervisor
  participant Sandbox as Shared sandbox

  API->>KV: Read route override (5s timeout)
  alt No override
    KV-->>API: No suffix
    API->>API: Select base shared sandbox
    API->>DO: Persist immutable assignment snapshot
  else shared-slot-v1
    KV-->>API: Return failover suffix
    API->>API: Derive deterministic alternate sandbox
    API->>DO: Persist immutable assignment snapshot
  else Read error or timeout
    KV--xAPI: Read fails or exceeds 5s
    break Allocation fails closed
      API--xAPI: Do not register session
    end
  else Unknown suffix
    KV-->>API: Return unsupported suffix
    break Assignment validation fails closed
      API--xAPI: Do not register session
    end
  end

  Note over DO,Sandbox: Alarm invocation: delivery authorization
  DO->>Queue: Drain pending message
  Queue->>Runtime: Attempt delivery
  Runtime->>Sandbox: listProcesses
  Sandbox--xRuntime: Typed discovery timeout
  Runtime->>DO: Persist timeout evidence #1 and stop_needed lease
  Runtime--xQueue: Return typed sandbox-connect failure
  Queue->>DO: Persist message retry deadline (+5s)
  Note over DO,Sandbox: This alarm invocation ends before cleanup runs

  Note over DO,Sandbox: Later alarm: physical cleanup
  DO->>Supervisor: Run cleanup maintenance
  Supervisor->>Sandbox: stopWrappers / listProcesses
  alt Absence confirmed
    Sandbox-->>Supervisor: Wrapper absent
    Supervisor->>DO: Release lease and clear incomplete evidence
  else Cleanup remains unresolved
    alt Fresh listProcesses timeout
      Sandbox--xSupervisor: Typed cleanup timeout
      Supervisor->>DO: Persist timeout evidence #2
    else Other cleanup failure
      Sandbox--xSupervisor: Still present or generic inspection failure
    end

    alt Attempts remain
      Supervisor->>DO: Persist cleanup retry deadline
      Note over DO,Supervisor: Five total attempts: immediate, then after 5/10/10/10s
    else Attempt five fails
      Supervisor->>DO: Quarantine session and stop sandbox probes
    end

    opt Two fresh timeout observations exist
      alt Session is pinned to the base shared assignment
        Supervisor->>KV: Publish shared-slot-v1 (5s per KV operation)
        alt Publication succeeds
          KV-->>Supervisor: Override recorded
          Note over API,KV: Future sessions use the alternate sandbox
        else Publication fails and retries remain
          KV--xSupervisor: Error or timeout
          Supervisor->>DO: Persist next publication deadline (2/4/8s)
        else Fourth publication attempt fails
          KV--xSupervisor: Error or timeout
          Supervisor->>DO: Persist publication exhaustion and schedule no retry
        end
      else Alternate, isolated, or legacy assignment
        Supervisor->>DO: Persist failover as not applicable and do not write KV
      end
    end
  end

  Note over API,Sandbox: The existing session always remains pinned to its original sandbox
```

### Architecture decision

**Decision:** Store one supported derivation suffix in KV per deterministic shared route key, snapshot the resolved assignment in session metadata, and keep timeout evidence plus publication retry state in the session Durable Object's wrapper-supervision domain.

**Rationale:** KV keeps the one-way routing decision shared-identity scoped without adding another Durable Object class. Wrapper supervision owns the real sandbox observations and cleanup lifecycle, while pending messages retain only message delivery state. This separation prevents alarm timing, terminalization, or interruption from corrupting route publication.

**Consequences:** New sessions can move to one deterministic alternate sandbox without moving existing sessions or destroying the shared sandbox. KV is eventually consistent, so a new session may briefly retain the base assignment after publication. KV errors and unknown suffixes fail allocation closed, with a five-second caller timeout. Persistent publication failure is bounded and may leave the route unchanged after the final retry.

## Verification

- [ ] Manual end-to-end validation was not performed because reproducing this backend-only path requires controlled outer wrapper-discovery timeouts, cleanup failures, and KV failures in a provisioned Worker and sandbox environment.

## Visual Changes

N/A

## Reviewer Notes

- Rollout adds the Wrangler-provisioned `SHARED_SANDBOX_OVERRIDES` KV binding; there is no database or Durable Object migration.
- Review the fresh-timeout evidence boundary, persisted publication deadlines, and cleanup quarantine behavior closely. Generic failures, cleanup delays, isolated sessions, legacy sessions, and already-failed-over assignments must not publish another override.
- Cleanup quarantine stops wrapper probes after five attempts but does not destroy the shared sandbox. Explicit or retention deletion performs best-effort cleanup only for that session's sandbox resources.
- Failover is intentionally one-way. Recovery for an unhealthy alternate sandbox remains out of scope.


services/cloud-agent-next/wrangler.jsonc
```diff
@@ -230,6 +230,11 @@
       "localConnectionString": "postgres://postgres:postgres@localhost:5432/postgres",
     },
   ],
+  "kv_namespaces": [
+    {
+      "binding": "SHARED_SANDBOX_OVERRIDES",
```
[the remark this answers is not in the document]
eshurakov: No change needed here. This intentionally uses Wrangler automatic resource provisioning: Cloudflare documents that omitting the KV resource ID causes Wrangler to provision or inherit the binding during deploy. The branch uses Wrangler 4.65.0, newer than the 4.45.0 release that introduced this behavior. See https://developers.cloudflare.com/workers/wrangler/configuration/#automatic-provisioning.

services/cloud-agent-next/src/persistence/session-metadata.ts
```diff
@@ -125,16 +132,55 @@ const MetadataCallbackSchema = z
...
+    if (sandboxId.slice(0, 3) !== route.routeKey.slice(0, 3)) {
+      context.addIssue({
+        code: 'custom',
+        message: 'Shared sandbox route and assignment prefixes must match',
+        path: ['sandboxId'],
+      });
+    }
+    if (
+      (route.suffix === undefined && sandboxId !== route.routeKey) ||
+      (route.suffix !== undefined && sandboxId === route.routeKey)
```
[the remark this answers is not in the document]
eshurakov: Fixed in 7c3c936ae. `recordSessionReady()` now reuses the exact async route-assignment validator before replacing `workspace.sandboxId`, so the value must equal `deriveSharedSandboxId(routeKey, suffix)`. Added a Workers integration regression test that submits a different same-prefix sandbox ID and verifies the readiness write is rejected without changing persisted metadata.

services/cloud-agent-next/wrangler.test.jsonc
```diff
@@ -27,6 +27,7 @@
       "tag": "v5",
     },
   ],
+  "kv_namespaces": [{ "binding": "SHARED_SANDBOX_OVERRIDES" }],
```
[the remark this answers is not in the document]
eshurakov: No change needed here. Wrangler automatic provisioning also creates persistent local resources for `wrangler dev` when the resource ID is omitted, which is the intended test configuration. The Workers integration suite starts with this binding and passes. See https://developers.cloudflare.com/workers/wrangler/configuration/#automatic-provisioning.

services/cloud-agent-next/src/session/session-message-queue.ts
```diff
@@ -326,6 +336,16 @@ export async function flushNextPendingSessionMessage(params: {
     await deletePendingSessionMessageByMessageId(params.storage, message.messageId);
     return { type: 'delivered', remainingCount: Math.max(0, totalCount - 1) };
   }
+  if (message.deliveryDisposition === 'terminalization-pending') {
```
alex-alecu: Can this path honor a persisted publication retry deadline before returning `terminalization-pending` as exhausted? Because it runs before `shouldSkipPendingFlush`, a new-message or maintenance alarm can consume all 2/4/8-second KV retries immediately and abandon failover before KV recovers.

services/cloud-agent-next/src/session/session-message-queue.ts
```diff
@@ -406,7 +416,11 @@ export async function flushNextPendingSessionMessage(params: {
         message,
         message.lastFlushError ?? 'Sandbox connection failed',
         params.now,
-        { policy, code: 'SANDBOX_CONNECT_FAILED' }
+        {
+          policy,
+          code: 'SANDBOX_CONNECT_FAILED',
+          sandboxConnectFailureReason: message.lastSandboxConnectFailureReason,
```
alex-alecu: Should this path preserve the timeout reason and publish a permanent route override even though no second `listProcesses` call occurred? One transient timeout followed by a cleanup delay exhausts the retry and moves every future session for this route.

services/cloud-agent-next/src/session/session-message-queue.ts
```diff
@@ -967,6 +989,58 @@ export function createSessionMessageQueue(
...
+            logger
+              .withFields({
+                sessionId: getSessionIdForLogs(),
+                messageId: flushResult.message.messageId,
+                publishAttempts,
+                retryAt,
+                error: errorMessage,
+              })
+              .warn('Shared sandbox failover publication failed; retry scheduled');
+            return {
```
alex-alecu: Could this retry state be excluded from `interruptPendingQueuedMessages` before retaining it here? If `interruptExecution` runs during this window, interruption sees the already-failed state and throws; the persisted interrupt batch then fails at the start of every alarm, so publication never retries.

services/cloud-agent-next/src/shared-sandbox-route.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export async function resolveSharedSandboxAssignment(
+  store: SharedSandboxOverrideStore,
+  routeKey: SandboxId
+): Promise<SharedSandboxAssignment> {
+  if (!isGeneratedSharedSandboxId(routeKey)) {
+    throw new Error('Shared sandbox route key must be a generated shared sandbox ID');
+  }
+
+  const suffix = await store.get(routeOverrideKey(routeKey));
```
alex-alecu: Is it acceptable for a KV outage or rate limit to block every shared session allocation, including routes that have never failed over? This turns a recovery mechanism into a synchronous availability dependency on the session-start path.

services/cloud-agent-next/src/session/session-message-queue.ts
eshurakov: Fixed in 2936a3beb. Failover publication no longer reuses the pending-message row. Wrapper supervision persists an independent sandbox recovery publication state with nextAttemptAt, and maintenance skips publication until that deadline; alarm scheduling includes the persisted recovery deadline.

services/cloud-agent-next/src/session/session-message-queue.ts
eshurakov: Fixed in 2936a3beb. Pending messages no longer retain or copy sandbox timeout provenance. Durable sandbox recovery counts only fresh typed listProcesses timeout observations: delivery starts an evidence episode, and a later cleanup inspection timeout can complete the threshold. A cleanup delay or generic cleanup failure does not trigger failover.

services/cloud-agent-next/src/session/session-message-queue.ts
eshurakov: Fixed in 2936a3beb. Route publication is no longer represented by a retained terminalization-pending message. The message terminalizes and leaves the pending queue independently, while wrapper supervision owns persisted publication retries, so interruption cannot capture or poison that recovery work.

services/cloud-agent-next/src/shared-sandbox-route.ts
eshurakov: Fail-closed allocation is intentional here: a read error cannot distinguish an absent override from an active failover, and falling back to the base could route a new session into a quarantined sandbox. 2936a3beb bounds every override read and write to five seconds, so KV failure rejects promptly instead of leaving session creation pending indefinitely.

services/cloud-agent-next/src/session/wrapper-runtime-state.ts
```diff
@@ -78,16 +164,20 @@ export type WrapperLeaseEvent =
...
   nextInstanceGeneration: 1,
 });
 
 export async function getWrapperLease(storage: DurableObjectStorage): Promise<WrapperLease> {
-  const parsed = wrapperLeaseSchema.safeParse(await storage.get(WRAPPER_LEASE_KEY));
-  return parsed.success ? parsed.data : emptyWrapperLease();
+  const stored = await storage.get(WRAPPER_LEASE_KEY);
+  if (stored === undefined) return emptyWrapperLease();
+  const parsed = wrapperLeaseSchema.safeParse(stored);
+  if (!parsed.success) throw new Error('Invalid persisted wrapper lease');
```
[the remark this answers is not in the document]
eshurakov: Fixed in 8f5977ca7. An invalid persisted wrapper_lease is now repaired once into a durable cleanup quarantine rather than throwing or falling back to an empty lease. The repair preserves a valid stored generation when possible, blocks delivery, schedules no cleanup probes, remains rollback-safe as stop_needed, and can be removed by the existing quarantined-session deletion path. Added a regression test that verifies the repaired lease is persisted and stable across reads.

services/cloud-agent-next/src/session/wrapper-runtime-state.ts
```diff
@@ -78,16 +164,44 @@ export type WrapperLeaseEvent =
...
+  const quarantined: WrapperLease = {
+    state: 'stop_needed',
+    nextInstanceGeneration: generation.success ? generation.data.nextInstanceGeneration : 1,
+    target: { kind: 'session' },
+    reason: 'observation-failed',
+    requestedAt: now,
+    nextAttemptAt: now + CLEANUP_EXHAUSTED_ROLLBACK_FENCE_MS,
+    attempts: WRAPPER_STOP_MAX_ATTEMPTS,
+    lastError: 'Invalid persisted wrapper lease',
+    exhaustedAt: now,
```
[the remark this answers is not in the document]
eshurakov: Fixed in b7401a4eb. Cleanup quarantine still prevents further wrapper probes, but explicit or retention deletion now makes one best-effort per-session sandbox cleanup call before discarding Durable Object state. That path runs workspace cleanup plus sandbox.deleteSession(sessionId), never destroys the shared sandbox, and a sandbox-side failure does not dead-letter deletion. The integration regression test verifies quarantined deletion skips physical stop retries while invoking sandbox session deletion once.

alex-alecu (approved):

eshurakov:
performed a general smoke test locally

PR #4194, opened 2026-06-23
feat(auth): enforce parent organization SSO

RSO:
## Summary

- Add a canonical effective organization SSO policy resolver that supports direct-parent inheritance and fails closed for ambiguous or misconfigured SSO policy.
- Enforce SSO at sign-in, magic-link creation, invitation creation/acceptance, WorkOS JIT membership, admin membership mutation, and subscription-created membership paths while preserving bot/service-account exemptions.
- Add invitation authentication metadata, credential-pepper rotation on SSO activation/conversion, SSO-domain advisory locking, and regression coverage for inherited child-organization SSO behavior.

## Verification

- Tried magic links for SSO-enabled domains -> always end up in SSO flow
- Tried to invite a member to a child org of an SSO-enabled organization -> always enforces SSO

## Reviewer Notes

- Review focus areas: fail-closed SSO policy resolution, WorkOS authority selection for child organizations, credential revocation/pepper rotation, and membership mutation locking.
- This PR intentionally avoids adding a unique normalized `sso_domain` index because existing legacy duplicate domains must fail closed without creating rollout risk.


apps/web/src/routers/organizations/organization-admin-router.ts
```diff
@@ -573,7 +576,46 @@ export const organizationAdminRouter = createTRPCRouter({
...
+      });
+    }
+
+    const userDomain = getLowerDomainFromEmail(existingUser.google_user_email);
+    if (
+      !existingUser.is_bot &&
+      ssoPolicy.status === 'required' &&
+      userDomain === ssoPolicy.domain
+    ) {
+      const workosProvider = await db.query.user_auth_provider.findFirst({
```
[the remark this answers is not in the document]
RSO: Yeah, when we ever move domains from one org to another we need to be very careful.

apps/web/src/routers/admin-router.ts
```diff
@@ -476,20 +478,21 @@ export const adminRouter = createTRPCRouter({
...
+          throw new TRPCError({ code: 'NOT_FOUND', message: 'User not found' });
+        }
 
-        if (ssoProvider) {
-          throw new TRPCError({
-            code: 'FORBIDDEN',
-            message:
-              'Cannot reset to magic link login for SSO users. The user must authenticate through their organization SSO provider.',
-          });
+        const domain = getLowerDomainFromEmail(user.google_user_email);
```
marius-kilocode, line 486: Why is this google specific? What does `google_user_email` represent?

RSO: It represents tech debt 😄, someone, before I joined Kilo, thought it was a good idea to name the column `google_user_email` instead of `email` because we only had google sign in at that time! We just never changed the column name, but `google_user_email` is just `email`

apps/web/src/lib/user/server.ts
```diff
@@ -713,15 +711,18 @@ const authOptions: NextAuthOptions = {
...
-          // Exclude fake-login from SSO checks - it should always work regardless of domain configuration
-          const redir =
-            accountInfo.provider != 'workos' &&
-            accountInfo.provider != 'fake-login' &&
-            (await doesOrgWithSSODomainExist(domainToCheck));
-
-          if (redir) {
-            return ssoSignInRedirectUrl(domainToCheck);
+          // Fake login is intentionally exempt in supported non-production environments.
+          if (accountInfo.provider !== 'workos' && accountInfo.provider !== 'fake-login') {
```
marius-kilocode, line 715: Is the `server.ts` the component that should contain logic like this?

marius-kilocode (approved):

apps/web/src/lib/user/server.ts
RSO: `server` in this sense indicates where the code should run (server or client), not that it's serving anything. So yeah, I think it's the right place, although the file could use some splitting-up

PR #4206, opened 2026-06-23
feat(usage): add ask usage

alex-alecu:
## Summary
- Added a private Ask Usage flow for admins so they can ask questions about their own Kilo usage.
- The flow starts a normal billable Cloud Agent session only after the admin clicks Start analysis.
- Usage answers come from the native MCP dataset tool with short-lived admin access, strict personal scope, and trusted cards and charts instead of model-written UI.

## Verification
- Sign in as a Kilo organization admin and open `/aks`.
- Click Start analysis and confirm the chat starts without asking for a repository.
- Confirm the answer shows cards or charts from returned usage data, and ask one follow-up in the same session.
- Sign in as a non-admin and confirm `/aks` is not available.

## Visual Changes

https://www.loom.com/share/e71a620693d24c298f43f526987e2d4b

## Reviewer Notes
The Ask Usage session uses the public KiloMan repository only because Cloud Agent sessions still require a repository. The runtime agent is locked to the usage dataset tool, and starting a new analysis is the renewal path when the short-lived MCP token expires.

### Best practices followed:
- Main inspiration https://posthog.com/newsletter/building-ai-agents
- React rendering should be pure and idempotent; components derive JSX from props and React commits only required DOM changes: <https://react.dev/learn/render-and-commit> and <https://react.dev/reference/rules/components-and-hooks-must-be-pure>.
- Vercel's generative UI pattern connects typed tool results to allowlisted React components rather than rendering model-generated code or markup: <https://ai-sdk.dev/docs/ai-sdk-ui/generative-user-interfaces>.
- MCP `registerTool` supports an `outputSchema`; when an output schema is present, the tool callback should return `structuredContent` and the SDK validates output against that schema.
- React warns that untrusted raw HTML through `dangerouslySetInnerHTML` creates XSS risk: <https://react.dev/reference/react-dom/components/common#dangerously-setting-the-inner-html>.
- MCP Apps are appropriate for portable third-party interactive apps only when the host implements sandboxed iframe, CSP, resource, and bridge semantics. They are unnecessary for this first-party static dataset renderer: <https://modelcontextprotocol.io/extensions/apps/overview>


apps/web/src/lib/kilo-datasets/query.ts
```diff
@@ -0,0 +1,887 @@
...
+        group: true,
+        searchable: true,
+      },
+      sourceVersion: {
+        expression: sql`${sql.identifier('source_version')}`,
+        type: 'string',
+        nullable: false,
+        group: true,
+      },
+      gitUrl: {
```
pandemicsyn, line 518: Must-fix: `gitUrl` is now a public dataset field backed by raw `cli_sessions.git_url`/`cli_sessions_v2.git_url`. Remotes can contain embedded credentials (`https://token@host/repo.git`) or private repository metadata, and this MCP path returns them to the model and UI. Remove this field for the MVP, or expose only a sanitized remote value that strips userinfo before it can be filtered, grouped, searched, described, or returned.

apps/web/src/lib/native-mcp/oauth/native-token-service.ts
```diff
@@ -0,0 +1,309 @@
...
+        'Authorization code was already consumed',
+        400
+      );
+    }
+    const accessToken = await mintAccessToken({
+      userId: code.kilo_user_id,
+      clientId: client.client_id,
+      scopes: code.granted_scopes,
+    });
+    const refreshToken = await issueRefreshToken({
```
pandemicsyn, line 191: Must-fix: this always issues a native MCP refresh token, but the native path has no revocable grant/expiry model and `/mcp` bearer verification only rechecks user eligibility. Once a third-party client gets one admin approval it can rotate refresh tokens indefinitely, and clients that did not register `refresh_token` still receive long-lived credential material. Either omit native refresh tokens for the MVP, or add a native grant/revocation/expiry record and issue refresh tokens only when the client declares that grant.

apps/web/src/app/api/mcp-gateway/oauth/token/route.ts
```diff
@@ -47,6 +47,23 @@ async function exchangeToken(request: NextRequest, route?: ScopedConnectRoute) {
...
+    nativeRefreshToken = await services.nativeMcpTokenService.hasRefreshToken(
+      parsed.data.refresh_token
+    );
+  }
+  if (isNativeMcpResource(parsed.data.resource, services.config.appBaseUrl) || nativeRefreshToken) {
+    const result = await services.nativeMcpTokenService.exchangeToken({
+      request: parsed.data,
+      headers: request.headers,
+    });
+    return NextResponse.json(result);
```
pandemicsyn, line 65: Must-fix: native token responses skip `withTokenResponseHeaders`, so bearer access/refresh tokens are returned without `Cache-Control: no-store` and `Pragma: no-cache`. Wrap this branch with the same token response headers as the gateway branch and cover it in the native token route test.

packages/session-ingest-contracts/src/rpc-contract.ts
```diff
@@ -332,6 +332,7 @@ export const sdkToolStateSchema = z.discriminatedUnion('status', [
     status: z.literal('completed'),
     input: sdkMetadataSchema,
     output: z.string(),
+    structuredContent: z.unknown().optional(),
```
pandemicsyn, line 335: nit: preserving arbitrary `structuredContent` as `z.unknown()` bypasses the existing string `output` truncation path. A buggy or malicious tool can send a huge/deep JSON value that is stored, replayed over WebSockets, and rendered without bounds. Add bounded JSON validation/trimming at ingest/trim time, or preserve only allowlisted structured schemas such as the Ask Usage dataset output.

pandemicsyn (changes requested):
Local :robot: found a few items.

PR #4209, opened 2026-06-23
feat(autocomplete): make promoted Inception FIM free

markijbema:
## Summary

- Extend the Inception promotion to FIM requests for `inception/mercury-edit-2`, which supports the KiloCode chat autocomplete FIM fallback from Kilo-Org/kilocode#11592.
- Add `INCEPTION_PROMO_MODEL` so the promo is scoped to the Mercury Edit 2 model instead of all Inception models.
- Preserve market-cost reporting while zeroing billable cost for promoted FIM and Edit requests.
- Add FIM route and usage-accounting coverage for promo enabled/disabled, non-promo models, and BYOK.

## Verification

- [ ] N/A - relying on CI for automated API route and billing coverage.

## Visual Changes

N/A

## Reviewer Notes

The cloud-facing model ID is `inception/mercury-edit-2`; KiloCode's `kilo/inception/mercury-edit-2` is the client/gateway stable model ID and is not accepted directly by this cloud route.

### DIFF
--- a/apps/web/src/lib/constants.ts
+++ b/apps/web/src/lib/constants.ts
@@ -56,6 +56,7 @@ export const APP_URL = resolveAppUrl({
 export const TRIAL_DURATION_DAYS = 14;
 
 export const AUTOCOMPLETE_MODEL = 'codestral-2508';
+export const INCEPTION_PROMO_MODEL = 'inception/mercury-edit-2';
 export const INCEPTION_PROMO_RUNNING = true;
 
 export const ENABLE_DEPLOY_FEATURE = true;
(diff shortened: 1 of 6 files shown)


apps/web/src/lib/constants.ts
```diff
@@ -56,6 +56,7 @@ export const APP_URL = resolveAppUrl({
 export const TRIAL_DURATION_DAYS = 14;
 
 export const AUTOCOMPLETE_MODEL = 'codestral-2508';
+export const INCEPTION_PROMO_MODEL = 'inception/mercury-edit-2';
```
chrarnoldus, line 59: this is the same string for FIM and next edit?

chrarnoldus (approved):

apps/web/src/lib/constants.ts
markijbema: yeah, i use next-edit as well via fim

PR #4211, opened 2026-06-23
feat(api): add usage email filters and members endpoint

evanjacobson:
## Summary

External API consumers previously had to rely on internal user IDs and did not have a documented way to discover organization members for filtering usage analytics. This adds email-oriented usage filtering/display, a public organization members endpoint, and OpenAPI/Swagger coverage so clients can build against the API without depending on private UI assumptions.

Get Members API: /api/v1/organizations/<id>/members

<details>
<summary>Details</summary>

- Adds email-based include/exclude filters and opt-in email display for usage analytics user dimensions while keeping user ID display as the default.
- Adds a public organization members route that returns active and invited members without exposing invite tokens or invite URLs.
- Extends the existing allowlisted OpenAPI generation pattern to include both tRPC usage procedures and the new REST endpoint.

</details>

## Verification

No manual UI verification; this is API/schema behavior with no visual interaction changes.

## Visual Changes

N/A

## Reviewer Notes

Focus review on email-to-user identity scoping for usage analytics, the public member response PII boundary, and the OpenAPI shape for mixed tRPC/REST paths.

### DIFF
--- a/apps/web/src/lib/openapi/trpc-openapi.ts
+++ b/apps/web/src/lib/openapi/trpc-openapi.ts
@@ -1,9 +1,16 @@
 import * as z from 'zod';
 import {
+  publicRestOpenApiRoutes,
   publicTrpcOpenApiProcedures,
+  type RestOpenApiRoute,
   type TrpcOpenApiProcedure,
 } from '@/lib/openapi/trpc-registry';
 import { TrpcErrorResponseSchema, trpcSuccessResponseJsonSchema } from '@/lib/trpc/transport';
+
+const RestErrorResponseSchema = z.object({
+  error: z.string(),
+  message: z.string().optional(),
+});
 
 type JsonSchema = Record<string, unknown>;
 
@@ -36,6 +43,10 @@ function pathForProcedure(procedure: TrpcOpenApiProcedure): `/api/trpc/${string}
 
 function successResponseSchema(data: JsonSchema): JsonSchema {
   return trpcSuccessResponseJsonSchema(data);
+}
+
+function jsonResponseSchema(data: JsonSchema): JsonSchema {
+  return data;
 }
 
 function errorResponse(description: string) {
@@ -82,6 +93,20 @@ function requestShapeForProcedure(procedure: TrpcOpenApiProcedure) {
   };
 }
 
+function pathParametersForRoute(route: RestOpenApiRoute) {
+  if (!route.pathParameters || route.pathParameters.length === 0) return {};
+
+  return {
+    parameters: route.pathParameters.map(parameter => ({
+      name: parameter.name,
+      in: 'path',
+      required: true,
+      description: parameter.description,
+      schema: zodToJsonSchema(parameter.schema),
+    })),
+  };
+}
+
 function operationForProcedure(procedure: TrpcOpenApiProcedure) {
   return {
     operationId: procedure.procedurePath.replaceAll('.', '_'),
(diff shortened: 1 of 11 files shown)


apps/web/src/routers/usage-analytics-router.ts
```diff
@@ -761,27 +1118,35 @@ export const usageAnalyticsRouter = createTRPCRouter({
...
       return {
-        timeseries: rows.map(row => ({
-          datetime: row[0] ?? '',
-          value: toSafeNumber(row[1]),
-          label: input.splitBy ? (row[2] ?? undefined) : undefined,
-        })),
+        timeseries: rows.map(row => {
+          const rawLabel = toStringValue(row[2]);
+          const label = input.splitBy
+            ? dimensionDisplayValue(input.splitBy, filters, userEmailsById, rawLabel)
```
St0rmz1, line 1142: **Moderate, Functional:** The email mapping correctly avoids exposing raw identifiers, but it happens after Snowflake has grouped the timeseries by raw `kilo_user_id`. If one person has usage under both their canonical ID and a legacy `oauth/...` ID, the response contains two points with the same timestamp and email instead of one combined point. Clients may render duplicate points or overwrite one of them, which makes the visible total incomplete. Could we join the scoped email map before aggregation and group by the mapped email and time bucket, similar to the breakdown query? A test with canonical and OAuth rows for the same email and timestamp would cover this case.

apps/web/src/routers/usage-analytics-router.ts
```diff
@@ -937,15 +1361,18 @@ export const usageAnalyticsRouter = createTRPCRouter({
...
+      const userEmailsById = userEmailsForDisplay(filters, scopedUserEmailMaps);
+
       return {
         rows: rows.map(row => {
           const dimensions: Record<string, string> = {};
           for (const d of requestedDims) {
             const raw = row[dimIndexMap[d]];
-            dimensions[d] = typeof raw === 'string' ? raw : '';
+            const value = toStringValue(raw);
+            dimensions[d] = dimensionDisplayValue(d, filters, userEmailsById, value);
```
St0rmz1, line 1372: **Moderate, Functional:** Mapping the returned user dimension to email avoids exposing raw identifiers, but the table query still groups and applies `LIMIT` using raw `kilo_user_id` values. A person with usage under both a canonical ID and a legacy `oauth/...` ID can therefore produce duplicate rows with the same visible dimensions. Those duplicate identity rows can also consume the limit and exclude rows for other users. Could we join the scoped email map in this query, group by the mapped email with the other requested dimensions, and apply the limit after that aggregation? Tests covering identity aggregation and a constrained limit would help preserve the expected table semantics.

apps/web/src/lib/openapi/trpc-openapi.ts
```diff
@@ -107,6 +132,59 @@ function operationForProcedure(procedure: TrpcOpenApiProcedure) {
...
+      },
+      '403': {
+        ...errorResponse('Access denied'),
+        content: {
+          'application/json': {
+            schema: zodToJsonSchema(RestErrorResponseSchema),
+          },
+        },
+      },
+      '500': {
```
St0rmz1, line 176: **Low, Documentation:** The shared REST operation documents the common error responses, but this members endpoint can also return HTTP 404. `organizations.withMembers` raises `NOT_FOUND` when the organization does not exist, and `handleTRPCRequest` preserves that status. Without a documented 404, generated clients and API consumers may treat a valid endpoint response as unexpected. Could the REST route metadata support route specific responses and add a 404 using `RestErrorResponseSchema` for this endpoint? An OpenAPI assertion for the members operation would keep the contract in sync.

St0rmz1 (changes requested):

RSO:
Why did you not add this as a tRPC endpoint, like the rest? I'm not sure it makes sense to have some endpoints in the API docs as tRPC and then some others as REST.

PR #4219, opened 2026-06-23
feat(enterprise) send and manage org recommendations email

St0rmz1:
## Summary

Adds a weekly recommendations email for Enterprise organizations and the controls to manage it. Organization owners can turn the email on or off from a new Email preferences section on the organization page. When it is on, a weekly cron emails the organization's owners a short summary of how many features are set up and the top open recommendations, linking back to the Feature adoption view.

- New `recommendations_digest_enabled` flag in the organization settings JSONB (`schema-types.ts`). No new table is added.
- New owner-only, Enterprise-gated `updateRecommendationsDigest` tRPC mutation that toggles the flag, plus the `useUpdateRecommendationsDigest` hook. The toggle writes a single JSONB key with `jsonb_set` (or removes it) instead of rewriting the whole settings object.
- New Email preferences card on the organization dashboard (owner-only) with two rows: the existing low balance alerts (which still opens its configuration modal) and the new weekly recommendations email (an on/off toggle).
- Send path: a `recommendationsDigest.html` template and a `sendRecommendationsDigestEmail` helper in `email.ts`, a `recommendations-digest.ts` library that builds each org's payload from `getOrganizationRecommendations`, resolves the owner recipients, and sends, and an `enterprise-recommendations-digest` cron registered in `vercel.json` to run Mondays at 09:00 UTC.
- Orgs with no open recommendations are skipped (no "all good" email). Before each send, a per recipient, per week marker is claimed in `transactional_email_log` so cron overlap or retries do not produce duplicates. The claim is released if the send fails so a later run can retry.
- The dispatch query filters opted-in orgs in SQL (`settings ->> 'recommendations_digest_enabled' = 'true'`) rather than loading every Enterprise org.
- Admin preview: a fixture for the new template in `email-testing-router.ts` so it renders on the admin email testing page, plus a row in `emails/AGENTS.md`.
- Tests for the toggle mutation, the digest builder, owner recipient resolution, and the duplicate-claim and retry behavior.

## Verification

- [x] As an owner of an Enterprise org, toggle the weekly recommendations email on and off and confirm the setting persists.
- [x] Preview the "Kilo: Your weekly recommendations" template on the admin email testing page (`/admin/email-testing`).

## Visual Changes

Two surfaces changed: the new Email preferences card on the organization page (owner-only), and the new recommendations email template. Screenshots to add:
Screenshot 2026-06-24 at 5 51 39 AM
Screenshot 2026-06-24 at 5 52 38 AM

## Reviewer Notes

- Recipients are the organization's owners, resolved at send time. There is no stored recipient list and no per recipient unsubscribe; owners turn the email off with the toggle.
- Access: reads and the toggle are Enterprise only, and the toggle is owner only.
- Idempotency reuses the existing `transactional_email_log` (no new table). The key is `organizationId:userId:weekMonday`, so it does not contain recipient email addresses. It is best effort: a send that succeeds at the provider but fails to record the marker would not be retried that week.


apps/web/vercel.json
```diff
@@ -75,6 +75,10 @@
     {
       "path": "/api/cron/cleanup-review-memory",
       "schedule": "0 */12 * * *"
+    },
+    {
+      "path": "/api/cron/enterprise-recommendations-digest",
+      "schedule": "0 9 * * 1"
```
RSO, line 81: Just a thought: If we run this on Sundays at this time, we may be able to schedule the emails in Mailgun to be delivered at 9 am on Mondays in the user's local timezone.

apps/web/src/app/api/cron/enterprise-recommendations-digest/route.ts
```diff
@@ -0,0 +1,35 @@
...
+import { sentryLogger } from '@/lib/utils.server';
+
+if (!CRON_SECRET) {
+  throw new Error('CRON_SECRET is not configured in environment variables');
+}
+
+export async function GET(request: Request) {
+  const authHeader = request.headers.get('authorization');
+  const expectedAuth = `Bearer ${CRON_SECRET}`;
+  if (authHeader !== expectedAuth) {
```
RSO: Shouldn't this be a `crypto.timingSafeEqual`?

RSO (approved):
A screenshot of the email would be appreciated in the PR description 🙏

apps/web/vercel.json
St0rmz1: I like this, but would prefer to defer it

- We don't store a per-user timezone. I checked the schema: the only timezone fields are cron_timezone for KiloClaw scheduled actions, nothing on the user profile. Local time delivery needs per recipient tz, which is net new capture.
- sendViaMailgun has no delivery time plumbing today, so the message builder would need a new optional param

PR #4253, opened 2026-06-25
feat(cost-insights): Introduce Cost Insights feature

jeanduplessis:
## Summary

Adds owner-scoped Cost Insights for personal users and organizations so authorized users can inspect Credit spend, configure Spend Alerts, review alert episodes, and act on Cost Suggestions.

- Adds Cost Insights dashboard, activity, Ask Kilo, and settings routes with responsive personal and organization UI.
- Adds normalized hourly spend rollups, coverage tracking, repair/backfill jobs, and atomic updates across Credit-spend paths.
- Adds anomaly and rolling 24-hour threshold evaluation, active episode state, notification delivery, retention cleanup, and cron entry points.
- Adds Coding Plan and Kilo Pass Cost Suggestions based on observed spend, with dismissal history and independent enablement.
- Adds owner-scoped authorization, current-recipient revalidation, server-side activity pagination, captured contributor evidence, and numeric sidebar review badges.
- Adds generated database migrations, seed data, Storybook coverage, and regression tests for rollups, evaluation, presentation, routing, and billing integrations.

## Verification

- Not manually verified in browser in this session.

## Visual Changes

N/A - screenshots were not captured as part of this update.

## Reviewer Notes

- Review spend-writer transaction boundaries carefully: Credit spend and its hourly rollup must commit atomically.
- Review anomaly coverage and repair behavior around incomplete historical windows and current partial hours.
- Review organization authorization and notification recipient revalidation for owners and billing managers.
- Cost Suggestions remain advisory and independent from Spend Alert enablement; they do not alter billing or spend behavior.
- Activity history uses server-side filtering and pagination, while alert events resolve current actor labels from captured driver IDs.

### Run this branch locally

Prerequisites: Docker Desktop must be running, `tmux` must be installed, and `.env.local` must contain a local `POSTGRES_URL`. Use Node from `.nvmrc` and pnpm from `package.json`.

1. Fetch and check out the branch. For a first checkout, use the tracking command; if the branch already exists locally, use `git switch jdp/cost-insights && git pull --ff-only` instead.

```bash
git fetch origin jdp/cost-insights
git switch --track origin/jdp/cost-insights
nvm install
nvm use
corepack enable
pnpm install
```

2. Configure local environment variables if this clone is not already configured. Team members with Vercel access can use the first path; otherwise use the interactive local setup.

```bash
# Team-member/Vercel path
vercel link --project kilocode-app
vercel env pull

# Or local-only setup
pnpm dev:setup-env

# Generate service-specific local env files after either path
pnpm dev:env
```

3. Start PostgreSQL and apply this branch's migrations. Re-run migration after every pull that adds migration files.

```bash
docker compose -f dev/docker-compose.yml up -d --wait postgres
pnpm drizzle migrate
```

4. Seed the Cost Insights fixture before opening the UI.

```bash
pnpm dev:seed cost-insights:spend-evidence
```

5. Start the app.

```bash
pnpm dev:start
```

Default web URL is `http://localhost:3000`. If another Kilo dev stack is running, generate env and start this worktree with a stable automatic offset, then read the actual Next.js port from status output:

```bash
KILO_PORT_OFFSET=auto pnpm dev:env
KILO_PORT_OFFSET=auto pnpm dev:start
KILO_PORT_OFFSET=auto pnpm dev:status --json
```

### What the seed creates

`pnpm dev:seed cost-insights:spend-evidence` creates a deterministic, owner-scoped fixture containing:

- 90 days of personal and organization Variable Credit spend.
- Monthly KiloClaw Scheduled Credit spend.
- Current-hour anomaly spikes and rolling 24-hour threshold crossings.
- Active Spend Alert banners, Coding Plan and Kilo Pass Cost Suggestions, notification rows, and activity history.
- An organization named `[seed:cost-insights] Northstar Labs` with owner, billing manager, and member-attributed spend drivers.
- Matching hourly rollups and complete coverage metadata, so all `24h`, `7d`, `30d`, and `90d` views are populated.

The command prints `personalLoginPath`, `organizationLoginPath`, owner IDs, organization ID, record counts, and the database target. Prefix a printed login path with `http://localhost:<nextjs-port>`.

Seed identities:

Cost Insights v1 is admin-only during initial rollout. The owner and billing-manager fixture users below are seeded with `is_admin = true`, so they can open Cost Insights and exercise the owner/billing-manager management paths. The contributor fixture is not an admin and is not authorized to open organization Cost Insights.

| Purpose | Email |
|---|---|
| Personal Spend owner and organization owner, Kilo admin | `cost-insights-owner@example.com` |
| Organization billing manager, Kilo admin | `cost-insights-billing-manager@example.com` |
| Organization spend contributor fixture, non-admin | `cost-insights-member@example.com` |

Organization ID: `4f2fc143-4b30-4c8a-878b-df89c89c6790`.

With the default port, use these direct fake-login URLs:

```text
Personal dashboard:
http://localhost:3000/users/sign_in?fakeUser=cost-insights-owner%40example.com&callbackPath=%2Fcost-insights

Organization dashboard as owner:
http://localhost:3000/users/sign_in?fakeUser=cost-insights-owner%40example.com&callbackPath=%2Forganizations%2F4f2fc143-4b30-4c8a-878b-df89c89c6790%2Fcost-insights

Organization dashboard as billing manager:
http://localhost:3000/users/sign_in?fakeUser=cost-insights-billing-manager%40example.com&callbackPath=%2Forganizations%2F4f2fc143-4b30-4c8a-878b-df89c89c6790%2Fcost-insights

Non-admin contributor access check:
http://localhost:3000/users/sign_in?fakeUser=cost-insights-member%40example.com&callbackPath=%2Forganizations%2F4f2fc143-4b30-4c8a-878b-df89c89c6790%2Fcost-insights
```

Wait for fake-login account creation to finish before evaluating the page. The admin owner and admin billing manager can change settings and review items. The non-admin contributor exists to exercise contributor attribution and should be blocked from organization Cost Insights.

### Suggested local checks

- Confirm personal and organization dashboards show anomaly and threshold banners, suggestions, spend metrics, and top drivers.
- Switch spend evidence among `24h`, `7d`, `30d`, and `90d`; zero-spend buckets should have zero-height bars.
- Open Activity, switch each filter, and paginate through retained events.
- Use Manage threshold from a threshold banner to open the matching settings section; adjust or disable the threshold there and confirm the banner clears, settings update, and activity records the change.
- Dismiss a suggestion and acknowledge alerts; confirm sidebar review count decreases.
- Open organization settings as the admin owner and admin billing manager, then repeat as the non-admin contributor to confirm organization Cost Insights is blocked.

### Seed safety and reset behavior

- The seed refuses `USE_PRODUCTION_DB=true` and rejects any `POSTGRES_URL` whose host is not loopback (`localhost`, `127.0.0.1`, or `::1`).
- Rerunning the seed is the normal reset path. It replaces only this fixture's stable users, organization, spend evidence, rollups, events, suggestions, and notification rows.
- Seeded users use placeholder Stripe customer IDs. Use fake login for Cost Insights pages; do not open Stripe-backed billing pages with these users.
- To restore mutated Cost Insights state after testing, rerun `pnpm dev:seed cost-insights:spend-evidence`.
- For a completely fresh local database, stop the app and run the destructive reset below. This removes all local app data, not only Cost Insights fixtures.

```bash
pnpm dev:stop
pnpm dev:db:reset
pnpm drizzle migrate
pnpm dev:seed cost-insights:spend-evidence
pnpm dev:start
```


apps/web/src/lib/cost-insights/repository.ts
```diff
@@ -0,0 +1,841 @@
...
+    .limit(1);
+  return organization?.name ?? 'Organization';
+}
+
+export async function deleteExpiredCostInsightEvents(
+  database: CostInsightDatabase,
+  retentionCutoff: string
+): Promise<number> {
+  const rows = await database
+    .delete(cost_insight_events)
```
St0rmz1, line 1097: **Moderate, Functional:** This unconditional retention delete can remove an event that still represents an active, unacknowledged alert. The foreign key is set to null while threshold state can remain active, so the banner disappears, acknowledgment is no longer possible, and evaluation will not create a replacement until the threshold recovers and crosses again. Please preserve the active episode's presentation and identity independently from retained event history, and add a retention test for an unreviewed episode older than 90 days.

apps/web/src/lib/cost-insights/notifications.ts
```diff
@@ -0,0 +1,239 @@
...
+
+async function claimPendingDeliveries(
+  database: CostInsightDatabase,
+  limit: number
+): Promise<CostInsightClaimedDeliveryRow[]> {
+  const result = await database.execute<CostInsightClaimedDeliveryRow>(sql`
+    WITH claimed AS (
+      SELECT delivery.id
+      FROM cost_insight_notification_deliveries delivery
+      WHERE delivery.status IN ('pending', 'failed')
```
St0rmz1: **Moderate, Functional:** Claims only select `pending` and `failed` rows before changing them to `sending`. If the process stops after the claim commits but before recording an outcome, that delivery remains in `sending` permanently because `claimed_at` is never used for recovery. Please make stale `sending` rows eligible for reclamation after a bounded lease, subject to the attempt limit, and add coverage for interruption after claim.

apps/web/src/lib/cost-insights/evaluation.ts
```diff
@@ -0,0 +1,573 @@
...
+  options: { asOf?: string } = {}
+): Promise<CostInsightEvaluationSummary> {
+  const asOf = options.asOf ?? new Date().toISOString();
+  const currentHourStart = floorUtcHour(new Date(asOf));
+  const topDriverEnd = addHours(currentHourStart, 1);
+  const suggestionWindowEnd = topDriverEnd;
+  const suggestionWindowStart = addDays(suggestionWindowEnd, -7);
+
+  const config = await getCostInsightOwnerConfig(database, owner);
+  const currentHourSpend = await getOwnerCurrentHourSpend(database, owner);
```
St0rmz1: **Moderate, Functional:** `currentHourStart` is derived from the captured `asOf`, but `getOwnerCurrentHourSpend` independently uses database `CURRENT_TIMESTAMP`. An hour rollover between these operations can combine the next hour's amount with the previous hour's drivers, snapshot interval, baseline, and deduplication key. It also makes historical `asOf` evaluations read live spend. Please pass the authoritative evaluation interval into the spend query and use the same boundary for all anomaly evidence.

apps/web/src/routers/cost-insights-router.ts
```diff
@@ -0,0 +1,454 @@
...
+    spend7DayThresholdMicrodollars = parseSpendThresholdUsd(params.input.spend7DayThresholdUsd);
+    spend30DayThresholdMicrodollars = parseSpendThresholdUsd(params.input.spend30DayThresholdUsd);
+  } catch (error) {
+    throw new TRPCError({
+      code: 'BAD_REQUEST',
+      message: error instanceof Error ? error.message : 'Invalid spend threshold.',
+    });
+  }
+
+  const { previous, current } = await updateCostInsightOwnerConfig(db, params.owner, {
```
St0rmz1: **Moderate, Functional:** This configuration update commits before active state is cleared and the required configuration or disablement event is created. A later failure can return an error while leaving settings changed, active episodes inconsistent with those settings, or required history missing. Please perform the configuration write, episode state changes, and event insertion in one transaction. External evaluation work can run after commit through a recoverable path.

apps/web/src/components/cost-insights/ask-kilo/CostInsightsAskKiloView.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+import { useEffect, useRef, useState, type FormEvent } from 'react';
+import { BarChart3, ChevronDown, ChevronUp, Send } from 'lucide-react';
+
+import { Button } from '@/components/ui/button';
+import { Input } from '@/components/ui/input';
+import { Label } from '@/components/ui/label';
+import { useCostInsightsTracking } from '../useCostInsightsTracking';
+
+const askKiloChartData = [
```
St0rmz1: **Moderate, Functional:** This fixed dataset is used by the production personal and organization routes, and every submitted question receives the same fixed dates and dollar analysis. The UI presents these values as the current owner's cost trend without loading owner spend or identifying them as sample data, which can mislead billing decisions. Please either connect this surface to authorized owner data or render it as a clearly disabled preview that cannot be mistaken for live financial information.

apps/web/src/lib/cost-insights/repository.ts
```diff
@@ -0,0 +1,841 @@
...
+            );
+  const [acknowledged] = await database
+    .update(cost_insight_owner_states)
+    .set(reviewValues)
+    .where(and(eq(cost_insight_owner_states.id, state.id), activeEpisode))
+    .returning({ id: cost_insight_owner_states.id });
+
+  if (!acknowledged) return false;
+
+  await createCostInsightEvent(database, {
```
St0rmz1: **Moderate, Functional:** The episode is marked reviewed before the required `alert_reviewed` event is inserted. If event insertion fails, the mutation rejects but the alert remains reviewed, and a retry cannot create the missing history because the unreviewed predicate no longer matches. Please wrap the conditional state update and event insertion in one transaction, with a failure test confirming both changes roll back together.

apps/web/src/lib/cost-insights/repository.ts
```diff
@@ -0,0 +1,841 @@
...
+      and(
+        eq(cost_insight_active_suggestions.id, params.suggestionId),
+        costInsightOwnerWhere(params.owner, cost_insight_active_suggestions),
+        isNull(cost_insight_active_suggestions.dismissed_at)
+      )
+    )
+    .returning();
+
+  if (!suggestion) return null;
+  await createCostInsightEvent(database, {
```
St0rmz1: **Moderate, Functional:** The suggestion is persisted as dismissed before the required `suggestion_dismissed` event is inserted. If event insertion fails, the suggestion remains hidden and a retry cannot repair the missing history because `dismissed_at IS NULL` no longer matches. Please perform the dismissal update and event insertion in one transaction, with a failure test confirming both changes roll back together.

apps/web/src/components/cost-insights/overview/SpendEvidenceCard.tsx
```diff
@@ -0,0 +1,207 @@
...
+}: {
+  data: CostInsightsDashboardData;
+  range: SpendRange;
+}) {
+  const viewerTimeZone = useViewerTimeZone();
+  const rawEvidence = range === data.range ? data.evidence : data.evidenceByRange[range];
+  const evidence = rawEvidence.map(point => ({
+    ...point,
+    label:
+      point.periodStart && (range === '1h' || range === '24h' || range === '7d')
```
St0rmz1: **Moderate, Functional:** Viewer time zone formatting is limited to `1h`, `24h`, and `7d`, so the `30d` and `90d` ranges retain labels generated in UTC. Those long range points also omit `periodStart`, preventing the client from localizing them. This conflicts with the Cost Insights requirement that UI timestamps use the viewer's current time zone. Please retain an appropriate period timestamp for these points and format their labels for the viewer, with regression coverage around midnight offsets.

apps/web/src/lib/cost-insights/jobs.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export async function runCostInsightHourlySweep(
+  database: CostInsightRootDatabase
+): Promise<CostInsightHourlySweepSummary> {
+  const owners = await listEnabledCostInsightOwners(database);
+  const failedOwners: CostInsightHourlySweepSummary['failedOwners'] = [];
+
+  for (const owner of owners) {
+    try {
+      await evaluateCostInsightsForOwner(database, owner);
```
St0rmz1: **Moderate, Functional:** The hourly sweep evaluates each owner only at the current time. Anomaly evaluation then inspects only the hour containing that time, so if post-spend evaluation failed for a burst near the end of the previous hour, the completed anomalous hour is never reconsidered. This prevents the sweep from catching that missed evaluation. Please have the sweep explicitly evaluate the just-completed hour for missed anomaly episodes while preserving current-time rolling threshold evaluation and episode deduplication.

apps/web/src/scripts/db/exa-usage-log-indexes.ts
```diff
@@ -0,0 +1,169 @@
...
+    for (const index of plan.indexes) {
+      console.log(
+        JSON.stringify({
+          mode: 'execute-index',
+          schemaName: plan.schema_name,
+          partitionName: plan.partition_name,
+          indexName: index.name,
+        })
+      );
+      await fromDb.execute(sql.raw(index.statement));
```
St0rmz1: **Moderate, Performance:** An interrupted `CREATE INDEX CONCURRENTLY` can leave an index relation with `indisvalid = false`. Retrying these `IF NOT EXISTS` statements sees the existing name and skips rebuilding it, so the operator can report success while the partition still lacks a usable query index. Please inspect `pg_index.indisvalid` for each planned index and drop or rebuild invalid indexes before treating the operation as complete.

dev/seed/cost-insights/spend-evidence.ts
```diff
@@ -0,0 +1,1779 @@
...
+      await tx
+        .delete(cost_insight_owner_hour_totals)
+        .where(
+          or(
+            eq(cost_insight_owner_hour_totals.owned_by_user_id, PERSONAL_OWNER_ID),
+            eq(cost_insight_owner_hour_totals.owned_by_organization_id, ORGANIZATION_ID)
+          )
+        );
+      await tx
+        .delete(cost_insight_rollup_coverage)
```
St0rmz1: **Moderate, Functional:** This owner-scoped fixture deletes the global rollup coverage row, then later recreates it from the selected fixture mode even though only the fixture owners were populated and verified. On an existing local database this can erase valid global coverage or claim completeness for unrelated owners with missing rollups, causing missing data to be treated as trustworthy zero spend. Please avoid changing global coverage in the normal fixture path, or require an explicit destructive mode that verifies all canonical data before restoring it.

dev/seed/cost-insights/spend-evidence.ts
```diff
@@ -0,0 +1,1779 @@
...
+  console.log('  healthy           Matching rollups and complete 90-day coverage.');
+  console.log('  repairable-drift  Complete coverage with missing, late, and stale rollups.');
+  console.log('  unknown-taxonomy  Healthy data plus one dry-run-only taxonomy diagnostic.');
+  console.log('  degraded-late     Late data plus unresolved degraded interval for inspection.');
+  console.log('');
+  console.log('Default: bootstrap. Reruns replace only this fixture data and v1 coverage state.');
+}
+
+function parseRollupMode(args: string[]): RollupMode {
+  if (args.length === 0) return 'bootstrap';
```
St0rmz1: **Low, Functional:** The documented no-argument setup command selects `bootstrap`, which removes completed historical fixture rollups and limits coverage to the current hour. That does not match the PR instructions stating that the same command creates complete rollups for all evidence ranges. Please either make `healthy` the default for the documented UI fixture or update the instructions to pass `--rollup-mode healthy` or run the required backfill.

apps/web/src/components/cost-insights/overview/CostInsightsDashboardView.tsx
```diff
@@ -0,0 +1,227 @@
...
+          onExploreThisHour={showThisHour}
+        />
+      ))}
+      {data.suggestions.map(suggestion => (
+        <SuggestionCard
+          key={suggestion.id}
+          suggestion={suggestion}
+          canManage={canManage}
+          onCta={() => onSuggestionCta?.(suggestion)}
+          onDismiss={() => onSuggestionDismiss?.(suggestion.id)}
```
St0rmz1, line 118: **Low, Functional:** The dismissal mutation's pending state is not passed through here, so this action remains enabled while a request is in flight. Repeated activation can issue concurrent requests and produce duplicate success feedback and invalidations. Please pass the pending suggestion ID into the view, disable the matching action until settlement, and expose `aria-busy`.

apps/web/src/lib/cost-insights/presenter.ts
```diff
@@ -0,0 +1,764 @@
...
+      '24h': mapDrivers(params.owner, topDriversByRange['24h'], actorLabels),
+      '7d': mapDrivers(params.owner, topDriversByRange['7d'], actorLabels),
+      '30d': mapDrivers(params.owner, topDriversByRange['30d'], actorLabels),
+      '90d': mapDrivers(params.owner, topDriversByRange['90d'], actorLabels),
+    },
+    alerts,
+    suggestions: formatActiveCostInsightSuggestions(activeSuggestions),
+    lastEvaluatedAt: dashboardState.state?.lastEvaluatedAt ?? null,
+    baselineMode: anomalyPolicy.mode,
+    eventPreview,
```
St0rmz1, line 849: **Low, Functional:** The live dashboard data never supplies `memberLimitsHref`, although `TopDriversCard` renders the member daily limits action when this field exists and Storybook fixtures include it. As a result, the organization dashboard never exposes this action. Please populate the organization-specific members destination for eligible organization viewers and add presenter coverage.

apps/web/src/app/api/cron/cost-insights-hourly/route.ts
```diff
@@ -0,0 +1,36 @@
...
+        (authHeader ? 'Invalid authorization header' : 'Missing authorization header')
+    );
+    return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
+  }
+
+  const summary = await runCostInsightHourlySweep(db);
+
+  return NextResponse.json(
+    {
+      success: true,
```
St0rmz1: **Low, Functional:** This returns HTTP success and `success: true` even when `failedOwners` is nonempty or notification delivery reports failures. The details remain in the response summary, but cron monitoring based on status or the success field will treat a partial or total failure as healthy. Please derive the success signal from the summary, or explicitly capture partial failures if retaining a 200 response is intentional.

apps/web/src/routers/cost-insights-router.ts
```diff
@@ -0,0 +1,454 @@
...
+  spendAlertsEnabled: z.boolean(),
+  anomalyAlertsEnabled: z.boolean(),
+  costSuggestionsEnabled: z.boolean(),
+  spendThresholdUsd: z.string().nullable(),
+  spend7DayThresholdUsd: z.string().nullable(),
+  spend30DayThresholdUsd: z.string().nullable(),
+});
+
+const AcknowledgeCostInsightAlertSchema = z.object({
+  alertKind: z.enum(['anomaly', 'threshold', 'threshold_7d', 'threshold_30d']),
```
pandemicsyn, line 43: Must-fix: Acknowledgment identifies only the alert kind. If the page is stale and a newer same-kind episode becomes active before this mutation runs, it acknowledges the newer episode rather than the one the user reviewed. Include the displayed event ID in the mutation and require it to match the active event ID.

apps/web/src/lib/exa-usage.ts
```diff
@@ -87,54 +90,91 @@ export async function recordExaUsage(params: {
...
+      costMicrodollars,
+    });
+
+    await captureCostInsightSpend(tx, {
+      owner: organizationId
+        ? { type: 'organization', id: organizationId }
+        : { type: 'user', id: userId },
+      actorUserId: userId,
+      occurredAt,
+      amountMicrodollars: costMicrodollars,
```
pandemicsyn, line 133: Must-fix: This casts an untrusted upstream response instead of validating it. Non-numeric, non-finite, or oversized values can reach the asynchronous billing path and fail after usage was served. Parse costDollars with a runtime schema and require the converted microdollars to be a positive safe integer.

apps/web/src/lib/cost-insights/presenter.ts
```diff
@@ -0,0 +1,764 @@
...
+  const elapsedDays = (new Date(end).getTime() - new Date(start).getTime()) / MS_PER_DAY;
+  return Math.max(1, Math.round(elapsedDays));
+}
+
+function hourlyEvidence(points: OwnerHourlySpend[], range: SpendRange): SpendEvidencePoint[] {
+  if (range === '1h' || range === '24h' || range === '7d') {
+    return points.map(point => ({
+      label: formatHourLabel(point.hourStart, range),
+      periodStart: point.hourStart,
+      variableUsd: microdollarsToUsd(point.variableMicrodollars ?? 0),
```
pandemicsyn: Must-fix: Uncovered rollup hours are represented as null, but this converts them to zero spend. Coverage gaps therefore appear as trustworthy $0 bars and lower aggregated 30d/90d totals. Preserve an unavailable state in the presentation model rather than coalescing incomplete evidence to zero.

apps/web/src/lib/cost-insights/presenter.ts
```diff
@@ -0,0 +1,764 @@
...
+  owner: CostInsightSpendOwner,
+  events: ListedCostInsightEvent[],
+  actorLabels: ReadonlyMap<string, string> = new Map()
+): CostInsightEvent[] {
+  return events.map(event => ({
+    id: event.id,
+    type: event.eventType === 'alert_reviewed' ? 'reviewed' : event.eventType,
+    title: event.title,
+    description: event.description,
+    occurredAt: event.occurredAt,
```
pandemicsyn: Must-fix: occurredAt is forwarded directly from a mode: 'string' Postgres timestamp, and lastEvaluatedAt is forwarded similarly. Production rows may use Postgres timestamp text that strict clients reject. Normalize all DB-backed timestamps with new Date(value).toISOString() before returning them through tRPC.

apps/web/src/lib/cost-insights/repository.ts
```diff
@@ -0,0 +1,841 @@
...
+
+  const activeState = state[0];
+  const alertCount =
+    (activeState?.activeAnomalyEventId && !activeState.activeAnomalyReviewedAt ? 1 : 0) +
+    (activeState?.activeThresholdEventId && !activeState.thresholdReviewedAt ? 1 : 0) +
+    (activeState?.active7DayThresholdEventId && !activeState.threshold7DayReviewedAt ? 1 : 0) +
+    (activeState?.active30DayThresholdEventId && !activeState.threshold30DayReviewedAt ? 1 : 0);
+  const suggestionCount =
+    (config?.cost_suggestions_enabled ?? true) ? (suggestions[0]?.value ?? 0) : 0;
+  return alertCount + suggestionCount;
```
pandemicsyn: Must-fix: Active Cost Suggestions contribute to the sidebar attention badge here. The Cost Insights contract reserves attention for unreviewed Spend Alerts and says suggestions must not be presented as alerts or warnings. Return only the unreviewed alert count.

apps/web/src/components/cost-insights/settings/CostInsightsSettingsView.tsx
```diff
@@ -0,0 +1,287 @@
...
+                  id="spend-alerts-enabled"
+                  className="relative before:absolute before:inset-x-0 before:-inset-y-2.5"
+                  checked={data.enabled}
+                  disabled={disabled}
+                  onCheckedChange={enabled => onChange?.({ enabled })}
+                />
+              </div>
+            </section>
+
+            {data.enabled && (
```
pandemicsyn: Must-fix: Turning Spend Alerts off hides the saved anomaly and threshold configuration. Users cannot inspect or edit settings that will be reused on re-enable, despite disabled-state changes being save-only. Keep the sub-options visible while alerts are off.

apps/web/src/lib/cost-insights/policy.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  const [wholePart, centsPart = ''] = trimmed.split('.');
+  const dollars = Number.parseInt(wholePart, 10);
+  const cents = Number.parseInt(centsPart.padEnd(2, '0') || '0', 10);
+  const totalCents = dollars * 100 + cents;
+  if (!Number.isSafeInteger(totalCents) || totalCents <= 0) {
+    throw new Error('Spend threshold must be greater than $0.00.');
+  }
+  return usdToMicrodollarsFromCents(totalCents);
```
pandemicsyn: Must-fix: The parser verifies totalCents before multiplying by 10,000, so an accepted value can become an unsafe microdollar integer and fail as an internal database error. Validate the final microdollar value and return a user-facing validation error.

packages/db/src/schema.test.ts
```diff
@@ -533,6 +533,25 @@ describe('database schema', () => {
...
+        'anomaly_alert',
+        'threshold_crossed',
+        'alert_reviewed',
+        'suggestion_created',
+        'suggestion_dismissed',
+        'disabled',
+      ],
+      CostInsightAlertKind: ['anomaly', 'threshold', 'threshold_7d', 'threshold_30d'],
+      CostInsightSuggestionKind: ['coding_plan', 'kilo_pass'],
+      CostInsightNotificationStatus: ['pending', 'sending', 'sent', 'failed'],
```
pandemicsyn: Must-fix: The runtime enum and generated database constraint include skipped, but this expected enum list does not. The stable-enum schema test will fail until skipped is added here.

apps/web/src/components/cost-insights/overview/SpendEvidenceCard.tsx
```diff
@@ -0,0 +1,207 @@
...
+                  <ChartGridLine position="bottom-0" label="$0" />
+                </div>
+                {evidence.map((point, index) => {
+                  const pointTotal = point.variableUsd + point.scheduledUsd;
+                  const totalHeight = spendBarHeightPercent(pointTotal, maxSpend);
+                  const scheduledShare = percentOf(point.scheduledUsd, pointTotal);
+                  return (
+                    <Tooltip key={point.label}>
+                      <TooltipTrigger asChild>
+                        <button
```
pandemicsyn, line 144: Must-fix: The default seven-day chart renders 168 ordinary buttons, creating 168 sequential keyboard stops solely for tooltips. Use roving focus or a chart-level accessible summary/table so keyboard users can inspect the data without tabbing through every hour.

packages/db/src/schema.ts
```diff
@@ -2669,6 +2683,688 @@ export const organizations = pgTable(
...
+    event_type: text().$type<CostInsightEventType>().notNull(),
+    alert_kind: text().$type<CostInsightAlertKind>(),
+    suggestion_kind: text().$type<CostInsightSuggestionKind>(),
+    active_suggestion_id: uuid().references(() => cost_insight_active_suggestions.id, {
+      onDelete: 'set null',
+    }),
+    actor_user_id: text().references(() => kilocode_users.id, { onDelete: 'set null' }),
+    title: text().notNull(),
+    description: text().notNull(),
+    snapshot: jsonb().$type<CostInsightEventSnapshot>().default({}).notNull(),
```
pandemicsyn, line 3309: Suggested/nit: $type<CostInsightEventSnapshot>() provides compile-time typing only and does not validate persisted JSONB. Readers immediately trust nested arrays, amounts, enums, and timestamps, so malformed or legacy data can break dashboard and notification reads. Parse snapshots with a runtime schema at the persistence boundary.

apps/web/src/components/cost-insights/CostInsightsSettingsClient.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+  const {
+    mutate: updatePersonalSettings,
+    isPending: personalUpdatePending,
+    isError: personalUpdateError,
+  } = useMutation(
+    trpc.costInsights.updateSettings.mutationOptions({
+      onSuccess: () => {
+        toast.success('Cost Insights settings saved');
+        void invalidateCostInsights();
```
pandemicsyn: Suggested/nit: The success callback discards the invalidation promise, so the mutation reports success and re-enables controls before settings, dashboard, activity, and badge queries finish refreshing. Return or await invalidateCostInsights() from both success callbacks.

apps/web/src/app/(app)/components/SidebarMenuList.tsx
```diff
@@ -108,7 +109,14 @@ export default function SidebarMenuList({
                   </SidebarMenuButton>
                 )}
                 {item.badge && (
-                  <SidebarMenuBadge className="bg-brand-primary text-primary-foreground peer-hover/menu-button:text-primary-foreground peer-data-[active=true]/menu-button:text-primary-foreground right-4 !top-1/2 h-4 min-w-0 !-translate-y-1/2 rounded-full px-1.5 text-[10px] font-bold tracking-wide uppercase ring-1 ring-brand-primary/30">
+                  <SidebarMenuBadge
```
pandemicsyn, line 123: Suggested/nit: This numeric badge is rendered as a bare number outside the menu link, giving assistive technology no explanation such as "2 Spend Alerts need review." Add an accessible badge label or equivalent screen-reader text to the menu item.

apps/web/src/lib/cost-insights/evaluation.ts
```diff
@@ -0,0 +1,573 @@
...
+      sql`SELECT pg_catalog.pg_advisory_xact_lock(pg_catalog.hashtextextended(${lockKey}, 0))`
+    );
+    return await evaluateCostInsightsForOwnerLocked(tx, owner, options);
+  });
+}
+
+export function scheduleCostInsightEvaluationAfterSpend(owner: CostInsightSpendOwner): void {
+  if (process.env.NODE_ENV === 'test') return;
+  setTimeout(() => {
+    void evaluateCostInsightsForOwner(db, owner)
```
pandemicsyn: Performance warning: Every spend launches a full owner evaluation. Bursts create one transaction per spend, all queueing on the same advisory lock and repeating overlapping range queries while consuming database connections. Use a durable owner-keyed dirty marker so one worker coalesces pending spends into one evaluation.

apps/web/src/lib/cost-insights/jobs.ts
```diff
@@ -0,0 +1,50 @@
...
+export type CostInsightHourlySweepSummary = {
+  evaluatedOwners: number;
+  failedOwners: Array<{ owner: CostInsightSpendOwner; error: string }>;
+  notifications: Awaited<ReturnType<typeof dispatchPendingCostInsightNotifications>>;
+};
+
+export async function runCostInsightHourlySweep(
+  database: CostInsightRootDatabase
+): Promise<CostInsightHourlySweepSummary> {
+  const owners = await listEnabledCostInsightOwners(database);
```
pandemicsyn: Performance warning: The hourly cron materializes every enabled or suggestion-enabled owner and evaluates them serially in one request. With tens of thousands of owners, one slow fallback stalls all later owners and the run cannot reliably finish. Process keyset-paginated batches with checkpoints, a time budget, and bounded concurrency.

apps/web/src/lib/cost-insights/canonical-sources.ts
```diff
@@ -0,0 +1,943 @@
...
+      SUM(${microdollar_usage.cost})::text AS total_microdollars,
+      COUNT(*)::text AS spend_record_count
+    FROM ${microdollar_usage}
+    LEFT JOIN ${microdollar_usage_metadata}
+      ON ${microdollar_usage_metadata.id} = ${microdollar_usage.id}
+    LEFT JOIN ${feature}
+      ON ${feature.feature_id} = ${microdollar_usage_metadata.feature_id}
+    LEFT JOIN ${api_kind}
+      ON ${api_kind.api_kind_id} = ${microdollar_usage_metadata.api_kind_id}
+    WHERE ${microdollar_usage.created_at} >= ${range.startInclusive}
```
pandemicsyn, line 572: Performance warning: These canonical scans filter the 800M-row microdollar_usage table by organization and time, but the schema has only separate created_at and organization_id indexes. Multi-day organization fallbacks can build huge bitmaps or scan all organization history. Add an online partial composite (organization_id, created_at) index before enabling these paths.

apps/web/src/lib/cost-insights/canonical-sources.ts
```diff
@@ -0,0 +1,943 @@
...
+      ${credit_transactions.organization_id} AS owned_by_organization_id,
+      ${coding_plan_terms.user_id} AS actor_user_id,
+      ${coding_plan_terms.plan_id} AS plan_id,
+      ${coding_plan_subscriptions.provider_id} AS provider_id,
+      ${coding_plan_terms.kind} AS term_kind,
+      SUM(-${credit_transactions.amount_microdollars})::text AS total_microdollars,
+      COUNT(*)::text AS spend_record_count
+    FROM ${coding_plan_terms}
+    INNER JOIN ${credit_transactions}
+      ON ${credit_transactions.id} = ${coding_plan_terms.credit_transaction_id}
```
pandemicsyn, line 634: Performance warning: This filtered credit_transactions result joins coding_plan_terms through credit_transaction_id, but that foreign-key column has no index. Large canonical queries can scan or hash the entire terms table. Add an index on coding_plan_terms(credit_transaction_id) through a generated migration.

apps/web/src/lib/cost-insights/evaluation.ts
```diff
@@ -0,0 +1,573 @@
...
+        params.asOf,
+        params.alertKind
+      );
+    }
+    return false;
+  }
+
+  if (crossingActive) return false;
+
+  const evidence = await getOwnerRollingDriverEvidenceExact(params.database, {
```
pandemicsyn, line 309: Performance warning: A first 7-day or 30-day threshold crossing groups the full raw canonical window while the owner evaluation lock remains held. Large owners can time out while blocking later evaluations. Build exact evidence from hourly driver rollups plus raw boundary fragments, selecting the top five in SQL.

packages/db/src/cost-insights-rollups.ts
```diff
@@ -0,0 +1,435 @@
...
+        ${values.driver.productKey}::text AS product_key,
+        ${values.driver.featureKey}::text AS feature_key,
+        ${values.driver.modelOrPlanKey}::text AS model_or_plan_key,
+        ${values.driver.providerKey}::text AS provider_key,
+        ${values.driver.actorUserId}::text AS actor_user_id,
+        ${values.amountMicrodollars}::bigint AS amount_microdollars,
+        ${values.spendRecordCount}::bigint AS spend_record_count,
+        ${lockKey}::text AS lock_key
+    ), owner_hour_lock AS MATERIALIZED (
+      SELECT pg_catalog.pg_advisory_xact_lock(
```
pandemicsyn: Performance warning: Every live spend takes an exclusive owner-hour advisory lock and updates the same owner/category total row. A busy organization therefore serializes billing behind one lock and hot tuple. Benchmark lock waits before rollout and use shared-live/exclusive-maintenance coordination or sharded accumulators if contention is material.

pandemicsyn (changes requested):
Found additional correctness issues and non-blocking performance warnings after excluding feedback already present on the PR. Database performance warnings are advisory and do not individually block merging.

apps/web/src/lib/cost-insights/repository.ts
jeanduplessis: Fixed in b4b660f75. Active alert state now keeps stable episode IDs and compact alert snapshots independent from retained Cost Insight Events, so 90-day retention can delete event history without removing banner presentation or acknowledgment identity. Added regression coverage around retained active episodes.

apps/web/src/components/cost-insights/overview/CostInsightsDashboardView.tsx
jeanduplessis: Fixed in b4b660f75. The overview client now passes the pending suggestion ID through to the dashboard and suggestion card; the matching dismiss action is disabled during settlement and exposes aria-busy to avoid duplicate requests/toasts.

apps/web/src/lib/cost-insights/presenter.ts
jeanduplessis: Fixed in b4b660f75. Live presenter data now supplies memberLimitsHref for eligible organization dashboards only when the existing member daily-limit controls are actually available to the viewer, with presenter coverage.

apps/web/src/routers/cost-insights-router.ts
jeanduplessis: Fixed in b4b660f75. Acknowledge mutations now include the displayed event/episode ID, and the repository update requires that ID to match the current active episode before marking reviewed. Added personal and organization regression coverage for stale-page same-kind episodes.

apps/web/src/components/cost-insights/overview/SpendEvidenceCard.tsx
jeanduplessis: Fixed in b4b660f75. The spend chart now uses roving focus: one bar is tabbable, arrow keys/Home/End move through periods, and unavailable periods have explicit accessible labels. This removes the 168 sequential Tab stops while preserving keyboard inspection.

apps/web/src/lib/exa-usage.ts
jeanduplessis: Fixed in b4b660f75. The Exa route now parses the upstream response through a runtime schema, requires finite positive costDollars when present, converts to microdollars, and rejects non-positive or unsafe integer conversions before recording usage. Added regression tests for malformed, non-finite, and oversized values.

apps/web/src/app/(app)/components/SidebarMenuList.tsx
jeanduplessis: Fixed in b4b660f75. Numeric sidebar badges are hidden from assistive tech and the menu item now includes screen-reader text describing the review count, e.g. '2 items need review'.

apps/web/src/lib/cost-insights/canonical-sources.ts
jeanduplessis: Fixed in 2d48162fa. Added generated schema/migration index idx_microdollar_usage_org_created_at on (organization_id, created_at) with organization_id IS NOT NULL partial predicate for organization fallback scans.

apps/web/src/lib/cost-insights/canonical-sources.ts
jeanduplessis: Fixed in 2d48162fa. Added generated schema/migration index IDX_coding_plan_terms_credit_transaction on coding_plan_terms.credit_transaction_id for canonical credit transaction joins.

packages/db/src/schema.ts
jeanduplessis: Fixed in 2d48162fa. Added CostInsightEventSnapshotSchema runtime validation in packages/db, validate snapshots on write, parse/degrade persisted snapshots on read, and skip notification dispatch for malformed persisted snapshots with regression coverage.

apps/web/src/lib/cost-insights/evaluation.ts
jeanduplessis: Fixed in 2d48162fa. Threshold driver evidence now uses hourly driver rollups for the complete interior and canonical raw fragments only at window boundaries, with a canonical fallback when rollup coverage is incomplete. Added integration coverage for the hybrid path.

apps/web/src/lib/cost-insights/notifications.ts
jeanduplessis: Fixed in b4b660f75. Stale notification deliveries in sending are now reclaimable after a bounded lease, exhausted stale claims are terminalized, and interruption-after-claim coverage was added.

apps/web/src/routers/cost-insights-router.ts
jeanduplessis: Fixed in b4b660f75. Configuration writes, episode clearing, and required configuration/disabled events now run in one transaction; external evaluation happens after commit.

apps/web/src/lib/cost-insights/evaluation.ts
jeanduplessis: Fixed in b4b660f75. Evaluation now derives one authoritative interval from asOf and passes that interval through anomaly spend/evidence queries instead of reading CURRENT_TIMESTAMP independently.

apps/web/src/lib/cost-insights/repository.ts
jeanduplessis: Fixed in b4b660f75. Suggestion dismissal now updates the suggestion and writes suggestion_dismissed history in one transaction, with rollback coverage for event-write failure.

apps/web/src/lib/cost-insights/repository.ts
jeanduplessis: Fixed in b4b660f75. Alert acknowledgment now performs the conditional state update and alert_reviewed event insert in one transaction, with rollback coverage for event-write failure.

apps/web/src/components/cost-insights/ask-kilo/CostInsightsAskKiloView.tsx
jeanduplessis: Fixed in b4b660f75. Ask Kilo cost-insights surface no longer presents the static sample as live owner financial analysis; it is gated/presented as preview behavior instead of production spend evidence.

apps/web/src/components/cost-insights/overview/SpendEvidenceCard.tsx
jeanduplessis: Fixed in b4b660f75. Long-range evidence points retain period timestamps and client labels are localized with the viewer time zone, including regression coverage around offset-sensitive dates.

apps/web/src/lib/cost-insights/jobs.ts
jeanduplessis: Fixed in b4b660f75 and hardened in 2d48162fa. Hourly sweep can recover the just-completed hour for missed anomaly episodes while preserving current-time rolling threshold evaluation and dedupe behavior.

apps/web/src/scripts/db/exa-usage-log-indexes.ts
jeanduplessis: Fixed in b4b660f75. The EXA usage-log index script now detects invalid concurrent indexes via pg_index.indisvalid and rebuilds/drops invalid relations before reporting partition index success.

dev/seed/cost-insights/spend-evidence.ts
jeanduplessis: Fixed in b4b660f75. Normal fixture seeding no longer mutates global rollup coverage for owner-scoped data; destructive coverage behavior is isolated from the standard path.

dev/seed/cost-insights/spend-evidence.ts
jeanduplessis: Fixed in b4b660f75. The documented fixture path/default behavior now matches healthy complete rollups for evidence ranges rather than bootstrap-only partial coverage.

apps/web/src/app/api/cron/cost-insights-hourly/route.ts
jeanduplessis: Fixed in 2d48162fa. Cron response now exposes processed counts and partial failure metadata so monitoring can distinguish fully healthy runs from runs with failed owner evaluations or notification failures.

apps/web/src/lib/cost-insights/presenter.ts
jeanduplessis: Fixed in b4b660f75. DB-backed timestamps returned through Cost Insights presentation are normalized with Date(...).toISOString() before crossing the tRPC/client contract.

apps/web/src/lib/cost-insights/presenter.ts
jeanduplessis: Fixed in b4b660f75. Presentation now preserves unavailable rollup evidence instead of coalescing coverage gaps to zero spend, so incomplete 30d/90d data does not appear as trustworthy zero bars/totals.

apps/web/src/components/cost-insights/settings/CostInsightsSettingsView.tsx
jeanduplessis: Fixed in b4b660f75. Spend Alert sub-options remain visible while alerts are disabled so saved anomaly/threshold settings remain inspectable and editable before re-enable.

apps/web/src/lib/cost-insights/repository.ts
jeanduplessis: Fixed in b4b660f75. Sidebar attention count now uses only unreviewed Spend Alerts; active Cost Suggestions no longer contribute alert attention.

apps/web/src/lib/cost-insights/policy.ts
jeanduplessis: Fixed in b4b660f75. Threshold parsing validates the final microdollar value after cents-to-microdollars conversion and returns user-facing validation instead of allowing unsafe DB values.

apps/web/src/components/cost-insights/CostInsightsSettingsClient.tsx
jeanduplessis: Fixed in b4b660f75. Settings success callbacks await/return Cost Insights query invalidation so controls are not re-enabled before dashboard/activity/badge refresh completes.

apps/web/src/lib/cost-insights/jobs.ts
jeanduplessis: Fixed in 2d48162fa. Hourly cron now uses leased keyset-paginated owner sweep checkpoints with a deadline and bounded concurrency, so large owner sets resume across invocations instead of materializing/evaluating all owners serially.

packages/db/src/cost-insights-rollups.ts
jeanduplessis: Fixed in 2d48162fa. Live rollup capture now uses shared advisory coordination while maintenance remains exclusive, reducing owner-hour serialization risk; added concurrency-oriented rollup coverage.

packages/db/src/schema.test.ts
jeanduplessis: Fixed in b4b660f75. Stable enum expectations include skipped so the runtime enum, database constraint, and schema test agree.

apps/web/src/lib/cost-insights/evaluation.ts
jeanduplessis: Fixed in b4b660f75 and hardened in 2d48162fa. Spend-triggered work uses durable owner-keyed dirty markers with bounded claim processing, coalescing bursts into owner evaluations instead of launching one full evaluation per spend.

packages/db/src/migrations/0176_opposite_tiger_shark.sql
```diff
@@ -0,0 +1,316 @@
...
+CREATE INDEX "IDX_cost_insight_degraded_intervals_unresolved" ON "cost_insight_rollup_degraded_intervals" USING btree ("start_hour","end_hour_exclusive") WHERE "cost_insight_rollup_degraded_intervals"."resolved_at" is null;--> statement-breakpoint
+CREATE INDEX "IDX_coding_plan_terms_credit_transaction" ON "coding_plan_terms" USING btree ("credit_transaction_id");--> statement-breakpoint
+CREATE INDEX "IDX_credit_transactions_user_negative_created" ON "credit_transactions" USING btree ("kilo_user_id","created_at") WHERE "credit_transactions"."amount_microdollars" < 0;--> statement-breakpoint
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 3d90cd2bf. Regenerated the Cost Insights migration without the credit_transactions negative-spend indexes, added a separate credit-transaction-indexes operator script that creates them with CREATE INDEX CONCURRENTLY and validates/rebuilds interrupted invalid indexes, plus script coverage.

packages/db/src/migrations/0176_opposite_tiger_shark.sql
```diff
@@ -0,0 +1,316 @@
...
+CREATE INDEX "IDX_credit_transactions_org_negative_created" ON "credit_transactions" USING btree ("organization_id","created_at") WHERE "credit_transactions"."amount_microdollars" < 0 AND "credit_transactions"."organization_id" IS NOT NULL;--> statement-breakpoint
+CREATE INDEX "idx_exa_usage_log_org_created" ON "exa_usage_log" USING btree ("organization_id","created_at") WHERE "exa_usage_log"."organization_id" IS NOT NULL;--> statement-breakpoint
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 3d90cd2bf. Regenerated the migration without the plain parent-table exa_usage_log organization index so existing/current Exa partitions continue to use the concurrent per-partition provisioning path instead of a locking transactional migration.

apps/web/src/lib/cost-insights/evaluation.ts
```diff
@@ -0,0 +1,790 @@
...
+  variableMicrodollars: number;
+  anomalyPolicy: Awaited<ReturnType<typeof getCostInsightAnomalyPolicy>>;
+  topDrivers: OwnerTopSpendDriver[];
+}): Promise<boolean> {
+  if (params.variableMicrodollars < params.anomalyPolicy.thresholdMicrodollars) {
+    return false;
+  }
+
+  const dashboardState = await getCostInsightDashboardState(params.database, params.owner);
+  if (dashboardState.state?.activeAnomalyHourStart === params.hourStart) {
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 3d90cd2bf. The active anomaly hour comparison now normalizes the DB-backed timestamp with Date(...).toISOString() before comparing it to the canonical evaluation hour, so the short-circuit works with production-shaped Postgres timestamp text.

apps/web/src/components/cost-insights/shell/CostInsightsAlertBar.tsx
```diff
@@ -0,0 +1,39 @@
...
+export function CostInsightsAlertBar({
+  owner,
+  alertCount,
+}: {
+  owner: CostInsightsOwner;
+  alertCount: number;
+}) {
+  const reviewHref =
+    owner.type === 'organization'
+      ? '/organizations/acme-cost-insights/cost-insights'
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 3d90cd2bf. CostInsightsAlertBar now receives reviewHref from its caller instead of hardcoding a placeholder organization route, and the Storybook story supplies an explicit fixture route.

apps/web/src/routers/cost-insights-router.ts
```diff
@@ -0,0 +1,333 @@
...
+      if (acknowledged) {
+        trackCostInsightsAlertAction({
+          ...personalTrackingContext(ctx.user.id),
+          action: 'acknowledge',
+          alertKind: input.alertKind,
+        });
+      }
+      return { success: true };
+    }),
+  disableThreshold: adminProcedure.mutation(async ({ ctx }) => {
```
St0rmz1: **Low, Design:** The `disableThreshold` mutation (personal here, and the organization equivalent) is not referenced by any client component; a full-tree scan finds it only in the two routers and their tests. The presenter emits only `acknowledge`, `view_spend`, and `manage_threshold` actions for threshold banners, so no UI path reaches this endpoint.

Beyond being dead surface, it is also asymmetric: `disableOwnerThreshold` nulls only `spend_threshold_microdollars` and calls `clearCostInsightThresholdEpisode(database, owner, null)` with the default `alertKind = 'threshold'` (the 24h window). If someone later wires a disable control onto a 7d or 30d banner and points it here, it will silently disable the 24h threshold instead of the one the user acted on, since the mutation carries no `alertKind`.

Suggestion: either remove the unused `disableThreshold` mutations, or add an `alertKind` input and thread it through both the config field being nulled and `clearCostInsightThresholdEpisode` so all three windows are handled symmetrically before it is wired to UI.

apps/web/src/routers/organizations/organization-cost-insights-router.ts
```diff
@@ -0,0 +1,249 @@
...
+    distinctId: userId,
+    userId,
+    ownerType: 'organization',
+    organizationId,
+    authorizedRole,
+  };
+}
+
+async function ensureOrgManageAccess(ctx: TRPCContext, organizationId: string) {
+  if (!ctx.user.is_admin) {
```
St0rmz1: **Low, Design:** This `if (!ctx.user.is_admin)` branch is unreachable. Every procedure in this router is built on `adminProcedure`, which already throws `FORBIDDEN` unless `ctx.user.is_admin`, so by the time `ensureOrgManageAccess` runs the user is always an admin. The same applies to the non-admin path in `resolveOrgReadContext` (the `ensureOrganizationAccess` call after the `if (ctx.user.is_admin)` return).

This is not a security hole: the reachable admin branch correctly requires a direct owner or billing-manager role before allowing mutations, matching the spec rule that admins must not change customer settings without owner or billing-manager authority. The concern is that the dead branch implies non-admins can reach these paths, which is misleading and could invite an incorrect "loosen the gate" change later.

Suggestion: drop the unreachable non-admin branches, or if they are kept deliberately in case the procedure base changes, add a short comment saying so plus a test asserting the admin gate fires first.

apps/web/src/app/api/cron/cost-insights-hourly/route.ts
```diff
@@ -0,0 +1,67 @@
...
+if (!CRON_SECRET) {
+  throw new Error('CRON_SECRET is not configured in environment variables');
+}
+
+export const maxDuration = 300;
+
+export async function GET(request: Request) {
+  const authHeader = request.headers.get('authorization');
+  const expectedAuth = `Bearer ${CRON_SECRET}`;
+  if (authHeader !== expectedAuth) {
```
St0rmz1: **Nit, Design:** The bearer token check uses `authHeader !== expectedAuth`, a non-constant-time string comparison, which is a timing side channel in principle. This matches the existing convention for cron routes in this repo (the sibling retention route does the same), so it is consistent rather than a regression.

Since both cron routes are new here, a shared helper using `crypto.timingSafeEqual` would be a small hardening win. Not blocking.

St0rmz1:
**Low, Documentation:** The "Run this branch locally" instructions say the seeded owner (`cost-insights-owner@example.com`) and billing manager (`cost-insights-billing-manager@example.com`) can view and manage organization Cost Insights, and that the Kilo admin is a read-only inspector. Per the spec (Authorization rules 1 to 3) and the actual code, v1 is Kilo-admin-only: personal routes use `adminProcedure` (requires `is_admin`), and the organization layout does `if (!isGlobalAdmin) notFound()`. The seeded owner and billing-manager emails are not `@admin.example.com`, so those non-admin logins will hit notFound / FORBIDDEN rather than a manageable dashboard. The read-only admin login is actually the working path, which is the reverse of what the instructions imply.

A reviewer following these steps will conclude the feature is broken when it is behaving to spec. Suggestion: update the local-run narrative so the owner and billing-manager fixtures only gain access when they are also Kilo admins, and make the primary working path a Kilo admin who also holds an owner or billing-manager role in the seeded org.

St0rmz1 (approved):

apps/web/src/routers/cost-insights-router.ts
jeanduplessis: Fixed in a36f12f15. Removed the unused personal and organization disableThreshold mutations and their router coverage, leaving threshold disablement to the settings flow reached through Manage threshold so 24h/7d/30d behavior stays symmetric.

apps/web/src/routers/organizations/organization-cost-insights-router.ts
jeanduplessis: Fixed in 980402be8. Removed the unreachable non-admin branches from the organization Cost Insights router; the router now reflects the adminProcedure gate directly while still requiring a direct owner or billing-manager role before organization mutations.

apps/web/src/app/api/cron/cost-insights-hourly/route.ts
jeanduplessis: Fixed in fc50d0f16. Added a shared timingSafeEqual-based cron authorization helper, wired both Cost Insights cron routes through it, and added route/helper tests for invalid authorization.

jeanduplessis:
Updated the PR description for the local-run documentation note. It now states that Cost Insights v1 is admin-only during initial rollout, clarifies that the seeded owner and billing-manager fixtures are Kilo admins, replaces the stale read-only-admin URL with a non-admin contributor access check, and updates the suggested local checks to use Manage threshold via settings.

packages/db/src/migrations/0178_gorgeous_cassandra_nova.sql
```diff
@@ -0,0 +1,313 @@
...
+CREATE INDEX "IDX_cost_insight_degraded_intervals_unresolved" ON "cost_insight_rollup_degraded_intervals" USING btree ("start_hour","end_hour_exclusive") WHERE "cost_insight_rollup_degraded_intervals"."resolved_at" is null;--> statement-breakpoint
+CREATE INDEX "IDX_coding_plan_terms_credit_transaction" ON "coding_plan_terms" USING btree ("credit_transaction_id");--> statement-breakpoint
```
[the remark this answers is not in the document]
jeanduplessis: Coding plans table is very small, the impact is negligible.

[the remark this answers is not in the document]
jeanduplessis: Fixed in f30206dc4. I removed this index from the generated Drizzle migration path and moved it into the new `cost-insights-source-indexes` operator script, where it is created with `CREATE INDEX CONCURRENTLY IF NOT EXISTS` and validated/rebuilt if a prior concurrent attempt left it invalid.

packages/db/src/migrations/0178_gorgeous_cassandra_nova.sql
```diff
@@ -0,0 +1,313 @@
...
+CREATE INDEX "idx_microdollar_usage_org_created_at" ON "microdollar_usage" USING btree ("organization_id","created_at") WHERE "microdollar_usage"."organization_id" is not null;
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in f30206dc4. I removed `idx_microdollar_usage_org_created_at` from the generated migration/schema declaration and added it to the new `cost-insights-source-indexes` operator script so the 800M-row `microdollar_usage` table gets the index through `CREATE INDEX CONCURRENTLY IF NOT EXISTS`, with catalog verification and invalid-index rebuild handling.

packages/db/src/database-url.ts
```diff
@@ -29,12 +29,18 @@ export function getDatabaseClientConfig(postgresUrl: string): ClientConfig {
...
+
   if (process.env.DATABASE_CA) {
     clientConfig.ssl = {
       ca: process.env.DATABASE_CA.replace(/\\n/g, '\n'),
       rejectUnauthorized: true,
       servername: databaseUrl.hostname,
     };
+  } else if (sslMode === 'require') {
+    clientConfig.ssl = {
+      rejectUnauthorized: false,
```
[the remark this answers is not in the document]
jeanduplessis: Removed temporary hack.

pandemicsyn (approved):

PR #4279, opened 2026-06-26
feat(scripts): expire internal team free credits

iscekic:
## What

One-off maintenance script (`apps/web/src/scripts/d2026-06-26_expire-internal-free-credits.ts`) that expires free credit grants belonging to internal team accounts.

## Selection

Users whose `kilocode_users.google_user_email` ends in `@kilocode.ai` or `@kilo.ai` (case-insensitive suffix match, so it's independent of the nullable backfilled `email_domain` column). Per matched user, the affected set is every credit transaction that is:

- `is_free = true`
- `organization_id IS NULL` (personal)
- `amount_microdollars > 0` (a grant, not an adjustment/expiration row)
- not already expired (no `credits_expired` / `orb_credit_expired` / `orb_credit_voided` record referencing it)

This includes grants **with or without** an existing `expiry_date`.

## Behavior

For each affected grant, set `expiry_date = now` and expire it through `processLocalExpirations()`, so balance accounting matches the production expiration cron exactly:

- Fully-spent grants expire `$0` → effectively left untouched.
- Partially-spent grants expire only their unspent remainder.

If expiring a user's grants drives their balance below zero (e.g. spend previously clawed back via `total_microdollars_acquired`), a one-off `accounting_adjustment` credit (`is_free`, positive) restores the balance to exactly `$0`. No user ends negative.

## Mechanics

- Dry run by default; `--execute` performs writes. `--yes` skips the confirmation prompt; `--batch-size` / `--concurrency` tunable.
- Emits a per-user JSONL log, a mutation log (for rollback), and an error log under `src/scripts/output/`.
- No schema changes. Reuses `computeExpiration` / `processLocalExpirations` from `@/lib/creditExpiration`.

## Notes

- The corrective `accounting_adjustment` is itself a free, non-expiring credit, so re-running the script would re-target it. Intended as a one-off.

### DIFF
--- a/apps/web/src/scripts/d2026-06-26_expire-internal-free-credits.ts
+++ b/apps/web/src/scripts/d2026-06-26_expire-internal-free-credits.ts
@@ -1,0 +1,468 @@
+/**
+ * Immediately expires ALL free credit grants (with or without an existing expiry
+ * date) belonging to internal team accounts — users whose email ends in
+ * @kilocode.ai or @kilo.ai — while keeping their balances non-negative.
+ *
+ * Per user:
+ *   - Every free, personal, positive credit grant that hasn't already been
+ *     expired is tagged with expiry_date = now and then expired (the unspent
+ *     remainder is removed from the balance via a credits_expired transaction).
+ *       · Fully-spent grants expire $0 → effectively left alone.
+ *       · Partially-spent grants expire their unspent remainder.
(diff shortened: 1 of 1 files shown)


iscekic:
Script already ran against prod. @pedroheyerdahl has run output details in Slack DMs.

emilieschario:
Approved in concept, DNR

emilieschario (approved):

PR #4316, opened 2026-06-30
feat(notifications): add upgrade nudge for legacy extension users (version < 7)

markijbema:
## Summary

- Adds a new `extensionVersionBelow` field to `KiloNotification` to enable version-gated server-side filtering of notifications by extension major version.
- Updates the `/api/users/notifications` route to extract the client's extension version from `X-KiloCode-Version` / `User-Agent` headers and passes it to `generateUserNotifications`.
- Adds a new `legacy-upgrade-june-2026` notification targeting users on extension version < 7 (kilocode-legacy users), informing them that the legacy extension reaches end of life on **July 31, 2026** and directing them to the end-of-life notice. Current kilocode users (version ≥ 7) never see this notification.

## Verification

- Install an older (version < 7) kilocode-legacy extension build and call `/api/users/notifications` — the `legacy-upgrade-june-2026` notification should appear with the EOL messaging and link to the GitHub notice.
- Call the endpoint with a current kilocode extension (version ≥ 7) or with no version headers — the notification should **not** appear.

## Visual Changes

N/A

## Reviewer Notes

- Filtering is server-side: the route extracts the version and passes it into `generateUserNotifications`; the extension does not need to understand the new field.
- `extensionVersionBelow: 7` means any encoded float < 7.0 (i.e. major version 0–6); since minor/1000 + patch/1_000_000 < 1, this is safe for all realistic version strings.
- When version headers are absent the notification is hidden (conservative, avoids showing it to non-extension clients).
- Notification title and message are consistent with the [Legacy IDE Extensions End of Life](https://github.com/Kilo-Org/kilocode-legacy#legacy-ide-extensions-end-of-life) section in the kilocode-legacy README (EOL date: July 31, 2026).

---

Built for [Mark IJbema](https://kilo-code.slack.com/archives/D0A5633FAF2/p1782809669707179?thread_ts=1777492079.769449&cid=D0A5633FAF2) by [Kilo for Slack](https://kilo.ai/slack)

### DIFF
--- a/apps/web/src/lib/notifications.ts
+++ b/apps/web/src/lib/notifications.ts
@@ -41,12 +41,12 @@ const normalUnconditionalNotifications: KiloNotification[] = [
   //if you just want a simple straightforward global message, add it here.
   {
     id: 'legacy-upgrade-june-2026',
-    title: 'A New Version of Kilo Code Is Available',
+    title: 'Legacy Kilo Code Extension: End of Life July 31, 2026',
     message:
-      "You're using an older version of Kilo Code. Upgrade to the latest version to get new models, faster performance, and the best experience.",
+      'This extension reaches end of life on July 31, 2026 — no further updates, bug fixes, or security patches after that date. Switch to the current Kilo Code extension for continued support.',
     action: {
-      actionText: 'Get the Latest Version',
-      actionURL: 'https://marketplace.visualstudio.com/items?itemName=kilocode.kilo-code',
+      actionText: 'See End of Life Notice',
+      actionURL: 'https://github.com/Kilo-Org/kilocode-legacy#legacy-ide-extensions-end-of-life',
     },
     showIn: ['extension'],
     extensionVersionBelow: 7,
--- a/apps/web/src/lib/notifications.ts
+++ b/apps/web/src/lib/notifications.ts
@@ -31,10 +31,26 @@ export type KiloNotification = {
   showIn?: ('extension' | 'extension-native' | 'cli')[];
   // ISO 8601 timestamp after which this notification should no longer be shown
   expiresAt?: string;
-  // When specified, only show to extension clients whose numeric version (major + minor/1000 + patch/1_000_000) is strictly below this value.
-  // Use an integer major version (e.g. 7 means "below 7.x.x"). Hidden when the client version is unknown.
+  // When specified, only show to clients whose numeric version (major + minor/1000 + patch/1_000_000) is strictly below this value.
+  // Use an integer major version (e.g. 7 means "below 7.x.x"). An unknown/absent client
+  // version is treated as an old (pre-versioning) client and IS shown — legacy extensions
+  // send no version headers, so absence of data means "old", not "hide".
   extensionVersionBelow?: number;
 };
+
+/**
+ * Decide whether a version-gated notification should be shown to a client.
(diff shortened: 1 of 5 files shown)


markijbema:
This doesnt work, the legacy extension doesnt send the header this pr expect

markijbema:
I'll add headers to the new extension, and then send to old agents without headers

apps/web/src/lib/userAgent.test.ts
```diff
@@ -51,3 +55,21 @@ describe('getKiloCodeVersionNumber', () => {
...
+    expect(isLegacyKiloExtensionUserAgent('axios/0.27.2')).toBe(true);
+  });
+
+  test('does not match current extension, CLI, bot, or missing User-Agents', () => {
+    expect(isLegacyKiloExtensionUserAgent(null)).toBe(false);
+    expect(isLegacyKiloExtensionUserAgent(undefined)).toBe(false);
+    expect(isLegacyKiloExtensionUserAgent('')).toBe(false);
+    expect(isLegacyKiloExtensionUserAgent('opencode-kilo-provider/7.1.0')).toBe(false);
+    expect(isLegacyKiloExtensionUserAgent('opencode-kilo-provider')).toBe(false);
+    expect(isLegacyKiloExtensionUserAgent('Kilo-Code/5.1.0')).toBe(false);
```
chrarnoldus: I think the function should be renamed because this actually is the user-agent for LLM calls.

markijbema: we could also make it true there right?

chrarnoldus: right, do we have strong reasons to believe the new extension will never use axios?

markijbema: I think we should remove this notification asap (say before august)

chrarnoldus: can we rename this to isLegacyKiloExtensionNotificationsUserAgent then

apps/web/src/lib/notifications.ts
```diff
@@ -33,11 +31,39 @@ export type KiloNotification = {
...
+}
+
 const normalUnconditionalNotifications: KiloNotification[] = [
   //If you need to check or personalize the notification, see examples at the bottom of this file
   //if you just want a simple straightforward global message, add it here.
+  {
+    id: 'legacy-upgrade-june-2026',
+    title: 'Kilo Code 5.x: End of Life July 31, 2026',
+    message:
+      'Version 5.x of the Kilo Code extension reaches end of life on July 31, 2026 — no further updates, bug fixes, or security patches after that date. Upgrade to the latest version for continued support.',
```
lambertjosh:
```suggestion
      'Kilo Code extension version 5.x reaches end of life on July 31, 2026. After that date, it will no longer receive updates, bug fixes, or security patches. Upgrade to the latest version for continued support.',
```

How about this for slightly smoother language.

lambertjosh:
I couldn't get this to popup for me locally, for some reason. 

I did get this notification (also with another 4 about the github star giveaway, the CLI, etc.) We should remove it in favor of the EOL announcement (basically do a swap).

image

chrarnoldus (approved):

markijbema:
That other notification is hardcoded in the extension. I don't think it will make much of a difference whether we touch it; people will have discarded it months ago, and they otherwise will need to manually update to have it disappear

PR #4325, opened 2026-06-30
CloudAgent - Support CLI-configured models in remote sessions

eshurakov:
## Summary

Support CLI-configured models when chatting with live remote CLI sessions from Cloud Agent web and mobile. Previously the model picker was limited to Kilo Gateway models. This change introduces a session-scoped, owner-fenced `list_models` protocol v1, validates and normalizes the SDK `ProviderListResponse` into a provider-aware catalog, and wires it into the web and mobile model pickers while preserving the Gateway fallback for older CLIs.

## Verification

- Manually verified the web Cloud Chat model picker lists provider-grouped CLI models for a live remote session.
- Manually verified selecting a CLI model sends the structured `{ providerID, modelID }` payload.
- Manually verified an older CLI still falls back to the Kilo Gateway model list.

## Visual Changes

N/A

## Reviewer Notes

- Scope is live remote CLI sessions only; cloud-hosted Cloud Agent sessions remain Kilo Gateway-only.
- `UserConnectionDO` now fences catalog requests by session owner, limits pending commands, and rejects oversized catalog payloads.
- Observed model, catalog default, and explicit override are kept separate; only explicit override provenance affects sends.


apps/web/src/lib/cloud-agent-sdk/session-manager.ts
```diff
@@ -532,6 +578,84 @@ function createSessionManager(config: SessionManagerConfig): SessionManager {
...
+  // A web-picked override should stop applying once we see live proof the
+  // CLI actually ran a message on a different model — otherwise the picker
+  // gets stuck showing a choice that's no longer what's being sent. Gated on
+  // `remoteHistoryReplaying` so a reconnect's replayed history (which can
+  // predate the override) can't wipe a selection that just hasn't been used
+  // yet.
+  function clearOverrideIfDiverged(model: ModelSelection): void {
+    if (remoteHistoryReplaying) return;
+    const override = store.get(remoteModelOverrideAtom);
+    if (override && !modelRefsEqual(override.selection.model, model.model)) {
```
jeanduplessis: Remote override divergence only compares provider/model, so a live message with the same model but a different variant leaves the stale override active. Since `send()` later reuses `remoteModelOverride.selection.variant`, the picker and future sends can keep applying an old variant after the CLI proves it ran another one. Consider clearing when `!modelSelectionsEqual(override.selection, model)` instead of only `!modelRefsEqual(...)`.

apps/web/src/components/cloud-agent-next/hooks/useSessionModels.ts
```diff
@@ -0,0 +1,419 @@
...
+  remoteContextLengthByProviderAndModel?: ContextLengthByProviderAndModel;
+};
+
+export function resolveGatewayOrganizationId(
+  fetchedSessionData: Pick<FetchedSessionData, 'organizationId'> | null,
+  routeOrganizationId: string | undefined,
+  sessionIdFromParams: string | null
+): string | undefined {
+  if (fetchedSessionData) return fetchedSessionData.organizationId ?? undefined;
+  return sessionIdFromParams ? undefined : routeOrganizationId;
```
jeanduplessis: For an existing org-owned remote session, `fetchedSessionData` is initially null and `sessionIdFromParams` is present, so this returns `undefined`. If an older CLI is in legacy mode, `useOrganizationModels(undefined, true)` can load the personal/default Gateway catalog and enable the picker before the persisted org ID hydrates. That can expose/send fallback models outside org model policy. Consider disabling legacy Gateway fallback until `fetchedSessionData` resolves for existing sessions, or using the route org provisionally and revalidating once persisted ownership loads.

apps/mobile/src/components/agents/model-selector.tsx
```diff
@@ -107,3 +190,195 @@ export function ModelSelector({
...
+            <Icon size={14} color={colors.warn} />
+            <Text selectable className="flex-1 text-xs leading-4 text-foreground">
+              {notice.message}
+            </Text>
+            {notice.retry ? (
+              <Pressable
+                onPress={onRetry}
+                accessibilityRole="button"
+                accessibilityLabel="Retry loading CLI models"
+                className="flex-row items-center gap-1 rounded-full bg-secondary px-2 py-1 active:opacity-70"
```
jeanduplessis, line 374: This retry `Pressable` is a compact `px-2 py-1` chip with no `hitSlop` or min target. On mobile, this falls below the 44px touch target expected for actionable controls, making the recovery path hard to hit when CLI model loading fails. Add `hitSlop` or `min-h-[44px] min-w-[44px]` while keeping the visual chip compact.

apps/mobile/src/lib/hooks/use-session-model-options.ts
```diff
@@ -0,0 +1,377 @@
...
+  const catalog = input.remoteModelState.catalog;
+  if (!catalog) {
+    throw new Error('CLI catalog is required for v1 model options');
+  }
+
+  let opaqueIndex = 0;
+  const options = sortRemoteModelCatalogProviders(catalog.providers).flatMap(provider =>
+    provider.models.map(model => {
+      const option: SessionModelOption = {
+        id: `remote-model-${opaqueIndex}`,
```
jeanduplessis, line 243: Remote CLI catalog options use order-based IDs like `remote-model-0`, but favorites are persisted by `option.id`. If catalog ordering changes or another CLI/session has a different catalog, that saved favorite can point at a different provider/model. Consider storing favorites by stable remote identity such as `providerID:modelID`, or disabling favorites for remote CLI catalog options until stable keys exist.

jeanduplessis (approved):
Approved per request. I left inline findings for the verified issues from the pinned-head review.

PR #4328, opened 2026-06-30
fix(auth): harden device-auth poll (atomic consume + blocked re-check)

markijbema:
## Summary

Follow-up to #4314 (block rotates pepper). That PR was Workstream B; this is **Workstream A** — server-side hardening of the device-authorization poll path in `apps/web/src/lib/device-auth/device-auth.ts`. It closes two findings from the device-auth review:

- **Finding 1 — non-atomic single-use consume.** `pollDeviceAuthRequest` read the approved request, minted a token, then issued an unconditional `status = 'expired'` update. Two concurrent polls could each pass the read and each mint a long-lived API token from one approval. Replaced with a guarded compare-and-swap (`UPDATE ... SET status='expired' WHERE code = ? AND status='approved' RETURNING kilo_user_id`); the token is minted only by the caller that won the swap, and losers normalize to `expired`.
- **Finding 2 — token minted after the user is blocked.** Authorization was only checked at approval time; the later poll minted a long-lived API token with no re-check. A user could approve while in good standing, get blocked, then poll and still receive a fresh token. The poll now re-validates the user at mint time via `blocked_reason` and `isUserBlacklistedByDomain`, returning `denied` instead of minting. Because the approval is consumed first, a blocked user can't retry the code into a token.

No schema change, no migration. The poll API route already maps `denied` -> 403 and `expired` -> 410, so no route change was needed.

## Verification

- `apps/web/src/lib/device-auth/device-auth.test.ts` (24 tests pass), including two new cases:
  - concurrent polls on one approval mint **exactly one** token (the other gets `expired`);
  - a user blocked between approval and poll gets `denied` with no token, and a retry returns `expired`.
- Existing single-use / pending / denied / expired poll tests still pass.
- The domain-blacklist branch shares the same `if` as the blocked-reason check; I did not add a dedicated test for it because `getBlacklistedDomains` is Redis-cached (60s) and not reliably seedable in a unit test.

## Visual Changes

N/A

## Reviewer Notes

- The blocked re-check returns `denied` while the row is set to `expired` by the consume; both are terminal/non-retryable, and `denied` (403) is the more informative signal to the polling device.
- Two atomic commits: Finding 1 (atomic consume) then Finding 2 (blocked/blacklist re-check).

### DIFF
--- a/apps/web/src/lib/device-auth/device-auth.ts
+++ b/apps/web/src/lib/device-auth/device-auth.ts
@@ -3,6 +3,7 @@ import { db } from '@/lib/drizzle';
 import { device_auth_requests, kilocode_users } from '@kilocode/db/schema';
 import { eq, and, lt, sql } from 'drizzle-orm';
 import { generateApiToken } from '@/lib/tokens';
+import { isUserBlacklistedByDomain } from '@/lib/user/server';
 import { randomInt } from 'node:crypto';
 
 const CODE_LENGTH = 8;
@@ -201,6 +202,14 @@ export async function pollDeviceAuthRequest(code: string): Promise<{
 
   if (!user) {
     throw new Error('User not found');
+  }
+
+  // Re-check authorization at mint time. Authorization is verified when the
+  // user approves the request, but they may have been blocked or blacklisted
+  // between approval and this poll. The approval is already consumed above, so
+  // a now-blocked user cannot retry this code into a fresh long-lived token.
+  if (user.blocked_reason || (await isUserBlacklistedByDomain(user))) {
+    return { status: 'denied' };
   }
 
   const token = generateApiToken(user, { deviceAuthRequestCode: code });
--- a/apps/web/src/lib/device-auth/device-auth.ts
+++ b/apps/web/src/lib/device-auth/device-auth.ts
@@ -176,11 +176,27 @@ export async function pollDeviceAuthRequest(code: string): Promise<{
     return { status: request.status as 'pending' | 'denied' };
   }
 
-  // For approved requests, fetch user and generate token
+  // Atomically consume the approval before minting: a single guarded
+  // compare-and-swap flips 'approved' -> 'expired' and returns the row only to
+  // the caller that won. This prevents concurrent polls from each minting a
+  // long-lived token from one approval.
+  const consumed = await db
+    .update(device_auth_requests)
+    .set({ status: 'expired' })
+    .where(and(eq(device_auth_requests.code, code), eq(device_auth_requests.status, 'approved')))
+    .returning({ kiloUserId: device_auth_requests.kilo_user_id });
+
+  if (consumed.length === 0 || !consumed[0].kiloUserId) {
+    // Lost the race to another poller (or the status changed) — normalize to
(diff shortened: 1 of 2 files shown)


apps/web/src/lib/device-auth/device-auth.ts
```diff
@@ -176,26 +177,42 @@ export async function pollDeviceAuthRequest(code: string): Promise<{
...
   if (!user) {
     throw new Error('User not found');
   }
 
-  const token = generateApiToken(user, { deviceAuthRequestCode: code });
+  // Re-check authorization at mint time. Authorization is verified when the
+  // user approves the request, but they may have been blocked or blacklisted
+  // between approval and this poll. The approval is already consumed above, so
+  // a now-blocked user cannot retry this code into a fresh long-lived token.
+  if (user.blocked_reason || (await isUserBlacklistedByDomain(user))) {
```
St0rmz1, line 211: **Low, Security:** When this re-check trips, the function returns `{ status: 'denied' }` silently. The canonical auth path (`validateUserAuthorization` / `authError` in `lib/user/server.ts:1077`) emits an `AUTH-FAIL 403` warn line for blocked and blacklisted users, so those denials are visible in backend logs. This branch has no equivalent, so a blocked or blacklisted user who approved before being blocked and then polls (exactly the abuse case this branch defends against) leaves no server-side trace.

This is backend observability only; it does not change the response the device receives. Suggest logging before the return, matching the existing `AUTH-FAIL` format and distinguishing the two reasons, e.g.:

```ts
if (user.blocked_reason || (await isUserBlacklistedByDomain(user))) {
  const reason = user.blocked_reason ? 'blocked' : 'blacklisted-domain';
  console.warn(`AUTH-FAIL 403 (${kiloUserId}): device-auth poll denied (${reason})`);
  return { status: 'denied' };
}
```

apps/web/src/lib/device-auth/device-auth.ts
```diff
@@ -176,26 +177,42 @@ export async function pollDeviceAuthRequest(code: string): Promise<{
...
   }
 
-  // For approved requests, fetch user and generate token
+  // Atomically consume the approval before minting: a single guarded
+  // compare-and-swap flips 'approved' -> 'expired' and returns the row only to
+  // the caller that won. This prevents concurrent polls from each minting a
+  // long-lived token from one approval.
+  const consumed = await db
+    .update(device_auth_requests)
+    .set({ status: 'expired' })
```
St0rmz1, line 186: **Nit, Functional:** For the blocked/blacklisted case this row is set to `expired`, but `pollDeviceAuthRequest` returns `denied` (line 211), which the route maps to 403 while a consumed or race-lost code returns `expired` / 410. Two things worth a look:

1. The persisted row state (`expired`) does not match the response (`denied`), which can be confusing when debugging from the database.
2. More substantively, that distinct 403-vs-410 lets a polling client tell "you were blocked" apart from "code is gone." If you'd rather not signal the block to the device, returning `expired` here (still consumed, still terminal) would make the blocked case indistinguishable from a normal consumed code while the backend log from the comment above preserves the real reason for operators.

Not blocking; the current behavior is intentional per the description.

PR #4334, opened 2026-06-30
feat(kiloclaw) openclaw workspace download

St0rmz1:
## Summary

Adds an "Export OpenClaw Workspace" feature so a user can download their workspace text from a KiloClaw instance and move it to another OpenClaw host. The export covers the markdown that defines an assistant (persona, instructions, and the `memory/` tree) plus a generated list of installed skills. It deliberately leaves out the large binary skill and canvas payloads, along with credentials, channels, sessions, and config.

The flow spans the controller (builds the archive), the Durable Object and worker (proxy the bytes), and the web app (authenticated download route plus a new card in claw settings). Output is a `.tar.gz` or a `.zip`, and the `.zip` can be AES-256 password protected.

## Changes

- Controller: new `POST /_kilo/files/export-openclaw-workspace` endpoint and `openclaw-export.ts`. Walks `~/.openclaw/workspace`, collects every `.md` file (excluding `skills/` and `canvas/`), and appends a generated `INSTALLED_SKILLS.md` built from each skill's `SKILL.md` frontmatter. Enforces per-file, total-size, and file-count caps, skips symlinks, and canonicalizes the workspace root. Builds `.tar.gz` (tar-stream + gzip) or `.zip` (@zip.js/zip.js, optional AES-256).
- Capability gating: adds `files.export-openclaw-workspace` to the controller capability list so clients detect support instead of inferring it from the version.
- Worker and Durable Object: `exportOpenclawWorkspace` returns the archive bytes with metadata headers (file count, total bytes, skipped count). Old controllers without the route return `controller_route_unavailable`, and the UI hides the feature when the capability is absent.
- Worker error handling: allow-list `openclaw_export_*` codes in the platform error sanitizer so export error messages survive sanitization.
- Web: authenticated `POST /api/claw/openclaw-export` route handler that resolves the active personal or org instance, calls the internal client, and streams the archive as a download. New `OpenclawExportCard` (format selector, optional password) grouped with the existing import card under a new "Import & Export" settings section.
- Passphrase is sent in the POST body, used only to encrypt, and never logged.
- Spec: documents the endpoint, text-only scope, caps, metadata headers, and zip-only encryption in `.specs/kiloclaw-controller.md`.
- Adds `@zip.js/zip.js` and `tar-stream` to the controller package.

## Verification

- [ ] On a running instance with workspace content (incl. a skill), export as `.tar.gz`, extract, and confirm the markdown tree and `INSTALLED_SKILLS.md` are present and skills/canvas payloads are excluded.
- [ ] Export as a password-protected `.zip` and confirm it opens with the password in a standard archive tool.

## Visual Changes

| Before | After |
|---|---|
|  |  |

## Reviewer Notes

- Scope is intentionally text-only. Skill and canvas payloads can be multiple GB on 10 GB volumes, so the export carries markdown plus a skill inventory for reinstall, not the skill code itself.
- `TOOLS.md` is exported as-is and contains KiloClaw-specific tool instructions; the card warns these may not apply on another host.
- Encryption is zip-only; a password combined with `format: tar.gz` is rejected at the API.
- `SKILL.md` reads are size-capped so a large skill manifest cannot exhaust controller memory.


jeanduplessis (approved):
LGTM

PR #4355, opened 2026-07-01
docs(agents): consolidate AGENTS.md guidance into skills and rules

pandemicsyn:
## Summary

Clean up

Key moves:

- **Cloudflare Worker/DO conventions** duplicated across services consolidated into the repo-local `durable-objects` skill (`references/repo-conventions.md`): DO call retries (`withDORetry`), DO stub helpers, sub-module splitting, IO boundaries, DB clients. The module-scope DB-client-caching rule moved into `workers-best-practices`.
- **Coding standards** consolidated into existing and kinda duplicated `.kilo/rules/coding-style.md` (reconciling the `as`-operator rule to the nuanced version); per-package restatements removed. Stale `.kilocode/rules/...` path fixed.
- **GDPR/PII** deduped (kept once in `.kilo/rules/gdpr-pii.md`). Theres also a pre-existing ci check.
- **PR conventions** moved into a `/cloud-pr` command. We already had one a pr slash command at some point and I'm guessing most folks have their own workflow.
- **Spec index** moved into a new `specs` skill (rebuilt from disk — all `.specs/` except `template.md`); per-service spec pointers added to `kiloclaw`, `kiloclaw-billing`, `mcp-gateway`. 
- **Domain Context** rescoped to state `CONTEXT.md` covers only the Code Reviewer + Security Agent domains.
- CI-enforced Markdown Tables prose removed from root `AGENTS.md`. 
- Plus some additional trimming like the .plans instruction.

Root `AGENTS.md` shrank from 134 to ~75 lines.

## Visual Changes

N/A


AGENTS.md
```diff
@@ -21,14 +21,15 @@ scripts/          CI and one-off scripts
 
 ## Domain Context
 
-Before changing domain behavior, read `CONTEXT.md`.
-Use canonical terms from `CONTEXT.md` in code, docs, task descriptions, tests, and agent outputs.
-Do not introduce synonyms for existing concepts unless updating `CONTEXT.md` first.
-Do not duplicate the full context contract inside `AGENTS.md`.
+`CONTEXT.md` is the domain-language and ownership contract for the **Code Reviewer** and **Security Agent** domains only (review execution/analytics, Security Findings, Security Sync, notifications, remediation, and their email delivery — mainly `apps/web/src/lib/{code-reviews,security-agent}/`, `services/security-sync/`). It does not cover other areas of the monorepo.
+
+When working in those domains, read `CONTEXT.md` first and use its canonical terms in code, docs, task descriptions, tests, and agent outputs. Do not introduce synonyms for its concepts without updating `CONTEXT.md` first, and do not duplicate the full contract inside `AGENTS.md`.
```
pandemicsyn, line 26: Previously its seems like we instructed the agent to load this all the time. But this doc actually seems like it only applies to Code/Security reviewer. I'm guessing because the agent confuses the terminology used by the two? 

This might work better as a skill or something? But  hopefully this cuts down on eager loading.

.agents/skills/durable-objects/references/repo-conventions.md
```diff
@@ -0,0 +1,84 @@
+# Repo Conventions — Worker/DO Services in This Monorepo
```
jeanduplessis, line 1: Comment/question, not review feedback: 

Something I ran into in the on-call repo was the split between where we put custom-authored skills and 3rd party skills, and does it even matter. For instance we tend to use `npx skills` as our "skill package manager" for 3rd party skills and install that into .agents/skills as its mostly supported across all agents.

In the on-call repo we had a mixture of custom and 3rd party skills and ended up with:
```
- .agents/skills → 3rd party skills
- .kilo/skills → custom skills
```

It feels duplicative though to have skills in two places and maybe we should just put it all in .agents/skills like it seems we're going for in this repo.

No strong opinions, just think it would be good to be consistent across all repos so we don't have to think about it.

.agents/skills/durable-objects/SKILL.md
```diff
@@ -34,6 +34,9 @@ Fetch the relevant doc page when implementing features.
 - `./references/rules.md` - Core rules, storage, concurrency, RPC, alarms
 - `./references/testing.md` - Vitest setup, unit/integration tests, alarm testing
 - `./references/workers.md` - Workers handlers, types, wrangler config, observability
+- `./references/repo-conventions.md` - This repo's Worker/DO conventions: DO call
```
jeanduplessis, line 37: Ok this is actual review feedback and here's where we should make a decision on how we want to handle 3rd party vs custom skills.

If we want to use `npx skills` as our "skill package manager" to easily be able to update skills to their latest version, then I think we should withhold from making any changes to those skills like we are doing here.

An alternative: No 3rd party skills in repo by default
- We could have a recommended list of skills that might be useful but require the user to install it separately.
- This puts extra burden on the developer to make sure they have the right "tools" in place when working on the repo, but one could argue that's how we've done things in the past with which IDE you use, etc.

@marius-kilocode might have some thoughts on this as I believe he favors a more bare repo with limited agent pre-configuration to allow for more flexibility in personal development workflows and control.

.agents/skills/specs/SKILL.md
```diff
@@ -0,0 +1,35 @@
+---
+name: specs
+description: Business-rule specs in .specs/ govern KiloClaw billing/lifecycle/controller/data model/Composio, MCP Gateway auth, model experiments, Security Agent, subscription center, team/enterprise seat billing, Impact affiliate/referrals, Kilo Pass, organization SSO, Stripe early fraud warnings, and coding plans. Load before making ANY change (bug fix, feature, refactor, or review) to a domain covered by one of these specs, to read the authoritative rules first.
```
jeanduplessis:
```suggestion
description: Business-rule specs for KiloClaw billing/lifecycle/controller/data model/Composio, MCP Gateway auth, model experiments, Security Agent, subscription center, team/enterprise seat billing, Impact affiliate/referrals, Kilo Pass, organization SSO, Stripe early fraud warnings, and coding plans. Load when you need context about the business requirements that guided the implementation.
```

This is a change @RSO and I discussed previously where we should be careful to have the agent think these specs are the only source of truth (as apposed to the current code implementation). We should also be careful to not have the agent load all of these ALL the time and have it get confused. It tended to load these too eagerly in the past. I'd be okay with the agent knowing about these specs, but would suggest we don't force them to be loaded all the time by the agent. The developer could deliberately load a spec into context if needed.

.agents/skills/specs/SKILL.md
```diff
@@ -0,0 +1,35 @@
...
+---
+
+# Business-Rule Specs
+
+Specs in `.specs/` are the authoritative source of truth for the business rules and
+invariants of the domains they cover. Before making **any** change to a covered
+domain — including bug fixes, new features, refactors, or reviews — you **must**
+first read the relevant spec. Implementation mechanics (route names, columns, retry
+cadence) belong in plans and code; the spec owns the rules and invariants.
```
jeanduplessis:
```suggestion
Specs in `.specs/` are context as to the original business intention, rules and
invariants of the domains they cover. Consult them for context and flag inconsistencies
to the user if instructions or changes will cause deviations from the original intent. 
```

.kilo/command/cloud-pr.md
```diff
@@ -0,0 +1,30 @@
+---
```
jeanduplessis, line 1: I think we might be good to delete this command. I imagine most folks have their own workflow for this and the repo template serves as a guide for external contributors.

services/kiloclaw-billing/AGENTS.md
```diff
@@ -4,6 +4,10 @@
 
 `services/kiloclaw-billing` owns the KiloClaw billing lifecycle worker.
 
+## Specs
+
+This service is governed by `.specs/kiloclaw-billing.md` and `.specs/kiloclaw-billing-lifecycle.md`. Read them (and load the `specs` skill) before changing billing behavior here.
```
jeanduplessis:
```suggestion
Load the `specs` skill to access the business rules and invariants of the KiloClaw billing service when you need context about the business requirements that guided the implementation. 
```

AGENTS.md
jeanduplessis: This is an artifact of the /grill-with-docs skill. 
It think it has long-term value, but it needs to be back-filled with context about the rest of the repo.
We might also want to consider whether we want this to live separately in each app/service/package and maybe just have a context map at the root of the repo?

RSO (approved):
I think this is an improvement. I do want to second some of @jeanduplessis's comments regarding the wording of the specs. If we tell the bot that the specs are the authoritative source of truth, it will be reluctant to suggest a change when my prompt conflicts with a spec. Usually, when this happens, the bot will twist itself into all kinds of weird corners to somehow figure out how to comply with my prompt while keeping the spec as is.

Personally, I'd much rather have the bot just make a change to the spec file when this happens, because that makes it very clear to me that what I'm asking conflicts with something we decided earlier. It also fits much more with our iterative approach to software development, as opposed to the waterfall-y approach of treating specs as set-in-stone.

.agents/skills/durable-objects/references/repo-conventions.md
pandemicsyn: i actually struggled with this more broadly as well - because i also have the actual official cloudflare skill + cloudflare mcp set up globally. 

I'm down for copying the setup in the on-call repo.

.agents/skills/durable-objects/SKILL.md
pandemicsyn: Mentioned this above as well but:

>i actually struggled with this more broadly as well - because i also have the actual official cloudflare skill + cloudflare mcp set up globally.

I'd be up for this.  I already have my own review agents, pr skills, etc. So this doesn't feel un-natural. 

Some of these custom additions for Cloudflare specifically could also go in REVIEWS.md. e.g.  insisting on withDORetry

.kilo/command/cloud-pr.md
pandemicsyn: Makes sense.

jeanduplessis (approved):

.kilo/command/cloud-pr.md
marius-kilocode: @jeanduplessis I very much agree.

services/kiloclaw-billing/AGENTS.md
```diff
@@ -4,6 +4,10 @@
 
 `services/kiloclaw-billing` owns the KiloClaw billing lifecycle worker.
 
+## Specs
+
+Load the `specs` skill to access the business rules and invariants of the KiloClaw billing service when you need context about the business requirements that guided the implementation.
```
marius-kilocode, line 9: Why to we have to force the model to load specs or even use the specs skill everytime. I want to decide when I load requirements into context, I don't expect this to be a default behavior. I find in general that we should remove that from all AGENTS.md files.

services/kiloclaw/AGENTS.md
```diff
@@ -4,6 +4,10 @@
 
 KiloClaw is a Cloudflare Worker that runs per-user OpenClaw AI assistant instances on Fly.io Machines. The CF Worker handles auth, config management, and proxies HTTP/WebSocket traffic to each user's Fly Machine via Fly Proxy.
 
+## Specs
+
+This service is governed by `.specs/kiloclaw-controller.md`, `.specs/kiloclaw-datamodel.md`, and `.specs/kiloclaw-composio.md`. Read the relevant spec (and load the `specs` skill) before changing controller/machine lifecycle, the instance/subscription data model, or Composio credential handling.
```
marius-kilocode, line 9: Specifications are no governing business rules and should not act as such. We can define business rules in context but should never enforce them to be as part of the agents rule IMO.

AGENTS.md
```diff
@@ -81,78 +56,14 @@ Schema is in `packages/db/src/schema.ts`. Migrations live in `packages/db/src/mi
 - **After a rebase that conflicts on migration files:** delete all migration files, snapshots, and journal entries that were added on the branch, then re-run `pnpm drizzle generate` to regenerate a clean migration from the current schema diff. Re-append any backfill SQL afterward.
 - Prefer a single migration per feature branch when the code has not yet been deployed to production. If multiple migrations accumulated during development, squash them by deleting all branch-local migrations and regenerating.
```
marius-kilocode, line 57: Could this be not part of a database skill?

AGENTS.md
```diff
@@ -81,78 +56,14 @@ Schema is in `packages/db/src/schema.ts`. Migrations live in `packages/db/src/mi
...
-## Plans
+## Stripe Subscription Schedules
 
-When writing implementation plans, always save them to the `.plans/` directory at the repo root. This is the designated location for all planning documents — do not place them elsewhere in the repo.
+When using `subscriptionSchedules.create()` with `from_subscription`, Stripe prohibits setting `metadata` in the same call (it copies metadata from the subscription automatically). Set custom metadata (e.g., `origin` tags) in the subsequent `subscriptionSchedules.update()` call instead.
 
 ## Git Safety
-
 - **Never** use `--force`, `--no-verify`, or any other flag that bypasses git hooks or safety checks without explicit user approval.
 - If a hook or check fails, diagnose the issue and either fix it or ask the user how to proceed — do not silently skip it.
```
marius-kilocode, line 69: Hooks are developer preference. Many disable them. So why should this rule apply for everyone?

PR #4274, opened 2026-06-26
feat(gateway): expose efficient auto routing models

johnnyeric:
## Summary

- Adds `autoRouting.models` metadata to the `kilo-auto/efficient` model in `/api/openrouter/models`.
- Populates the list from the published benchmark routing table using concrete model ids already present in the model response.
- Keeps other auto models unchanged because they are not benchmark-table backed.

## Verification

- Ran local web with a mock benchmark routing table and verified `curl -s http://localhost:3000/api/openrouter/models | jq '.data[] | select(.id=="kilo-auto/efficient" or .id=="kilo-auto/free") | {id, autoRouting}'` returned concrete `autoRouting.models`.

```
  {
    "id": "kilo-auto/efficient",
    "autoRouting": {
      "models": [
        "anthropic/claude-sonnet-4.6",
        "openai/gpt-5.5"
      ]
    }
  }
  {
    "id": "kilo-auto/free",
    "autoRouting": {
      "models": [
        "stepfun/step-3.7-flash:free"
      ]
    }
  }
```

- Comman

## Visual Changes

N/A

## Reviewer Notes

- `autoRouting.models` is omitted when no benchmark routing table is published or no concrete candidate ids are present.
- Virtual `kilo-auto/*` candidates are excluded from the exposed model list.

PR #4396, opened 2026-07-05
feat(models): expose auto-routing choices in models API

johnnyeric:
## Summary

Expose the concrete models behind the virtual **Auto Efficient** (`kilo-auto/efficient`) and **Auto Free** (`kilo-auto/free`) routing choices in the `/api/openrouter/models` response, so the model selector can show which models auto-routing picks among.

- Adds `addAutoRoutingModels()`, which annotates the auto models with `autoRouting.models` — the concrete candidate ids from the benchmark routing table (efficient) and free-model resolution (free), intersected with the models actually available to the caller.
- Applied on **both** paths: org (`getAvailableModelsForOrganization`) and public/BYOK (route handler).
- The benchmark routing table is cached in-process (5-min TTL, serves last-known-good on fetch failure via `createCachedFetch`), so it is not an uncached admin-worker round-trip per request and a transient worker outage does not blank the choices.
- Adds the **first direct test coverage** for `getAvailableModelsForOrganization` (previously only ever mocked).

This is the server-side counterpart that lets us drop the client-side merge fallback in `kilo-gateway`.

## Verification

Ran local web (`pnpm dev`) against a docker Redis seeded via `/api/cron/sync-providers` (935-model catalog), with a mock benchmark routing table served on the local worker port, same approach as #4274.

```
curl -s http://localhost:3000/api/openrouter/models
  | jq '.data[] | select(.id=="kilo-auto/efficient" or .id=="kilo-auto/free") | {id, autoRouting}'
```

```json
{
  "id": "kilo-auto/efficient",
  "autoRouting": { "models": ["anthropic/claude-opus-4.8", "openai/gpt-5.5"] }
}
{
  "id": "kilo-auto/free",
  "autoRouting": { "models": ["stepfun/step-3.7-flash:free"] }
}
```

## Visual Changes

N/A - server-side API change; the model-selector rendering already lives on `main` in the extension.

### DIFF
--- a/apps/web/src/app/api/openrouter/models/route.test.ts
+++ b/apps/web/src/app/api/openrouter/models/route.test.ts
@@ -1,15 +1,7 @@
 import { beforeEach, describe, expect, test } from '@jest/globals';
 import { NextRequest } from 'next/server';
 import type { OpenRouterModel } from '@/lib/organizations/organization-types';
-import { getEnhancedOpenRouterModels } from '@/lib/ai-gateway/providers/openrouter';
-import { getUserFromAuth } from '@/lib/user/server';
-import { getDirectByokModelsForUser } from '@/lib/ai-gateway/providers/direct-byok';
-import { listAvailableExperimentModels } from '@/lib/ai-gateway/experiments/list-available-experiment-models';
-import { addUserByokAvailability, getUserByokProviderIds } from '@/lib/ai-gateway/byok';
-import { getAvailableModelsForOrganization } from '@/lib/organizations/organization-models';
 import { GET } from './route';
-import { getBenchmarkRoutingTable } from '@/lib/ai-gateway/auto-routing-benchmark-admin-client';
-import { getAutoFreeCandidates } from '@/lib/ai-gateway/auto-model/resolution';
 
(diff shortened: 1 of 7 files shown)


chrarnoldus:
My guess is you also have to add these changes here for the organization-specific endpoint: https://github.com/Kilo-Org/cloud/blob/1acc95868f5c9d76f605405583ed2d4274d73700/apps/web/src/routers/organizations/organization-settings-router.ts#L165

but wouldn't it make more sense to add the extra metadata the first time the auto model definitions are created? https://github.com/Kilo-Org/cloud/blob/2f13d9636bd4abc3d2f7809a141143ffc2e2d96b/apps/web/src/lib/ai-gateway/providers/openrouter/index.ts#L36

johnnyeric:
Good catch, added to the organization endpoint too (that was the original point, but missed after refactoring).

On the second part, I looked into it and there's an issue: org/feature filtering happens later and only drops top-level model ids - it doesn't touch the nested autoRouting.models. Adding the choices at definition time would let the org endpoint list candidates that were filtered out for that org (deny-listed models would show up as choices routing can never pick). Added a test locking this invariant in.

Let me know if I am missing something.

chrarnoldus (approved):

PR #4411, opened 2026-07-06
feat(api): add organization members endpoint

evanjacobson:
## Summary

Public API consumers did not have a documented way to list organization members and pending invites, which made external organization administration workflows depend on private app surfaces.

This adds a public `GetMembers` endpoint for accessible organizations and documents it in the generated OpenAPI surface. The implementation follows the existing v1 route, API-token authorization, response serialization, and OpenAPI registry patterns used by the rest of the public API.

<details>
<summary>Details</summary>

- Returns active members and pending invitations for an organization the caller can access.
- Omits invite tokens and invite URLs from the public response.
- Keeps this PR scoped to the members endpoint and docs only; usage analytics changes are not included.

</details>

## Verification

- [ ] No manual UI verification; this is an API-only change.

## Visual Changes

N/A

## Reviewer Notes

<details>
<summary>Automated checks run locally</summary>

- `pnpm --filter web test --runTestsByPath 'src/app/api/v1/organizations/[id]/members/route.test.ts'`
- `pnpm --filter web test -- 'src/app/api/docs/route.test.ts' src/lib/openapi/trpc-openapi.test.ts`
- `pnpm format`

</details>


apps/web/src/lib/organizations/organization-types.ts
```diff
@@ -116,6 +141,28 @@ type ActiveMember = {
...
+  InvitedOrganizationMemberSchema,
+]);
+
+export const PublicInvitedOrganizationMemberSchema = InvitedOrganizationMemberSchema.omit({
+  inviteToken: true,
+  inviteId: true,
+  inviteUrl: true,
+});
+
+export const PublicOrganizationMemberSchema = z.discriminatedUnion('status', [
```
St0rmz1, line 162: **Low, Security:** This public schema strips the invite token, invite id, and invite URL, which is exactly right. One thing worth a second look: because `PublicOrganizationMemberSchema` reuses `ActiveOrganizationMemberSchema` as-is, the public response still includes `dailyUsageLimitUsd` and `currentDailyUsageUsd` for every active member (see the fields at lines 126-127).

Why it matters: `listPublic` is gated by `organizationMemberProcedure`, so any member of the org (not just an owner or billing manager) can call it with an API token and read every colleague's daily spend limit and how much they have spent today. This matches what the internal `organizations.withMembers` procedure already exposes to the same audience, so it is not a regression. But this PR promotes those fields onto the documented public API surface, and the description says usage analytics is out of scope, so it is worth confirming this is intentional rather than inherited by accident.

Suggestion: if per-member usage and limits should not be part of the public contract, add them to the `.omit()` set (or build a dedicated public active-member schema without them) so the payload matches the stated scope. If they are intended, no change needed; a one-line comment noting the deliberate inclusion would help future readers.

apps/web/src/routers/organizations/organization-members-router.ts
```diff
@@ -77,6 +80,21 @@ async function getDirectOrganizationRole(
 }
 
 export const organizationsMembersRouter = createTRPCRouter({
+  listPublic: organizationMemberProcedure
+    .input(OrganizationIdInputSchema)
+    .output(PublicOrganizationMembersSchema)
+    .query(async ({ input }) => {
+      const organization = await getOrganizationById(input.organizationId);
```
St0rmz1: **Low, Functional:** The `organizationMemberProcedure` middleware runs `ensureOrganizationAccess` before this resolver executes. For a non-admin caller, that middleware throws `UNAUTHORIZED` (401) whenever no membership row exists, which includes the case where the organization does not exist at all. That means this `getOrganizationById` check and the `NOT_FOUND` it throws are only ever reachable by admins, who are granted the `owner` role without an existence check.

Why it matters: a regular API-token consumer coding against this endpoint will never observe the 404 for a missing organization; they will get a 401 instead. The generated OpenAPI docs now advertise a 404 for this route (see `trpc-registry.ts`), so external consumers may write handling for a status code they will not receive. It is also a redundant database round-trip on the common path, since membership already guarantees the organization exists for non-admins.

Suggestion: either drop the `getOrganizationById` lookup and the documented 404 (rely on the membership check that already runs), or keep the check but document the 404 as an admin-only outcome so the contract matches observable behavior. This is safe either way; it is mainly about matching the documented contract to reality.

apps/web/src/lib/openapi/trpc-registry.ts
```diff
@@ -78,4 +84,17 @@ export const publicTrpcOpenApiProcedures = [
...
-];
+  {
+    procedurePath: 'organizations.members.listPublic',
+    method: 'get',
+    tags: ['Organizations'],
+    summary: 'Return organization members',
+    description:
+      'Returns active and invited members for an organization the authenticated user can access. Invite tokens and invite URLs are omitted from the response.',
+    errorResponses: {
+      '404': 'Resource not found',
```
St0rmz1: **Low, Documentation:** This opts the `listPublic` route into a documented 404 response. As noted on the resolver, the 404 is only reachable by admin callers; a regular API-token consumer requesting a missing or inaccessible organization gets a 401 from the membership middleware, never this 404.

Why it matters: the OpenAPI document is the contract external consumers build against. Advertising a 404 that public callers cannot actually trigger is misleading.

Suggestion: reconcile this with whatever you decide for the resolver check. If the redundant lookup is removed, drop this 404 entry too; if it stays, consider a description that makes the admin-only nature clear.

St0rmz1 (approved):

PR #4442, opened 2026-07-07
feat(cloud-agent-next): cap tool subprocess memory via cgroup slices

eshurakov:
## Summary

Confines tool subprocesses (and, separately, the kilo server) to dedicated cgroup v2 memory slices in the cloud-agent sandbox container, so a runaway tool (`tsc` on a large monorepo, a parallel build) is OOM-killed in seconds with a visible error instead of pinning the whole container at its memory ceiling — where cache-reclaim thrash freezes every process for hours without ever triggering the kernel OOM killer.

**New wrapper module `wrapper/src/tool-cgroup.ts`** (with `src/shared/tool-cgroup-env.ts` for shared env keys) creates two sibling cgroups at the cgroup-fs root and runs a periodic async sweep over `/proc` that classifies strict descendants of the wrapper into tool work vs. the kilo-server chain, migrating each into its slice:

- `kilo-tools` — `memory.oom.group=1`, so the whole tool tree dies atomically and a multi-process build can't escape the cap by splitting memory across workers.
- `kilo-server` — `memory.oom.group=0` (hardcoded), so a fresh tool child that inherits this slice for up to one sweep interval can't take the server down with it; the kernel kills the biggest resident instead.

Modes (`TOOL_CGROUP_MODE`): `off` (default), `observe` (migrate + report usage, no cap), `enforce` (`memory.max = MemTotal − reserve`). Each slice also takes an optional `cpu.weight` so a parallel build pinning every core can no longer starve the kilo server's event loop. `mode=off` neutralizes stale caps left by a previous wrapper generation in a reused container.

**Rollout gating:** the Worker passes `TOOL_CGROUP_*` knobs through to the wrapper only for orgs listed in `TOOL_CGROUP_ORG_IDS` (`*` = all), defaulting to `""` (off for everyone). The wrapper-client filters keys to an allowlist and validates shell entries before interpolating them into both the process env and the inlined shell command.

**Architectural changes:**

- The kilo server is now explicitly treated as a separate child process (the `main.ts` comment is corrected from "in-process" to spawned via `createKilo()`/cross-spawn), enabling independent OOM-kill and respawn.
- New wrapper-side periodic `/proc` sweeper (async, `unref`'d timers) on the single event loop.
- New `isKiloServerUnreachableError` + deduped `restartKiloRuntime` (W5): when a prompt/command fails because the kilo server process is gone (ECONNREFUSED/RESET/EPIPE, "fetch failed"), the wrapper fire-and-forget respawns the server so the next attempt hits a live server instead of repeating the same connection failure. Application-level errors (bad model, unknown command) do not trigger a restart.

`/health` now reports an optional `toolCgroup` snapshot. The feature degrades gracefully: a read-only cgroup fs (e.g. devcontainers/local dev) returns null and the wrapper runs exactly as before this existed.

## Verification

- Deployed to a test Cloudflare sandbox environment and confirmed that running `typecheck` in a small container in the cloud repo now kills the tool process with an OOM (the target incident class) instead of freezing the container.
- [ ]

## Visual Changes

N/A

## Reviewer Notes

- Feature ships **off by default** (`TOOL_CGROUP_ORG_IDS=""`); rollout is per-org. Review the `TOOL_CGROUP_ORG_IDS`/`isOrgInList` gating in `cloudflare-agent-sandbox.ts`.
- Process classification in `isKiloServerProcess` relies on an exact `kilo`/`.kilo` basename immediately followed by `serve`; a future kilo re-exec with a different entrypoint basename would miss classification — see `KILO_SERVER_BASENAMES`.
- `restartKiloRuntime` is deduped via `inFlightRuntimeRestart` to keep concurrent failed requests from racing to respawn the server.
- Design rationale doc `MEMORY_CGROUPS_PLAN.md` is referenced in comments (W2/W4/W5 markers) but is not present in the repo — confirm whether it should be added or lives elsewhere.
- No UI changes; the only externally visible surface is the optional `toolCgroup` field on the wrapper `/health` response.


services/cloud-agent-next/wrapper/src/main.ts
```diff
@@ -277,6 +287,27 @@ async function main() {
     materializePromptAttachments,
     onSessionBound: (feedPolicy: SessionBoundFeedPolicy) =>
       globalFeedManager.onSessionBound(feedPolicy),
+    toolCgroupHealth: () => toolCgroup?.health() ?? null,
+    // Deduped: concurrent failed requests must share one restart rather than each
+    // respawning the kilo server and racing to mutate closeKiloServer/kiloClient.
+    // No restarts during shutdown — a server spawned after handleShutdown snapshots
+    // activeRuntimeStartups would never be closed and leak past the wrapper's exit.
+    restartKiloRuntime: (): Promise<void> => {
```
RSO, line 295: > A server that reliably dies on startup (or is being repeatedly OOM-killed because a server limit was set too low) yields one-restart-per-failed-request indefinitely.

RSO (approved):
Change seems fine, though I do wonder whether it wouldn't make more sense to implement some of this stuff on the CLI side, as it seems like we're opening ourselves up to potential bugs later on in future CLI versions where we accidentally classify non-tool processes as tool processes and kill them.

services/cloud-agent-next/wrapper/src/main.ts
eshurakov: kilo server won't be capped by default, only processes that it starts

eshurakov:
@RSO yes makes sense! but as a first step it would be nice to test it with cloud agent only

PR #4445, opened 2026-07-08
feat(session-ingest): push notifications for remote session attention signals

eshurakov:
## Summary

Session-ingest now sends a mobile push when a remote root session needs attention: the assistant finishes a turn (`status -> idle`) or the session is waiting for user input (`status -> question` / `permission`). This gives VS Code and Agent Manager users a signal to return to a remotely controllable session.

Key behavior:
- `SessionIngestDO.ingest()` emits attention signals alongside metadata changes. Completed signals include an excerpt from the latest completed assistant message; needs-input signals use a fixed prompt.
- `queue-consumer` collects signals across ingest chunks, persists metadata first, and dispatches pushes best-effort through the notifications service. Signals from already-committed chunks are also dispatched after a later chunk fails, because retries will not re-emit those transitions.
- Pushes are eligible only for remote-controllable root sessions with a live CLI heartbeat. Recent local CLI activity suppresses the push.
- The web app registers visibility-gated presence for the canonical Kilo session. The notifications service suppresses pushes when that session is actively open in a visible web tab.
- Existing Cloud Agent push payloads and `completed` status are reused for compatibility with deployed notification schemas.

## Verification

- [x] Manually verified with the CLI and local Cloud Agent, including that no push is sent while the session is focused in the web app.
- [x] `pnpm --filter notifications test -- src/lib/notifications-service-cloud-agent.test.ts`
- [x] `pnpm --filter notifications typecheck`
- [x] `pnpm --filter notifications lint`
- [x] `pnpm --filter cloudflare-session-ingest typecheck`
- [x] `pnpm --filter cloudflare-session-ingest lint`
- [x] `pnpm --filter @kilocode/event-service test -- src/__tests__/presence.test.ts`
- [x] `pnpm --filter @kilocode/event-service typecheck`
- [x] `pnpm --filter web typecheck`
- [x] `git diff --check`

## Visual Changes

N/A

## Reviewer Notes

- First-status guard: an idle transition only produces a completed signal after the DO has already recorded a status, preventing historic backfills from notifying about old turns.
- The completed-message lookup is bounded to the 50 newest message rows and skips R2-offloaded rows.
- Web foreground suppression uses presence rather than websocket subscriptions because subscriptions are not visibility-gated.
- Push dispatch is non-fatal: failures are logged and never block queue acknowledgement or retry behavior.


services/notifications/src/dos/NotificationChannelDO.ts
```diff
@@ -35,6 +35,10 @@ const ACCEPTED_BOOKKEEPING_RETRY_DELAY_MS = 30_000;
 
 export class NotificationChannelDO extends DurableObject<Env> {
   async dispatchPush(input: DispatchPushInput): Promise<DispatchPushOutcome> {
+    return this.dispatchPushCore(input);
```
iscekic: Not needed?

services/notifications/src/lib/cloud-agent-session-push.ts
```diff
@@ -39,14 +39,23 @@ async function dispatchSessionPush(
...
 
+  const log = (outcome: string) =>
+    console.log('Cloud agent session notification', {
+      userId,
+      cliSessionId,
+      outcome,
+    });
+
   if (!session) {
+    log('missing_session');
```
iscekic: Leftover logs?

services/session-ingest/src/dos/SessionIngestDO.ts
```diff
@@ -343,6 +389,59 @@ export class SessionIngestDO extends DurableObject<Env> {
...
+          lt(ingestItems.item_id, range.end)
+        )
+      )
+      .orderBy(ingestItems.ingested_at, ingestItems.id)
+      .all();
+
+    // Parts offloaded to R2 (oversized items) store '{}' inline; skip them rather
+    // than fetching from R2 — this is a best-effort excerpt, not a full transcript.
+    return buildAssistantExcerpt(
+      rows.filter(row => !row.item_data_r2_key).map(row => row.item_data)
```
iscekic: Should we filter at the query level?

services/session-ingest/src/dos/SessionIngestDO.ts
```diff
@@ -343,6 +389,59 @@ export class SessionIngestDO extends DurableObject<Env> {
...
+        item_data: ingestItems.item_data,
+        item_data_r2_key: ingestItems.item_data_r2_key,
+      })
+      .from(ingestItems)
+      .where(eq(ingestItems.item_type, 'message'))
+      .orderBy(desc(ingestItems.ingested_at), desc(ingestItems.id))
+      .limit(COMPLETED_MESSAGE_SCAN_LIMIT)
+      .all();
+    for (const row of rows) {
+      if (row.item_data_r2_key) continue;
```
iscekic: Filter at query level?

services/session-ingest/src/queue-consumer.ts
```diff
@@ -260,6 +300,94 @@ function createIngestChunker(
...
+  if (signals.length === 0) return;
+  const parentSessionId = mergedChanges.has('parentId')
+    ? (mergedChanges.get('parentId') ?? null)
+    : sessionRow.parent_session_id;
+  if (
+    !isEligibleForRemoteSessionAttention({
+      parentSessionId,
+    })
+  ) {
+    console.log('Skipping attention signal dispatch (ineligible session)', {
```
iscekic: Raw console log, here and below

iscekic (approved):
Mostly nits, LGTM

PR #4453, opened 2026-07-08
feat(web): hide personal surface for users with disabled personal account

markijbema:
## What

Builds on #4447 (`personal_account_disabled`). When a user's personal account is disabled — i.e. the login exists only because they were invited to an org — we stop surfacing the personal-account UI and default them into their organization instead.

Approach: **we don't link these users to personal pages**, but we don't hard-block them either. Direct navigation to a personal route still works and renders the page (with the personal sidebar), it's just no longer reachable through our own navigation.

## Changes

- **Default landing** (`getProfileRedirectPath`): users with a disabled personal account are always sent into an organization (their oldest, deterministic ordering) regardless of org count; falls back to `/connected-accounts` if they somehow belong to no orgs. This is a default-destination choice, not a block.
- **Org switcher**: hides the "Personal" entry for these users.
- **Footer user menu** (`SidebarUserFooter`): the avatar/name row becomes a dropdown containing **Connected Accounts** + **Sign out** — the home for user-global identity surfaces, shown in both the personal and org sidebars.
- **Personal sidebar**: drops the Connected Accounts item (now in the footer for everyone).
- **Sidebar selection** (`AppSidebar`): on `/connected-accounts` (the one personal route we still link to, from the footer), these users get their default org's sidebar instead of the personal one. Every other personal route falls through to the normal personal sidebar so the page renders.

## What this intentionally does NOT do

- No catch-all server-side redirect guard blocking personal routes. Direct navigation to `/profile`, `/subscriptions`, `/cost-insights`, etc. is allowed and renders normally; those routes are simply not linked for these users.
- The org menu stays strictly org-scoped and untouched — Connected Accounts lives in the footer, not the org Account group.

## Design notes

- Connected Accounts is genuinely *user-global* (primary email + linked auth providers), not personal-account-scoped, which is why it moves to the footer user menu.
- Terminology is kept tied to the `personal_account_disabled` field (no "invited-only" synonym) so the concept stays greppable.

## Testing

- Added `getProfileRedirectPath` tests: disabled personal account with multiple orgs (→ an org) and with zero orgs (→ `/connected-accounts`).
- `pnpm --filter web typecheck`, `server.test.ts` (56/56), `pnpm --filter web lint` (clean), `pnpm format`.

### DIFF
--- a/apps/web/src/app/(app)/components/AppSidebar.tsx
+++ b/apps/web/src/app/(app)/components/AppSidebar.tsx
@@ -54,10 +54,10 @@ export default function AppSidebar(props: React.ComponentProps<typeof Sidebar>) 
   const searchParams = useSearchParams();
   const setupStep = searchParams.get('step');
   const { open, setOpenMobile, setOpenTransient } = useSidebar();
-  const isInvitedOnly = Boolean(user?.personal_account_disabled);
+  const personalAccountDisabled = Boolean(user?.personal_account_disabled);
   const { data: organizations, isPending: isOrganizationsPending } = useQuery(
     trpc.organizations.list.queryOptions(undefined, {
-      enabled: isInvitedOnly && !currentOrgId,
+      enabled: personalAccountDisabled && !currentOrgId,
       trpc: { context: { skipBatch: true } },
     })
   );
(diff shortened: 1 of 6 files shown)


apps/web/src/app/(app)/components/AppSidebar.tsx
```diff
@@ -44,11 +47,36 @@ function extractOrgWastelandId(pathname: string): { orgId: string; wastelandId:
...
   const searchParams = useSearchParams();
   const setupStep = searchParams.get('step');
   const { open, setOpenMobile, setOpenTransient } = useSidebar();
+  const personalAccountDisabled = Boolean(user?.personal_account_disabled);
+  // The only personal route we still link to for these users (footer user menu).
+  // For that route we keep them in their org sidebar instead of the personal one;
+  // other personal routes are not linked but remain accessible with the personal
+  // sidebar if reached directly.
+  const isLinkedAccountRoute =
+    pathname === '/connected-accounts' || pathname.startsWith('/connected-accounts/');
```
RSO: I think we should also include `/install` and `/learn` and move those to the new dropdown as well

RSO (approved):

apps/web/src/app/(app)/components/AppSidebar.tsx
markijbema: Done in 4d788a2 — moved Install and Learn into the footer user menu alongside Connected Accounts, and removed the Start group from the personal sidebar. They're now available in both the personal and org sidebars.

PR #4460, opened 2026-07-08
feat(web): show model-specific data policies

chrarnoldus:
## Summary
- preserve model-route training and prompt-retention policies from the OpenRouter frontend endpoint
- show model-specific `Trains` and `Retains prompts` values in model and provider detail tables
- retain provider-level policy badges and filters as provider defaults, with a note that individual models can differ
- fall back to provider policy for snapshots created before endpoint policy fields were stored

## OpenRouter evidence
Amazon Bedrock reports `retainsPrompts: false` at the provider level, while its `anthropic/claude-fable-5` route reports `endpoint.data_policy.retainsPrompts: true` and `retentionDays: 30`. The endpoint payload also exposes model-route `training`.

## Validation
- `pnpm format`
- `pnpm --filter @kilocode/db typecheck`
- changed-file web lint: 0 warnings/errors
- standalone model-policy parser/helper assertions
- `git diff --check`

Full web typecheck/lint exceeded the 120-second sandbox limit. The focused Jest test reached repository global setup but could not run because Postgres/Docker is unavailable in this environment.


chrarnoldus:
CleanShot 2026-07-09 at 08 54 45@2x

marius-kilocode (approved):

PR #4469, opened 2026-07-09
feat(cloud-agent-next): consume Kilo opaque session capabilities

eshurakov:
### Why

Cloud Agent sandboxes currently receive the raw, long-lived Kilo user token in plaintext across many carriers — `KILOCODE_TOKEN`, `KILO_AUTH_CONTENT`, `auth.json`, provider `apiKey`/`kilocodeToken`, and the wrapper-to-Worker `workerAuthToken`. Any compromise or accidental logging inside the untrusted sandbox exposes a reusable user credential. This PR contains that token by issuing a short-lived opaque `kka1.*` Kilo capability per dispatch and redeeming it per-request at the existing managed-SCM outbound interceptor, so contained sandboxes never see the raw token.

### What was done

- Issue one opaque `kka1.*` Kilo capability per wrapper dispatch (via `git-token-service`) when the session uses managed SCM containment, and replace every sandbox-visible carrier of the raw token (`KILOCODE_TOKEN`, `KILO_AUTH_CONTENT`, `auth.json`, provider `apiKey`/`kilocodeToken`/`baseURL`) with that capability. The raw token now lives only in trusted Worker/DO/broker metadata.
- Extend the managed-SCM outbound interceptor (`sandbox-outbound.ts`) to classify and redeem `kka1.*` capabilities through a new route-aware `redeemKiloSessionCapability` broker policy, reusing the same per-container egress interception already used for GitHub/GitLab capabilities. The capability carries normalized targets so runtime policy cannot drift.
- Replace the raw Kilo JWT used as `workerAuthToken` with a fenced `wrapper_dispatch_ticket` JWT (4h lifetime) carrying the dispatch fence (cloud-agent session + `kiloSessionId` + `wrapperRunId`/`wrapperGeneration`/`wrapperConnectionId`). Wrapper-owned routes (`/ingest`, `/kilo-global-ingest`, `/logs`) validate the ticket and reject fence mismatches; a legacy raw-Kilo-JWT path keeps already-bound wrappers working until their next dispatch.
- Authorize the session-ingest bootstrap: a bounded clone-based parse of `POST /api/session` extracts the bootstrap `sessionId` and permits it only when it matches the capability-bound Kilo session. Existing exact-path ingest routes (`/api/session/<id>/{export,import,ingest}`) continue to cover historical full-sync and incremental batches.
- Wrapper `auth.json` is now written from the materialized `KILOCODE_TOKEN` capability and refreshed on every ready request; the log uploader reads the current dispatch ticket lazily (it outlives any single ticket) and aborts active uploads before swapping uploaders.
- Fallback: uncontained sessions and DIND sandboxes (which don't set HTTPS interception) fall back to the real token and authenticate directly.
- Presentation boundary: `MessageBubble` renders the Worker-published sanitized string assistant errors instead of throwing when the error is a plain string.

### High-level architecture

```mermaid
sequenceDiagram
  participant DO as SessionService (DO, trusted)
  participant GTS as git-token-service (trusted broker)
  participant SB as Sandbox wrapper/SDK (untrusted)
  participant INT as Outbound interceptor (egress)
  participant UP as Kilo upstream (backend/provider/ingest)

  rect rgb(220,240,220)
  Note over DO,GTS: Trusted: raw user token never leaves here
  DO->>GTS: issueKiloSessionCapability(userToken, targets, containerId)
  GTS-->>DO: opaque kka1.* capability
  end

  DO->>SB: KILOCODE_TOKEN / KILO_AUTH_CONTENT / auth.json = capability<br/>workerAuthToken = wrapper_dispatch_ticket (fenced)
  SB->>INT: Kilo SDK outbound (Bearer kka1.*)
  INT->>GTS: redeemKiloSessionCapability(capability, containerId, method, url, bootstrapSessionId?)
  GTS-->>INT: real authorization + routeClass
  INT->>UP: forward with real Authorization
  UP-->>INT: response
  INT-->>SB: response

  alt bootstrap POST /api/session
    INT->>INT: parse body.sessionId, permit only if == capability-bound session
  end
  alt uncontained / DIND sandbox
    DO->>SB: KILOCODE_TOKEN = raw userToken (direct egress, no interception)
  end
```

Two distinct credentials are introduced, on purpose:

- **`kka1.*` Kilo capability** — sandbox → Kilo upstream egress, redeemed per-request at the outbound interceptor (the headline containment change).
- **`wrapper_dispatch_ticket` JWT** — wrapper → Worker routes only (`/ingest`, `/kilo-global-ingest`, `/logs`), minted with `NEXTAUTH_SECRET` and carrying the dispatch fence so a stolen ticket cannot be replayed against a different session/connection.

### Architecture decision

**Decision:** Use one opaque `kka1.*` Kilo capability everywhere the sandbox previously received the raw Kilo token, issued once per dispatch and redeemed per-request through the existing route-aware managed-SCM broker.

**Context:** The sandbox is an untrusted execution environment that must present a Kilo credential to call backend API, provider-model, and session-ingest routes. The raw user token is long-lived and crossed many plaintext carriers inside the sandbox. The repo already has a managed-SCM containment + per-container outbound-interception mechanism for GitHub/GitLab capabilities, gated by `MANAGED_SCM_CONTAINMENT_ORG_IDS`.

**Rationale:** Reusing the existing outbound interception and the `git-token-service` broker (with `SCM_SESSION_CAPABILITY_ENCRYPTION_KEY`) keeps the raw token in trusted metadata only, leaves the Kilo CLI unchanged, and adds no new secret. A per-dispatch capability with a 4h lifetime (intentionally no background refresh — a new trusted dispatch mints a fresh one) bounds exposure, and the route-aware allowlist narrows what a leaked capability can reach.

**Alternatives considered:**

- **Short-lived raw Kilo token rotation inside the sandbox.** Rejected because the raw token would still be plaintext and reusable for its lifetime inside the sandbox, and it establishes no route-aware allowlist.

**Consequences:** Contained sessions get strong token containment but depend on the outbound interceptor being able to intercept egress (ports 80/443); DIND sandboxes and uncontained orgs fall back to the raw token. The wrapper-to-Worker auth surface changes from a raw Kilo JWT to a fenced dispatch ticket, so already-running wrappers rely on a legacy-JWT path until their next dispatch. Route classification is strict: the provider `baseURL` baked into the capability targets must match what the sandbox actually calls, or valid capabilities are rejected as `upstream_not_allowed`.

## Verification

- [x] Standard and DIND outbound probes against a real contained sandbox: capability redemption, HTTPS interception, response streaming, request cancellation, CA propagation, and egress cleanup.
- [x] Deployed test-Worker smoke: a fresh contained `kg` session reported `OPAQUE_CAPABILITY_PRESENT`, and Logpush confirmed `kka1` redemption by the containment container with 200 responses on backend API, organization-model discovery, and provider-model routes.

Session-history persistence through the ingest bootstrap path was discovered broken during that smoke (Kilo's import bootstrap emits `POST /api/session`, which the capability policy rejected as `upstream_not_allowed`) and fixed in this commit. It has unit/integration coverage but has **not** been re-verified end-to-end against a deployed Worker — that is the main open verification item.

## Visual Changes

Minor: `MessageBubble` now renders the Worker-published sanitized string assistant errors instead of throwing when the error field is a plain string. No screenshot; this is a fallback rendering path for error strings already produced by the Worker sanitization boundary.

## Reviewer Notes

- **Open verification:** the ingest bootstrap fix needs a deployed re-test to confirm session history is persisted to ingest. The redemption/forwarding paths above were verified before that fix and are unaffected by it.
- **Rollout order:** requires the `git-token-service` deployment that implements `issueKiloSessionCapability` / `redeemKiloSessionCapability`. Containment is gated by `MANAGED_SCM_CONTAINMENT_ORG_IDS`; the `wrangler.jsonc` change setting it to `*` is the local `env.dev` block only (for branch testing), not production.
- **Compatibility:** a legacy raw-Kilo-JWT `workerAuthToken` path keeps already-bound wrappers working until their next dispatch; DIND sandboxes and uncontained orgs fall back to the raw token and authenticate directly.
- **Strict route matching:** the sandbox's provider `baseURL` must match the capability's baked-in targets, or valid capabilities are rejected as `upstream_not_allowed`.
- **Debug flag:** `LOG_REJECTED_KILO_URLS=1` is local-debug-only and logs full rejected URLs (which may contain tokens); it must never be set on a deployed Worker.
- Package unit, integration, wrapper, typecheck, lint, and formatting checks pass.


services/cloud-agent-next/src/session-service.ts
```diff
@@ -1343,9 +1346,10 @@ export class SessionService {
       SESSION_ID: sessionId,
       SESSION_HOME: sessionHome,
       [PNPM_STORE_ENV_VAR]: PNPM_STORE_DIR,
-      // Inject Kilocode credentials (with override support)
-      KILOCODE_TOKEN: kilocodeToken,
-      KILO_AUTH_CONTENT: JSON.stringify({ kilo: { type: 'api', key: originalToken } }),
+      // Opaque Kilo capability — redeemed for the real credential at the outbound interceptor
+      KILOCODE_TOKEN: kiloCapability,
```
pandemicsyn, line 1360: Must-fix: This capability is exposed to sandbox code, but redemption currently treats almost every path under `backendBaseUrl` as `backend_api` and substitutes the full user bearer token. The sandbox can therefore call a credential-minting route such as `POST /api/organizations/:id/user-tokens` and receive an exportable organization JWT, defeating credential containment. Replace the backend catch-all with an explicit method-and-path allowlist for the Kilo runtime operations that are actually required.

services/cloud-agent-next/src/session-service.ts
```diff
@@ -1811,6 +1822,68 @@ export class SessionService {
...
+      );
+    }
+    // The provider base URL baked into the capability's targets MUST match the
+    // baseURL the sandbox actually calls — otherwise the outbound interceptor's
+    // route classification rejects the request as `upstream_not_allowed`, even
+    // though the capability itself is valid. Never let getSaferEnvVars derive
+    // this independently from env.KILO_OPENROUTER_BASE.
+    return {
+      capability: issued.value.capability,
+      providerBaseUrl: derivedTargets.targets.providerBaseUrl,
```
pandemicsyn, line 1895: Must-fix: Only the derived provider target is returned to runtime construction. `sessionIngestBaseUrl` is normalized from `localhost` to `host.docker.internal` in the capability, while `KILO_SESSION_INGEST_URL` remains the original `localhost` value. The interceptor compares origins exactly, so local contained restore/import/ingest calls fail with `upstream_not_allowed`. Return and materialize the exact derived ingest target for normal and devcontainer runtime environments too.

services/cloud-agent-next/src/server.ts
```diff
@@ -432,19 +517,34 @@ app.put(
...
       return c.text(authResult.error, 401);
     }
-    if (authResult.userId !== userId) {
+    if (authResult.claims.userId !== userId) {
       return c.text('Token does not match session user', 403);
     }
 
+    const kiloSessionId = new URL(c.req.url).searchParams.get('kiloSessionId');
+    if (!kiloSessionId) {
+      return c.text('Missing kiloSessionId parameter', 400);
```
pandemicsyn, line 530: Must-fix: This unconditional parameter requirement breaks the legacy-wrapper compatibility path added by this PR. Already-running wrappers authenticate with a legacy Kilo JWT but use the old log-upload URL without `kiloSessionId`, so they now receive 400. For `legacy_kilo_token` claims only, accept the old URL shape and resolve the authoritative Kilo session from current session access; keep the query/claim check for dispatch tickets.

services/cloud-agent-next/src/session-service.ts
```diff
@@ -1811,6 +1822,68 @@ export class SessionService {
...
+      cloudAgentSessionId: params.cloudAgentSessionId,
+      kiloSessionId: params.kiloSessionId,
+      outboundContainerId: getOutboundContainerId(env, params.sandboxId, {
+        managedScmContainment: true,
+      }),
+      userToken: params.userToken,
+      targets: derivedTargets.targets,
+    });
+    if (!issued.success) {
+      throw ExecutionError.invalidRequest(
```
pandemicsyn: Must-fix: Every issuance failure is converted to non-retryable `INVALID_REQUEST`, including `rpc_error` returned when `git-token-service` is temporarily unavailable. That can terminalize queued work on a transient service outage. Preserve `INVALID_REQUEST` for permanent target/configuration failures, but map RPC and availability failures to a retryable execution error.

pandemicsyn (approved):

services/cloud-agent-next/src/session-service.ts
eshurakov: Fixed in 775c382e3: `issueKiloSessionCapability` now returns `sessionIngestBaseUrl` and it is materialized into both the normal runtime env (`getSaferEnvVars`) and the devcontainer restore env (`getDevContainerRestoreEnv`), so contained restore/import/ingest calls match the interceptor origin baked into the capability rather than the raw localhost value. Covered by an updated assertion in `session-service.test.ts`.

services/cloud-agent-next/src/server.ts
eshurakov: Fixed in 775c382e3: the `kiloSessionId` query requirement is now scoped to `wrapper_dispatch_ticket` claims only. For legacy `legacy_kilo_token` uploads the authoritative `kiloSessionId` is resolved from `requireCurrentSessionAccess`, and a 400 is returned only if neither source provides one. Covered by a new test in `server.test.ts`.

services/cloud-agent-next/src/session-service.ts
eshurakov: Fixed in 775c382e3: transient issuance failures (`rpc_error`, `service_not_configured`, `temporarily_unavailable`) now map to retryable `ExecutionError.workspaceSetupFailed` via `isRetryableKiloCapabilityIssuanceFailure`; permanent target/config failures remain `invalidRequest`. Covered by a new test in `session-service.test.ts`.

PR #4480, opened 2026-07-09
feat(session-ingest): add feature-flagged direct ingest

pandemicsyn:
## Summary

- Add a feature-flagged, off-by-default direct ingest path for conservative `POST /api/session/:sessionId/ingest` payloads so eligible requests can persist through one synchronous `SessionIngestDO.ingest()` RPC before returning.
- Add fail-closed rollout configuration, deterministic user bucketing, bounded request buffering, strict buffered validation, direct fallback through existing R2 staging and queue enqueue, and terminal structured ingest events.
- Preserve the legacy queue path for gate misses, missing or invalid `Content-Length`, oversized/offloaded or multi-chunk envelopes, and direct RPC failures.
- Add Workers-runtime coverage for the maximum direct RPC envelope and regenerate Worker bindings for the checked-in off-by-default config.
- Switch session-ingest Wrangler type generation to `--include-runtime=false`, matching services such as Gastown so checked-in binding diffs exclude bundled runtime declarations.

CLI change is in https://github.com/Kilo-Org/kilocode/pull/12117
Little bit of prior work to break out stuff is in https://github.com/Kilo-Org/cloud/pull/4472

## Verification

Local manual e2e (dev only, gate enabled for a single allowlisted dev user via `.dev.vars`): tested the direct path against a local `wrangler dev`. Auth token and `kiloUserId` redacted.

```
# 1) create session
POST /api/session
-> 200 {"id":"ses_86f8…","ingestPath":"/api/session/ses_86f8…/ingest"}

# 2) ingest with ?v=2 (3 valid items: session / message / part)
POST /api/session/ses_86f8…/ingest?v=2
-> 200 {"success":true}

# 3) export immediately after the POST (read-your-writes)
GET /api/session/ses_86f8…/export
-> 200 {"info":{"title":"local direct test"},
        "messages":[{"info":{"id":"msg_1","role":"assistant"},
                     "parts":[{"id":"prt_1","messageID":"msg_1","text":"hello from direct path"}]}],
        "sessionDiff":[]}
```

Structured terminal event logged by the worker for that ingest:

```
{
  event: 'direct_ingest_ok',
  ingestRequestId: '0c1fa152-…',
  sessionId: 'ses_86f8…',
  ingestVersion: 2,
  declaredBytes: 216,
  actualBytes: 216,
  durationMs: 8,
  items: 3,
  metadataChanges: 1
}
```

What this confirms:
- Direct path taken (`direct_ingest_ok`, not `direct_ingest_legacy`): one synchronous DO RPC, no R2 staging or queue send on the happy path.
- `ingestVersion: 2` accepted and recorded — validates the paired kilocode `?v=2` client bump end-to-end.
- Read-your-writes: the export immediately after the POST returns the title, message, and part.
- `declaredBytes === actualBytes` (216): the bounded counting reader measured the body exactly.
- `metadataChanges: 1`: the session title projected to Postgres (best-effort, off the response path).

Prod behavior is unchanged: the gate ships off (`DIRECT_INGEST_PERCENT=0`, empty allowlist); the allowlist entry used above exists only in a local gitignored `.dev.vars`.

## Visual Changes

N/A

## Reviewer Notes

- This PR is stacked on #4472 and should stay based on `florian/chore/session-ingest-direct-prep` until PR 1 merges.
- Rollout defaults to off: `DIRECT_INGEST_PERCENT=0`, empty allowlist, and a 4 MiB hard cap. Invalid config closes the gate.
- Direct RPC uses one attempt only; ambiguous DO failures fall back durably via the existing staging helper and log `direct_ingest_fallback`.
- Direct happy path intentionally performs no R2 staging, no queue send, and no permanent item R2 writes.
- `worker-configuration.d.ts` now contains only project bindings; runtime types come from `@cloudflare/workers-types` in `tsconfig.json`. (matches gas town, and should help cut down on future churn i think)

## wtf is this actually doing?

Session ingest is how the CLI/cloud agent sends session data to cloud, including session metadata, messages, parts, diffs, and status changes. 

With the feature flag off, which is the default, requests follow the existing durable path: stage the request body in R2, enqueue a queue message, then let the queue consumer validate, chunk, and write items into the per-session Durable Object. 

When the feature is enabled for a user and the request is small enough, the route validates the buffered body and writes it to the Durable Object once before returning 200, so reads immediately after the POST can see the new data. Anything outside that conservative envelope falls back to the existing R2 plus queue path.

```mermaid
flowchart TD
  A[POST session ingest] --> B[Validate session access]
  B --> C{Feature gate and envelope eligible?}
  C -- No --> L[Legacy: stage body in R2]
  L --> Q[Queue message]
  Q --> W[Consumer validates and chunks]
  W --> D[SessionIngestDO persists items]
  C -- Yes --> V[Buffer <= 4 MiB and validate]
  V -- Malformed --> R400[400, no writes]
  V -- Empty or no valid items --> R200[200 no-op]
  V -- Offload or multi-chunk --> L
  V -- Eligible --> D2[One SessionIngestDO RPC]
  D2 -- Accepted --> M[Best-effort metadata update]
  D2 -- Tombstoned --> R404[404]
  D2 -- Throws --> L
  M --> OK[200]
```


St0rmz1:
**Low, Functional:** The direct path surfaces synchronous rejections that the legacy queue path did not. An under-declared `Content-Length` returns `413 payload_too_large` (`direct-ingest.ts`:160), malformed JSON returns `400 malformed_json` (:170), and a tombstoned session returns `404 session_not_found` (:252). On the legacy path these cases effectively always returned `200 {success:true}`, with malformed or deleted-session writes resolved asynchronously by the consumer.

For selected users this changes what the ingest client can observe. If the CLI/cloud-agent treats a non-200 ingest response as a hard error (aborting the sync or surfacing an error), behavior differs from today for that cohort. Blast radius is limited by the allowlist and 0% default, so this is mainly a rollout gate: please confirm the ingest client tolerates 400/413/404 before widening the percentage.

services/session-ingest/src/ingest/direct-ingest.ts
```diff
@@ -0,0 +1,406 @@
...
+  );
+  if (!buffered.ok) {
+    logEvent('warn', {
+      event: 'direct_ingest_reject',
+      ...eventBase(request, startedAt, {
+        declaredBytes: contentLength,
+        actualBytes: null,
+        items: null,
+      }),
+      reason: 'declared_bytes_exceeded',
```
St0rmz1, line 171: **Nit, Functional:** This log hardcodes `reason: 'declared_bytes_exceeded'`, but `readBoundedStream` can return `limit: 'configured_cap'` as well as `'declared_bytes'`. It is unreachable today because the route already rejects `contentLength > maxBytes` earlier as `oversized_body`, so the label is accurate for now. If that earlier guard ever moves, this telemetry would silently mislabel a cap overflow. Consider deriving the reason from `buffered.limit` so the log stays correct regardless of the upstream guards.

St0rmz1 (approved):

services/session-ingest/src/ingest/direct-ingest.ts
pandemicsyn: fixed in https://github.com/Kilo-Org/cloud/pull/4480/commits/649eabd537e134fd06f16cc0c7693d4ead8bc2ca

iscekic (approved):
LGTM, bot found these, feel free to disregard 🤖

Reviewed the diff vs the stacked base. `worker-configuration.d.ts` excluded (generated). Nothing blocking — the happy path, fail-closed config, one-attempt RPC + durable fallback, and default-off rollout all check out.

### Standards (all judgement calls)

1. **Dead dual-limit path — `bounded-stream-reader.ts`.** The `configuredCap` param is dead for the sole caller: `direct-ingest.ts` already rejects `contentLength > config.maxBytes` before calling, and `parseDirectIngestConfig` caps `maxBytes <= INGEST_CHUNK_MAX_BYTES`, so `declaredBytes <= configuredCap` always. The `totalBytes > configuredCap` branch and `limit: totalBytes > configuredCap ? 'configured_cap' : 'declared_bytes'` can never report `configured_cap` in prod. Suggest collapsing to a single `declaredBytes` bound and dropping `configured_cap` from the result union + downstream reject mapping.
2. **Duplicated unsigned-int contract.** `/^(0|[1-9]\d*)$/` + `Number.isSafeInteger` is implemented twice — `parseUnsignedInteger` (rollout) and `parseContentLength` (direct-ingest). Could share one helper.
3. **Data clump.** `(request, startedAt, metrics)` thread together through `legacy`/`fallbackAfterDirectFailure`/`logFallback`/`eventBase` — candidate for one bundle.
4. **Minor:** `logEvent(level, event) { console[level](event) }` is pure delegation but sits next to direct `console.error`/`console.warn` in the same file — inconsistent. `legacy()` names the destination, not the action (`fallbackToQueue()` reads better).

### Spec

1. **413 on under-declared bodies diverges from "oversized → legacy".** When the actual streamed body exceeds the declared `Content-Length`, `readBoundedStream` returns `too_large` and the handler returns 413 (`direct-ingest.ts:178`), dropping the payload. The pre-existing R2 path didn't enforce Content-Length and returned 200, so an under-declaring client that previously persisted now gets rejected — and this 413 outcome isn't in the decision flow. Worth confirming it's acceptable vs routing to legacy fallback.
2. **Telemetry richer than the spec's three-event list** (`direct_ingest_reject`/`parse_reject`/`noop`/`tombstone` etc.). Underlying outcomes are specced, so likely intentional — just flagging.
3. **Minor:** `Content-Length: 0` eligible request goes legacy (`empty_body`), creating an empty R2 object + queue message instead of a direct 200 no-op.

pandemicsyn:
Addressed the latest review feedback in `c8877d3c2`:

- Removed the unreachable `configuredCap` branch/result from `readBoundedStream`; the route already rejects declarations above the configured cap, so the reader now enforces only the declared byte count. Overflow cancellation/lock-release coverage remains.
- Confirmed the existing CLI/cloud-agent ingest client treats 400/404/413 as terminal non-retryable responses without aborting the session; this is documented as a pre-ramp invariant, and the gate remains off by default.
- Kept the richer terminal-event taxonomy intentionally for rollout observability.
- Kept `Content-Length: 0` on the legacy path to preserve pre-rollout behavior rather than introducing another selected-client response difference.
- Did not apply the unsigned-int/helper/data-clump/naming suggestions: they are stylistic, add cross-module churn, and do not improve this PR’s correctness.

Validation: `pnpm format`; 398 session-ingest tests; package typecheck; package lint; `git diff --check`.

services/session-ingest/src/ingest/direct-ingest.ts
```diff
@@ -0,0 +1,434 @@
...
+async function fallbackAfterDirectFailure(
+  request: DirectIngestRequest,
+  r2Key: string,
+  bytes: Uint8Array,
+  metrics: DirectIngestMetrics,
+  startedAt: number,
+  directError: unknown
+): Promise<DirectIngestResponse> {
+  try {
+    await stageAndEnqueue(request.env, directFallbackQueueParams(request, r2Key), bytes);
```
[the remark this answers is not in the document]
pandemicsyn: Valid known limitation, pre-existing in the queue retry path, outside this PR’s read-after-write guarantee. Direct items remain durable and immediately readable; metadata projection remains best effort. Track separately if fallback telemetry indicates material occurrence.

PR #4494, opened 2026-07-10
feat(code-reviewer) update object shape to handle future review types

St0rmz1:
## Summary

Plumbs a forward-shaped `reviewAgents` object end-to-end along the code-reviewer -> cloud-agent path so the pipeline can carry per-agent model selections. Today every review produces a single `standard` agent that mirrors the session's model and thinking effort; the extra shape (`reviewType`, `aggregationStrategy`, and additional `agents[]` entries) is carried but not yet consumed by execution, in preparation for an upcoming council (multi-agent) review mode.

## Changes

- Added `ReviewAgentSelection` and `ReviewAgentsConfig` types in `apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts` and mirrored them in `services/code-review-infra/src/types.ts` (on `CodeReview` and `CodeReviewRequest`).
- In `prepareReviewPayload`, introduced a single `standardModel` source (`config.model_slug || DEFAULT_CODE_REVIEW_MODEL`) used for both `sessionInput.model` and `reviewAgents.agents[0].model` so the two cannot drift, and attached `reviewAgents` to the returned payload.
- Threaded `reviewAgents` through the `POST /review` handler in `services/code-review-infra/src/index.ts` into the orchestrator's `start` params.
- Persisted `reviewAgents` in `CodeReviewOrchestrator` state and added a log line recording `reviewType`, agent count, and agent roles for observability ahead of council mode.
- Added a test asserting a standard review builds exactly one `standard` agent whose `model` and `thinkingEffort` match the session input.

## Verification

- [ ] No manual testing performed. The change is forward plumbing that is not yet consumed by execution (only `agents[0]`, which mirrors existing session behavior, is used today); covered by the added unit test.

## Visual Changes

N/A

## Reviewer Notes

- All new fields are optional, so the change is backward-compatible with in-flight reviews and older payloads.
- `reviewType`, `aggregationStrategy`, and any `agents[]` beyond `agents[0]` are intentionally unused for now and exist only to carry council-mode data.

### DIFF
--- a/apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
+++ b/apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
@@ -306,12 +306,15 @@ export async function prepareReviewPayload(
           }
         );
         const authToken = generateApiToken(user, { botId: 'reviewer' });
+        // Single source for the standard reviewer's model so the session input and the
+        // forward-shaped `reviewAgents[0]` can never drift apart.
+        const standardModel = config.model_slug || DEFAULT_CODE_REVIEW_MODEL;
         const sessionInput: SessionInput = {
           gitUrl: `https://bitbucket.org/${workspaceSlug.data}/${repositorySlug.data}.git`,
           kilocodeOrganizationId: owner.id,
           prompt,
           mode: DEFAULT_CODE_REVIEW_MODE as 'code',
-          model: config.model_slug || DEFAULT_CODE_REVIEW_MODEL,
+          model: standardModel,
           variant: config.thinking_effort ?? undefined,
           upstreamBranch: review.head_ref,
           platform: PLATFORM.BITBUCKET,
@@ -322,6 +325,20 @@ export async function prepareReviewPayload(
           bitbucketIntegrationId: integration.id,
           bitbucketPullRequestId: review.pr_number,
           bitbucketExpectedHeadSha: expectedHeadSha.data,
+        };
+
+        // Forward-shaped agent selections, built for every review (see GitHub/GitLab
+        // path below). Today this is always a single 'standard' agent mirroring the
+        // session's model/effort; council mode will populate one entry per specialist.
+        const reviewAgents: ReviewAgentsConfig = {
+          reviewType: 'standard',
+          agents: [
+            {
+              role: 'standard',
+              model: standardModel,
+              thinkingEffort: config.thinking_effort ?? null,
+            },
+          ],
         };
 
         logExceptInTest('[prepareReviewPayload] Prepared Bitbucket payload', {
@@ -340,6 +357,7 @@ export async function prepareReviewPayload(
           sessionInput,
           owner,
           repositorySize: null,
+          reviewAgents,
         };
       }
       case PLATFORM.GITHUB:
--- a/apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
+++ b/apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
@@ -22,6 +22,10 @@ import {
   fetchGitHubRepositorySize,
 } from '@/lib/integrations/platforms/github/adapter';
 import type { GitHubAppType } from '@/lib/integrations/platforms/github/app-selector';
+import type {
+  ReviewAgentSelection,
+  ReviewAgentsConfig,
+} from '@kilocode/worker-utils/review-agents';
 import {
   findKiloReviewNote,
   fetchMRInlineComments,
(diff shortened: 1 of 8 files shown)


apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
```diff
@@ -133,6 +162,12 @@ export type CodeReviewPayload = {
   previousCloudAgentSessionId?: string;
   /** Provider-reported repository storage size, formatted for log correlation. */
   repositorySize?: string | null;
+  /**
+   * Forward-shaped review agent selections. Built for every review (standard = a single
+   * `'standard'` agent). Only `agents[0]` is consumed today; the rest is plumbing for
+   * council mode.
+   */
+  reviewAgents?: ReviewAgentsConfig;
```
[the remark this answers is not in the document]
St0rmz1: Fixed in f3d3f0d. The Bitbucket branch now builds the same forward-shaped `reviewAgents` before its early return in `prepareReviewPayload`. It derives a single `standardModel` (`config.model_slug || DEFAULT_CODE_REVIEW_MODEL`) that feeds both `sessionInput.model` and `reviewAgents.agents[0].model`, so the two cannot drift, and attaches `reviewAgents` to the returned payload. Also extended the Bitbucket unit test to assert the `reviewAgents` shape and that `agents[0].model === sessionInput.model`.

services/code-review-infra/src/code-review-orchestrator.ts
```diff
@@ -920,8 +922,21 @@ export class CodeReviewOrchestrator extends DurableObject<Env> {
           ? undefined
           : params.previousCloudAgentSessionId,
       repositorySize: params.repositorySize,
+      reviewAgents: params.reviewAgents,
```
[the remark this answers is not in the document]
St0rmz1: Fixed in f3d3f0d. `retryFreshAfterInfraFailure` now forwards `reviewAgents: this.state.reviewAgents` into `retryStub.start(...)`, alongside `repositorySize`, so a cross-attempt infra retry preserves the persisted selection (members/strategy) instead of dropping it.

services/code-review-infra/src/index.ts
```diff
@@ -103,6 +103,7 @@ app.post('/review', async (c: Context<HonoEnv>) => {
         skipBalanceCheck: body.skipBalanceCheck,
         previousCloudAgentSessionId: body.previousCloudAgentSessionId,
         repositorySize: body.repositorySize,
+        reviewAgents: body.reviewAgents,
```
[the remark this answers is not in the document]
St0rmz1: Addressed in f3d3f0d, then hardened further in the latest push. We opted for a defensive-logging approach rather than full HTTP-boundary validation, to stay consistent with the endpoint's existing untyped `c.req.json()` boundary. The root invariant is that nothing between `saveState()` and `setAlarm()` in `start()` may throw. The observability log there is now total: it no longer assumes `reviewAgents.agents` is an array, and it reads roles with `agent?.role`, so a malformed authenticated payload (`{"reviewAgents":{}}`, non-array `agents`, or null entries like `{"agents":[null]}`) can never throw and strand a queued review.

apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
```diff
@@ -122,6 +122,35 @@ export type SessionInput = {
...
+
+/**
+ * Review agent configuration sent along the code-reviewer -> cloud-agent path.
+ *
+ * NOTE (forward plumbing): today the pipeline only builds and consumes a single
+ * `'standard'` agent (`agents[0]`). `reviewType`, `aggregationStrategy`, and additional
+ * `agents[]` entries are carried end-to-end for the imminent council mode but are not yet
+ * consumed by execution. See the standard/council fork in prepare-review-payload.
+ */
+export type ReviewAgentsConfig = {
```
pandemicsyn: Suggested/nit: This cross-service wire contract is duplicated here and in `services/code-review-infra/src/types.ts`, so council-mode additions can drift independently. Define a shared Zod schema in `@kilocode/worker-utils` and infer/import both producer and consumer types from it.

apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
pandemicsyn: Suggested/nit: `prepareReviewPayload` now constructs `reviewAgents` on every return path, but this optional type allows a future producer branch to silently omit it. Make the field required on `CodeReviewPayload`; keep it optional only on the Worker request and persisted state for rolling-deployment and in-flight compatibility.

services/code-review-infra/src/code-review-orchestrator.ts
```diff
@@ -1025,6 +1044,7 @@ export class CodeReviewOrchestrator extends DurableObject<Env> {
       skipBalanceCheck: this.state.skipBalanceCheck,
       previousCloudAgentSessionId: undefined,
       repositorySize: this.state.repositorySize,
+      reviewAgents: this.state.reviewAgents,
```
pandemicsyn, line 1047: Suggested/nit: This retry propagation has no integration assertion, so it can regress while payload-construction tests remain green. Extend the existing `retry-fresh` Worker test to seed `reviewAgents` and assert that the retry Durable Object preserves it.

pandemicsyn (approved):

The remaining comments are non-blocking contract and coverage improvements. The previously identified runtime-validation concern is not repeated because it is already covered by the resolved thread at `services/code-review-infra/src/index.ts:106`.

apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
St0rmz1: Done in 420aa77. Extracted the wire contract into a single shared module `@kilocode/worker-utils/review-agents` that defines Zod schemas (`ReviewAgentSelectionSchema`, `ReviewAgentsConfigSchema`) and infers the types from them. Both sides now import from it: `apps/web/.../prepare-review-payload.ts` and `services/code-review-infra/src/types.ts` (which re-exports for existing `./types` importers). The two hand-maintained copies are gone, so council-mode additions can no longer drift independently.

apps/web/src/lib/code-reviews/triggers/prepare-review-payload.ts
St0rmz1: Done in 420aa77. `reviewAgents` is now required on `CodeReviewPayload` (the producer output), so the compiler enforces that every `prepareReviewPayload` return path sets it and a future branch cannot silently omit it. It stays optional on the Worker request (`CodeReviewRequest`) and persisted state (`CodeReview`) for rolling-deployment and in-flight compatibility, as suggested.

services/code-review-infra/src/code-review-orchestrator.ts
St0rmz1: Done in 420aa77. Extended the existing `retry-fresh starts a new retry attempt...` Worker integration test to seed `reviewAgents` on the failed attempt state and assert the fresh retry Durable Object persists it, alongside the existing `repositorySize` assertion. The propagation is now covered at the integration level, so it can no longer regress while payload-construction tests stay green.

PR #4501, opened 2026-07-10
feat(kilo-pass): add Kilo Pass for Organizations

jeanduplessis:
## Intent

Kilo Pass for Organizations needs an entitlement model that cannot be represented safely by personal Kilo Pass subscriptions.

An organization purchase must cover every paid seat, distribute pooled Credit capacity between the parent and direct child organizations, preserve purchased terms, and remain correct when Stripe events arrive late or more than once. It must also recover from allocation conflicts and interrupted issuance without duplicating Credits.

This change is successful when:

- purchased pass capacity always matches paid seat count;
- every purchased pass resolves to the parent or a direct child;
- a paid service window is issued exactly once;
- allocation and term changes do not rewrite existing issuance;
- qualifying spend unlocks each bonus at most once;
- owners and billing managers can understand and repair blocked states;
- regular members cannot access agreement or allocation controls.

## Solution

The implementation introduces a Kilo Pass org agreement as the source of truth for organization entitlement. Personal Kilo Pass remains separate.

The main design choices are:

- Stripe confirms payment and carries the seat add-on, but does not define entitlement terms.
- Immutable term versions preserve the commercial rules used by each issuance.
- Versioned allocation plans assign capacity to direct children; the remainder always belongs to the parent.
- Immutable issuance snapshots preserve the allocation, amounts, and bonus threshold for each service window.
- Transactional processing and stable identities prevent duplicate grants during retries.
- Blocked and overallocated agreements fail closed and retain an explicit repair path.
- The organization subscription page owns purchase, allocation, cancellation, and recovery controls.

ADR 0003 records the full source-of-truth and lifecycle decisions.

## Risk and rollout

This is a high-risk change because it affects billing, organization authorization, persistent data, Credit issuance, and automated processing.

The main risks are:

- duplicate or missing Credit grants from replayed or delayed Stripe events;
- seat and pass quantities drifting apart;
- incorrect spend attribution between parent and child organizations;
- unauthorized access to agreement and allocation controls;
- an overallocated or failed agreement becoming impossible to repair;
- rolling back application code after the additive migration has been applied.

## Validation 

The following user journeys were verified to be correct:
- Purchasing the subscription from the organization subscription page.
- Paid-invoice activation.
- Future pass allocation updates.
- Paid seat increase and prorated parent supplement.
- Seat decrease into overallocated state, reconciliation, and retry.
- Cancellation through the paid-through boundary.

### Visual evidence

| Surface              |  Screenshot  |
| --------------- | ------------- |
| Organisation dashboard | CleanShot 2026-07-30 at 12 43 25@2x|
| Subscriptions page | Helium 2026-07-30 13 16 25 |
| Kilo Pass signup | Helium 2026-07-30 13 16 38|
| Kilo Pass Subscription  | Helium 2026-07-30 12 48 27 |
| Distribute passes | Helium 2026-07-30 12 48 55 |
| Upcoming changes | CleanShot X 2026-07-30 12 52 05 |

## Durable decisions

Human reviewers should pay particular attention to these decisions:
- Purchased pass capacity always equals paid seat count.
- Direct-child allocations are explicit; all remaining capacity defaults to the parent.
- Paid Stripe invoices authorize entitlement; browser returns do not. Waits for webhook confirmation.
- Current issuance snapshots are immutable.
- Allocation and term changes normally apply at future commercial or issuance boundaries.
- Delayed processing recreates the original service window instead of shifting it.
- Refunds and disputes suspend future processing without automatically clawing back pooled Credits.


apps/web/src/lib/kilo-pass-org/service.ts
```diff
@@ -0,0 +1,1805 @@
...
+    const paidUntil =
+      existing.paid_until && asDate(existing.paid_until) > input.paidUntil
+        ? asDate(existing.paid_until)
+        : input.paidUntil;
+    const pendingCapacityEffectiveAt = existing.next_capacity_effective_at
+      ? asDate(existing.next_capacity_effective_at)
+      : null;
+    const appliesPendingCapacity =
+      pendingCapacityEffectiveAt !== null && input.paidFrom >= pendingCapacityEffectiveAt;
+    const replacesPendingCapacity = input.paidSeatCount > existing.purchased_pass_capacity;
```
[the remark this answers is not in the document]
jeanduplessis: Verified the full seat-change path before changing this. A decrease is represented by lowering the same Stripe seat and org-pass item quantities with proration disabled; a later increase updates those same provider quantities immediately. Therefore the confirmed increase supersedes the pending decrease, and retaining the old internal next capacity would diverge from Stripe at renewal. I kept the behavior, added an explanatory comment, and added a regression test covering 25 -> scheduled 10 -> paid increase 30, asserting the pending decrease is cleared. Commit: a97453bf9.

apps/web/src/lib/ai-gateway/processUsage.ts
```diff
@@ -484,6 +483,17 @@ async function insertUsageTransaction(
       coreUsageFields,
       metadataFields
     );
+    if (coreUsageFields.organization_id && coreUsageFields.cost > 0) {
```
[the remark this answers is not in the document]
jeanduplessis: Verified and fixed in a97453bf9. Negative organization rows now call mutateOrganizationUsage in the insertion transaction, restoring organizations.microdollars_used, microdollars_balance, and organization_user_usage while remaining excluded from Kilo Pass qualifying spend. Added a regression test for +500 followed by -200, asserting the organization and member daily usage settle at 300.

apps/web/src/components/subscriptions/org-kilo-pass/view-shared.tsx
```diff
@@ -0,0 +1,318 @@
...
+                      {allocation.passCount} passes ·{' '}
+                      {formatOrgPassMoney(allocation.baseCreditsUsd ?? 0)} monthly Credits
+                    </span>
+                    {allocation.hasProratedCredits ? (
+                      <InfoTooltip
+                        label={`About ${allocation.organizationName}'s prorated Credit amount`}
+                        content={`This amount includes prorated Credits for passes added partway through the billing cycle. Prorated Credits are based on the days remaining in the current period. A full monthly period includes ${formatOrgPassMoney(fullMonthlyCreditsPerPassUsd)} in base Credits per pass.`}
+                      />
+                    ) : null}
+                    {allocation.supplementCreditsUsd
```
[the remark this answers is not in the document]
jeanduplessis: Verified: the API exposes only aggregated base Credits plus hasProratedCredits, and no production mapper could populate supplementCreditsUsd. Removed the unused field and unreachable rendering branch in a97453bf9; the existing prorated-credit tooltip remains the supported presentation.

iscekic (commented):
I didn't have time to take a look at this today. In the meantime here are some bot findings, feel free to discard any if irrelevant.

apps/web/src/components/subscriptions/org-kilo-pass/OrgKiloPassDetailView.tsx
```diff
@@ -0,0 +1,502 @@
...
+  const hasChildOrganizations = nextAllocations.some(allocation => allocation.kind === 'child');
+  const hasUpcomingAssignmentChanges =
+    hasCurrentIssuance &&
+    hasFutureDistribution &&
+    haveAssignmentChanges(currentAllocations, nextAllocations, nextTotalPasses);
+  const dialogAllocations = editingAllocations ?? nextAllocations;
+  const dialogChildTotal = getDirectChildTotal(dialogAllocations);
+  const dialogParentPasses = Math.max(0, nextTotalPasses - dialogChildTotal);
+  const dialogReductionRequired = Math.max(0, dialogChildTotal - nextTotalPasses);
+  const isDialogOverallocated = dialogReductionRequired > 0 || condition?.kind === 'overallocated';
```
iscekic: **P1 — Overallocated reconciliation deadlocks in this dialog.**

`isDialogOverallocated` ORs `dialogReductionRequired > 0` with `condition?.kind === 'overallocated'`, and Save is `disabled={isOverallocated || ...}`. When the agreement's server-side `processingCondition` is `overallocated` — the exact state this dialog exists to repair — the stale server condition keeps `isDialogOverallocated` true **even after the user edits child allocations down to a valid total** (`dialogReductionRequired` becomes 0). The condition only clears after a successful save + invalidation, but Save can never be clicked.

There is no alternate UI path: the ConditionAlert action opens this same dialog, and "Reset pass assignments" restores the current (over-total) allocation, which the server correctly rejects. The service layer handles reconciliation fine (`updateAllocation` validates against `next_purchased_pass_capacity ?? purchased_pass_capacity` and reclaims blocked runs — service tests cover it), so this is purely a view bug, but it blocks the ADR-mandated repair path and the PR's own validated journey ("seat decrease into overallocated state, reconciliation, and retry").

Likely fix: derive the disabled state only from `dialogReductionRequired > 0`; the server already rejects over-assignment. Note the existing view tests can't catch this — they use `renderToStaticMarkup`, and the Radix `Dialog` portals, so dialog content never renders in the test environment.

apps/web/src/lib/kilo-pass-org/stripe-adapter.ts
```diff
@@ -0,0 +1,426 @@
...
+    .map(item => ({ price: item.price.id, quantity: item.quantity ?? 1 }));
+  const existingSchedule = subscription.schedule;
+  const schedule =
+    typeof existingSchedule === 'string'
+      ? await stripe.subscriptionSchedules.retrieve(existingSchedule)
+      : existingSchedule;
+  const hasRemovalScheduled = schedule?.phases.some(phase =>
+    phase.items.every(item => item.price !== passItem.price.id)
+  );
+  if (hasRemovalScheduled) return;
```
iscekic: **P1 — Canceling Kilo Pass silently no-ops while a billing-cycle change is scheduled; the customer is charged again and the cancellation silently reverses.**

The seat billing-cycle change flow also uses a subscription schedule, and its phase 2 swaps the pass item's price to the new cadence (`getStripePriceIdForKiloPass(tier, targetCycle)`). A cycle change always changes cadence, so the phase-2 pass price always differs from `passItem.price.id`, making `phase.items.every(item => item.price !== passItem.price.id)` **true** — `hasRemovalScheduled` misfires and this function returns without scheduling anything.

Sequence: cycle change scheduled → user cancels Kilo Pass → Stripe untouched → `organizationKiloPassService.cancel` still sets DB `cancel_at_period_end` (UI shows "Cancellation scheduled") → at renewal the cycle change applies *and* the pass item keeps billing at the new cadence price → `invoice.paid` → `activatePaidAgreement` unconditionally sets `state = Active` and issues credits. The user is charged for a cancelled product and the cancellation undoes itself.

(The reverse ordering fails safe but confusingly: a pending pass-cancellation schedule blocks cycle changes with "A billing cycle change is already scheduled" — misleading message.)

Suggest tagging the cancellation schedule with `metadata.origin` like the cycle-change flow does, and detecting foreign schedules by origin/shape — either merge the pass removal into the existing phases or throw (like `resume`'s `SCHEDULE_REWRITE_UNSAFE`) instead of inferring schedule intent from item prices.

apps/web/src/routers/organizations/organization-kilo-pass-router.ts
```diff
@@ -0,0 +1,376 @@
...
+    throw new TRPCError({
+      code: 'CONFLICT',
+      message: 'Pass assignments were changed elsewhere. Refresh and try again.',
+    });
+  }
+  throw error;
+}
+
+export const organizationKiloPassRouter = createTRPCRouter({
+  usage: organizationMemberProcedure.output(UsageOutputSchema).query(async ({ input }) => {
```
iscekic: **P2 — Member-visible Kilo Pass data contradicts the domain contract added in this PR.**

`usage` is `organizationMemberProcedure` (any role), and `KiloPassUsageCard` renders on the org dashboard for all members, exposing pass counts, tier terms, and bonus progress (only the Manage button is role-gated). The router test "allows a member to read current Kilo Pass usage without billing access" enshrines this.

But the CONTEXT.md added in this PR says: "Do not expose pass counts, tiers, bonus rules… to regular members", and ADR 0003: "Regular members see generic organization Credit balance and transaction language without Kilo Pass labels, counts, tiers, or bonus mechanics." The implementation plan's validation journey #10 also lists member access denial / generic Credit visibility.

Code and contract disagree — worth an explicit decision: gate the endpoint + card to billing roles, or amend CONTEXT.md/ADR to bless member-visible usage.

apps/web/src/lib/kilo-pass-org/hierarchy-guard.ts
```diff
@@ -0,0 +1,64 @@
...
+    .innerJoin(
+      kilo_pass_org_allocation_plan_rows,
+      eq(kilo_pass_org_allocation_plan_rows.allocation_plan_id, kilo_pass_org_allocation_plans.id)
+    )
+    .where(
+      and(
+        eq(kilo_pass_org_agreements.parent_organization_id, organization.parentOrganizationId),
+        ne(kilo_pass_org_agreements.state, 'ended'),
+        eq(kilo_pass_org_allocation_plan_rows.allocation_container_organization_id, organizationId),
+        gt(kilo_pass_org_allocation_plan_rows.pass_capacity, 0)
```
iscekic, line 113: **P2 — This guard checks *all* historical allocation plans, so a child can never be released once it has ever had a nonzero allocation.**

Plans for past windows are retained (by design, for snapshot audit), so after `initial=5 → later plan=0`, this query still finds the old row and blocks detach/reparent/delete forever — contradicting CONTEXT.md's repair path ("must first set allocation to zero first; already granted pooled credits remain with child").

The new admin tests cover `initial=1` (blocked), `initial=0/future=1` (blocked), and `initial=0/future=0` (allowed), but never `initial=1, future=0` — the reconcile-then-detach scenario. If the strict reading (nonzero initial *or* future, forever) is intended, the contract's "set allocation to zero first" sentence is misleading; otherwise this should evaluate only the currently-effective plan (plus any pending future plan).

apps/web/src/lib/kilo-pass-org/service.ts
```diff
@@ -0,0 +1,1807 @@
...
+        agreementId: agreement.id,
+        effectiveAt: effectiveWindowStart,
+        version: latest.version + 1,
+        actorId: input.actorUserId,
+        allocations,
+      });
+      if (blockedRun) await reclaimBlockedProcessingRun(tx, blockedRun.id);
+      await tx
+        .update(kilo_pass_org_agreements)
+        .set({ processing_condition: KiloPassOrgProcessingCondition.Ready })
```
iscekic: **P3 — `updateAllocation` unconditionally resets `processing_condition` to `ready`.**

An ordinary allocation edit silently clears `suspended_for_review` (the dispute/refund payment-review suspension), bypassing the ADR's manual-review intent — and once manual legacy agreements exist, it would also flip `manual` to automated processing without the required explicit renewal transition. Consider clearing only `blocked`/`overallocated`/`failed` here. (`scheduleOrganizationPassCapacity` has the same overwrite behavior for `suspended_for_review`.)

apps/web/src/lib/kilo-pass-org/service.ts
```diff
@@ -0,0 +1,1807 @@
...
+              eq(kilo_pass_org_processing_runs.agreement_id, agreement.agreementId),
+              eq(kilo_pass_org_processing_runs.state, KiloPassOrgProcessingRunState.Running)
+            )
+          );
+        await tx
+          .update(kilo_pass_org_agreements)
+          .set({ processing_condition: KiloPassOrgProcessingCondition.Failed })
+          .where(eq(kilo_pass_org_agreements.id, agreement.agreementId));
+      });
+      throw error;
```
iscekic: **P3 — Cron failure isolation.** Rethrowing here aborts the whole issuance run: agreements after the failing one (ordered by `paid_until`) lose a cycle. It self-heals next run (the failed agreement becomes `failed` and is excluded from the `ready` filter), but continuing to the next agreement and reporting the failure would avoid collateral delay for unrelated customers. Also note the join above doesn't exclude soft-deleted parent organizations, and issuance depends on `organizations.created_by_kilo_user_id` still existing — an admin-deleted parent (or deleted owner user) keeps monthly issuance running / turns it into a permanent `failed` state.

apps/web/src/routers/organizations/organization-subscription-router.ts
```diff
@@ -336,7 +355,15 @@ export const organizationsSubscriptionRouter = createTRPCRouter({
...
 
       const purchase = latestPurchase;
-      return await handleUpdateSeatCount(purchase.subscription_stripe_id, newSeatCount, total);
+      const result = await handleUpdateSeatCount(
+        purchase.subscription_stripe_id,
+        newSeatCount,
+        total
+      );
+      if (result.success && newSeatCount < total) {
+        await scheduleOrganizationPassCapacity({ organizationId, paidSeatCount: newSeatCount });
```
iscekic, line 364: **P3 — Error UX after a successful seat change.** `scheduleOrganizationPassCapacity` throws `active agreement has no paid-through date` for a `pending_payment` agreement (purchased, payment not yet confirmed). The Stripe seat change has already succeeded at this point, so the user sees a misleading 500 for a state they can't act on. Consider returning early (or treating it as no-op) for `pending_payment` agreements.

apps/web/src/app/(app)/organizations/[id]/subscriptions/kilo-pass/activation/page.tsx
```diff
@@ -0,0 +1,24 @@
+import { OrganizationByPageLayout } from '@/components/organizations/OrganizationByPageLayout';
```
Comment (iscekic) [inline apps/web/src/app/(app)/organizations/[id]/subscriptions/kilo-pass/activation/page.tsx]: **P3 — Dead/unreachable surface.** Nothing links to `/activation` in the shipped flow (the adapter never returns `kind: 'checkout'` and no Checkout Session is created — 3DS is handled inline via `handleNextAction` and the user lands on the detail page). Same for `/review` (redirect-only), the unused `checkout` result kind, `retryAfterSeconds` (always null), and service-state union members (`requires_action`, `activating`) the service never produces. Harmless, but it's maintenance surface and confusing for future readers — consider removing or wiring it up.

apps/web/src/lib/kilo-pass-org/bonus-repair.ts
```diff
@@ -0,0 +1,65 @@
...
+  nowIso: string
+): Promise<{ examined: number; recordedMisses: number }> {
+  const snapshots = await tx
+    .select()
+    .from(kilo_pass_org_issuance_snapshots)
+    .where(
+      and(
+        eq(kilo_pass_org_issuance_snapshots.bonus_mode, KiloPassOrgBonusMode.AfterBase),
+        isNull(kilo_pass_org_issuance_snapshots.bonus_unlocked_at),
+        lt(kilo_pass_org_issuance_snapshots.window_end, nowIso)
```
iscekic, line 26: **P3 — Unbounded sweep.** Every run re-selects *all* expired, locked, `after_base` snapshots ever created (including ones already recorded as missed — the idempotency key dedupes the write but not the scan) and locks them `FOR UPDATE`. Fine at current scale, but this grows without bound; consider a marker (e.g. a `repair_completed_at` or filtering on audit-record existence) or a lookback window.

apps/web/src/components/subscriptions/org-kilo-pass/OrgKiloPassGroup.tsx
```diff
@@ -0,0 +1,87 @@
...
+      <OrgKiloPassSubscriptionCard
+        organizationId={organizationId}
+        paidSeatCount={agreement?.paidSeatCount ?? organization?.seat_count ?? 0}
+        agreement={
+          agreement && terms && (!isEnded || showTerminal)
+            ? {
+                status: commercialState ?? 'pending_payment',
+                tierName: terms.tierName,
+                paidSeatCount: agreement.paidSeatCount,
+                price: `${formatOrgPassMoney(terms.pricePerPassUsd * agreement.paidSeatCount)}/${commercialState === 'active' ? 'month' : 'period'}`,
```
iscekic: **P3 — Minor display accuracy.** For annual agreements this renders the monthly-equivalent price with a `/month` suffix, while the detail view correctly shows "Annual" billing. Consider using the cadence-aware label (or `/mo equiv.`) to avoid implying monthly billing. Relatedly, the org Kilo Pass billing history lists shared invoices whose totals include seat charges — the copy is accurate, but the amounts aren't pass-only, which may confuse during reconciliation.

apps/web/src/lib/kilo-pass-org/service.ts
```diff
@@ -0,0 +1,1807 @@
...
+        allBonusUnlocked: false,
+        hasProratedCredits: false,
+      });
+    }
+    const currentWindowEnded = latestSnapshot ? asDate(latestSnapshot.end) <= now : false;
+    const [latestRun, pendingTransitions] = await Promise.all([
+      latestRunForAgreement(row.agreement.id),
+      db
+        .select({
+          id: kilo_pass_org_term_transitions.id,
```
iscekic: **P3 — Ops surface gaps vs ADR 0003 (batch note).**

- `kilo_pass_org_term_transitions` is read here (`pendingTermTransitions`) but never written anywhere — no code path can schedule a term/tier change, so the "Terms change scheduled" UI is dead today. Presumably intentional scope, but it's dead schema + UI; consider hiding the section until a writer exists.
- The ADR promises "one deduplicated email to current parent owners and billing managers" for blocked windows — no email code exists in this PR; blocked state is only visible on the subscription page.
- `kilo_pass_org_audit_records` is only written by the bonus-miss paths. Cancel / resume / allocation changes / suspensions record no actor+reason audit, despite the ADR's mutation-audit requirement.
- There is no explicit control to clear `suspended_for_review` (admins get a read-only card; the only exits are indirect — an allocation edit or run retry flipping condition to `ready`).

apps/web/src/lib/kilo-pass-org/service.ts
```diff
@@ -0,0 +1,1807 @@
...
+      existing.paid_until && asDate(existing.paid_until) > input.paidUntil
+        ? asDate(existing.paid_until)
+        : input.paidUntil;
+    const pendingCapacityEffectiveAt = existing.next_capacity_effective_at
+      ? asDate(existing.next_capacity_effective_at)
+      : null;
+    const appliesPendingCapacity =
+      pendingCapacityEffectiveAt !== null && input.paidFrom >= pendingCapacityEffectiveAt;
+    // Seat increases update the same provider quantity that held any scheduled decrease,
+    // so the confirmed higher quantity supersedes that pending decrease.
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in e13d58487. Paid capacity now compares against the pending capacity, so 25 -> scheduled 10 -> paid 12 clears the stale pending decrease. Added regression coverage that verifies capacity 12, cleared pending fields, and a subsequent allocation update.

apps/web/src/components/subscriptions/org-kilo-pass/OrgKiloPassDetailView.tsx
jeanduplessis: Fixed in e13d58487. Dialog save validity now depends only on the edited reduction requirement, not the stale server overallocated condition. Added focused coverage proving a repaired allocation enables Save.

apps/web/src/lib/kilo-pass-org/stripe-adapter.ts
jeanduplessis: Fixed in e13d58487. Kilo Pass cancellation schedules are tagged with an origin, verified by origin and shape, and foreign schedules now fail closed with SCHEDULE_REWRITE_UNSAFE instead of silently no-oping. Paid invoice activation also preserves cancel_at_period_end. Added billing-cycle schedule regression coverage.

apps/web/src/routers/organizations/organization-kilo-pass-router.ts
jeanduplessis: Fixed in e13d58487. The usage endpoint now requires parent billing authority, and the dashboard card is rendered only for owners/billing managers. Router tests now verify billing-role access and regular-member denial.

apps/web/src/lib/kilo-pass-org/hierarchy-guard.ts
jeanduplessis: Fixed in e13d58487. The hierarchy guard now evaluates the latest authoritative allocation plan rather than every historical row. Added the missing initial 1 -> future 0 -> reparent allowed regression test.

apps/web/src/lib/kilo-pass-org/service.ts
jeanduplessis: Fixed in e13d58487. Recovery transitions now preserve manual and suspended_for_review across allocation, seat-capacity, activation, issuance, and retry paths. Payment review has an explicit audited admin clear action that recomputes ready vs overallocated.

apps/web/src/lib/kilo-pass-org/service.ts
jeanduplessis: Fixed in e13d58487. Issuance failures are isolated per agreement and reported while later agreements continue; soft-deleted parents are excluded, and recurring issuance resolves a current billing authority rather than relying solely on the creator.

apps/web/src/routers/organizations/organization-subscription-router.ts
jeanduplessis: Fixed in e13d58487. Capacity scheduling returns a no-op for agreements without paid-through entitlement, including pending_payment, so a successful Stripe seat change is not followed by a misleading 500.

apps/web/src/app/(app)/organizations/[id]/subscriptions/kilo-pass/activation/page.tsx
Comment (jeanduplessis) [reply apps/web/src/app/(app)/organizations/[id]/subscriptions/kilo-pass/activation/page.tsx]: Fixed in e13d58487. Removed the unreachable activation/review routes, activation components/state mapper, checkout URL result variant, status polling endpoint, retryAfterSeconds, and dead term-transition presentation.

apps/web/src/lib/kilo-pass-org/bonus-repair.ts
jeanduplessis: Fixed in e13d58487. Expired bonus repair is now bounded to 250 skip-locked rows and marks every examined snapshot complete, so completed historical rows are not rescanned indefinitely.

apps/web/src/components/subscriptions/org-kilo-pass/OrgKiloPassGroup.tsx
jeanduplessis: Fixed in e13d58487. Annual agreements now show a monthly-equivalent value explicitly marked as billed annually. Billing history also states that totals include seat and Kilo Pass charges.

apps/web/src/lib/kilo-pass-org/service.ts
jeanduplessis: Addressed the batch in e13d58487: removed the unwritable term-transition UI, added durable leased/deduplicated blocked-window notifications to current owners/billing managers, added audit records for allocation/capacity/cancel/resume/retry/payment-review changes, and added an explicit reason-required admin control to clear suspended_for_review.

apps/web/src/lib/kilo-pass-org/stripe-adapter.ts
```diff
@@ -0,0 +1,441 @@
...
+  if (schedule) {
+    const expectedRetainedPrices = new Set(retainedItems.map(item => item.price));
+    const removalPhase = schedule.phases[1];
+    const hasRemovalScheduled =
+      schedule.metadata?.origin === ORGANIZATION_KILO_PASS_CANCELLATION_ORIGIN &&
+      schedule.phases.length === 2 &&
+      removalPhase?.items.length === retainedItems.length &&
+      removalPhase.items.every(item => expectedRetainedPrices.has(String(item.price)));
+    if (hasRemovalScheduled) return;
+    throw new Error('SCHEDULE_REWRITE_UNSAFE');
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 86e637a4f. Cancellation now recognizes an owned resumed schedule whose second phase still contains the pass item and safely rewrites it back to a removal schedule. Added cancel/resume/cancel regression coverage.

apps/web/src/lib/kilo-pass-org/stripe-adapter.ts
```diff
@@ -0,0 +1,441 @@
...
+  }
+  const target = await stripe.subscriptionSchedules.create({ from_subscription: subscription.id });
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 86e637a4f. A single-phase metadata-free schedule is adopted only when its items exactly match the live subscription, then tagged and rewritten. Foreign schedule shapes still fail closed. Added orphaned-create recovery coverage.

apps/web/src/lib/kilo-pass-org/service.ts
```diff
@@ -0,0 +1,2018 @@
...
+          appliesPendingCapacity || replacesPendingCapacity
+            ? null
+            : existing.next_capacity_effective_at,
+        paid_from: input.paidFrom.toISOString(),
+        paid_until: paidUntil.toISOString(),
+        activation_provider_event_id:
+          existing.activation_provider_event_id ?? input.providerEventId,
+      })
+      .where(eq(kilo_pass_org_agreements.id, input.agreementId));
+    if (
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 86e637a4f. Parent supplement creation is now gated out for manual and suspended_for_review conditions, matching the activation issuance guard. Added regression coverage proving a suspended paid increase does not create a supplement.

apps/web/src/components/organizations/OrganizationDashboard.tsx
```diff
@@ -166,6 +167,9 @@ export function OrganizationDashboard({
               <OrganizationAdminMembers organizationId={organizationId} />
             </LockableContainer>
             <SeatUsageCard organizationId={organizationId} />
+            {canManageOrganizationBilling(currentRole) && (
+              <KiloPassUsageCard organizationId={organizationId} />
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 86e637a4f. The dashboard now renders Kilo Pass usage only for billing roles on a top-level parent organization, so child dashboards do not issue the parent-only query.

apps/web/src/lib/kilo-pass-org/notifications.ts
```diff
@@ -0,0 +1,151 @@
...
+    try {
+      const result = await sendKiloPassOrgBlockedEmail(candidate.email, {
+        organizationId: candidate.organizationId,
+        organizationName: candidate.organizationName,
+      });
+      if (!result.sent) {
+        failed++;
+        await database
+          .update(kilo_pass_org_notification_deliveries)
+          .set({ status: 'pending', lease_expires_at: null })
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 86e637a4f. Claims now have a three-attempt cap, permanent SendResult failures become terminal failed rows, and retry resets are scoped to unsent sending rows. Reset errors are captured without aborting the cron.

apps/web/src/lib/kilo-pass-org/hierarchy-guard.ts
```diff
@@ -0,0 +1,76 @@
...
+      eq(kilo_pass_org_allocation_plans.agreement_id, kilo_pass_org_agreements.id)
+    )
+    .where(
+      and(
+        eq(kilo_pass_org_agreements.parent_organization_id, organization.parentOrganizationId),
+        ne(kilo_pass_org_agreements.state, 'ended')
+      )
+    )
+    .orderBy(
+      desc(kilo_pass_org_allocation_plans.effective_window_start),
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 86e637a4f. The hierarchy guard now selects the newest plan version, matching plan supersession semantics when blocked-window reconciliation writes a newer version at an earlier effective window.

apps/web/src/lib/kilo-pass-org/service.ts
```diff
@@ -0,0 +1,2018 @@
...
+    await recordAudit(tx, {
+      agreementId: agreement.id,
+      action: 'payment_review_suspended',
+      reason: 'provider_adverse_payment_event',
+      before: { processingCondition: agreement.processing_condition },
+      after: {
+        processingCondition: KiloPassOrgProcessingCondition.SuspendedForReview,
+        paymentReviewRequiredAt: suspendedAt,
+      },
+      idempotencyKey: `kpo:payment-review:${agreement.id}:${suspendedAt}`,
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 86e637a4f. Repeated adverse events now preserve the original payment_review_required_at and return without appending another audit row; the audit identity is stable per agreement.

apps/web/src/app/api/cron/kilo-pass-org-issuance/route.ts
```diff
@@ -0,0 +1,30 @@
...
+if (!CRON_SECRET) throw new Error('CRON_SECRET is not configured in environment variables');
+
+export async function GET(request: Request) {
+  if (request.headers.get('authorization') !== `Bearer ${CRON_SECRET}`) {
+    sentryLogger('cron', 'warning')('SECURITY: Invalid organization Pass issuance authorization');
+    return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
+  }
+  const summary = await runOrganizationPassIssuanceCron(db);
+  // Dispatch after issuance so newly created blocked-run deliveries are included.
+  const notifications = await dispatchOrganizationPassBlockedNotifications(db);
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 86e637a4f. Issuance failures are logged before notification dispatch. Notification dispatch errors and any issuance/notification failure now return success:false with HTTP 500; added route coverage.

apps/web/src/lib/kilo-pass-org/service.ts
```diff
@@ -0,0 +1,2024 @@
...
+    await recordAudit(tx, {
+      agreementId: agreement.id,
+      action: 'payment_review_suspended',
+      reason: 'provider_adverse_payment_event',
+      before: { processingCondition: agreement.processing_condition },
+      after: {
+        processingCondition: KiloPassOrgProcessingCondition.SuspendedForReview,
+        paymentReviewRequiredAt: suspendedAt,
+      },
+      idempotencyKey: `kpo:payment-review:${agreement.id}`,
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 766c93675. State-based early return deduplicates retries within one suspension, while the audit key again includes suspendedAt so a later suspension after admin clearance records independently without a unique-key rollback.

apps/web/src/lib/kilo-pass-org/hierarchy-guard.ts
```diff
@@ -0,0 +1,73 @@
...
+      kilo_pass_org_allocation_plans,
+      eq(kilo_pass_org_allocation_plans.agreement_id, kilo_pass_org_agreements.id)
+    )
+    .where(
+      and(
+        eq(kilo_pass_org_agreements.parent_organization_id, organization.parentOrganizationId),
+        ne(kilo_pass_org_agreements.state, 'ended')
+      )
+    )
+    .orderBy(desc(kilo_pass_org_allocation_plans.version))
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 766c93675. The guard scopes to the current non-ended agreement and checks the plan effective now, all future plans, and plans effective for unresolved snapshotless runs. Added coverage distinguishing an effective zero plan from a merely scheduled future zero.

apps/web/src/lib/kilo-pass-org/stripe-adapter.ts
```diff
@@ -0,0 +1,476 @@
...
+    target = await stripe.subscriptionSchedules.create({ from_subscription: subscription.id });
+  }
+  if (!target) throw new Error('SCHEDULE_REWRITE_UNSAFE');
+  await stripe.subscriptionSchedules.update(target.id, {
+    metadata: { origin: ORGANIZATION_KILO_PASS_CANCELLATION_ORIGIN },
+    end_behavior: 'release',
+    phases: [
+      {
+        items: currentItems,
+        start_date: Math.floor(period.start.getTime() / 1000),
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 766c93675. Re-cancellation of an owned resumed schedule now preserves the active/latest phase start and end boundaries instead of re-anchoring from the subscription period. Added an after-renewal boundary regression test.

apps/web/src/lib/kilo-pass-org/stripe-adapter.ts
```diff
@@ -0,0 +1,476 @@
...
+      removalPhase !== undefined &&
+      scheduleItemsMatch(removalPhase.items, retainedItems);
+    if (hasRemovalScheduled) return;
+    const isResumedSchedule =
+      isOwnedSchedule &&
+      schedule.phases.length === 2 &&
+      removalPhase !== undefined &&
+      scheduleItemsMatch(removalPhase.items, currentItems);
+    const isOrphanedCreate =
+      !schedule.metadata?.origin &&
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 766c93675. Orphan adoption now requires an entirely empty metadata object in addition to exact single-phase item matching. Added coverage rejecting matching schedules with foreign metadata.

apps/web/src/lib/kilo-pass-org/service.ts
```diff
@@ -0,0 +1,2024 @@
...
+    const validation = validateAllocation(
+      input.paidSeatCount,
+      input.initialAllocations.map(a => a.passCapacity)
+    );
+    if (!validation.valid) throw new Error(`invalid initial allocation: ${validation.reason}`);
+    const [existing] = await tx
+      .select({ id: kilo_pass_org_agreements.id })
+      .from(kilo_pass_org_agreements)
+      .where(eq(kilo_pass_org_agreements.provider_subscription_id, input.providerSubscriptionId));
+    if (existing) return { agreementId: existing.id, created: false };
```
iscekic: This lookup matches by `provider_subscription_id` with no state filter, so it also returns an **ended** agreement — and nothing clears `provider_subscription_id` when an agreement ends (`handleOrganizationKiloPassSubscriptionEvent` only sets `state: 'ended'`).

Restart flow: after a scheduled cancellation removes the add-on at period end, the owner clicks "Add Kilo Pass again" (offered by `OrgKiloPassDetail`/`OrgKiloPassGroup`). `createOrganizationKiloPassCheckout` then gets `created: false` for the ended row, so no new agreement, allocation plan, or initial allocations are persisted — and `handleOrganizationKiloPassInvoicePaid` looks the agreement up with `ne(state, 'ended')`, finds nothing, and returns `false`. Net effect: Stripe re-adds the add-on and bills the customer every period, but activation and issuance never happen and there is no self-serve repair path.

I think two layers need fixing: (1) this early return should ignore ended agreements (and/or checkout should refuse while any non-ended agreement exists), and (2) the unique partial index `UQ_kilo_pass_org_agreements_provider_subscription` covers ended rows too, so inserting a fresh agreement for the same seat subscription would violate it — provider ids likely need clearing (or the index scoping to non-ended) when an agreement ends. A regression test for cancel → repurchase → invoice.paid would be very worthwhile.

apps/web/src/lib/kilo-pass-org/stripe-adapter.ts
```diff
@@ -0,0 +1,480 @@
...
+    initialAllocations: input.allocations.map(allocation => ({
+      organizationId: allocation.childOrganizationId,
+      passCapacity: allocation.passCount,
+    })),
+  });
+  const price = getStripePriceIdForKiloPass({
+    tier: input.tier as KiloPassTier,
+    cadence: cadence as KiloPassCadence,
+  });
+  const updated = await stripe.subscriptions.update(subscription.id, {
```
iscekic, line 131: `createOrganizationKiloPassCheckout` never refuses when the org already has a non-ended (e.g. active) agreement. `createPendingAgreement` returns the existing agreement as `created: false`, but the Stripe-side effect still runs: `subscriptions.update` with `items: [{ price, quantity }]` and no item `id` **adds** a new subscription item — Stripe does not merge items by price. A user who navigates directly to `/subscriptions/kilo-pass/setup` while active (the page renders normally; only the group card hides the CTA) ends up with two pass add-on items and gets double-billed every period.

Suggest failing closed before mutating the subscription: if a non-ended agreement exists for the parent org (or a pass item already exists on the subscription), throw a user-facing error instead of adding another item.

apps/web/src/lib/kilo-pass-org/consumption.ts
```diff
@@ -0,0 +1,206 @@
...
+  const identity = sourceIdentity(input);
+  const [credit] = await tx
+    .insert(credit_transactions)
+    .values({
+      kilo_user_id: input.kiloUserId,
+      organization_id: input.organizationId,
+      amount_microdollars: -input.amountMicrodollars,
+      is_free: false,
+      description: `Organization consumption (${input.source})`,
+      credit_category: identity,
```
iscekic, line 122: This inserts one `credit_transactions` row per AI-gateway usage record (and per Exa charge) for every organization, whether or not a Kilo Pass agreement exists. Two concerns:

1. `getCreditTransactionsForOrganization` and `getAdminCreditTransactionsForOrganization` both return the latest 100 org transactions. After this lands, a moderately busy org's list (the member-visible mobile credit activity screen and the admin credit transactions view) will be wall-to-wall `Organization consumption (ai-gateway)` rows, pushing actual grants and purchases out of view. Consider filtering `kpo:consumption:%` categories from those list queries.
2. Volume: this duplicates `microdollar_usage` into one of the largest tables at request rate. The ADR's canonical debit ledger is a reasonable goal, but it may belong in its own table (or reference usage rows) rather than `credit_transactions`.

apps/web/src/lib/kilo-pass-org/service.ts
```diff
@@ -0,0 +1,2024 @@
...
+      parentOrganizationId: organizations.id,
+      fallbackUserId: organizations.created_by_kilo_user_id,
+    })
+    .from(kilo_pass_org_agreements)
+    .innerJoin(organizations, eq(kilo_pass_org_agreements.parent_organization_id, organizations.id))
+    .where(
+      and(
+        inArray(kilo_pass_org_agreements.state, [
+          KiloPassOrgAgreementState.Active,
+          KiloPassOrgAgreementState.CancelAtPeriodEnd,
```
iscekic, line 1918: A `pending_payment` agreement whose invoice never reaches `paid` (card declined on the initial add-on invoice, or an abandoned `requires_action`) has no exit: this cron only selects `active`/`cancel_at_period_end`, `reconcileOrganizationKiloPassPayment` only succeeds when the latest invoice is actually paid, and no event expires the agreement. It can sit in `pending_payment` indefinitely — with the add-on item still on the subscription, and the `pending:<sub>` placeholder in `provider_seat_add_on_item_id` if binding never happened. Should these age out to `ended` after a grace period, or at least surface in an ops/admin view?

docs/adr/0003-org-kilo-pass-source-of-truth.md
```diff
@@ -0,0 +1,205 @@
...
+
+Kilo Pass purchase, status, and allocation controls live on existing organization subscription page beside seat controls. Personal Subscription Center does not manage org Kilo Pass.
+
+Authorized parent organization owners and billing managers see Kilo Pass terminology in management and audit surfaces. Regular members see generic organization Credit balance and transaction language without Kilo Pass labels, counts, tiers, or bonus mechanics.
+
+Child sub-org owners do not gain Kilo Pass allocation authority from child ownership. They may see resulting usable credits through existing balance surfaces, while parent owners and billing managers retain agreement and allocation control.
+
+A child sub-org cannot detach, reparent, archive, or delete while it has nonzero initial or future Kilo Pass allocation. A parent organization owner or billing manager must first set allocation to zero. Already granted pooled credits remain with child because shared balance cannot safely identify source-specific remainder for clawback.
```
iscekic: This says a child cannot detach/reparent/archive/delete while it has a nonzero *initial or future* allocation, but `assertOrganizationHierarchyChangeAllowed` implements — and its tests enshrine (`allows detaching a child with zero allocation despite issued credits`) — a narrower rule: current plan + future plans + plans covering unresolved runs, explicitly excluding historical issued allocations. The implemented rule seems right; worth rewording here (and the matching CONTEXT.md relationship) so the contract matches the code.

iscekic (commented):
Here are some from my bot, I'm reading the PR in parallel.

iscekic (approved):
Nothing too major from me, feel free to pick what you want to fix.

iscekic:
Damn bot deleted my review comments while posting its own - let me restore them real quick. 🤦

apps/web/src/components/subscriptions/org-kilo-pass/selection.ts
iscekic, line 1: Prefer parsing with zod

apps/web/src/lib/kilo-pass-org/calculations.ts
iscekic, line 1: I love this line, it's such an AI-ism. If we ever hit this branch, we're either living like kings or hyperinflation has killed us all.

apps/web/src/lib/kilo-pass-org/notifications.ts
iscekic, line 1: We could make this a push notification too

apps/web/src/lib/kilo-pass-org/service.ts
iscekic, line 1: Might make sense to break up this file

apps/web/src/lib/kilo-pass-org/stripe-adapter.ts
iscekic, line 1: I've been bitten by casting stuff to Stripe types before - I'd have an agent double check all these in actual runtime.

Then again, I've also been bitten by types directly returned from the stripe sdk, so ymmv.

apps/web/src/lib/kilo-pass-org/stripe-adapter.ts
```diff
@@ -0,0 +1,480 @@
...
+      currentPhase !== undefined &&
+      scheduleItemsMatch(currentPhase.items, currentItems);
+    if (!isResumedSchedule && !isOrphanedCreate) throw new Error('SCHEDULE_REWRITE_UNSAFE');
+  } else {
+    target = await stripe.subscriptionSchedules.create({ from_subscription: subscription.id });
+  }
+  if (!target) throw new Error('SCHEDULE_REWRITE_UNSAFE');
+  const resumedPhase =
+    schedule?.metadata?.origin === ORGANIZATION_KILO_PASS_CANCELLATION_ORIGIN
+      ? schedule.phases.at(-1)
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 34311a7bd. Cancellation now selects the phase containing the subscription item current period (falling back to schedule.current_phase/current period) instead of phases.at(-1). Added regression coverage for cancel → resume → cancel in the same period and for a genuinely renewed subscription.

apps/web/src/lib/kilo-pass-org/service.ts
jeanduplessis: Fixed in 34311a7bd. Agreement creation now ignores ended rows, provider identities are unique only among non-ended agreements, and provider event/payment-review lookups also ignore ended rows. Added coverage for creating a fresh agreement on the same seat subscription after the prior agreement ends.

apps/web/src/lib/kilo-pass-org/stripe-adapter.ts
jeanduplessis: Fixed in 34311a7bd. Checkout now fails closed before Stripe mutation if the subscription already contains a pass add-on, and also aborts if durable agreement creation reports a concurrent non-ended agreement. Both paths have regression tests and do not call subscriptions.update.

apps/web/src/lib/kilo-pass-org/consumption.ts
jeanduplessis: Addressed in 34311a7bd. Request-level qualifying debit rows are now written only for current/planned allocation containers of a non-ended agreement; ordinary org usage updates balances without duplicating into credit_transactions. Member/admin Credit activity queries also filter kpo:consumption rows before their 100-row limit. Added coverage for the no-agreement path.

apps/web/src/lib/kilo-pass-org/service.ts
jeanduplessis: Fixed in 34311a7bd. Recoverable payment failures remain pending, while invoice.voided and invoice.marked_uncollectible now remove the unpaid add-on and end the pending agreement, allowing a clean retry. Added adapter and Stripe event-routing coverage and documented the lifecycle.

docs/adr/0003-org-kilo-pass-source-of-truth.md
jeanduplessis: Fixed in 34311a7bd. The ADR and CONTEXT.md now match the implemented contract: current effective, future-effective, and unresolved-run allocations block hierarchy changes; superseded historical allocations and issued credits do not.

apps/web/src/components/subscriptions/org-kilo-pass/selection.ts
jeanduplessis: Fixed in 34311a7bd. Session storage is now parsed with a Zod schema via safeParse, with coverage for valid and malformed persisted selections.

apps/web/src/lib/kilo-pass-org/calculations.ts
jeanduplessis: Keeping the safe-integer overflow guard as defensive arithmetic validation. No behavior change needed; resolving this non-actionable observation.

apps/web/src/lib/kilo-pass-org/notifications.ts
jeanduplessis: Agreed this could be useful, but push requires a new notifications contract, preference semantics, worker dispatch handling, and mobile routing. The current ADR promises deduplicated email, so I am keeping push as a follow-up rather than expanding this PR.

apps/web/src/lib/kilo-pass-org/service.ts
jeanduplessis: Agreed on the maintainability concern. I kept this correctness-focused follow-up scoped rather than combining it with a broad module split; lifecycle, issuance, and query projection are good extraction boundaries for a follow-up refactor.

apps/web/src/lib/kilo-pass-org/stripe-adapter.ts
jeanduplessis: Checked against the installed Stripe 19.3.1 runtime types/API in 34311a7bd. The cast was hiding a removed top-level Invoice.payment_intent field on the Clover API. Checkout now expands and reads latest_invoice.confirmation_secret directly with no Stripe invoice cast; schedule phase access uses the SDK runtime shape and has updated tests.

apps/web/src/lib/kilo-pass-org/stripe-adapter.ts
```diff
@@ -0,0 +1,526 @@
...
+    },
+    expand: ['latest_invoice.confirmation_secret'],
+  });
+  const passItem = organizationPassItem(updated);
+  await bindProviderSeatAddOnItem({
+    agreementId: pending.agreementId,
+    providerSeatAddOnItemId: passItem.id,
+  });
+  const invoice = typeof updated.latest_invoice === 'object' ? updated.latest_invoice : null;
+  if (invoice?.confirmation_secret?.client_secret) {
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in e1b53ca3b. Checkout now expands the invoice payment PaymentIntent and returns payment_action only when its actual status is requires_action and a confirmation secret is available. Hard-declined requires_payment_method invoices remain pending. Added regression coverage for both statuses.

apps/web/src/lib/kilo-pass-org/stripe-adapter.ts
```diff
@@ -0,0 +1,526 @@
...
+  if (invoice.status !== 'void' && invoice.status !== 'uncollectible') return false;
+  const reference = invoice.parent?.subscription_details?.subscription;
+  const subscriptionId = typeof reference === 'string' ? reference : reference?.id;
+  if (!subscriptionId) return false;
+  const [agreement] = await db
+    .select()
+    .from(kilo_pass_org_agreements)
+    .where(
+      and(
+        eq(kilo_pass_org_agreements.provider_subscription_id, subscriptionId),
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in e1b53ca3b. Terminal cleanup now requires the invoice to contain a line for the agreement bound Kilo Pass add-on item before deleting the item or ending the agreement. Added coverage showing terminal pass invoices clean up while seat-only terminal invoices are ignored.

apps/web/src/lib/kilo-pass-org/consumption.ts
```diff
@@ -0,0 +1,250 @@
...
+ */
+export async function recordOrganizationConsumption(
+  tx: DrizzleTransaction,
+  input: RecordOrganizationConsumptionInput
+): Promise<OrganizationConsumptionResult> {
+  if (!Number.isSafeInteger(input.amountMicrodollars) || input.amountMicrodollars <= 0) {
+    return { recorded: false, organizationUsage: noAlert };
+  }
+
+  const [relevantAgreement] = await tx
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in e1b53ca3b. The hot-path lookup is now two separately indexable queries: a parent lookup using the existing partial parent index, followed only on miss by a child lookup starting from allocation rows. Added a partial positive-container index and regenerated the branch-local 0201 migration. Tests cover parent, positively allocated child, zero-capacity child, and unrelated org paths.

apps/web/src/lib/kilo-pass-org/stripe-adapter.ts
```diff
@@ -0,0 +1,535 @@
...
+      ...subscription.metadata,
+      type: ORGANIZATION_KILO_PASS_METADATA_TYPE,
+      organizationId: input.organizationId,
+      kiloUserId: input.actorUserId,
+      tier: input.tier,
+      cadence,
+    },
+    expand: [
+      'latest_invoice.confirmation_secret',
+      'latest_invoice.payments.data.payment.payment_intent',
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in c919ce09d. The subscription update now expands only latest_invoice.confirmation_secret. PaymentIntent status is loaded through a separate invoicePayments.list call using the valid four-level data.payment.payment_intent expansion. Provider lookup failures fail open to pending, while only requires_action returns a browser action. Added requires-action, hard-decline, and lookup-failure coverage.

apps/web/src/lib/kilo-pass-org/stripe-adapter.ts
```diff
@@ -0,0 +1,535 @@
...
+    .where(
+      and(
+        eq(kilo_pass_org_agreements.provider_subscription_id, subscriptionId),
+        eq(kilo_pass_org_agreements.state, KiloPassOrgAgreementState.PendingPayment)
+      )
+    )
+    .orderBy(desc(kilo_pass_org_agreements.created_at))
+    .limit(1);
+  if (!agreement) return false;
+  if (!invoiceLineForSubscriptionItem(invoice, agreement.provider_seat_add_on_item_id ?? '')) {
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in c919ce09d. Terminal cleanup now recovers pending:<subscription> agreements by requiring a known Kilo Pass-priced invoice line, matching the subscription item by that exact price, removing it, and ending the agreement. Seat-only invoices remain ignored. Added bound, unbound, and seat-only regression coverage.

apps/web/src/lib/kilo-pass-org/service.test.ts
```diff
@@ -0,0 +1,1348 @@
...
+      expect.arrayContaining([
+        expect.objectContaining({ organizationId: parent.id, passCount: 0 }),
+        expect.objectContaining({ organizationId: child.id, passCount: 1 }),
+      ])
+    );
+  });
+
+  it('reclaims a blocked original window after seat capacity reconciliation and issues it once', async () => {
+    const retryWindow = {
+      start: new Date('2027-07-01T00:00:00.000Z'),
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in c919ce09d. The blocked/reclaimed test now derives its anchor, issuance windows, paid-through boundary, and cron time from the current clock via nextIssuanceBoundary/monthlyWindowFromOriginalAnchor, removing the absolute-date tripwire.

PR #4526, opened 2026-07-14
fix(ai-gateway): add Meta provider

chrarnoldus:
## Summary
- preserve provider names and data policies when synthesizing Kilo-exclusive inference providers
- classify Muse models under Meta without introducing a provider module dependency cycle
- deduplicate synthetic providers by slug while retaining one metadata object per provider

## Validation
- `pnpm --filter web run typecheck`
- `pnpm --filter web run dependency-cycle-check`
- targeted oxlint and oxfmt checks
- `git diff --check`


chrarnoldus:
preparation for https://github.com/Kilo-Org/cloud/pull/4524

lambertjosh (approved):

PR #4555, opened 2026-07-15
Update preferred AI model variants

chrarnoldus:
## Summary
- keep regular GPT-5.6 Sol preferred ahead of the discounted Martian variant
- include discounted Martian Sol only while its model status is public
- remove GPT-5.6 Terra from preferred models
- use regular Claude Opus 4.8 and Qwen 3.7 Plus variants
- prefer discounted DeepSeek V4 Pro while its model status is public, otherwise use regular V4 Pro
- update the focused preferred-model test for conditional inclusion, exclusion, and Sol ordering

## Validation
- git diff --check
- targeted oxfmt check for both changed files
- automated Kilo review: no issues found on prior revisions
- remaining validation runs in CI


marius-kilocode:
What happens if training opt-in was disabled?

chrarnoldus:
If training/data collection is disabled, the discounted Sol and discounted DeepSeek entries are marked with `requires-data-collection`. Teams organizations that deny collection have those models filtered from the available model catalog, and direct requests are rejected with `data_collection_required` by the gateway check. Regular Sol remains immediately above discounted Sol, and regular DeepSeek V4 Pro is the fallback only when the discounted model itself is not public; users with collection disabled should select the regular model.

marius-kilocode (approved):

PR #4603, opened 2026-07-16
Mobile UX bundle: scroll-to-bottom, mic sizing, live remote titles, search clear

iscekic:
## Summary

Four small, independent mobile UX fixes:

1. **Scroll-to-bottom button** in agent conversations — fades in bottom-right once scrolled up past ~100px, tap scrolls to the latest message. Never shows on short/empty conversations, and resets correctly when switching sessions.
2. **Mic icon unification** — the composer's mic button now matches the size and shape of its sibling controls (paperclip/stop/send) in both the agent chat composer and the new-session prompt, while preserving a ≥44pt touch target at every size. The kilo-chat composer is untouched.
3. **Stale Remote session names** — `GET /sessions/active` now overlays the persisted `cli_sessions_v2.title` onto CLI heartbeat snapshots when a meaningful DB title exists, falling back to the heartbeat title on a missing row, null/empty/whitespace title, or a DB failure (never blanks the Remote section). In-app renames also invalidate `activeSessions.list` so the Remote row updates without waiting for the next poll.
4. **Search clear (X) button** on the Agents page — appears while the search field has text, clears only the search query and dismisses the keyboard, without touching any selected platform/project filters. The existing empty-state "Clear search"/"Clear filters" CTA keeps its broader behavior.

## Testing

- Automated: extended unit tests for the title-overlay logic (including DB-failure and fallback cases), the scroll-to-bottom at-bottom detection, and a new pure `session-search-state` module covering the debounce and both clear paths.
- Manual on-device (iOS simulator): verified all four items, including an on-device before/after repro of the stale-title bug via a live CLI session and an org-scoped Remote row.

### DIFF
--- a/apps/mobile/src/components/agents/session-message-list.tsx
+++ b/apps/mobile/src/components/agents/session-message-list.tsx
@@ -1,11 +1,14 @@
 import { FlashList, type FlashListRef, type ListRenderItem } from '@shopify/flash-list';
 import { type OlderMessagesError } from 'cloud-agent-sdk';
+import { ChevronDown } from 'lucide-react-native';
 import { useCallback, useEffect, useRef } from 'react';
-import { type ViewStyle } from 'react-native';
+import { Pressable, View, type ViewStyle } from 'react-native';
+import Animated, { FadeIn, FadeOut } from 'react-native-reanimated';
 
 import { useSessionListAutoScroll } from '@/components/agents/use-session-list-auto-scroll';
 import { SessionPaginationHeader } from '@/components/agents/session-pagination-header';
 import { shouldTriggerOlderMessagesLoad } from '@/components/agents/session-message-list-state';
+import { useThemeColors } from '@/lib/hooks/use-theme-colors';
 
 const listStyle = { flex: 1 } satisfies ViewStyle;
 const listContentContainerStyle = { paddingVertical: 8 } satisfies ViewStyle;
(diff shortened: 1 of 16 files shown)


apps/mobile/src/components/agents/session-message-list.tsx
```diff
@@ -98,42 +104,66 @@ export function SessionMessageList<T>({
...
+          entering={FadeIn.duration(200)}
+          exiting={FadeOut.duration(150)}
+          pointerEvents="box-none"
+          className="absolute bottom-4 right-4"
+        >
+          <Pressable
+            accessibilityRole="button"
+            accessibilityLabel="Scroll to bottom"
+            onPress={scrollToLatestAnimated}
+            className="h-10 w-10 items-center justify-center rounded-full border border-border bg-card shadow-lg shadow-black/25 active:opacity-70"
```
St0rmz1, line 161: **Low, Design:** This button renders at `h-10 w-10` (40x40) with no `hitSlop`, which falls short of the 44pt touch target guideline. The mic button change in this same PR explicitly pairs visual size with `hitSlop` to guarantee 44pt at every size; consider doing the same here, e.g. adding `hitSlop={{ top: 2, bottom: 2, left: 2, right: 2 }}`.

St0rmz1 (approved):

apps/mobile/src/components/agents/session-message-list.tsx
iscekic: Good catch. Since this PR is already merged, the fix landed on #4623 (small enough to piggyback): added `hitSlop={2}` to reach the 44pt target.

PR #4623, opened 2026-07-18
feat(auto-routing): add model-switch telemetry to decision logging

iscekic:
Stacked on #4611; retarget to `main` once it merges.

## Problem

We cannot measure real model switching in kilo-auto/efficient today: the `auto_routing_decision` log event has no incumbent field (`sticky: false` fires even when the model did not change), and successful decisions are sampled at ~1%, which decimates within-session switch sequences. This telemetry gates planned policy improvements (threshold hardening, near-miss hysteresis, route-flip debounce).

## Changes

- **`incumbentModel`** — the sticky model considered for this decision (null when none).
- **`switched`** — the incumbent existed and the decided model differs from it.
- **`switchReason`** — returned by `computeDecision` on the decision (additive optional contract field): `threshold` (incumbent denied, off the route, or below the accuracy bar), `cost` (the mode's switch condition triggered), `capability` (ejected by #4611's modality/context filters), null when there was no incumbent or it was kept.
- **`routeChanged`** — this decision's route key vs the incumbent's; the sticky cache entry now stores the route key alongside the model. Null when unknowable (no incumbent, or a pre-routeKey cache entry — old entries still parse).
- **Sampling bypass** — decisions with `switched: true` always log; `sampled: false` marks bypassed lines (also retries/failures) so downstream rate math only scales sampled rows.
- **`contextTokens`** — `constraints.promptTokensEstimate` when the gateway sent it (available since #4611), null otherwise; no new plumbing.

## Notes / assumptions

- In `best_accuracy` mode the accuracy-gap switch condition is reported as `cost` — it is the same "eligible incumbent, but the switch pays off" branch as the cost factor.
- All fields are additive; no schema/contract breaks, no new datastores.

## Verification

- `services/auto-routing`: vitest 138 passed (new coverage: switchReason threshold/cost/capability, routeChanged incl. legacy sticky entries, sampling bypass on switch, drop of routine unsampled decisions), typecheck, lint clean.
- `packages/auto-routing-contracts`: vitest 83 passed, typecheck, lint clean.

### DIFF
--- a/packages/auto-routing-contracts/src/normalize.ts
+++ b/packages/auto-routing-contracts/src/normalize.ts
@@ -390,11 +390,15 @@ function collectModalities(value: unknown, out: Set<string>): void {
   }
 
   // Guards against callers that omit the `type` discriminator — the
-  // presence of a known media field is itself sufficient signal.
+  // presence of a known media field is itself sufficient signal. We do NOT
+  // treat a bare `file` key as a signal: it collides with ordinary tool
+  // output (e.g. a `read_file` result whose `content` is `{ file: ... }`),
+  // and the typed `file`/`input_file`/`document` cases above already cover
+  // the real Responses/Anthropic file shapes.
   if ('image_url' in value || 'input_image' in value) {
     out.add('image');
   }
-  if ('input_file' in value || 'file' in value) {
+  if ('input_file' in value) {
     out.add('file');
   }
 }
--- a/services/auto-routing/src/decision-engine.ts
+++ b/services/auto-routing/src/decision-engine.ts
@@ -179,9 +179,11 @@ export function computeDecision(
       tableVersion: table.version,
       reasoningEffort: incumbent.reasoningEffort ?? null,
       sticky: true,
+      switchReason: null,
     };
   }
 
+  const switched = incumbentModel !== null && incumbentModel !== freshPick.model;
   return {
     model: freshPick.model,
     taskType: classification.taskType,
@@ -190,5 +192,18 @@ export function computeDecision(
     tableVersion: table.version,
     reasoningEffort: freshPick.reasoningEffort ?? null,
     sticky: false,
+    // 'cost': the incumbent was eligible but the mode's switch condition
+    // (cost factor / accuracy gap) made the fresh pick worth it;
+    // 'capability': the modality/context filters ejected it from the route;
+    // 'threshold': it is denied, off the route, or below the accuracy bar.
+    switchReason: !switched
+      ? null
+      : incumbent
+        ? incumbent.meetsThreshold
+          ? 'cost'
+          : 'threshold'
+        : routeCandidates.some(c => c.model === incumbentModel)
+          ? 'capability'
+          : 'threshold',
   };
 }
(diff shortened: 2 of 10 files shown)


packages/auto-routing-contracts/src/normalize.ts
```diff
@@ -333,3 +333,262 @@ function isSensitiveKey(key: string) {
...
+      out.add('file');
+    }
+  }
+
+  // Guards against callers that omit the `type` discriminator — the
+  // presence of a known media field is itself sufficient signal.
+  if ('image_url' in value || 'input_image' in value) {
+    out.add('image');
+  }
+  if ('input_file' in value || 'file' in value) {
```
St0rmz1: **Low, Functional:** This fallback guard treats any record with a literal `file` key as requiring the `file` modality. Since `content`/`parts` are recursed into unconditionally, a tool result whose `content` happens to be a structured object with a `file` property (e.g. from a `read_file`-style tool) would be misdetected as an image/file request. That forces the turn onto a document-capable model it doesn't actually need, adding cost. Consider requiring a sibling `url`/`data`/`source` field before treating an untyped `file` key as a media signal, or dropping this fallback for `file` since the typed cases (`input_file`, `document`) above already cover the real Responses/Anthropic shapes.

services/auto-routing/src/decision-engine.ts
```diff
@@ -29,19 +41,113 @@ function pickFreshCandidate(
...
+//     side behaviour for image support.
+//   * Unknown context length is NOT proof of unfitness: a candidate whose
+//     row is missing `context_length` keeps its rank inside the eligible
+//     set. Only candidates with a known, provably-too-small context are
+//     excluded.
+//   * When every candidate's known context is provably too small, fall
+//     back to the candidates sharing the maximum known context so a
+//     large-but-still-too-small model is preferred over a slightly-smaller
+//     one we know cannot fit either.
+function applyCapabilityFilters(
```
St0rmz1, line 107: **Low, Design:** `codingPlanSatisfiesConstraints` in `decide.ts` re-implements the same policy as `applyCapabilityFilters` here: unknown capability data fails when an enforced modality is required, unknown context length keeps its rank, and a known-too-small context fails. The two implementations currently agree, but a future policy change (e.g. adding an enforced modality, or adjusting the unknown-context rule) has to be made in both places or the coding-plan short-circuit and benchmark routing will silently diverge. Consider extracting a single shared predicate (e.g. `modelSatisfiesConstraints(caps, constraints)`) that both call.

St0rmz1 (approved):

packages/auto-routing-contracts/src/normalize.ts
iscekic: Dropped the bare `file`-key fallback so structured tool output (e.g. a `read_file` result whose content is `{ file: ... }`) no longer misclassifies as a document request. Typed `input_file`/`document` cases still cover the real provider shapes; added a regression test.

services/auto-routing/src/decision-engine.ts
iscekic: Extracted `satisfiesRequiredModalities` / `contextProvablyTooSmall` / `modelSatisfiesConstraints` into decision-engine as the single policy source; both `applyCapabilityFilters` and the coding-plan short-circuit now call them.

PR #4637, opened 2026-07-20
feat(ai-gateway): prefix openrouter/ model display names with OpenRouter

chrarnoldus:
When a model id starts with `openrouter/` but its display name does not contain 'OpenRouter', prefix the display name with `OpenRouter ` in `formatName` (apps/web/src/lib/ai-gateway/providers/openrouter/index.ts).

The prefix is applied before any markers (`($$$$)`, `(new)`, `(retires ...)`) are appended, and skipped when the name already contains 'OpenRouter'. Added test coverage for the new behavior.


chrarnoldus:
CleanShot 2026-07-20 at 14 34 49

johnnyeric (approved):

PR #4639, opened 2026-07-20
feat(session-ingest): gate remote session pushes by user

eshurakov:
## Summary
- replace the disabled remote-session attention push flag with an exact single-user rollout gate
- forward the optional REMOTE_SESSION_ATTENTION_PUSH_USER_ID through queued and direct ingest dispatches
- document the setting and cover disabled, non-selected, and selected-user behavior

## Validation
- pnpm exec vitest run src/remote-session-notifications.test.ts
- pnpm run typecheck
- pnpm run lint
- pnpm -w exec oxfmt --check changed files
- pnpm run test (498 passed; one existing R2-offload route test timed out at 5 seconds)


services/session-ingest/src/remote-session-notifications.ts
```diff
@@ -45,7 +43,7 @@ export async function dispatchRemoteSessionAttentionSignal(
   params: { kiloUserId: string; sessionId: string; signal: AttentionSignal },
   deps: DispatchRemoteSessionAttentionDeps
 ): Promise<DispatchRemoteSessionAttentionOutcome> {
-  if (!REMOTE_SESSION_ATTENTION_PUSH_ENABLED) {
+  if (deps.remoteSessionAttentionPushUserId?.trim() !== params.kiloUserId) {
```
St0rmz1, line 46: **Low, Functional:** This trims `remoteSessionAttentionPushUserId` before comparing it to `params.kiloUserId`, but does not guard against `params.kiloUserId` being empty. If `kiloUserId` were ever an empty string and the env var were unset or empty, `''.trim() === ''` would be true and the gate would incorrectly pass instead of suppressing. There's no code path today that produces an empty `kiloUserId`, so this is a defensive gap rather than an active bug. Suggest adding an explicit guard, e.g. `if (!params.kiloUserId || deps.remoteSessionAttentionPushUserId?.trim() !== params.kiloUserId)`, so the empty string case is not incidentally treated as a match.

St0rmz1 (approved):

PR #4657, opened 2026-07-21
feat(cloud-agent): update admin health page

eshurakov:
## Summary

Distinguish Cloud Agent failures that need platform action from failures that need user action, and make the distinction visible and filterable on the existing `/admin/cloud-agent-next` health page.

- Shared exhaustive classifier in `@kilocode/worker-utils` maps structured lifecycle facts (stage/code, workspace subtype, assistant reason, provider ownership) to `failure_responsibility` (`platform`/`user`/`unknown`) plus a stable low-cardinality `failure_reason`. Ambiguous failures stay `unknown` rather than being guessed.
- Assistant terminal classification preserves explicit `payment_required` and `model_missing` codes instead of collapsing them into `assistant_error`.
- Queue reports carry the optional fields and `report-store` persists them idempotently. New nullable `failure_responsibility`/`failure_reason` columns on `cloud_agent_session_runs` and `cloud_agent_sessions` with check constraints (both null or both present; failed rows only) and a partial reporting index on failed runs (migration `0191_lyrical_nitro`). Legacy rows read as `unknown/unclassified` via COALESCE; no backfill.
- Admin health overview returns platform/user/unknown failure counts across run and setup events, a `platformFailureRate` (platform failures over completed + platform failures) alongside the existing all-failure rate, and responsibility/reason on top errors; drill-down filters by the exact aggregated dimensions.
- Health page replaces the five-card row with an observed-outcome distribution on a shared denominator (per the included design doc), defaults the error table to platform failures while keeping user/unknown visible, and adds a responsibility filter plus responsibility/reason columns.

## Verification

- [x] Manually verified the `/admin/cloud-agent-next` health page: observed-outcome summary, platform/user/unknown counts, responsibility filter, and error drill-down

## Visual Changes

Screenshot 2026-07-21 at 21 49 36

## Reviewer Notes

- Additive rollout: nullable columns and optional queue-report fields; consumers tolerate missing classification and legacy rows surface as `unknown/unclassified`.
- `unknown` is a first-class result meant to be driven down over time; interruptions are never assigned responsibility and are excluded from failure-rate denominators.
- Classification lives in the reporting domain; the admin UI never parses diagnostic prose, and no raw provider/setup content enters the new fields.


packages/db/src/schema.ts
```diff
@@ -5729,22 +5758,16 @@ export const cloud_agent_session_runs = pgTable(
       table.failure_code,
       table.terminal_at
     ),
+    index('IDX_cloud_agent_session_runs_responsibility_reason_terminal')
```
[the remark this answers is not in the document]
eshurakov: Fixed in f2d7b96a4. Added `.concurrently()` to the index builder and regenerated the migration (squashed into `0193_bitter_lightspeed.sql`), which now emits `CREATE INDEX CONCURRENTLY` wrapped in `COMMIT;`/`BEGIN;` statement-breakpoints matching the 0163/0170 precedent. Verified with `pnpm drizzle check`, the schema-to-snapshot drift check from `schema.test.ts`, and a full `pnpm drizzle:verify-bootstrap` run against a scratch database.

packages/worker-utils/src/cloud-agent-failure.ts
```diff
@@ -56,6 +56,198 @@ export const WORKSPACE_FAILURE_SUBTYPES = [
...
+  return { responsibility, reason };
+}
+
+function classifyWorkspaceFailure(
+  subtype: WorkspaceFailureSubtype | undefined
+): CloudAgentFailureClassification {
+  switch (subtype) {
+    case 'git_authentication_failed':
+      return classified('user', 'source_control_authentication');
+    case 'git_pack_corrupt':
```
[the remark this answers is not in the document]
eshurakov: Agreed — fixed in d069859bb. `git_pack_corrupt` now maps to `unknown`/`source_control_network` alongside the other transport-side git subtypes (clone/checkout timeout, network failed). The wrapper patterns for this subtype (`early eof`, `invalid index-pack output`, `bad object`, pack corruption) are transport/server-side symptoms rather than user repo-state, so per the PR's principle of keeping ambiguous failures `unknown` it no longer attributes the failure to the user. Added a classifier test pinning the mapping.

packages/db/src/migrations/0193_bitter_lightspeed.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE "cloud_agent_session_runs" DROP CONSTRAINT "cloud_agent_session_runs_failure_classification_check";--> statement-breakpoint
```
pandemicsyn, line 1: Must-fix: This migration drops the `failure_stage`/`failure_code` database constraint and never adds an updated replacement. `$type` does not protect direct SQL or future writers, so invalid pairs can be persisted and corrupt health aggregation. Add a replacement check covering all valid pairs, including the new `payment_required` and `model_missing` cases, and add regression coverage.

services/cloud-agent-next/src/session/safe-failure-projection.ts
```diff
@@ -69,34 +71,81 @@ export function workspaceFailureMessage(subtype: WorkspaceFailureSubtype): strin
...
+export type AssistantFailureClassification = {
+  reason: CloudAgentAssistantFailureReason;
+  safeMessage: string;
+  providerOwnership: CloudAgentProviderOwnership;
+  terminalCode?: 'payment_required' | 'model_missing';
+};
+
+export function classifyAssistantFailure(
+  source: unknown,
+  defaultProviderOwnership: CloudAgentProviderOwnership = 'unknown'
```
pandemicsyn, line 83: Must-fix: The default is `unknown`, and all production terminalization call sites use it unless the error text contains `[BYOK]`. Managed-provider 401/503 failures therefore cannot reach the new platform classifications and are excluded from the default platform view. Derive ownership from the admitted model/provider context and pass it to the classifier, with managed-provider coverage.

pandemicsyn (approved):

packages/db/src/migrations/0193_bitter_lightspeed.sql
eshurakov: That's intentional because that check grew to be super big and it gets hard to manage

services/cloud-agent-next/src/session/safe-failure-projection.ts
eshurakov: Valid — fixed in 05ddcf8.

Derivation happens at the shared terminalize boundary (`terminalizeMessageOnce` in `session-message-state.ts`) rather than at each call site: every production terminalization funnels through it, and it's where the admitted model (`admissionSnapshot.agent.model`, with the legacy fallbacks) is always loaded. The ingest call site has no model context, so a call-site fix would have needed new DO-context plumbing for the same result.

Rule: assistant failures keep `byok` when the error text carries the `[BYOK]` marker; otherwise an admitted model implies `managed` (all runs dispatch through the managed kilo provider — `normalizeKilocodeModel` prefixes `kilo/` and the kilo client defaults `providerID: 'kilo'`); no model context stays `unknown`. Non-assistant failures are untouched.

Managed 401/503 now reach `platform/managed_provider_authentication` and `platform/managed_provider_unavailable`. Covered by new unit tests at the boundary plus a supervisor-level test for the managed-503 scenario.

PR #4671, opened 2026-07-22
fix(session-ingest): preserve attachment capability in live heartbeats

iscekic:
## Summary

- project each CLI connection capability onto every live `sessions.heartbeat` session row
- preserve `true`, `false`, and absent capability semantics
- keep live heartbeat behavior aligned with `activeSessions.list`

## Why

The merged remote-attachment flow initially enables the mobile paperclip from the active-session snapshot, then the next live heartbeat disables it because the SDK reads capability from each session row while session-ingest only sent it at the envelope level.

Companion CLI PR: Kilo-Org/kilocode#12394

## Verification

- 573 session-ingest tests
- session-ingest typecheck
- session-ingest lint
- `git diff --check`
- independent review: no findings

### DIFF
--- a/apps/mobile/src/components/agents/mobile-session-manager-helpers.ts
+++ b/apps/mobile/src/components/agents/mobile-session-manager-helpers.ts
@@ -29,7 +29,7 @@ export async function buildRemoteAttachmentParts(
   const parts = await Promise.all(
     submission.files.map(async file => {
       const result = await trpcClient.cloudAgentNext.getAttachmentDownloadUrl.mutate({
-        messageUuid: submission.messageUuid,
+        messageUuid: submission.wire.path,
         filename: file.remoteName,
       });
       const mime = mimeForExtension(normalizeAttachmentExtension(file.remoteName));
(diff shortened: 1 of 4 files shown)


iscekic:
(bot) @kilocode-bot please re-review latest head 4ed258a655a9e364639c21809ac33fe1a9793e4c. Local E2E found and fixed a distinct upload-path/message-UUID R2 key mismatch, with a regression using intentionally different IDs.

iscekic:
(bot) Final local iOS E2E passed on exact head `4ed258a655a9e364639c21809ac33fe1a9793e4c` with Kilo CLI head `6750e1c9425c16ee7a2700f7bf3d259d118b0326`: capability remained available after heartbeat, `logo.png` uploaded, download presigning used the matching upload path, the CLI materialized it, and Auto Efficient / Qwen 3.7 Plus described it successfully. All 16 checks pass; Kilobot reports no issues and there are no unresolved threads.

jeanduplessis (approved):

PR #4728, opened 2026-07-23
feat(code reviewer) council hard exclusions

St0rmz1:
## Summary

Adds non-configurable, code-owned rules that prevent an automated (webhook) code
review from using the expensive Council review type for certain PRs, even when
the repo is opted in and the org is council-entitled. Draft, bot-authored, and
fork/cross-repo PRs are downgraded to a standard review. This closes the main
open-source abuse vector, where an external contributor could otherwise trigger
Council runs (and spend the org's council budget) by opening or pushing PRs.
It downgrades to standard; it does not skip the review.

## Changes

- `packages/worker-utils/src/code-review-council.ts`: add `CouncilHardExcludeReason`
  (`'draft' | 'bot-author' | 'fork-pr'`) and a pure `councilHardExcludeReason(facts)`
  that returns the first applicable reason or `null`. `determineAutomatedReviewType`
  now returns `standard` when council is unavailable, `standard` when a hard exclude
  applies, and `council` otherwise. The vetoes run after the entitlement/config/opt-in
  checks, so no config can opt a PR past them (fail closed). Extends
  `AutomatedReviewPrFacts` with `isBot` and `isFork`.
- `apps/web/.../github/webhook-schemas.ts`: parse the PR author's `user.type`
  (optional) so the bot signal comes from GitHub's authoritative account type
  rather than a login-string heuristic.
- `apps/web/.../github/webhook-handlers/pull-request-handler.ts`: build the PR facts
  (`isDraft`, `isBot` from `user.type === 'Bot'`, `isFork` from the resolved checkout
  ref, `author`), pass them into the decision, and log the exclude reason when council
  was available but vetoed.
- Tests: unit coverage for `councilHardExcludeReason` and the updated
  `determineAutomatedReviewType` (each exclude downgrades an otherwise-available
  council; a clean PR still gets council), plus handler integration tests asserting
  bot-authored and fork PRs create a `standard` review.

## Verification

- [ ] No manual testing for this PR. The change is pure decision logic covered by
  unit tests (worker-utils) and handler integration tests (web), both passing. The
  live webhook to Council execution and merge-gate flow was validated end to end in
  the predecessor PR that introduced automated Council reviews; this PR only narrows
  which PRs reach that path.

## Visual Changes

N/A

## Reviewer Notes

- Bot detection uses GitHub's `user.type === 'Bot'` and fork detection uses the
  existing checkout-ref resolution, both server-side and not author-controlled, so
  they cannot be spoofed to reach the council path.
- `draft` is already skipped earlier in the handler before this code runs, so its
  exclude is defensive and a no-op in practice; it is kept so the rule set is complete
  and self-contained.
- `councilHardExcludeReason` returns a reason (not a bool) to support surfacing "why
  no council" later without re-deriving it, and as the extension point for the next
  round of selection gates (author association, size, labels, budget), which are out
  of scope here.

### DIFF
--- a/apps/web/src/lib/integrations/platforms/github/webhook-schemas.ts
+++ b/apps/web/src/lib/integrations/platforms/github/webhook-schemas.ts
@@ -146,6 +146,10 @@ export const PullRequestPayloadSchema = z.object({
       id: z.number(),
       login: z.string(),
       avatar_url: z.string(),
+      // GitHub account type: 'User' | 'Bot' | 'Organization'. Used to hard-exclude bot-authored
+      // PRs (dependabot/renovate/etc.) from the council review path. Optional for forward/back
+      // compatibility with payloads/fixtures that omit it.
+      type: z.string().optional(),
     }),
     head: z.object({
       sha: z.string(),
--- a/apps/web/src/lib/integrations/platforms/github/webhook-schemas.ts
+++ b/apps/web/src/lib/integrations/platforms/github/webhook-schemas.ts
@@ -147,9 +147,12 @@ export const PullRequestPayloadSchema = z.object({
       login: z.string(),
       avatar_url: z.string(),
       // GitHub account type: 'User' | 'Bot' | 'Organization'. Used to hard-exclude bot-authored
-      // PRs (dependabot/renovate/etc.) from the council review path. Optional for forward/back
-      // compatibility with payloads/fixtures that omit it.
-      type: z.string().optional(),
+      // PRs (dependabot/renovate/etc.) from the council review path. GitHub always sends this on
+      // real PR webhooks; a missing value only occurs in partial payloads/fixtures. Defaults to
+      // 'User' (a regular, non-bot account) so an absent type fails open: the PR stays eligible for
+      // council rather than being treated as a bot. Explicit default over `.optional()` documents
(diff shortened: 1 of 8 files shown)


apps/web/src/lib/integrations/platforms/github/webhook-schemas.ts
```diff
@@ -146,6 +146,10 @@ export const PullRequestPayloadSchema = z.object({
       id: z.number(),
       login: z.string(),
       avatar_url: z.string(),
+      // GitHub account type: 'User' | 'Bot' | 'Organization'. Used to hard-exclude bot-authored
+      // PRs (dependabot/renovate/etc.) from the council review path. Optional for forward/back
+      // compatibility with payloads/fixtures that omit it.
+      type: z.string().optional(),
```
jrf0110: Is there an assumed type for cases where this field is omitted? Might be a good idea to add a default value

jrf0110 (approved):

apps/web/src/lib/integrations/platforms/github/webhook-schemas.ts
St0rmz1: Good call. Addressing

PR #4750, opened 2026-07-24
feat(extension): collapse long assistant code blocks

iscekic:
## Summary

Long code blocks (>15 lines) in **assistant** chat messages in the Kilo browser extension side panel now render collapsed once the message finishes streaming: an 8-line preview with a fade-out and a `Show more (N lines)` toggle. Tool panels, thinking blocks, and user messages are unchanged.

## Product decisions

- Scope: assistant-message markdown `pre` blocks only (fenced and indented uniformly; inline code untouched)
- Threshold: >15 lines (trailing-newline tolerant); default collapsed; per-block toggle; not persisted
- Streaming: while a message's own stream is active, its blocks render fully with no collapse chrome; collapse applies the moment that message's stream ends — even while later tool rounds of the same run continue — and previously finalized messages never flash open on a new run
- Appearance: 8-line preview + fade + `Show more (N lines)` / `Show less` with `aria-expanded`

## Implementation

- New pure module + `CollapsibleCodeBlock` component (`entrypoints/sidepanel/collapsible-code-block.ts(x)`); unconditional hooks so the streaming→collapsed handoff keeps hook order
- `MessageEvent` overrides ReactMarkdown's `pre` renderer for the assistant branch only, with a defensive fallback to the default renderer on unexpected child shapes
- Precise streaming signal: `streamingMessageIdAtomFamily` + core `onAssistantStreaming` start/end callback (start at the first content delta; end guaranteed by `try/finally` on success, abort, and error), wired through both turn runners; three clear points (core end signal, run `finally`, `abortConversationRun`) so stop/close/delete mid-stream never leaks a stale id
- `clearPerConversationAtoms` evicts the new family (covers a conversation whose only entry is a mid-first-run streaming id)

## Test plan

- Vitest: line counting (trailing newlines, CRLF, empty), 15/16 boundary, preview slicing, chrome decision matrix, atom eviction (incl. streaming-only conversation), core start/end/reject/non-streamed streaming-callback contract — 243 pass
- Chrome E2E (`tests/e2e/collapsible-code-blocks.test.ts`): scripted-SSE assistant message with 20-line + 18-line + 3-line fenced blocks; asserts collapsed previews, exact toggle labels, per-block independent expand/collapse with `aria-expanded`, short block plain, no horizontal overflow
- Firefox E2E: same flow mirrored in `firefox-selenium-e2e.ts` (scroll assertions omitted — no helper idiom in that suite)
- Mid-stream chrome state is unit-covered only: the scripted-SSE harness fulfills the whole body in one shot, so intermediate streaming paints are not stably observable. **Named-mock justification**: the mock is the extension's established `tests/e2e/kilo-api-fixture.ts` `mockKiloApi` harness — a real model call cannot deterministically produce an exact >15-line fenced block with a stable assertable line count and timing; every existing extension rendering E2E uses this harness per `apps/extension/AGENTS.md`.

## Gate

`verify` (typecheck + lint + format:check + 243 unit tests), `build`, `build:firefox`, `e2e:chrome` (53 passed), `e2e:firefox` (26/26), `pnpm format` — all green.

### DIFF
--- a/apps/extension/entrypoints/sidepanel/collapsible-code-block.tsx
+++ b/apps/extension/entrypoints/sidepanel/collapsible-code-block.tsx
@@ -1,0 +1,76 @@
+import { ChevronDown, ChevronUp } from 'lucide-react';
+import { useState } from 'react';
+import type { JSX } from 'react';
+import {
+  countCodeLines,
+  isCollapsible,
+  previewCode,
+  resolveCodeBlockChrome,
+} from './collapsible-code-block';
+
+interface CollapsibleCodeBlockProps {
+  readonly code: string;
+  readonly forceExpanded: boolean;
+  readonly languageClassName?: string | undefined;
+}
+
+const CodePre = ({
+  children,
+  languageClassName,
+}: {
+  readonly children: string;
+  readonly languageClassName: string | undefined;
+}): JSX.Element => (
+  <pre>
+    <code className={languageClassName}>{children}</code>
+  </pre>
+);
+
+export const CollapsibleCodeBlock = ({
+  code,
+  forceExpanded,
+  languageClassName,
+}: CollapsibleCodeBlockProps): JSX.Element => {
+  // Always call hooks: collapsible can flip false→true while streaming crosses the
+  // Threshold, and forceExpanded can flip true→false at finalize. Conditional hooks
+  // Would break order and drop expand state at that handoff.
+  const [expanded, setExpanded] = useState(false);
+  const collapsible = isCollapsible(code);
+  const chrome = resolveCodeBlockChrome({ collapsible, forceExpanded });
+
+  if (chrome === 'plain' || chrome === 'expanded-no-chrome') {
+    return <CodePre languageClassName={languageClassName}>{code}</CodePre>;
+  }
+
+  const lineCount = countCodeLines(code);
+  const displayCode = expanded ? code : previewCode(code);
+
+  return (
+    <div className="relative min-w-0">
+      <div className="relative min-w-0">
+        <CodePre languageClassName={languageClassName}>{displayCode}</CodePre>
+        {expanded ? null : (
+          <div
+            aria-hidden="true"
+            className="pointer-events-none absolute inset-x-0 bottom-0 h-10 rounded-b bg-gradient-to-t from-[rgb(9_9_11)] to-transparent"
+          />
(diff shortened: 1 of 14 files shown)


apps/extension/entrypoints/sidepanel/collapsible-code-block.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const displayCode = expanded ? code : previewCode(code);
+
+  return (
+    <div className="relative min-w-0">
+      <div className="relative min-w-0">
+        <CodePre languageClassName={languageClassName}>{displayCode}</CodePre>
+        {expanded ? null : (
+          <div
+            aria-hidden="true"
+            className="pointer-events-none absolute inset-x-0 bottom-0 h-10 rounded-b bg-gradient-to-t from-[rgb(9_9_11)] to-transparent"
```
pandemicsyn, line 55: Neon review finding (nit): The fade-out gradient uses an opaque `rgb(9 9 11)` end color, but the actual `<pre>` background (from `.agent-message-markdown :where(pre)`) is `rgb(9 9 11 / 70%)` layered over the assistant bubble's `bg-zinc-900`. The two colors are close but not identical, so the fade can show a faint seam against the real background instead of blending perfectly.

Suggested fix: Use the same semi-transparent background value (or a CSS variable shared with the `pre` rule) for the gradient's terminal color so the fade blends exactly with the block's actual background.

Generated by Neon. Edit or delete before submitting the review.

apps/extension/entrypoints/sidepanel/collapsible-code-block.tsx
```diff
@@ -0,0 +1,76 @@
...
+          />
+        )}
+      </div>
+      <button
```
pandemicsyn, line 59: Neon review finding (nit): The toggle button has `aria-expanded` but no `aria-controls`/id association with the code region it discloses, which is a minor accessibility gap for a disclosure widget (screen readers can't programmatically link the control to the controlled content).

Suggested fix: Give the `<pre>`/wrapper an id and add `aria-controls` on the button pointing to it, consistent with standard disclosure-widget patterns.

Generated by Neon. Edit or delete before submitting the review.

pandemicsyn (approved):
lgtm - couple nits inline it looks like - feel free to ignore.

PR #4751, opened 2026-07-24
feat(extension): show session cost in the context popover

iscekic:
## Summary

Shows the running total session cost next to context usage in the browser extension side panel. The existing `ContextDonut` popover gains a **Session cost** row (under the token summary, above "Compact now"), formatted `$X.XXXX` identical to the mobile app's `formatCost`.

- Client-side sum of streamed `usage.cost` from the gateway chat-completions stream; no backend changes.
- `usageSchema` accepts `cost: z.number().nullish()`: a chunk with no `cost` or `cost: null` contributes $0 and never breaks `prompt_tokens` parsing (free models, where the gateway strips cost fields, correctly read `$0.0000`).
- Cost is forwarded per completion via the existing `onUsage` (a tool-round turn sums multiple completions), accumulated into a per-conversation in-memory jotai atom with the same lifecycle as `contextUsageAtomFamily` (kept across close, evicted on delete and sign-out, reset on reload).
- Session cost survives manual and auto compaction (compaction resets only measured context usage).

## Test plan

- `pnpm --filter kilo-extension verify` — typecheck, lint, format:check, 230 vitest tests (new: stream-client cost parsing present/missing/`null`, runner per-completion forwarding incl. in-turn tool-round order, atom eviction/sign-out lifecycle, `formatSessionCost`/`addSessionCost` edge cases).
- `pnpm --filter kilo-extension e2e:chrome` — 53 passed; new E2E proves empty `$0.0000` state, in-turn tool-round accumulation (`$0.0130`), cross-turn accumulation (`$0.0140`), unchanged donut aria-label, and cost surviving manual compaction (`$0.0123`).
- `pnpm --filter kilo-extension build` / `build:firefox` / `e2e:firefox` — green (existing scenarios; the new row has Chrome-only E2E, matching the existing context-usage coverage).

### DIFF
--- a/apps/extension/entrypoints/sidepanel/agent-chat-panel.tsx
+++ b/apps/extension/entrypoints/sidepanel/agent-chat-panel.tsx
@@ -24,6 +24,7 @@ import {
   compactConversationEvents,
   hasCompactableHistory,
 } from '@/src/shared/agent-context-compaction';
+import type { TurnUsage } from '@/src/shared/agent-llm-turn-runner-core';
 import { defaultMode } from '@/src/shared/agent-chat-placeholder';
 import { getKiloApiBaseUrl } from '@/src/shared/auth';
 import type { StoredAuth } from '@/src/shared/auth';
@@ -474,7 +475,7 @@ export const AgentChatPanel = ({
       }
     };
     let currentRunHasUsage = false;
-    const updateRunUsage = (usage: { costUsd?: number; promptTokens: number }): void => {
+    const updateRunUsage = (usage: TurnUsage): void => {
       if (isCurrentRun()) {
         currentRunHasUsage = true;
         store.set(contextUsageAtomFamily(conversationId), { promptTokens: usage.promptTokens });
--- a/apps/extension/entrypoints/sidepanel/agent-chat-panel.tsx
+++ b/apps/extension/entrypoints/sidepanel/agent-chat-panel.tsx
@@ -10,6 +10,7 @@ import {
   evictConversationAtoms,
   remoteMcpStoreAtom,
   runningConversationIdsAtom,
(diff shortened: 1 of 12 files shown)


apps/extension/entrypoints/sidepanel/agent-chat-panel.tsx
```diff
@@ -469,10 +474,15 @@ export const AgentChatPanel = ({
       }
     };
     let currentRunHasUsage = false;
-    const updateRunUsage = (usage: { promptTokens: number }): void => {
+    const updateRunUsage = (usage: { costUsd?: number; promptTokens: number }): void => {
```
pandemicsyn: Neon review finding (minor): The inline usage type `{ costUsd?: number; promptTokens: number }` duplicates the `TurnUsage` interface exported from `agent-llm-turn-runner-core.ts`. This is a style nit, not a correctness issue, but future usage-field additions (e.g. completion tokens) risk drifting between the two type definitions since one isn't derived from the other.

Suggested fix: Import and reuse `TurnUsage` (or `OnTurnUsage`) from `@/src/shared/agent-llm-turn-runner-core` instead of re-declaring the shape inline for `updateRunUsage`.

pandemicsyn (approved):
lgtm

apps/extension/entrypoints/sidepanel/agent-chat-panel.tsx
iscekic: (bot) Fixed in 92bb4c12e — `updateRunUsage` now takes the exported `TurnUsage` interface from `@/src/shared/agent-llm-turn-runner-core` instead of re-declaring the shape inline, so future usage fields stay in sync.

PR #4763, opened 2026-07-24
fix(mobile): PR sheet chrome + session UX defect batch; sandbox CLI 7.3.63 for routed-model stamps

iscekic:
## Summary

Mobile UX defect batch (7 reproduced defects) + GitHub rate-limit investigation (report-only).

- **R1/R2 — PR formSheets**: the comment-composer, review-submit, and merge sheets did not paint their (already-existing) headers; a stray leading-edge `‹` floated over content; with the keyboard open the CTA footer rendered fully behind the keyboard. Root cause (device-backed): react-native-screens formSheet only honors a pinned header when the screen content's direct children are `[header, ScrollView]` — the old `View → ScrollView + sticky footer` tree made the ScrollView full-bleed over the header; the stray `‹` was the parent PR screen's "Go back" showing through a transparent top band. Fix: shared pinned `PrFormSheetHeader` outside the ScrollView, footer CTAs as trailing scroll content with `automaticallyAdjustKeyboardInsets` + `AppAwareKeyboardPaddingView`, plus density passes so all CTAs sit above the keyboard at both detents without scrolling.
- **R4 — file navigator**: long paths overlapped the viewed toggle. Fix: `min-w-0` chain, truncating shrinkable directory segment, non-shrinking basename, non-shrinking toggle with a guaranteed gap.
- **"Paste a link" hint**: removed the redundant helper state (the input placeholder already shows the example URL); dead slot state pruned.
- **R5 — new-session composer**: paperclip + mic rendered mid-height flanking the input. Fix: full-width input on top, one bottom action row (paperclip leading, voice trailing; voice-unavailable shows paperclip only).
- **R6 — cloud-agent preparation**: the fixed `Importing session…` footer row duplicated the in-transcript PreparationGroup. Fix: `shouldShowSessionFooterRow` gate — while `preparing`, suppress the footer only when the merged transcript carries a visible (non-no-op) preparation item; otherwise the footer keeps rendering (never a blank progress window).
- **R7 — per-message model label**: removed entirely; message info is now on long-tap via a new message-details sheet (Copy with immediate clipboard write, role, sent time, model row, cost/tokens block; VoiceOver/TalkBack rotor copy action unchanged).
- **R8 — auto model in Models section**: display layer — the context sheet's Models section never renders `kilo/kilo-auto/*` rows (render-only filter; totals/residual untouched; count derives from filtered rows). Data path — root cause was the sandbox-pinned CLI `7.3.54` predating the routed-model stamp feature (first released in `7.3.63`); bumped the pin in lockstep (Dockerfiles, wrangler `image_vars`, `KILO_CLI_VERSION`) and re-proved live: fresh `kilo-auto/efficient` sessions now stamp `step-finish.model` with the concrete routed id (e.g. `qwen/qwen3.7-plus`) at the ingest boundary and in the app-facing payload.

R3 (file-navigator missing header) was dropped: could not be reproduced — already fixed on baseline.

## GitHub rate-limit investigation (report only — no code changes)

| Claim | Evidence |
|---|---|
| All PR-review GitHub calls use per-user GitHub App user access tokens: the token response schema requires `expires_in` + `refresh_token` and refresh uses `grant_type: 'refresh_token'` — the GitHub App user-to-server pattern; each call builds an Octokit with the fetched per-user token; all 16 procedures wrap `withGitHubUserTokenRetry`, so the rate-limit bucket is per user and scales with user count; no installation-token/shared-bucket path exists in this feature | token schema + exchange `apps/web/src/lib/integrations/platforms/github/user-authorization.ts:13-18,72-93`; refresh grant `services/git-token-service/src/github-user-authorization-service.ts:502-511`; per-call fetch+Octokit `apps/web/src/lib/github-pr-review/retry.ts:63-70` + `client.ts:14-19`; 16 call sites `github-pr-review-router.ts:509-956` |
| GitHub-documented primary limits for user-to-server tokens: 5,000 REST req/h + 5,000 GraphQL points/h per user (EXTERNAL doc figure, not in-repo) | https://docs.github.com/en/rest/rate-limit + https://docs.github.com/en/graphql/overview/rate-limits-and-node-limits-for-the-graphql-api |
| Unbounded concurrent GraphQL burst: thread-comment follow-ups for up to 50 threads fire in one `Promise.all` | `github-pr-review-router.ts:653-673` + loop `399-429` |
| Tight pagination with no pacing | `apps/mobile/src/lib/pr-review/diff/pr-review-file-list-state.ts:197-204`; `octokit.paginate` in `listChecks` (`apps/web/src/routers/github-pr-review-router.ts:570-581`); `onEndReached` auto-pagination (`apps/mobile/src/components/pr-review/diff/pr-diff-file-list.tsx:343-347`) |
| Mergeability poller: each 3s tick = 1 tRPC overview refetch = 3 GitHub API requests server-side (pulls.get + repos.get + GraphQL), up to 10 ticks | `apps/mobile/src/components/pr-review/merge/pr-merge-section.tsx:49-50,131-142`; fan-out `apps/web/src/routers/github-pr-review-router.ts:515-537` |
| Invalidation multiplier per mutation family: comment/submit → overview + ALL loaded thread pages; merge/update/auto-merge → overview + checks + ALL loaded file pages; reply/resolve/reactions → thread pages only | `apps/mobile/src/lib/pr-review/use-pr-review-mutations.ts:34-41`; `apps/mobile/src/lib/pr-review/merge/use-pr-merge-mutations.ts:31-40`; `apps/mobile/src/lib/pr-review/discussion/use-review-discussion-mutations.ts:55-60` |
| Mobile auto-retries `TOO_MANY_REQUESTS` 2× against an exhausted bucket | `apps/mobile/src/lib/query-client.ts:7-26` (permanent-code set omits it) |
| Headerless 403 secondary limit misclassified as terminal `FORBIDDEN` | `apps/web/src/lib/github-pr-review/errors.ts:49-53,163-171` |
| `retryAfterEpochMs` dropped at the tRPC boundary | `apps/web/src/lib/github-pr-review/retry.ts:22-27` |
| No conditional requests/ETag; rate-limit headers read only after failure | `apps/web/src/lib/github-pr-review/errors.ts:146`; repo-wide grep confirms no If-None-Match in GitHub paths |
| `getFileLines` downloads whole raw file per 500-line window, no cache | `apps/web/src/routers/github-pr-review-router.ts:611-637`; `apps/mobile/src/lib/pr-review/diff/use-pr-diff-context-loader.ts:61-68` |

**Recommendations (no code in this PR):** bound GraphQL follow-up concurrency + pace pagination; plumb `retryAfter` to the client and suppress auto-retry on 429; classify headerless 403 secondary limits as retryable `TOO_MANY_REQUESTS`; consider ETag caching for overview/files.

## Notes

- The sandbox CLI pin bump (7.3.63) takes effect for new sandboxes; production images pick it up on the next image build/publish. Existing sandboxes stay on 7.3.54 until recycled.
- `default-slash-commands.generated.ts`: source marker bumped for the lockstep test; the catalog body was not regenerated (the generator's `kilo serve` step hung locally with 7.3.63) — content drift vs 7.3.63 is possible and tracked as follow-up.
- `providerMetadata.kilocode.routedModelID` is not emitted by the 7.3.63 stamp path; consumers read `part.model` (`{providerID, modelID}`), which `cloud-agent-sdk/part-utils#getStepFinishRoutedModel` already does.

### DIFF
--- a/apps/mobile/src/components/pr-review/merge/pr-merge-sheet-parts.tsx
+++ b/apps/mobile/src/components/pr-review/merge/pr-merge-sheet-parts.tsx
@@ -1,10 +1,15 @@
 // Form sub-components for the S8 merge sheet. Extracted out of
 // `pr-merge-sheet.tsx` to keep that file under the repo's 300-line limit.
 
-import { type RefObject } from 'react';
-import { Switch, TextInput, View } from 'react-native';
+import { type ReactNode, type RefObject } from 'react';
+import { Pressable, Switch, TextInput, View } from 'react-native';
+import * as Haptics from 'expo-haptics';
 
-import { PillGroup } from '@/components/security-agent/settings-pill-group';
+import {
+  PrFormSheetFooter,
+  useFormSheetKeyboardVisible,
+} from '@/components/pr-review/pr-form-sheet-chrome';
+import { Button } from '@/components/ui/button';
 import { Text } from '@/components/ui/text';
 import { useThemeColors } from '@/lib/hooks/use-theme-colors';
 import { cn } from '@/lib/utils';
@@ -13,8 +18,19 @@ import {
   PR_MERGE_DESCRIPTIONS,
 } from '@/lib/pr-review/merge/merge-blocked-reasons';
 import { type MergeMethodOption } from '@/components/pr-review/merge/pr-merge-icons';
+import { PrReviewReconnectNotice } from '@/components/pr-review/pr-review-reconnect-notice';
 
-export function MethodPicker({
+function shortMethodChipLabel(value: AllowedMergeMethod): string {
+  if (value === 'merge') {
+    return 'Merge';
+  }
+  if (value === 'squash') {
+    return 'Squash';
+  }
+  return 'Rebase';
+}
+
+function MethodPicker({
   methodOptions,
   method,
   isDisabled,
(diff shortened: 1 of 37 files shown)


iscekic:
(bot) @kilo-code-bot please review the latest head (57e0e8d6a) — the branch was updated with a clean merge of main.

apps/mobile/src/components/pr-review/merge/pr-merge-sheet-parts.tsx
```diff
@@ -81,19 +120,24 @@ export function CommitTitleField({
...
 }
 
-export function CommitMessageField({
+function CommitMessageField({
   messageRef,
   inputRef,
   isDisabled,
+  /** Half-detent / keyboard-open: keep a single-line-ish field so footer CTAs fit. */
+  compact = false,
 }: Readonly<{
```
pandemicsyn, line 129: `useFormSheetKeyboardVisible()` is now called independently in both `PrMergeSheet` (for `isHalfDetent`) and `CommitMessageField` (for `tight`), each registering its own pair of `Keyboard` listeners. Functionally harmless (both derive the same boolean) but it duplicates listener setup/teardown on every mount and is easy to lose sync with if the two call sites ever diverge in show/hide event names.

Suggested fix: Thread the keyboard-visible boolean down from `PrMergeSheet` as a prop instead of re-subscribing inside `CommitMessageField`, matching the single-source pattern already used for `isHalfDetent`.

pandemicsyn (commented):
lgtm, one nit inline from the robot.

pandemicsyn (approved):

PR #4769, opened 2026-07-25
Fix mobile attention, model-selection and PR-review defects

iscekic:
Fixes four defects found by an end-to-end verification run on iOS against a local stack, each confirmed by an independent code triage before any code was written.

## What was broken

**A live CLI session waiting on you showed no NEEDS INPUT badge.** Live rows take their status from the CLI heartbeat, which only ever reports `idle`/`busy`. The correct `question`/`permission` status was already in `cli_sessions_v2`, already fetched into the same hook, and discarded for live ids — so the badge could never reach the one state it exists for.

**Choosing a model in a cloud-agent session was silently ignored.** The write path stored React state plus a per-org and a per-user preference; the read path resolved from the session's stored `agent.model`, the streamed message model, or the first catalog entry. No overlap, and nothing wrote a per-session model. Sessions with a stored `agent.model` had the pick reverted within a frame, before any send could use it.

**The attention badge cleared on viewing rather than answering.** Opening a blocked session and backing out without answering cleared the badge while the CLI was still waiting — and because the ack store is in-memory only, it reappeared after an app restart. The app gave two different answers to the same question depending on whether the process had restarted.

**A CLI that died with a question pending left the row pinned at `question` forever**, producing a permanent NEEDS INPUT badge on a session that opens to an empty transcript. A row found locally had been stuck for ten hours.

Plus a copy fix: submitting *Request changes* on your own PR reported that you can't *approve* your own PR.

## Notable implementation points

The live-status overlay resolves **in the tRPC router**, not only in client helpers. The live cache refreshes with `fetchQuery`, which replaces the cached result wholesale, so sticky state held only on the client is wiped on every enrichment, reconnect and `cli.connected` refresh. Serving the resolved value also means this works for **already-released CLIs** that will never report attention on the heartbeat — it does not depend on the companion CLI change in Kilo-Org/kilocode.

The disconnect reset targets `retry`, not `idle`. `idle` is the only status the ingest attention path treats as completed, so a synthetic `idle` would risk a false "Task completed" push on every disconnect. The reset is awaited **before** `cli.disconnected` is broadcast, because the client reacts to that broadcast by refetching stored history and has no handler for `session.status.updated` on that cache — a fire-and-forget write would race the refetch.

Cloud-agent model persistence happens after **successful admission**, not at resolve time; persisting at resolve would record a model for a run that never started. Selecting a model without sending still falls back after leaving the screen — intended, since the chip reflects what actually ran, matching existing remote behaviour.

## Deliberately not changed

The `CLOUD` eyebrow (it is the repo name, not a platform badge), the focus-gated in-session model selector (it renders when the composer is focused), plain-text questions not raising needs-input (only structured `question`/`permission` events are attention states), and PR-review comment path assignment (paths are captured at line selection; a unit test already asserts it). All four were reported as bugs during verification and disproved by triage.

Rows already stuck at `question` from earlier disconnects are not backfilled — that is a separate one-off cleanup.

## Checks

`apps/mobile` typecheck, lint, format, knip, 1871 tests · `session-ingest` typecheck, 594 tests · `cloud-agent-next` typecheck, 2359 tests · `apps/web` typecheck, 9376 tests. Independent reviewer over the full diff: no findings.

Device E2E is running separately and will be reported in a comment.

### DIFF
--- a/apps/mobile/src/app/(app)/agent-chat/[session-id].tsx
+++ b/apps/mobile/src/app/(app)/agent-chat/[session-id].tsx
@@ -14,7 +14,6 @@ import { Button } from '@/components/ui/button';
 import { Text } from '@/components/ui/text';
 import { shouldRetryNotFoundOnSpawnedRoute } from '@/lib/spawned-not-found-retry';
 import { useTRPC } from '@/lib/trpc';
-import { useAckSessionAttentionOnOpen } from '@/lib/session-attention';
 
 export default function SessionDetailScreen() {
   const {
@@ -40,7 +39,6 @@ export default function SessionDetailScreen() {
   }>();
   const trpc = useTRPC();
   const router = useRouter();
-  useAckSessionAttentionOnOpen(sessionId);
   const sessionQuery = useQuery({
     ...trpc.cliSessionsV2.get.queryOptions(
       { session_id: sessionId },
--- a/apps/mobile/src/components/agents/session-detail-content.tsx
+++ b/apps/mobile/src/components/agents/session-detail-content.tsx
@@ -136,6 +136,7 @@ export function SessionDetailContent({
   const remoteModelState = useAtomValue(manager.atoms.remoteModelState);
   const observedModel = useAtomValue(manager.atoms.observedModel);
   const remoteModelOverride = useAtomValue(manager.atoms.remoteModelOverride);
+  const cloudAgentModelOverride = useAtomValue(manager.atoms.cloudAgentModelOverride);
   const availableCommands = useAtomValue(manager.atoms.availableCommands);
   const remoteCommandState = useAtomValue(manager.atoms.remoteCommandState);
   const contextUsage = useAtomValue(manager.atoms.contextUsage);
@@ -271,6 +272,7 @@ export function SessionDetailContent({
     modelOptions,
     selectedModel: sessionModels.selectedValue,
     selectedVariant: sessionModels.selectedVariant,
+    cloudAgentModelOverride,
   });
 
   const viewTrackedRef = useRef<string | null>(null);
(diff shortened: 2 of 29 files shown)


iscekic:
(bot) Device E2E complete on `426e13be0`. All acceptance criteria for this PR pass on an iOS simulator against a local stack.

| Defect | Result | Evidence |
|---|---|---|
| DEF-2 model selection | PASS | Chip holds `Auto Balanced`, survives leave/re-enter and cold relaunch; `model: 'kilo-auto/balanced'` in the cloud-agent-next log for the run that actually executed |
| DEF-3 live-row overlay | PASS | With the **stock released CLI** (which reports `busy` on the wire while the DB says `question`), the amber NEEDS INPUT badge appears, survives ~50s of heartbeats without flicker, and survives a tRPC refetch |
| DEF-4 ack-on-answer | PASS | Open + back out without answering keeps the badge, and it is still there after a cold relaunch; answering clears it |
| DEF-5 disconnect reset | PASS | Hard-kill with a pending question → status `retry` within ~1s, badge clears with no pull-to-refresh, other connection's sessions untouched, no completion push |
| DEF-6 422 copy | PASS | Request-changes on an own PR now reads "you can't **review** your own pull request", pending comments retained, switching to COMMENT still submits |

Two notes worth recording.

**DEF-5 initially failed E2E and the root cause was broader than this feature.** The reset code, its ordering and its identity handling were all correct — it was never reached. Under workerd, `ctx.getWebSockets()` still includes the closing socket during `webSocketClose`, so the stale-reconnect guard matched *itself* and returned early. That early return also skips ownership cleanup and the `cli.disconnected` broadcast, so on any real disconnect none of that ran either. The `Stale CLI socket closed (already replaced)` line had been appearing on ordinary disconnects; in hindsight that was the symptom. Fixed in `426e13be0`, with a test that leaves the closing socket listed as workerd does.

**The overlay is worth more than "covering old CLIs".** Even with a patched CLI, the DB flips to `question` within milliseconds while the heartbeat only fires every ~10s. The overlay covers that window too, so the two fixes complement each other rather than one superseding the other.

jeanduplessis (approved):

PR #4773, opened 2026-07-26
fix(mobile): agents session-list scroll behavior + session platform icons

iscekic:
## Summary

Batch B of the Mobile Agents session-list UX work — two confirmed defect fixes, one scroll-behavior hardening, and one feature, shipped as four coherent commits:

1. **Back-navigation scroll reset (defect, reproduced at the gate).** Scrolling deep into the Agents list, opening a session, answering a live question (ack), and going back reset the list to the top. Root cause: the `SectionList` was keyed on `${sortBy}:${attentionFocusRevision}`, so any attention-revision change while the frozen list was away remounted it on return and dropped the scroll offset. The remount was introduced by #4605 to refresh ack state in frozen cells, but rows already self-subscribe via `useSessionAttentionRevision()`. The list now keys only on `sortBy` (sort changes still remount, as before) and passes the focus-snapshotted attention revision as `extraData`, which re-renders visible cells after unfreeze without remounting.
2. **Agents tab re-tap scrolls to top (defect, reproduced).** Re-tapping the active Agents tab did nothing. Wired `useScrollToTop` from `@react-navigation/native` to the list ref — standard iOS tab-bar behavior, animated, composing with the existing haptics-only `tabPress` listener.
3. **Scroll-to-top on committed search query (hardening; the reported defect did not reproduce at the gate).** The repro gate could not confirm "committing a new search query resets scroll to top" as a baseline defect, and the gate evidence was inconclusive for the list content component, which has no baseline mechanism that resets the offset on query commit (no ref, no effect, no remount path). The behavior is therefore shipped explicitly: a `prevSearchQueryRef` effect scrolls the list to top (non-animated) when the committed, debounced query changes — never on initial mount, focus refetch, attention revision, sort remount, pagination, pull-to-refresh, or section-data identity changes with an unchanged query. The selection decision lives in the pure `shouldResetScrollOnCommittedQuery()` helper with unit coverage; the acceptance E2E runbook carries the search regression assertion.
[...]

PR #4772, opened 2026-07-26
fix(mobile): loading and skeleton polish

iscekic:
## Summary

Mobile loading polish, three items:

1. **Shared `Skeleton` shimmer** — a soft horizontal gradient sweep (`expo-linear-gradient`) is layered over the existing opacity pulse. The sweep is driven by one refcounted module-level Reanimated clock shared by every instance, so the sweep costs a single UI-thread animation no matter how many skeletons are on screen (each instance also keeps its per-instance pulse, now at a faster 1700 ms cadence). Reduce Motion renders the same static muted block as before. API unchanged: `Skeleton({ className })`.
2. **Session transcript pagination: no visible jump when an older page arrives** — the list header's transient loading skeleton (~80pt) collapsed in the same update that prepended the older page. FlashList's `maintainVisibleContentPosition` compensates prepended items but not the header height change, which produced a measured ~43pt upward jump of on-screen content at every page boundary (60fps frame analysis on the baseline). The placeholder is now suppressed at the render-model layer (an already-visible omitted-items banner stays stable through the load), and the fetch starts two viewports before the boundary so the prepend usually lands before the user reaches it. The identical pattern in the kilo-chat message list is fixed the same way, with the dead `isFetchingOlder` prop removed.
3. **Notifications settings loader matches the final layout** — category placeholders are now structurally identical to the settled rows (same container classes, content-sized skeleton lines, native-switch-sized 31×51 block), and the master Switch / "Enable notifications" CTA are gated on a new pure `deriveGateSettled` predicate (permission settled AND, when granted, both push-token queries fetched-or-errored). A fully enabled user no longer sees a transient Enable card or a Switch flashing a wrong value while the token queries settle.

## Test plan

**Automated:** full mobile suite green (226 files / 1881 tests), including the new `deriveGateSettled` truth table and updated pagination-header render-model tests. `oxfmt`, `tsgo`, `oxlint`, `knip`, `git diff --check` clean; `expo-doctor` introduces no new findings (the out-of-date-packages advisory pre-exists on main).

**On-device (iPhone 17 Pro simulator, dev client built from this branch):**

| Criterion | Result |
|---|---|
| Shimmer sweep moves across skeletons (frame analysis) | PASS — highlight at ≥3 distinct x-positions; no frame freeze |
| Reduce Motion → static skeletons; setting restored | PASS |
[...]

PR #4778, opened 2026-07-26
fix(mobile): stable active-session order, unified list scroll, reachable mono blocks

iscekic:
Three mobile fixes, stacked on `fix/mobile-loading-polish` (#4772).

**Base note:** this branch also carries a snapshot merge of `feat/mobile-session-list-ux` (#4773, open) at `0affde149`, because the unified-scroll item rebuilds the container its scroll behaviors attach to. The PR diff transiently shows #4773's commits (including its two small `apps/web` schema/SDK lines) until the bases merge; GitHub collapses them. This branch's own changes are under `apps/mobile/` plus one `apps/web` fix in `cloud-agent-sdk/session-manager.ts` (item 4).

## 1. Stable "Active now" ordering (`d8bc2405f`)

The Agents "Active now" tray and Home "Agent sessions" section reshuffled as data refreshed: nothing in the pipeline (ingest worker, router, live-sync updater, hook, consumers) imposed an order, and any refetch or websocket `setQueryData` could permute the array. Fixed once at the shared read boundary (`useAgentSessions`), which both surfaces inherit.

New pure `sortActiveSessionsByCreatedAt`:

1. Rows with no parseable `createdAt` sort **first**, tie-broken by `id` ascending.
2. Enriched rows sort by `createdAt` **descending** (newest first).
3. Equal `createdAt` → `id` ascending — a total order, so output never depends on input order (no reliance on `Array.prototype.sort` stability; input array never mutated).

**The precise stability claim (do not overstate):** a row lacking `createdAt` is placed at the top provisionally. When enrichment supplies `createdAt` (~1–2s later), the row moves to its true position — **at most one reposition, within one enrichment cycle of the row appearing.** For a brand-new session that position is already the top, so it does not move; a reactivated or orphaned session may move down once. **No row that has had a `createdAt` for longer than one enrichment cycle ever moves.** The reported defect (rows reshuffling on refresh) is that last sentence, and it is fully fixed; the bounded one-time reposition of a just-appeared row is an accepted residual, pinned by unit tests (A12/A13). A session the worker reports but that has no `cli_sessions_v2` row never gets a `createdAt` and pins to the top indefinitely — accepted as a deviation: it is perfectly stable, and burying a live session at the bottom is strictly worse UX. Reconciling worker/DB divergence is out of scope.

The comparator keys strictly on a **parseable `createdAt`** (via `parseTimestamp`; unparseable is treated as missing, never epoch 0) and deliberately not on `isEnriched`'s any-of-three-fields heuristic. Sorting at the client read boundary rather than the router is required because the live-sync owner mutates the query cache out-of-band and would overwrite any server-side `ORDER BY`.

Unit tests cover tie determinism, reversed-input identity, non-mutation, the unenriched buckets, and the two permitted one-time moves.

## 2. One continuous scroll on the sessions list page (`d3f30c526`)

The "Active now" tray was pinned above the history list, permanently consuming viewport. It now renders as the history `SectionList`'s `ListHeaderComponent` and scrolls away with it in one continuous gesture. `ScreenHeader`, filter chips, and the search header stay pinned above the list.

Single render site (mandatory): the `isLoading` early return is gone — loading renders the same `SectionList` with empty sections and skeletons as `ListEmptyComponent`, so the tray's local `+N more` expanded state survives the loading→rows transition. The remaining early returns (full-screen error, first-use empty) are gated on `!isLoading` so a cold open with an empty cache cannot flash "No sessions yet"; `ListEmptyComponent` is `isLoading ? skeletons : body-empty` so a tray-first load shows skeletons under the tray instead of "No past sessions". The gating decision lives in a pure `selectSessionListContentSurface` selector with focused tests.

Preserves #4773's behaviors: scroll restoration on back-navigation (baseline captured RESTORED pre-change), committed-search reset-to-top, tab re-tap scroll-to-top, pull-to-refresh and pagination — re-verified on device in this PR's E2E.

## 3. Reachable monospace blocks (`f674ff734`)

Long lines in tool-card mono blocks were cut at the right edge and **unreachable**: Yoga shrank the mono `Text` to the card width, so the horizontal `ScrollView`'s content equaled its viewport and there was nothing to scroll (drags did nothing). All 13 horizontal-scroll sites (plus the preparation output tail) are deduped into one `MonoScrollBlock`:

- `shrink-0` mono text keeps intrinsic width so the scroll has real content to deliver (the actual fix, not an affordance-only patch).
- Visible horizontal scroll indicator when content overflows.
- Gesture coexistence: `directionalLockEnabled` (iOS), `nestedScrollEnabled` (Android), `canCancelContentTouches` — a mostly-vertical drag keeps scrolling the conversation; a horizontal pan scrolls the block.
- Height pinned from the content's measured layout (keyed to the displayed text so grown content remeasures) to avoid the RN 0.83 Fabric horizontal-ScrollView ~10× height-mismeasure hazard; pre-change device baseline recorded (expanded read-card mono viewport 152pt on a 375×667pt screen) and re-measured in E2E.
- A visible "Truncated" marker when the existing 1000/2000-char cap applies (the cap was previously silent).

Command text (bash `$ …`, preparation setup command) intentionally stays soft-wrapped; markdown code fences are untouched (out of scope, they wrap intentionally); `kilo-chat/**` has zero diff.

## 4. Session-open empty flash — reproduced and fixed (`bc44e50ea`, `a6bc40301`)

A fourth reported defect — opening an in-progress session flashing an empty state — initially did not reproduce on the unmodified baseline and was dropped. That conclusion was wrong: the race only becomes visible when the messages page fetch is slower than the first CLI heartbeat. A committed E2E latency harness (`bc44e50ea`; `E2E_LATENCY_*` env knobs delaying tRPC queries / WS events, inert unless set) reproduces it on demand at 3s messages delay: skeleton (~0.5s) → **"No messages yet" (~3s)** → transcript, captured frame-level on video.

Root cause in `cloud-agent-sdk/session-manager.ts` (`a6bc40301`): `onFirstActivity` cleared `isLoading` on the first CLI heartbeat while the snapshot page fetch was still in flight. Fix: while a remote session's initial history replay is pending, `onFirstActivity` keeps `isLoading`; `onReplayComplete` (fired by the transport on every terminal snapshot outcome, success or failure) owns the clear. Regression tests cover the race plus the fallback and non-remote paths (171/171 green). This is the one `apps/web` change on this branch.

Device verification on this head with the same 3s delay in place (fresh verifier agent): non-empty session opens skeleton → transcript directly with zero "No messages yet" frames; a genuinely-empty session still reaches "No messages yet" (~3.5s, no skeleton hang). Evidence in the comments.

## Verification

- `pnpm format && pnpm typecheck && pnpm lint && pnpm check:unused && pnpm test` — green (230 files / 1924 tests).
- Device E2E on iPhone SE (iOS 26.5) against a local stack with three concurrent live CLI sessions: ordering across pull-to-refresh / websocket status changes / the enrichment window (A1–A5), unified-scroll behaviors (C1–C7), and mono-block reachability / gesture coexistence / height-parity (D1–D11) — results in the comments.
- Session-open fix E2E at `a6bc40301` with 3s messages latency in place: skeleton → transcript directly (zero empty-state frames) for a non-empty session; genuine-empty session still shows "No messages yet" — results in the comments.


iscekic:
(bot) E2E interim result (iPhone SE, iOS 26.5, local stack, 3 live CLI sessions): **Item 1 (ordering) ACCEPT** — A1-A5/A12 pass on both surfaces (order stable across pull-to-refresh ×3, websocket status changes, and the enrichment window; new sessions appear at top and stay). **Item 3 (unified scroll) ACCEPT** — C1-C7 pass (scroll restoration identical offsets, search/tab-re-tap to top, continuous tray+history scroll, expand survives refresh, pagination works expanded). **Item 4 (mono blocks) REJECT on D2** — wide mono content still lays out at viewport width (hierarchy shows no node wider than 375pt), so horizontal pans reveal no line tails; D1/D5/D6/D8/D9/D10 pass. Diagnosis: on Fabric the ScrollView content container stretches the text to viewport width so `shrink-0` has no effect; repair in flight using the `self-start` intrinsic-sizing pattern. Scoped re-verification of D2/D3/D4/D11 follows the repair.

iscekic:
(bot) E2E final result (iPhone SE, iOS 26.5, local stack, 3 live CLI sessions) — head `68ba3dee9`

**Item 1 (deterministic active-session order): ACCEPT** — A1–A5, A12 pass.
**Item 3 (unified session-list scroll): ACCEPT** — C1–C7 pass.
**Item 4 (reachable mono blocks): ACCEPT** — D2–D6, D10, D11 pass.

The interim D2 rejection (horizontal pans never reached the inner scroller — RN 0.83 Fabric nested-scroll negotiation failure inside the session FlashList) is resolved by `68ba3dee9`: the mono scroller now uses `react-native-gesture-handler`'s ScrollView with directional activation offsets (`activeOffsetX`/`failOffsetY` ±10). Scoped re-verification on device: wide output tails reachable (`part079`, `dir59/file.ts`, `…TAIL00_END00…TAIL04`), long-line edit diffs reachable with column alignment intact, vertical drags fail over to the conversation list, block horizontal offset preserved through vertical scrolls, block height materially unchanged vs pre-fix baseline (no 10× inflation), virtualization and pagination-to-top intact with multiple wide blocks on screen.

iscekic:
(bot) E2E follow-up at `a6bc40301`: the session-open empty flash is now **reproduced, fixed, and verified on device**.

**Reproduction (pre-fix, `68ba3dee9`)** — the earlier "did not reproduce" conclusion was wrong: it only becomes visible when the messages page fetch is slower than the first CLI heartbeat. With a new E2E latency harness (`E2E_LATENCY_MESSAGES_MS=3000`, delays `cliSessionsV2.getSessionMessagesPage` by 3s, committed as `bc44e50ea`), frame-level video of opening the live "CLI output test" session shows exactly the reported sequence: skeleton (~0.5s) → **"No messages yet" (~3s)** → transcript. Root cause traced to `cloud-agent-sdk/session-manager.ts`: `onFirstActivity` cleared `isLoading` on the first CLI heartbeat while the snapshot page fetch was still in flight, and nothing re-armed it.

**Fix (`a6bc40301`)** — while a remote session's initial history replay is pending, `onFirstActivity` no longer clears `isLoading`; `onReplayComplete` (which the transport fires on every terminal snapshot outcome, success or failure) now owns the clear. Regression tests cover the race and the fallback/non-remote paths; full `session-manager.test.ts` 171/171 green, `apps/web` typecheck green. Fresh reviewer pass over the fix diff: no valid findings.

**On-device verification (this head, same 3s delay in place, fresh verifier agent, iPhone SE iOS 26.5 simulator):**
- Non-empty session open ("CLI output test", ACTIVE NOW): skeleton → transcript **directly**; Vision OCR on every 4fps frame in the critical window found **zero** "No messages yet" frames; transcript content confirmed rendered. **PASS**
- Genuine-empty session open (`ses_061480d9…`, `history: null`): skeleton → "No messages yet" appears ~3.5s after open (does not hang on skeleton). **PASS**

Verdict: **VERIFIED**. No manual post-merge actions.

iscekic:
(bot) Standin code review at `a6bc40301` (Kilobot review is unavailable; a fresh `mobile-reviewer` agent reviewed `git diff origin/main...HEAD` instead, prioritizing this branch's own commits).

**Verdict: NO VALID FINDINGS.**

Reviewed all six own commits:

| Commit | Scope |
|---|---|
| `d8bc2405f` | `sortActiveSessionsByCreatedAt` + `useAgentSessions` boundary |
| `d3f30c526` | Active-now as `ListHeaderComponent` + `selectSessionListContentSurface` |
| `f674ff734` | `MonoScrollBlock` + tool-card / preparation adoption |
| `68ba3dee9` | RNGH `ScrollView` + `activeOffsetX` / `failOffsetY` |
| `bc44e50ea` | E2E tRPC/WS latency harness |
| `a6bc40301` | cloud-agent-sdk `isLoading` vs `remoteHistoryReplaying` |

Verified: latency harness is inert when env knobs are unset (`e2eFetch === fetch`, early WS return); base branches (#4772/#4773) contribute no blockers in the stacked surface; mobile unit suite green (1925 passed). Residual non-finding: the E2E-only URL prefix rule also matches other `cliSessionsV2.get*` procedures when latency env is set — harmless, production path stays identity.

This was the final review pass; no further review rounds are queued unless the head changes.

jeanduplessis (approved):

PR #4786, opened 2026-07-26
test(mobile): allow a locally built CLI in the E2E remote-cli harness

iscekic:
## What

`apps/mobile/e2e/remote-cli.sh` gains an opt-in `KILO_CLI_BIN` mode: when set to a locally built kilo binary, `prepare_env` skips the npm install of `@kilocode/cli`, prepends the binary's own directory to the `PATH` exported into `.cli-env` (the built CLI keeps sibling assets next to the executable), and fails loudly if the path is not an executable file.

Under the same opt-in, the generated `.cli-env` now also carries the flags the session-tail E2E depends on:

- `KILO_NO_DAEMON=1` — the TUI must not attach to a machine daemon, so the embedded worker owns the ingest queue (otherwise the verification is a false pass/fail);
- `KILO_DEBUG_SESSION_INGEST=true` — the `ingest sync` / `ingest flush ok` lines are the primary evidence channel;
- `KILO_REMOTE=1` — without it the session never mirrors live to the device.

They must be *emitted*: `prepare_env` rewrites `.cli-env` on every `prepare`/`start`, silently discarding anything hand-appended (hit for real during bug reproduction). With `KILO_CLI_BIN` unset, behaviour and the generated env file are byte-for-byte identical to before.

## Why this PR exists at all

The fix under verification is **CLI-side**: Kilo-Org/kilocode#12545 (flush the session ingest tail on shutdown). A CLI-side fix cannot be exercised on device while the harness can only ever install a *published* CLI — this harness change is the smallest thing that makes that verification reachable, and the only reason this cloud PR exists. The batch instruction said one PR from this worktree; the two-PR deviation is deliberate and stated in both PR bodies.

## Checks

`pnpm format && pnpm typecheck && pnpm lint && pnpm check:unused` in `apps/mobile`, plus `git diff --check` — all green. The script was additionally validated by dry-running the emitted-env branch with a stub `KILO_CLI_BIN` (default path unchanged). Live verification against the stack runs with the companion CLI PR's on-device E2E; results will be posted as a follow-up comment.

### DIFF
--- a/apps/mobile/e2e/remote-cli.sh
+++ b/apps/mobile/e2e/remote-cli.sh
@@ -29,6 +29,9 @@
 #
 # Env overrides:
 #   KILO_CLI_VERSION   npm version/tag of @kilocode/cli to install (default: latest)
+#   KILO_CLI_BIN       absolute path to a locally built kilo binary; skips npm install,
+#                      prepends its directory to PATH, and emits KILO_NO_DAEMON=1,
+#                      KILO_DEBUG_SESSION_INGEST=true, and KILO_REMOTE=1 into the env file
 #
 # Requires: node, tmux, npm, a running stack (see e2e/AGENTS.md). Never reads
 # .env files directly; the token is minted by the dev:seed command.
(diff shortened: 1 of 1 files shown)


iscekic:
(bot) **Standin review — round 1: `No findings.`** (review of record; head reviewed: `8dd6e45580315dc6933a816f3c7fd30f213d9942`, full diff vs `main`)

The standin reviewer (reviewer-bot substitute per the workflow override) statically traced the full diff against plan §5, `set -euo pipefail` semantics, quoting, PATH precedence, the heredoc/append mechanics, and default-path equivalence, and returned **no actionable findings**. The loop is converged on the exact latest head.

Residual testing risks recorded (no action): the new branches' first live exercise is the orchestrator-run on-device E2E by design; a *symlinked* `KILO_CLI_BIN` would resolve the wrong directory — the E2E handoff passes the real built-binary path and records its version stamp as the guard; the bad-path validation ordering (after the token mint) is cosmetic and pre-existing; the header says "absolute path" while relative paths also work (more permissive, no failure mode).

The on-device E2E for the companion fix (Kilo-Org/kilocode#12545) exercises this harness end to end; results will be posted here and there as `(bot)` comments.

iscekic:
(bot) **Harness exercised end to end by the companion on-device E2E: PASS.** The acceptance run for Kilo-Org/kilocode#12545 used this branch's `KILO_CLI_BIN` mode exactly as designed: `prepare_env` skipped the npm install, put the locally built binary's directory first on `PATH` (`kilo --version` resolved to `0.0.0-fix-cli-ingest-flush-on-shutdown-202607262141`), and the emitted `KILO_NO_DAEMON=1` / `KILO_DEBUG_SESSION_INGEST=true` / `KILO_REMOTE=1` held across every `prepare`/`start` rewrite of `.cli-env` — the TUI footer showed `◆ Remote` and the version stamp on every attempt, and the `ingest sync` / `ingest flush ok` lines were the primary evidence channel.

The full E2E matrix (11/11: `/exit` ×3, SIGINT ×3, SIGHUP ×3, delayed-`/exit` control ×2, plus the coalescing check) is posted on Kilo-Org/kilocode#12545. The default developer flow (no `KILO_CLI_BIN`) was left byte-for-byte identical and was not disturbed by the run.

jeanduplessis (approved):

PR #4803, opened 2026-07-27
feat(ai-gateway): add NVIDIA direct BYOK support

lambertjosh:
## Summary

Adds NVIDIA as a direct BYOK provider (`nvidia-byok`). Users save an NVIDIA Developer API key and select namespaced models such as `nvidia-byok/nvidia/nemotron-3-super-120b-a12b`; Cloud strips the outer prefix and calls NVIDIA's hosted OpenAI-compatible Chat Completions endpoint with bearer authentication.

NVIDIA BYOK already exists client-side in `kilocode`; this adds the corresponding Cloud routing and catalog support.

### Why direct instead of Vercel

Vercel serves NVIDIA-authored models through providers such as Baseten, DeepInfra, and Together. It does not expose NVIDIA as an inference-provider slug that can use an NVIDIA API key, so NVIDIA BYOK must route directly to `https://integrate.api.nvidia.com/v1`.

## Model catalog

NVIDIA uses models.dev as its model catalog, consistent with existing direct BYOK providers. Sync keeps models that are:

- not deprecated;
- marked as supporting tool calls;
- text input and text output.

models.dev is the catalog authority for this integration. Availability may lag NVIDIA, so a stale catalog entry can return an upstream error until models.dev is corrected. No static routable fallback is compiled into Cloud.

Reasoning variants also come from models.dev. Explicit `effort` values are validated through Cloud's shared reasoning schema and exposed as model variants. Toggle-only metadata does not expose variants because NVIDIA toggle transports differ by model; those models use their provider default.

## Request routing

NVIDIA supports Chat Completions only. The provider translates Kilo's nested reasoning effort into NVIDIA's top-level `reasoning_effort` and removes gateway/OpenRouter fields NVIDIA rejects, including `provider`, `providerOptions`, `transforms`, `reasoning`, `include_reasoning`, `safety_identifier`, and `prompt_cache_key`.

Credential testing follows the existing static test-model convention using `nvidia/nemotron-3-nano-30b-a3b`. No migration or new Cloud environment variable is required.

## Verification

- Focused provider, catalog, conversion, serialization, and provider-ID tests pass.
- Web typecheck and lint pass.
- Tested locally end to end with a real NVIDIA key: BYOK key validation, model sync/listing, non-streaming tool calls, streaming, reasoning-disabled requests, zero-cost `direct-byok` usage attribution, unsupported protocol rejection, and disabled-key behavior.
- Live probing was used to validate rejected request fields. No credentials are committed or logged.


lambertjosh:
Tested end to end against a local dev server with a real client-style request, which caught a bug the unit tests and direct API probes both missed.

**Bug found and fixed** (`44d52341c`): NVIDIA rejected every proxied request with `Validation: Unsupported parameter(s): \`safety_identifier\``. The gateway injects `safety_identifier`, `user`, and `prompt_cache_key` for caller attribution and cache hints, and NVIDIA validates unknown fields. These are now stripped for `nvidia-byok`, matching the existing `seed` provider, with regression coverage added.

**Verified locally** (fake user, personal NVIDIA key, local Postgres + serverless-redis-http):

- BYOK UI lists NVIDIA and shows the hosted-API notice with working doc/terms links.
- "Test" returns `API key test success. Provider: nvidia-byok. Model: nvidia/nemotron-3-nano-30b-a3b.`
- Provider sync stored 21 NVIDIA models; excluded models are absent, and context overrides applied (`nemotron-mini-4b` 4096, `llama-3.2-90b-vision` 32768).
- `GET /api/gateway/v1/models` returns 21 `nvidia-byok/...` models. Reasoning is advertised only where supported: Super `none/low/high`, Ultra `none/medium/high`, and Nano exposes no variants with `supported_parameters` limited to `max_tokens, temperature, tools`.
- Non-streaming tool call through the gateway succeeded (`read_file` with `{"path":"package.json"}`).
- Streaming with reasoning disabled succeeded.
- `openai/gpt-oss-120b` tool call succeeded with reasoning stripped.
- Usage logged as `provider=direct-byok` with `cost=0`.
- `POST /v1/responses` correctly rejected with `api_kind_not_supported`.
- Disabling the key removed all NVIDIA models and stopped routing.

Full suite state: 49 targeted tests, typecheck, and lint all pass.

lambertjosh:
image

apps/web/src/components/organizations/byok/BYOKKeysManager.tsx
```diff
@@ -631,6 +631,41 @@ export function BYOKKeysManager({ organizationId }: BYOKKeysManagerProps) {
...
+                {selectedProvider ===
+                  DirectUserByokInferenceProviderIdSchema.enum['nvidia-byok'] && (
+                  <Alert>
+                    <Info className="size-4" />
+                    <AlertDescription>
+                      <p>
+                        Requests send prompts and model outputs to NVIDIA&apos;s hosted API Catalog
+                        under your NVIDIA agreement. NVIDIA limits Developer Program endpoints to
+                        prototyping, research, development, and testing; serving production end
+                        users may require NVIDIA AI Enterprise licensing.
```
chrarnoldus: I don't think we want this, out terms should already say users agree to the terms of providers.

apps/web/src/lib/ai-gateway/providers/direct-byok/index.ts
```diff
@@ -24,6 +24,10 @@ function convertModel(
 ) {
   const id = formatDirectByokModelId(provider, model);
   const name = DIRECT_BYOK_PROVIDERS_META[provider.id] + ': ' + model.name;
+  const variants = getModelVariants(id, provider.id);
+  // NVIDIA's hosted endpoints reject reasoning fields on models that do not document
+  // them, so only advertise reasoning where verified reasoning controls exist.
+  const supportsReasoning = provider.id !== 'nvidia-byok' || variants !== undefined;
```
chrarnoldus: What does this do concretely?

If it's important, have you considered making it part of the `DirectByokModel`?

apps/web/src/lib/ai-gateway/providers/direct-byok/sync-direct-byok.ts
```diff
@@ -142,6 +176,29 @@ function modelsDevFetcher(
   };
 }
 
+const nvidiaFetcher: ProviderFetcher = {
+  providerId: 'nvidia-byok',
+  async fetch(ctx) {
+    const [response, catalog] = await Promise.all([
+      fetch('https://integrate.api.nvidia.com/v1/models'),
```
chrarnoldus: Why are we calculating the intersection? Why not trust models.dev?

apps/web/src/lib/ai-gateway/providers/openrouter/types.ts
```diff
@@ -86,6 +86,9 @@ export type OpenRouterChatCompletionRequest = OpenAI.Chat.ChatCompletionCreatePa
 
     // https://platform.minimax.io/docs/api-reference/text-openai-api#4-important-note
     reasoning_split?: boolean;
+
+    // OpenRouter extension; direct providers must remove it unless explicitly supported.
+    include_reasoning?: boolean;
```
chrarnoldus: Where does this come from, what client adds it?

apps/web/src/components/organizations/byok/BYOKKeysManager.tsx
lambertjosh: I removed it, it did seem fairly unique so I had left it and NVIDIA asked us to add one other disclaimer, but they obviously clicked through to get the API key anyway. This is just a reminder.

apps/web/src/lib/ai-gateway/providers/direct-byok/index.ts
lambertjosh: Removed since in a later change started stripping reasoning for models that don't support it, so this is unnecessary.

apps/web/src/lib/ai-gateway/providers/direct-byok/sync-direct-byok.ts
lambertjosh: models.dev advertises more than are currently available (22 out of 57 do not exist as of today) and NVIDIA's catalog does not include all the details.

apps/web/src/lib/ai-gateway/providers/openrouter/types.ts
lambertjosh: Our clients don't send it, but we advertise it in supported_parameters so a third party client could send and then get a 400 back from nvidia. The alternative is to stop advertising it for NVIDIA, which was removed in the comment above. As of now we advertise but remove, it may be cleaner to do the inverse, though. What do you think is best?

chrarnoldus: I think we should leave it out if we have no evidence it's important.

apps/web/src/lib/ai-gateway/providers/direct-byok/sync-direct-byok.ts
chrarnoldus: Isn't the opposite also true (models are missing from models.dev)?

I don't really like having provider-specific hacks for generic problems, like models.dev being out of date or /models not including details.

apps/web/src/components/organizations/byok/BYOKKeysManager.tsx
chrarnoldus: Last time I included an NVIDIA specific disclaimer I was told not to do that.

apps/web/src/lib/ai-gateway/providers/direct-byok/sync-direct-byok.ts
lambertjosh: Won't we show models that don't exist and don't work then to the user?

apps/web/src/lib/ai-gateway/providers/direct-byok/nvidia-byok.ts
```diff
@@ -0,0 +1,41 @@
...
+      return;
+    }
+
+    const reasoningEffort =
+      request.body.reasoning?.enabled === false
+        ? 'none'
+        : (request.body.reasoning_effort ?? request.body.reasoning?.effort);
+    const parsedReasoningEffort = ReasoningEffortSchema.safeParse(reasoningEffort);
+    if (parsedReasoningEffort.success) {
+      (request.body as { reasoning_effort?: string }).reasoning_effort = parsedReasoningEffort.data;
```
chrarnoldus: Why is this `as` necessary?

apps/web/src/lib/ai-gateway/providers/direct-byok/nvidia-byok.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    // NVIDIA rejects these with `Validation: Unsupported parameter(s)`.
+    delete request.body.provider;
+    delete request.body.providerOptions;
+    delete request.body.transforms;
+    delete request.body.reasoning;
+    delete request.body.safety_identifier;
+    delete request.body.prompt_cache_key;
+    // Older clients may still send this untyped OpenRouter field.
+    delete (request.body as { include_reasoning?: boolean }).include_reasoning;
```
chrarnoldus: Please remove this or define it on the body type if it's really necessary.

apps/web/src/lib/ai-gateway/providers/direct-byok/nvidia-byok.ts
```diff
@@ -0,0 +1,41 @@
...
+  },
+  models: cachedEnhancedDirectByokModelList({
+    providerId: 'nvidia-byok',
+    recommendedModels: [],
```
chrarnoldus: Please add a model here.

apps/web/src/lib/ai-gateway/providers/direct-byok/sync-direct-byok.ts
```diff
@@ -125,9 +139,57 @@ export function parseModelsDevProviderModels(entry: unknown): RawModel[] {
...
+        return effort.success
+          ? [[effort.data, { reasoning: { enabled: effort.data !== 'none', effort: effort.data } }]]
+          : [];
+      })
+    );
+  }
+  return {};
+}
+
+export function parseNvidiaProviderModels(entry: unknown): RawModel[] {
```
chrarnoldus: Please generalize this instead of making it NVIDIA only.

apps/web/src/lib/ai-gateway/providers/direct-byok/types.ts
```diff
@@ -13,6 +14,8 @@ export const DirectByokModelSchema = z.object({
   flags: z.array(DirectByokModelFlagSchema).readonly().optional(),
   context_length: z.number(),
   max_completion_tokens: z.number(),
+  supported_parameters: z.array(z.string()).optional(),
```
chrarnoldus: Can we add a flag `reasoning` instead, similar to kilo exclusive model?

apps/web/src/lib/ai-gateway/providers/direct-byok/sync-direct-byok.ts
chrarnoldus: Resolved using the generalized models.dev fetcher from main. NVIDIA now supplies its standard `/v1/models` URL to the same availability-intersection mechanism used by other providers; there is no NVIDIA-specific catalog parser.

apps/web/src/lib/ai-gateway/providers/direct-byok/sync-direct-byok.ts
```diff
@@ -125,9 +129,45 @@ export function parseModelsDevProviderModels(entry: unknown): RawModel[] {
...
 }
 
+export function parseNvidiaProviderModels(entry: unknown): RawModel[] {
+  const provider = ModelsDevProviderSchema.parse(entry);
+
+  return Object.values(provider.models)
+    .filter(
+      model =>
+        model.status !== 'deprecated' &&
+        model.tool_call === true &&
```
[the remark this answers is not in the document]
chrarnoldus: Resolved in `99938e909` through the generalized live availability intersection plus tool-call/text-input requirements. I also checked NVIDIA’s live catalog before pushing: the previously identified broken Gemma, Qwen, and Sarvam IDs are currently absent, so they are excluded without retaining a provider-specific denylist.

apps/web/src/lib/ai-gateway/providers/direct-byok/nvidia-byok.ts
```diff
@@ -0,0 +1,41 @@
...
+    const { request } = context;
+    if (request.kind !== 'chat_completions') {
+      return;
+    }
+
+    const reasoningEffort =
+      request.body.reasoning?.enabled === false
+        ? 'none'
+        : (request.body.reasoning_effort ?? request.body.reasoning?.effort);
+    const parsedReasoningEffort = ReasoningEffortSchema.safeParse(reasoningEffort);
```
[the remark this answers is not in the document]
chrarnoldus: Resolved in `99938e909`. The already-resolved `DirectByokModel` is now passed into the provider transform, and NVIDIA validates the parsed request effort against that model’s synced variant efforts. Unsupported efforts, including efforts on models with no matching variants, are stripped before forwarding.

apps/web/src/lib/ai-gateway/providers/direct-byok/nvidia-byok.ts
chrarnoldus: Resolved. The cast is gone. The gateway chat request boundary now explicitly permits the full shared reasoning-effort enum (plus inbound `null` compatibility), so NVIDIA can assign the schema-validated value directly.

apps/web/src/lib/ai-gateway/providers/direct-byok/nvidia-byok.ts
chrarnoldus: Resolved in `99938e909`. `nvidia/nemotron-3-nano-30b-a3b` is now the recommended/fallback model, matching the credential test model and current NVIDIA catalog metadata.

apps/web/src/lib/ai-gateway/providers/direct-byok/nvidia-byok.ts
chrarnoldus: Resolved by removing this deletion and its local type entirely. Direct BYOK no longer advertises `include_reasoning`, so there is no NVIDIA-specific handling for it.

apps/web/src/lib/ai-gateway/providers/direct-byok/sync-direct-byok.ts
chrarnoldus: Resolved after merging main. The NVIDIA-only parser was removed; the shared models.dev parser/fetcher now accepts generic availability, tool-call, text-input, context-override, and non-empty options. NVIDIA only supplies configuration values.

apps/web/src/lib/ai-gateway/providers/direct-byok/types.ts
chrarnoldus: Resolved by the preparatory changes merged into main. `reasoning` is now a shared `DirectByokModel` flag, populated from models.dev and used by conversion to advertise the reasoning parameter; the provider-specific `supported_parameters` field is gone.

apps/web/src/lib/ai-gateway/providers/direct-byok/sync-direct-byok.ts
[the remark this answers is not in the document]
chrarnoldus: Follow-up after the latest main merge: tool/text compatibility is now enforced globally for every models.dev-backed provider, with missing metadata treated optimistically. NVIDIA still uses the generalized live availability intersection. The current catalog audit finds eight affected live combinations, all NVIDIA models with explicit `tool_call: false`.

apps/web/src/lib/ai-gateway/providers/direct-byok/sync-direct-byok.ts
chrarnoldus: Follow-up: capability requirements are no longer fetcher options. Explicit `tool_call: false` and explicit input modalities without text are filtered directly by the shared models.dev parser for every provider; undefined metadata remains accepted.

apps/web/src/lib/ai-gateway/providers/direct-byok/sync-direct-byok.ts
```diff
@@ -125,9 +129,45 @@ export function parseModelsDevProviderModels(entry: unknown): RawModel[] {
...
+        model.modalities?.input?.includes('text') === true &&
+        model.modalities?.output?.includes('text') === true
+    )
+    .map(model => {
+      const reasoningEfforts = getNvidiaReasoningEfforts(model.id);
+      return {
+        id: model.id,
+        name: shortenDisplayName(model.name),
+        context_length: model.limit?.context,
```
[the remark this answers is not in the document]
chrarnoldus: Follow-up: the context-length override mechanism and NVIDIA overrides were intentionally removed in `56ae955b4` per owner direction. Synced context limits now come directly from models.dev.

apps/web/src/lib/ai-gateway/providers/direct-byok/nvidia-byok.ts
chrarnoldus: Follow-up: the recommendation and credential-test model are now `nvidia/nemotron-3-super-120b-a12b`, which models.dev describes for collaborative agents and high-volume reasoning workloads. It is currently present in NVIDIA’s live catalog.

apps/web/src/lib/ai-gateway/providers/direct-byok/sync-direct-byok.ts
```diff
@@ -137,7 +177,7 @@ function modelsDevFetcher(
       if (!entry) {
         throw new Error(`models.dev catalog missing ${catalogKey} entry`);
       }
-      return parseModelsDevProviderModels(entry);
+      return parseEntry(entry);
```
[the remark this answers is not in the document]
chrarnoldus: Follow-up: the empty-catalog rejection policy was intentionally removed in `56ae955b4` per owner direction. The generic sync behavior now applies unchanged to NVIDIA.

apps/web/src/lib/ai-gateway/providers/direct-byok/nvidia-byok.ts
chrarnoldus: Follow-up after OpenAI 6.49.0 landed on main: the SDK now exposes the full gateway effort union plus `null`, so the temporary `Omit<..., reasoning_effort>` request-type override has also been removed.

apps/web/src/lib/ai-gateway/providers/direct-byok/nvidia-byok.ts
```diff
@@ -0,0 +1,55 @@
...
+      delete request.body.reasoning_effort;
+    }
+
+    // NVIDIA rejects these with `Validation: Unsupported parameter(s)`.
+    delete request.body.provider;
+    delete request.body.providerOptions;
+    delete request.body.transforms;
+    delete request.body.reasoning;
+    delete request.body.safety_identifier;
+    delete request.body.prompt_cache_key;
```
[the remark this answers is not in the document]
chrarnoldus: Fixed in `91a657134`. `include_reasoning` is now explicitly modeled on the gateway chat request body, stripped without a cast for NVIDIA, and covered by the gateway-only-field regression test.

chrarnoldus (approved):

PR #4808, opened 2026-07-27
fix(extension): side panel polish — floating memory card, empty-state hint, analytics toggle colour, history filter, pinned new-tab button

iscekic:
## What this fixes

Five reported side-panel defects, each reproduced on the baseline (`28802f7ff`) before fixing (Chrome + Playwright, artifacts in the run record):

1. **Add to memory modal should float over instead of being on top.** The card was a flex sibling above the transcript and consumed its layout height (measured: conversation region top 106 → 360.39, height 291 → 36.61 with a pending draft). It is now a centred `role="dialog"` overlay with a backdrop.
2. **"Pick a tab and ask Kilo to inspect it." should go away once there's a message in the chat.** The greeting was a real seeded assistant event — persisted, and sent to the model as a prior turn. New conversations now start with zero events; the same string renders as the transcript's empty state and disappears after the first user message. Stored conversations whose first event is the legacy greeting are migrated on load (exact-text, first-position, assistant-only) and persisted back.
3. **Share usage analytics toggle lacked colour signalling on/off.** On-state track was `rgb(69,69,69)` vs off `rgb(51,51,51)`. The on state now uses the brand accent track with a dark knob.
4. **Empty conversations shouldn't be in history.** History now lists only conversations with at least one user message; empty conversations are removed from storage when they stop being open (close deletes them, load-time normalization drops stored empty non-open rows), and closing a tab evicts the per-conversation draft/usage/cost atoms even when the replacement blank tab reuses the conversation id.
5. **New tab button should float on the right.** The `+` button was the last child inside the horizontally scrolling tab strip (measured at x = 2087.53 in a 320px viewport). It is now a fixed-width sibling outside the strip, reachable at every scroll position.

## Simpler-shape decisions (from the accepted plan)

- **S1** — the memory card reuses the existing overlay shape (`fixed inset-0 … role="dialog" aria-modal="true"`) at `z-[25]`: above all panel chrome (context popover `z-20`, mode popover `z-10`), below Settings/History (`z-30`), so `Manage memories` still opens Settings visibly above the card.
- **S2** — the seeded greeting is deleted outright rather than filtered at render time, which also removes it from the first gateway request (asserted: one `user`, zero `assistant` messages).
- **S3** — History filters inside `getSortedStoredConversationHistory` via the existing `isStoredConversationEmpty`, the single funnel every History consumer routes through.
- **S4** — the existing hand-rolled switch is coloured in place; no component library, no shared `Switch` extraction.
- **S5** — item 5 is a flex restructure only: scroller `flex-1 min-w-0 overflow-x-auto`, button a `shrink-0` sibling. No sticky positioning or scroll shadows.
- **S6** — item 5's new geometry assertions are Chrome-only; the Firefox spec already covers tab creation and there is no browser-API split.
- **S7** — no new dialog semantics on the card (no Escape/backdrop-dismiss/trap/restore), matching the extension's two existing overlays; the one addition is initial focus on the note textarea.

Recorded consequence: an unsent in-memory draft in a closed *empty* tab is discarded (drafts were previously recoverable only by reopening from History — a path empty conversations no longer have; sent content is never affected).

## Verification

- `pnpm format`, `pnpm --filter kilo-extension verify` (typecheck, lint, 420 unit tests), `build`, `build:firefox` — green.
- `e2e:chrome` — 84 passed, 10 skipped, 0 failed. `e2e:firefox` — 33/33 workflows passed.
- New coverage: transcript empty state (visible, no CTA, gone after first message), first-request role composition, legacy greeting migration (modern/legacyEvents/user-message-safe), history filter + count + empty state, empty-tab close eviction (incl. id-reuse regression), floating-card layout invariance + focus + settings-above-card stacking, analytics switch on/off computed RGB, new-tab button pinned at both scroll ends.

No new dependencies, no manifest permission changes, no `DESIGN_TOKENS` changes.


iscekic:
(bot) @kilocode-bot please review the current head b79d6f754. It merges main into the branch; the apps/extension diff is byte-identical to 92f412d17, which you already reviewed with no findings.

jeanduplessis (approved):

PR #4822, opened 2026-07-28
feat(ci): add extension store release workflow

iscekic:
## Summary

**What:** a new `workflow_dispatch` GitHub Actions workflow, `.github/workflows/extension-store-release.yml` ("Extension Store Release" in the Actions sidebar).

**Why:** marketing needs store-submittable extension builds on demand, and the stores reject re-used version numbers — today producing them is a manual local build plus a manual `package.json` edit.

**How:** one button press runs two jobs:

1. `build-artifacts` (read-only token) — installs, gates on `pnpm --filter kilo-extension verify`, runs the unmodified `zip` / `zip:firefox` scripts, and uploads the Chrome MV3 zip, Firefox MV3 zip, and AMO sources zip as one artifact named `extension-store-<version>` (30-day retention). Before installing, a guard fails the run in seconds if any `extension-version-bump-*` branch still exists (a previous bump PR not yet merged), naming the stale branch and its recovery.
2. `bump-version` (runs only when the dispatched ref is `main`; the only job with write permissions) — patch-bumps `apps/extension/package.json` via `npm version patch --no-git-tag-version`, pushes branch `extension-version-bump-<new>`, and opens the bump PR (assigned to `iscekic`, reviewer `iscekic`) whose body explains the released version, the bump, and how to nudge its pending checks.

After the bump PR merges, the next press produces the next version with no manual edit (merged bump branches self-delete — `delete_branch_on_merge` is enabled on this repo, so the guard cannot false-trip on them). The two-job split keeps the job running third-party build tooling on a read-only token; the write-permission job installs nothing. The branch also carries two `.kilo_workflow/` infra commits (a `dispatch-role.sh` tmux fix and two workflow learnings), which are workflow tooling, not product surface.

## Verification

CI YAML — no product runtime path exists to click through; verified locally by an independent verifier round plus machine-parse checks:

- [x] YAML parses (pyyaml `safe_load`); every expression, `needs`, `if` guard, `permissions` block, artifact name/path, and SHA pin desk-checked against the acceptance criteria — all four action SHA pins byte-identical to `extension-ci.yml`; `actionlint` not available on this machine (noted, nothing installed).
- [x] `pnpm --filter kilo-extension zip` → `.output/kilo-extension-0.1.1-chrome.zip`; `pnpm --filter kilo-extension zip:firefox` → `kilo-extension-0.1.1-firefox.zip` + `kilo-extension-0.1.1-sources.zip` — all three paths exactly match the upload step's paths, and both built manifests' `version` equals `apps/extension/package.json`'s.
- [x] `npm version patch --no-git-tag-version` dry-run in a temp copy → single-line `version` diff (`0.1.1` → `0.1.2`), no other files, no lockfile churn.
- [x] `pnpm format:check` clean (oxfmt normalized comment spacing on three `permissions` lines; block-scalar heredoc untouched).

Inert by design until this merges — stated as fact, not run: real `workflow_dispatch`, artifact download from a run, the pending-bump guard trip, and bump-PR creation (`workflow_dispatch` is only offered for workflows on the default branch).

## Visual Changes

N/A

## Reviewer Notes

1. **Marketing needs repository write access** to see the "Run workflow" button — `workflow_dispatch` is unavailable to read-only members. Org-permissions action, outside this change.
2. **First-of-kind path — verify before the first real dispatch:** nothing in this repo has ever opened a PR from Actions (`gh pr create` has no in-repo precedent; pushes are proven via `kilo-app-release.yml` tags). If the org/repo setting "Allow GitHub Actions to create and approve pull requests" (Settings → Actions → General → Workflow permissions) is off — it is not API-readable with the current token (403, verified) — the `bump-version` job fails loudly *after* artifacts upload. Please confirm the setting once, ideally before or at first dispatch; if it is off, enable it and re-run the failed job (artifacts stay valid). Accepted residual risk if unconfirmed: the workflow is "artifacts only" until the first bump attempt.
3. **Optional human follow-up:** adding repository secret `VITE_POSTHOG_API_KEY` enables analytics in store builds; until then builds are analytics-less by design (same as today's local builds — `wxt.config.ts` only warns).
4. **Known limitation:** auto-created bump PRs show required checks as pending until a human nudges them (close/reopen or empty commit), because `GITHUB_TOKEN` PRs do not trigger workflow runs — the bump PR's own body says this to its reviewer.
5. Assigned to `iscekic`; review requested from `eshurakov` and `jeanduplessis` per cloud convention.


.github/workflows/extension-store-release.yml
```diff
@@ -0,0 +1,156 @@
...
+
+          1. Required checks show as pending: PRs opened with the default GITHUB_TOKEN
+             do not trigger workflow runs. Close and reopen this PR (or push an empty
+             commit) to start CI.
+          2. Review and merge. Only the version line changed; no product code.
+          EOF
+          gh pr create --base main \
+            --title "chore(extension): bump version to $NEW" \
+            --body-file /tmp/bump-pr-body.md \
+            --assignee iscekic \
```
pandemicsyn, line 155: The bump PR's assignee and reviewer are hardcoded to a single GitHub login (`iscekic`). If that person is unavailable, auto-created bump PRs will sit unreviewed and block the next store release (the build-artifacts guard fails while any `extension-version-bump-*` branch exists).

Suggested fix: Consider making the assignee/reviewer configurable (e.g. via a repo variable or a small owner list) so the workflow doesn't have a single-person bottleneck, or document the fallback procedure for someone else to take over the bump PR.

pandemicsyn (approved):

.github/workflows/extension-store-release.yml
pandemicsyn: feel free to ignore

iscekic: Yeah this is fine

PR #4823, opened 2026-07-28
feat(mobile): search input also filters the Active now tray

iscekic:
## Summary

**What:** The Agents session-list search input now also filters the pinned "Active now" tray, client-side, with the same semantics as the server-side history search: a case-insensitive literal substring match on the session's `title` OR `id`, in conjunction with the existing platform/project filters.

**Why:** The tray previously ignored the search text by deliberate design (PR #4626). A user explicitly asked to reverse that decision: when they search, they expect every session list on the screen — including the pinned active tray — to narrow. The code's own comments anticipated this reversal.

**How:**

- `selectPinnedActiveSessions` gains a required `searchQuery: string` param. The needle is trimmed/lowercased once, then applied as the first check inside the existing filter (`String.includes` on `title`/`id` — literal substring, LIKE wildcards match literally, mirroring the server's `position()` search). The platform/project logic below is unchanged.
- `AgentSessionListScreen` passes `effectiveSearchQuery` (the same committed, debounced value the history data source consumes), so the tray and history narrow together.
- `selectSessionListBodyModel` gains one guard: while searching with an empty history but a populated tray, it returns `render-list` instead of `filtered-empty` — tray matches are matches, so the body never claims "No sessions match" beneath a populated tray (that copy would be factually wrong on the feature's primary happy path: searching an active session's own title empties history by construction, since its stored row is excluded). The query-error branch (Retry CTA) and the filter-only branch (pre-existing deliberate decision) are unchanged.
- Comments/tests encoding the old "search is intentionally NOT a narrowing input" decision are updated (helper JSDoc, screen comment, body-model comments, both test suites — the flipped `@ts-expect-error` assertion is replaced by 8 positive search-semantics tests).

## Verification

- [x] **E2E on iOS simulator (iPhone 17 Pro Max, iOS 26.5, Maestro) against this worktree's local stack — `VERIFICATION PASSED.`** (round r3; run on merged head `dd976cf01`, and the later main-merge to the current head touched no mobile or owned files — clean merge, zero path overlap). Data: live active session "Bicycle pump haiku generation" plus stored "Zucchini haiku generation" and "New session". All five flows passed:
  1. Query `bicycle` (title substring, lowercase vs a capitalized title) → the tray row stays, ACTIVE NOW visible, and **no** "No sessions match" copy renders beneath the tray (tray-only-match state, AC1+AC3).
  2. Gibberish query `qxzjkv` → the tray hides; "No sessions match / Try a different search term." with a "Clear search" CTA appears (AC2, empty state).
  3. "Clear search" CTA (and the in-field X) → the full tray and full history return (AC4).
  4. Query `zucchini` → the stored list narrows to "Zucchini haiku generation" only, "New session" hidden, tray hides correctly — history search unchanged (AC5, regression).
  5. Session-id substring `0570dbf84` → the tray row stays (id-matching half of AC1).

  Feature-state matrix: happy **PASS**, tray-only-match **PASS**, empty **PASS**; retryable-unhappy skipped (pre-existing machinery untouched by this change; a deterministic search failure would require fault injection); non-retryable-unhappy N/A (structurally impossible — a pure client-side filter over an already-loaded list has no terminal failure mode). Post-run the worktree was byte-identical to baseline and the simulator/slot were released.
- [x] **Automated (rerun on the final merged head):** `pnpm format`, `pnpm typecheck`, `pnpm lint` (0 warnings / 0 errors), `pnpm check:unused`, and the mobile vitest suite (277 files / 2346 tests, incl. the new `search rules` helper tests and the new body-model guard tests) all pass in `apps/mobile`.

## Visual Changes

| Before | After |
|---|---|
| The search input narrows only stored history; the pinned "Active now" tray ignores it and keeps showing every active session. | The search input also narrows the tray: only active sessions whose title or id match stay pinned; with no match the tray hides; when only the tray matches, no contradictory "No sessions match" copy appears beneath it. |

No visual-design changes — tray, rows, and empty states reuse existing components; only which sessions appear changes. Simulator screenshots from the passing E2E run (tray before search, tray with a matching query, no-match empty state, restored tray) were captured and verified during verification.

## Reviewer Notes

- Focus areas: the matching semantics in `selectPinnedActiveSessions` (parity with the server's case-insensitive substring search), and the single new body-model guard (`isSearching && hasPinnedActive` → `render-list`).
- Accepted residuals (deliberate, from the approved plan): on query *refinement* there is a transient (one search round-trip) window where the tray filters by the new needle while history still shows the previous results under `keepPreviousData` — consistent with the history list's own stale-while-refetching design; true displayed-row lockstep was rejected as unjustified state for a sub-second window. Unicode case-folding beyond ASCII parity with Postgres `lower()` is out of scope.
- The filter-only (platform/project, no text) body-model branch is deliberately untouched — its "populated tray does not change the decision" behavior is a separate pre-existing product decision outside this change.
- **Pre-existing issue surfaced during review (not introduced here, suggested follow-up):** the tray's displayed title comes from the live heartbeat payload and `cliSessionsV2.rename` updates only the stored row — so renaming an *active* session never updates the title shown in the tray while it stays active. Consequence under this PR's tray filtering: searching the *new* name matches nothing visible (the stored row is excluded from history while active; the tray still shows the old heartbeat title). The root fix is backend: propagate the renamed title into the live session state, or overlay the stored title in the `activeSessions.list` enrichment. Full triage in the PR comments.


iscekic:
(bot) @kilocode-bot please review

iscekic:
(bot) Triage of the four takeaways, each verified against the code on the current head (`01b94342`):

1. **Trim divergence — not reachable; the suggested fix already exists.** The search controller trims at the commit boundary exactly as suggested: `commitSearchQuery(text.trim())` (`apps/mobile/src/components/agents/session-search-state.ts:93`); both clear paths commit `''`. The committed query is the single input to both consumers — the server fetch (`useAgentSessionSearch({ searchQuery, … })`) and the tray (`searchQuery: effectiveSearchQuery`, derived from that same committed state, `session-list-screen.tsx:177,190`). No untrimmed value reaches either side, so `"  login"` cannot match the tray but not history. A whitespace-only query commits as `''`, so `isSearching` is false (`session-list-screen.tsx:89`) — the empty-state suppression divergence cannot occur. The helper's own trim is defensive-only, by design.

2. **Renamed active session vanishing — scenario verified real, but the root cause is pre-existing and out of this PR's scope.** Both premises confirmed: the tray title comes from the live heartbeat payload (`activeSessions.list` → session-ingest `/api/sessions/active`; the router's DB enrichment overlays only status/platform/timestamps — never `title`), and `cliSessionsV2.rename` writes only `cli_sessions_v2.title` (`cli-sessions-v2-router.ts:1311`). So a mobile rename never updates the title the tray displays while the session is active — that display staleness exists today, with or without search. This PR's filter matches the corpus the UI actually displays: the session's *visible* title genuinely doesn't match the renamed string. The suggested fix (exclude only *pinned* ids from history) would change the deliberate history-exclusivity invariant (active sessions never appear in history — a recorded non-goal of this change) and would surface non-pinned active sessions in history — a product decision beyond this brief. The root fix is propagating the renamed title into the live session state, or overlaying the stored title in the active-sessions enrichment — backend follow-up, now flagged in the PR's Reviewer Notes.

3. **Filter-only branch — deliberate and recorded.** The new guard is gated on `isSearching` on purpose: the filter-only branch's "populated tray does not change the decision" behavior is a separate pre-existing product decision this brief does not reverse (plan non-goal; a body-model test pins it). Agreed it is follow-up material, not this PR.

4. **Test gaps — mostly covered or unreachable.** A whitespace-only query *is* tested ("whitespace-only query performs no narrowing", `session-list-helpers.test.ts`). Surrounding-whitespace input cannot reach the helper from the UI (commits are trimmed — see point 1). Empty-title behavior is pinned by the id-fallback test ("matches a session id substring, case-insensitively" — a non-matching title falls through to id). The tray's query source is one reviewed wiring line (`searchQuery: effectiveSearchQuery`, `session-list-screen.tsx:190`) implementing the deliberate pending-window design; no screen-render test harness exists in the owned paths, and the behavior is E2E-verified on a simulator.

No code changes result from this round. The posted summary (**No Issues Found | Recommendation: Merge**, covering all 6 PR files including the learnings doc) reviewed exactly the diff still on this PR — the head moves since (`99bdc009` → `01b94342`) are merges of `main` that change no PR file.

iscekic:
(bot) @kilocode-bot please review the current head 01b94342 — since your last summary the branch only merged latest main (the PR diff is unchanged: the same 6 files you approved).

pandemicsyn (approved):

PR #4825, opened 2026-07-28
fix(mobile): gate chat attachment size before materializing the file

iscekic:
## Summary

**What:** the mobile kilo-chat attachment composer now checks a picked file's size **before** reading the
file into memory, at a new mobile-only cap of 10 MiB, reusing the existing selection/toast helpers. Picking
an oversized or unmeasurable file is rejected with the existing rejection toast; only accepted files are
materialized, and multi-select files are materialized one at a time instead of all at once.

**Why:** picking a chat attachment used to materialize an arbitrarily large local file
(`fetch(file://)` + `Response.blob()` — two full copies: a JS `ArrayBuffer` and a `ByteArray` on the Java
heap in React Native's blob store) and only checked the size afterwards, against the shared 100 MB limit.
On a phone with a ~268 MB heap growth limit that is a real `OutOfMemoryError` class, and a 10-file
multi-select multiplied peak memory by 10.

**How:** the three pick functions now return normalized selections only (no network, no blob). The size gate
lives in `selectAllowedAttachments`, which was previously dead code, and now also rejects files whose size
is unknown or zero (fail-closed: `File.size` signals an unreadable file with `0`). Only then does the new
`materializeAttachment` read the file, one file at a time. The composer passes the same 10 MiB constant to
the queue's own post-hoc `maxBytes` check, so the two guards can never disagree.

### The Sentry crash cluster (KILO-APP-2J / KILO-APP-2R) — mechanism finding

The reported allocation (`NetworkEventUtil` → `Base64.encodeToString` → `StringFactory.newStringFromChars`
on an OkHttp dispatcher thread) was React Native's **network-inspector reporting** base64-encoding a large
**remote** HTTP response into a Java `String` (`NetworkEventUtil.kt:148`, reached from `NetworkingModule`'s
blob `responseHandlers` branch), with `enableNetworkEventReporting()` defaulting to `true` — in release
builds too. Under Expo SDK 55 that applied to *every* `fetch()`, because RN's `whatwg-fetch` sets
`responseType: 'blob'` on every request. The arithmetic matches: 54,298,696 bytes ≈ a 20.4 MB response,
220,681,432 bytes ≈ an 82.7 MB response (UTF-16 `String`, 2 bytes/char × 4/3 base64).

The chat attachment picker's `file://` fetch was **not** the crashing call: `BlobModule`'s `UriHandler`
serves non-`http(s)` URIs before OkHttp (`NetworkingModule.kt:309-311`), so a picker fetch cannot produce
the reported `RealCall$AsyncCall` → `Callback.onResponse` frames.

**Can `main` still hit the reported allocation? No.** Expo SDK 57 installs `expo/fetch` as `global.fetch`
(`EXPO_PUBLIC_USE_RN_FETCH` appears nowhere in this repo), which runs on its own OkHttp client and never
enters `NetworkingModule`. The only remaining first-party `XMLHttpRequest` (the signed upload PUT) uses the
default `text` response type. The probable crashing surface — labelled **unverified** — was a very large
tRPC/HTTP response on `agent-chat/[session-id]`, matching `view_names`. That allocation was framework code,
is already unreachable on `main`, and is not fixable here.

**What this PR fixes** is the related, still-live defect on `main`: unbounded materialization of picked
attachments (above). This PR is **not expected to close KILO-APP-2J or KILO-APP-2R** — their allocation site
is framework code `main` no longer reaches, so they should go inert on the next SDK 57 release for an
unrelated reason. Recommendation: a human resolves them once an SDK 57 build has field coverage. (No
`Fixes …` reference in any commit, deliberately.)

### The number, the trade-offs, and the limit

10 MiB is a product judgement: the binding constraint is *retained* bytes — up to 10 accepted blobs
(`MESSAGE_ATTACHMENT_MAX_COUNT`) stay resident in the blob store until the message is sent, so the worst
case is ~100 MiB against a ~268 MB heap growth limit. At 25 MiB the worst case would be 250 MiB — fatal on
the same device. It is more generous than the existing in-repo mobile precedent (cloud-agent attachments cap
at 5 MiB), and well inside the shared server limit (100 MB), so no server or web behaviour changes.

Accepted trade-offs, deliberately:

- A maximum-resolution camera capture (`quality: 1`) can exceed 10 MiB and will now be **rejected with the
  size toast** instead of uploading. Typical phone JPEGs (2–6 MB) are unaffected. No `quality` change or
  downscaling is included — that is a separate product change.
- A zero-byte or unmeasurable pick is now **rejected** (`Couldn't read <name>.`) instead of being uploaded
  as an empty attachment — matching the cloud-agent surface's decision to reject `size <= 0`.
- A multi-select now **keeps the files that succeeded** instead of discarding the whole pick when one file
  fails; each failed file is named in its own toast.
- Honest limit of any pre-read gate: an unreadable size stat combined with a picker that actively
  under-reports the size still falls through to `addFile`'s post-materialize check. That is a strictly
  smaller hole than before (no gate at all), and closing it fully would require reading the file — the thing
  the gate exists to avoid.

### Feature-state matrix — attachment size behaviour (kilo-chat composer)

| State | Required experience | How it is delivered |
|---|---|---|
| **Happy** | File within the limit is accepted, the chip shows upload progress, the message sends with the attachment. | `selectAllowedAttachments` returns it in `accepted`; `materializeAttachment` → `queue.addFile` → existing upload/chip/send path, unchanged. |
| **Unhappy, retryable** | The rejected file is named, the reason is specific, nothing is added, and the user can immediately pick another file. Three distinct messages, never merged: | 1. **Too large** → `<name> exceeds the 10 MB attachment limit.` 2. **Unmeasurable or empty** → `Couldn't read <name>.` 3. **Too many files** → the existing `You can attach up to 10 files.` In all three the composer state is untouched and the picker stays available, so retry is immediate. On a multi-select each file gets its own outcome: one bad file is named and skipped, the rest still attach. |
| **Unhappy, non-retryable** | — | **Not applicable.** Every selection-time rejection is a property of *the file the user chose* (its byte size, or that it could not be measured), so choosing a different file always recovers. There is no server verdict, no permission verdict, and no irreversible state at this stage. Terminal *upload* failures are a different, pre-existing chip state (`retryFile`) and are out of scope. |
| **Empty** | — | **Not applicable.** The gate only runs on a non-empty picker result; a cancelled pick returns `[]` and the composer is unchanged, so there is no empty surface to explain and no next-step CTA beyond the picker the user just dismissed. (A zero-*byte* file is the "unmeasurable or empty" retryable rejection above, not this state.) |

### Scope

Everything is inside `apps/mobile/src/components/kilo-chat/`. **No file under `packages/` changes** — the
shared `ATTACHMENT_MAX_BYTES` (100 MB) governs the server and the web client and is intentionally left
alone; `maxBytes` is a per-caller prop, so mobile can pass a smaller number without affecting web.

One environment fact for reading the E2E evidence: `services/kilo-chat/wrangler.jsonc:20` declares the media
bucket with `"remote": true`, so the happy-path upload below signs and PUTs against the **real**
`kilo-chat-media` R2 bucket, not a local one. That is pre-existing local kilo-chat behaviour, not something
this PR changes. Assertions B–D (the ones that verify this fix) are pure client behaviour and touch no
network.

## Verification

Unit coverage (vitest, colocated): accepted exactly at the 10 MiB boundary; rejected one byte above it with
the exact literal message `huge.bin exceeds the 10 MB attachment limit.`; mixed selection; capacity
truncation; size-rejection precedence over the capacity toast; unknown/`null`/zero size rejected as
unreadable; both picker paths return selections without calling `fetch`; `materializeAttachment` does read
bytes; and `Math.max` semantics both ways (measured over reported, reported over unreadable stat).

- [x] E2E (Android emulator, KiloClaw chat composer `chat/[sandbox-id]/[conversation-id]`, real device
  flow: login → KiloClaw tab → instance → new conversation → paperclip confirmed in the view hierarchy):
  **VERIFICATION PASSED** on the merged head. Two rounds were run; round 1 was environment-blocked (the
  `R2_*_KILOCHAT_MEDIA` Secrets-Store pair did not exist anywhere on the machine yet — every size-gate
  assertion still passed) and round 2 ran the full set once the pair was provisioned:
  - **A — happy path:** `small.bin` (1 MiB) picked → chip `small.bin, ready` → sent → attachment renders in
    the transcript (`Open small.bin`). Backend: `GET …/attachments/<id>/url 200`, `POST /v1/messages 201`.
  - **B — oversized rejected (the fix):** `big.bin` (12 MiB) picked → the toast reads verbatim
    **`big.bin exceeds the 10 MB attachment limit.`** No chip added, composer unchanged, app responsive on
    the same screen (screenshot captured at +2.2 s, corroborated by two bracketing frames).
  - **C — no crash, no silent drop:** logcat window free of `FATAL EXCEPTION`/`OutOfMemoryError`, no
    red-box/fatal JS in Metro, app PID stable across the whole window.
  - **D — recoverable:** immediately after the rejection, `small.bin` re-picked → chip `ready` → sent
    (`POST /v1/attachments/init 200` in 36 ms, `POST /v1/messages 201`) — the rejection leaves the composer
    fully usable.
  - The KiloClaw agent's reply to the random-bytes fixture errored (`Something went wrong… use /new`) — the
    agent turn on meaningless content, outside every acceptance criterion; the attachment pipeline itself
    returned 200/201 throughout.

## Visual Changes

None — this change gates file reads; it adds no UI surface. The only visible behaviour is the pre-existing
rejection toast (`<name> exceeds the 10 MB attachment limit.`), which now appears before any bytes are read
instead of after; E2E frames of it were captured and are described under Verification.

## Reviewer Notes

- The two `eslint-disable typescript-eslint/promise-function-async` comments in the picker were removed with
  the async signatures; the one on `pickAttachmentsFromSource` intentionally stays.
- The single `no-await-in-loop` disable in the composer marks the deliberate sequential materialize loop
  (bounding peak memory to one file).
- `addFilesWithinAttachmentCapacity` was deleted: the count limit now lives entirely inside
  `selectAllowedAttachments`, which already trims the selection to available capacity before any bytes are
  read.


iscekic:
(bot) @kilocode-bot please review

iscekic:
(bot) Triage of the three non-blocking nits from the 18:00Z review:

1. **Empty vs unreadable copy** — valid observation of the divergence from the cloud-agent precedent. The fail-closed rejection of zero-byte/unmeasurable picks is a deliberate, documented behaviour change of this PR; splitting a third `empty` reason changes user-facing copy and is deferred to a follow-up rather than expanding this fix's scope.
2. **Aggregated rejection toasts** — valid UX observation. The single-toast-first-failure shape is deliberate (multi-select keeps the files that succeeded; each materialize-time failure names its file). Aggregating pick-time rejections is a behaviour change beyond this fix, deferred to a follow-up.
3. **`response.ok` check in `materializeAttachment`** — reasonable defensive hardening; conditional on `file://` fetches resolving non-2xx rather than rejecting, which is not observed in E2E. Deferred to the same follow-up.
4. **Stale `existingCount` race** — pre-existing, as noted, and explicitly follow-up material: the 10-file cap bounding retained bytes predates this change.

No code change on this head; the approving summary stands on the reviewed tree.

iscekic:
(bot) @kilocode-bot please review

iscekic:
(bot) Triage of the two new non-blocking suggestions from the 20:23Z review of head 47015c1:

1. **`Math.max` NaN propagation** (`message-attachment-picker.ts:39`) — valid defensive point; the runtime types declare `file.size: number` and expo-file-system SDK 57 documents `0 if the file does not exist, or it cannot be read`, so the false-rejection path requires an out-of-contract `undefined`. Filtering candidates before the max is the more robust shape and is added to the same follow-up as the carried-over items; no code change on this approved head.
2. **Dead `fileSize` pass-through** (`:40`) — correct, harmless readability residue from the gate refactor; folded into the same follow-up.

The four previously triaged items (aggregated rejection toasts, empty-vs-unreadable copy split, `response.ok` guard, stale `existingCount` race — the one worth an issue) remain follow-up dispositions as recorded in the 18:05Z triage. No actionable comment is outstanding; the approving summary stands on the current head.

iscekic:
(bot) E2E evidence frames (round r1, Android emulator, KiloClaw chat composer) preserved at `/Users/igor/Projects/e2e-evidence/attach-oom-11fd/`: `big-pick-t2.png` — the rejection toast reading verbatim `big.bin exceeds the 10 MB attachment limit.`; `big-reject-1.png` — same rejection, earlier capture; `d-chip-failed.png` — the round-1 environment-blocked upload state (chip `failed` on the missing R2 media secret, since provisioned; round r2 then passed the full set, see Verification above).

jeanduplessis (approved):

PR #4846, opened 2026-07-28
feat(auto-routing): owner-configurable Efficient model pools

iscekic:
## Summary

**What.** `kilo-auto/efficient` can now be customized per owner: a personal user or an organization configures 1–10 exact `(managed model, canonical catalog variant)` pairs on the existing Auto routing card. Missing exact pairs are benchmarked on demand through the existing benchmark Worker; routing then decides only among ready, compatible entries and keeps the balanced fallback whenever none are usable.

**Why.** Owners want the Efficient auto model constrained to the exact models and variants they trust (same-model `xhigh` vs `max` stay distinct end-to-end) without losing the platform default when their pool cannot serve.

**How.**

- Shared contracts for Pool entries / Benchmark profiles with exact-pair identity (`packages/auto-routing-contracts`), ADR 0002 updates, canonical terms in CONTEXT.md.
- Global benchmark profile registry in the benchmark Worker: atomic all-or-nothing D1 admission (10 new/retried pairs per owner per rolling 24h, globally deduplicated), exact-pair composite keys with a rebuild-safe migration, profile-purpose runs on the existing single decider slot, and sparse custom routing tables assembled from ready/current profiles only.
- Auto-routing Worker: pools stored beside mode on the existing routing-settings Durable Object (mixed put/delete commits in one storage transaction), custom decide path failing closed on inactive/missing capability rows, org → personal → platform inheritance resolved independently per field.
- Gateway resolution serves the exact catalog variant (full settings, not effort-only mapping); a removed variant or unusable pool falls back to balanced. Legacy `reasoningEffort` decisions stay parseable.
- Web: validated settings API (enabled/visible/managed catalog entries only; virtual, experiment, hidden, denied, BYOK-only rejected) and the pool editor on the existing Auto routing card for personal and organization settings (owners/billing managers edit; members read-only). Deploy-order safe during the web-before-worker window: the card falls back to the legacy mode route (`poolSupported: false` hides pool controls; mode saves stay mode-only at every worker version; a mid-session rollback answers a retryable 503 instead of mis-saving). Malformed worker settings responses degrade to a graceful 502.

## Verification

- [x] **Local E2E (full local stack: nextjs + auto-routing + auto-routing-benchmark + D1, Chromium + curl + wrangler d1 + psql) — PASSED, all 10 acceptance criteria:**
  1. Personal/org pools save, clear, and inherit independently (org override served `claude-haiku-4.5`; clearing restored personal `claude-sonnet-4.5`, instrumented proof; mode and pool saves independent).
  2. Member read-only (disabled controls, PUT → 401); billing_manager flip restores full edit.
  3. Validation rejects hidden/virtual/missing-variant/unknown ids with specific reasons at API and UI.
  4. Same-model `xhigh` + `max` persist and render as distinct rows and distinct D1 PKs.
  5. Benchmarking state with 15s polling that transitions to Ready and stops at terminal.
  6. Failed → Retry re-admits (ledger +1); org-denied entry renders Unavailable with no retry and is removable; 10/24h quota rejects with `retryAt` copy, atomically (zero partial rows).
  7. Empty/inherited copy verified verbatim for personal and org cards.
  8. Load and save failures are retryable with Try again CTAs and recover.
  9. **Real routing assertion (real `kilo-auto/efficient` calls only):** seeded ready pool constrained the decision to pool entries (custom table, 18 routes, threshold-winner xhigh); stale profiles → balanced fallback with uncharged re-admission; no pool → default platform path.
  10. 19 state screenshots at 375px + desktop (personal + org variants of ready, benchmarking, failed/retry, unavailable, quota, empty, load-error, member read-only).
  - Deploy-order fallback matrix skipped as structurally unproducible in a single-tree local stack; covered by automated route/card tests including the rollback-503 path.
- [x] Automated coverage on the latest head: contracts 115 tests, benchmark 235, auto-routing 177, web focused 119 + admin-client 36; typecheck, lint, format, dependency-cycle, changes-only repo typecheck all clean; fresh cumulative review over the whole diff: no findings.

## Visual Changes

Local screenshots of the Auto routing card on `/profile` (desktop + 375px). Survey modal may appear for fresh seed users — dismiss it; the pool editor is underneath.

| Empty pool (desktop) | Add model |
|---|---|
| empty pool | add model |

| Pool card focus | Mobile 375px |
|---|---|
| pool card |  |

| Full profile after edit |
|---|
| after edit |

## Reviewer Notes

- Focus: exact-pair identity at every hop (picker → contracts → D1 composite keys → assembly → decide/sticky → gateway `getModelVariants`); admission vs save atomicity across the two workers (quota 429 with `retryAt`, all-or-nothing D1 admission, one DO storage transaction); the deploy-order fallback matrix (old web + new worker, new web + old worker, rollback mid-session).
- D1 migration 0005 rebuilds the profile tables with backfill — verify the generated SQL preserves legacy rows and exact-pair keys; 0006/0007 are additive.
- Accepted residuals: same-owner concurrent quota-boundary racing (documented in-code; quota is throttling, per-pair charge integrity is atomic); deploy-window pool saves answer a retryable 503 by design rather than risking a silent pool clear or resurrection.

### Follow-ups noted during review (non-blocking)

- `createWorkerAdminFetch` keeps the throw-on-malformed convention for the sibling admin routes; the settings client now sets the safer safeParse→502 precedent those can adopt.
- The new 502 mapping reports intermediary non-JSON failures (edge HTML error pages) as `502` rather than the upstream status; genuine worker error responses are always JSON, so this only affects edge cases that surface as the same retryable load/save error either way.
- Degrade paths have no log/Sentry signal yet (consistent with the route's pre-existing 502 branches).


apps/web/src/lib/ai-gateway/auto-routing-admin-client.ts
```diff
@@ -75,3 +90,77 @@ export function updateAutoRoutingMode(owner: {
...
+      status: response.status,
+      body: parsedError.success
+        ? parsedError.data
+        : { error: `Request failed: ${response.status}` },
+    };
+  }
+
+  return {
+    status: response.status,
+    body: AutoRoutingSettingsResponseSchema.parse(body),
```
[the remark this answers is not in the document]
iscekic: (bot) Fixed in 3239b9d78. `fetchAutoRoutingSettingsAdmin` now safeParses the success body and guards `response.json()`: a schema-mismatch or non-JSON 2xx resolves to `{status: 502, body: {error: 'Invalid worker settings response'}}` instead of throwing, so the route surfaces the intended graceful 502 through its existing error branch. Covered by new client tests (schema-mismatch 2xx, non-JSON 2xx, plus 200/404/429 regressions) and a fresh impl-reviewer pass with no findings.

apps/web/src/app/api/auto-routing/settings/route.ts
```diff
@@ -0,0 +1,269 @@
...
+
+function workerResultResponse(
+  result: { status: number; body: unknown },
+  owner: { userId: string; organizationId: string | null }
+): Promise<NextResponse> | NextResponse {
+  if (result.status >= 400) {
+    return workerErrorResponse(result);
+  }
+
+  const parsed = AutoRoutingSettingsResponseSchema.safeParse(result.body);
```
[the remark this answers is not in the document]
iscekic: (bot) Fixed in 3239b9d78 (same commit as the sibling thread). The client no longer throws on malformed success bodies: it returns a `{status: 502, body: {error: 'Invalid worker settings response'}}` result, which reaches this route handler and passes through the existing `result.status >= 400` branch — the graceful 502 is now reachable for both non-JSON and schema-mismatch 2xx responses. The route-level safeParse stays as defense-in-depth.

services/auto-routing-benchmark/src/profiles.ts
```diff
@@ -0,0 +1,555 @@
...
+    .where(
+      and(
+        eq(profileRequestEvents.owner_type, input.ownerType),
+        eq(profileRequestEvents.owner_id, input.ownerId),
+        gte(profileRequestEvents.admitted_at, windowStartIso)
+      )
+    )
+    .orderBy(asc(profileRequestEvents.admitted_at));
+
+  if (recentEvents.length + chargedAdmissions.length > PROFILE_ADMISSION_LIMIT) {
```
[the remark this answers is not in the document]
iscekic: (bot) Not changing this — it is the accepted residual risk documented during the plan and cumulative reviews, and this report cites no new evidence. The window is deliberately narrow: the quota is request throttling (10 new/retried pairs per owner per rolling 24h), not billing; the race is same-owner and self-inflicted (two concurrent saves), and the worst outcome is bounded over-admission of legitimate, globally deduplicated benchmark work. The integrity-critical guarantee — that one exact pair is never charged twice across owners or concurrent calls — IS atomic via the in-batch INSERT...SELECT WHERE NOT EXISTS guard and the composite PK. Moving the aggregate count into the batch would require per-row conditional count offsets on a hot path for no correctness gain; the accepted trade-off is recorded in the review history. If a future requirement makes quota billing-relevant, this check must become atomic first.

iscekic:
(bot) @kilocode-bot please review the latest head 3239b9d78 — it addresses your two malformed-response findings from the previous head; the third thread was replied to with rationale and resolved.

iscekic (commented):
Self-review pass complete — all 65 files read in full against the ADR invariants.

Verified end-to-end: exact-pair identity holds at every hop (picker → contracts XOR rules → D1 composite keys with rollback-safe rebuild/backfill → provenance-bound sparse assembly → fail-closed custom decide → variant-aware stickiness → gateway `getModelVariants`); admission-before-persist ordering with all-or-nothing quota and globally deduplicated charging; platform/profile single-slot sharing with platform priority; the deploy-order matrix (404→legacy synthesis with `poolSupported: false`, mode-only legacy saves, PUT 404→retryable 503, no legacy PUT fallback).

One finding, fixed directly in 061ff6fcc: the settings PUT contract parse spread the client body *after* auth-resolved owner identity, letting the body override `ownerType`/`ownerId` in `parsed.data` (benign today since downstream uses `owner.*`, but a future-trap contradicting the comment). Now auth identity always wins; 26/26 route tests pass.

Non-blocking nits noted but not actioned: `BenchmarkProfileOwnerTypeSchema`/`AutoRoutingModeOwnerTypeSchema` enum sync rests on a comment; a dead `unknown_variant` fallback branch in `validateSingleEntry`.

docs/adr/0002-auto-routing-efficient.md
```diff
@@ -41,41 +41,75 @@ Shared request-classification code (prompt, parsing, taxonomy, tier derivation,
 routing-table schema) lives in `packages/auto-routing-contracts` so the benchmark
 replays the exact code production runs.
 
+Owner-configurable **Efficient model pools** extend this design without a new
```
jeanduplessis: nit(non-blocking): I would have expected a new ADR to capture the change in auto routing architecture rather than editing an existing one. ADRs aren't living specifications. They are point in time architectural decisions.

services/auto-routing-benchmark/src/run.ts
```diff
@@ -400,17 +546,35 @@ export async function startRun(
     throw error;
   }
 
+  if (purpose === 'profile') {
+    try {
+      await markProfilesRunningForRun(env.BENCH_DB, runId, modelEntries, {
+        engineIdentity,
+        repetitions,
+      });
+    } catch (error) {
```
jeanduplessis, line 555: **Major — profile claim failures can repeatedly launch duplicate benchmark runs**

`markProfilesRunningForRun` failures are logged and suppressed here, so `startRun` continues and enqueues the profile run without confirming that its profile rows moved from `pending` to `running`.

When that run completes, `markProfilesReadyForRun` only updates rows with the matching `run_id` and `status = running`. The unclaimed rows therefore remain pending, and the subsequent `drainPendingProfileBatch` can start another run for the same entries. A transient D1 failure can leave profiles permanently pending while repeatedly consuming benchmark capacity.

Please abort startup when the claim transition fails: mark the inserted run failed and propagate the error, or atomically combine run insertion and profile claiming. The startup tests should also cover a rejected `markProfilesRunningForRun` and verify that no work is enqueued.

jeanduplessis (commented):
Review finding: profile claim failures can repeatedly launch duplicate benchmark runs.

jeanduplessis (changes requested):

services/auto-routing-benchmark/src/run.ts
iscekic: (bot) Fixed in 0e09154f5. When `markProfilesRunningForRun` rejects after `insertRun`, `startRun` now calls `failRunAndDrain` (marks the run failed, frees the shared decider slot, attempts drain) and rethrows — no queue work is enqueued. Covered by a new profile-runs test that asserts reject + `markRunFailed` + no `queueSendBatch`. Package tests: 236 passed.

docs/adr/0002-auto-routing-efficient.md
iscekic: (bot) Addressed in 0e09154f5. ADR 0002 restored to the main-branch baseline (point-in-time efficient routing decision). Owner-configurable Efficient model pools live in new ADR 0003, cross-linked to 0002.

jeanduplessis (approved):

PR #4872, opened 2026-07-29
feat(bitbucket) blobless wrapper

St0rmz1:
## Summary

  Completes blobless partial clone support for Bitbucket code review in the wrapper. GitHub and GitLab already use a blobless clone to stay under the clone timeout on large repositories. This extends the same treatment to Bitbucket, but only when the session origin carries an outbound capability (the `kbb1.` prefix), because a capability origin stays authenticated through the outbound interceptor and can serve the lazy blob fetches a partial clone defers. A raw-token Bitbucket origin is credential-stripped after bootstrap, so it still gets a full clone.

  This pairs with the Bitbucket outbound credential injection in #4868 (already merged), which mints those capabilities.

  ## Changes

  - `isBloblessReviewCloneEligible` now returns true for a Bitbucket review session when the origin token is a `kbb1.` capability, in addition to the existing GitHub and GitLab cases.
  - Added `hasBitbucketReviewCapability` and the `BITBUCKET_CAPABILITY_PREFIX` (`kbb1.`) constant.
  - `sanitizeBitbucketCodeReviewRemote` skips stripping the origin when it holds a capability, so the credential stays available for the partial clone's later lazy blob fetches. A raw workspace token is still stripped as before.
  - Test coverage for the contained Bitbucket path (blobless clone used, capability origin kept).
  - Corrected the `useBlobless` comment, which previously said only GitHub/GitLab qualify and described history retention incorrectly.

  ## Verification

  Manually confirmed that Bitbucket Cloud (`bitbucket.org`) honors `git clone --filter=blob:none`: a partial clone of a public bitbucket.org repo showed the server withholding blobs (promised objects, fetched lazily), with no "filtering not recognized by server" warning. This was the load-bearing assumption for the change.

  No end-to-end contained-session run yet. The path is dormant until the `BITBUCKET_TOKEN_CONTAINMENT_ORG_IDS` allow-list enables an org, so a live Bitbucket review will be smoke-tested after enabling post-deploy. Automated coverage: 49 wrapper session-bootstrap tests pass, including the new capability blobless case.

  - [ ] Live Bitbucket review after enabling the allow-list (post-deploy)

  ## Visual Changes

  N/A

  ## Reviewer Notes

  - Bitbucket blobless is intentionally gated on the `kbb1.` capability. Without a capability, the origin is credential-stripped and a partial clone's lazy fetches would fail, so those sessions keep a full clone.
  - The capability origin is deliberately left in place by `sanitizeBitbucketCodeReviewRemote` (it is scoped to one repo and useless outside the container), so blob fetches during the review authenticate through the outbound interceptor.
  - Fallback: if a server rejects `--filter` outright, the clone retries once as a full clone, so a review is never lost to the optimization.

### DIFF
--- a/services/cloud-agent-next/wrapper/src/session-bootstrap.ts
+++ b/services/cloud-agent-next/wrapper/src/session-bootstrap.ts
@@ -345,14 +345,30 @@ function isBitbucketReviewSession(
   return isCodeReviewSession(request) && repo?.kind === 'git' && repo.platform === 'bitbucket';
 }
 
+// Wire-format prefix of a Bitbucket outbound session capability (see the
+// git-token-service BitbucketSessionCapabilityCodec). A capability in the origin
+// stays authenticated through the outbound interceptor, unlike a raw token which
+// is stripped after bootstrap.
+const BITBUCKET_CAPABILITY_PREFIX = 'kbb1.';
+
+function hasBitbucketReviewCapability(request: WrapperSessionReadyRequest): boolean {
+  return (
+    isBitbucketReviewSession(request) &&
+    typeof request.repo.token === 'string' &&
+    request.repo.token.startsWith(BITBUCKET_CAPABILITY_PREFIX)
+  );
+}
+
 function isBloblessReviewCloneEligible(request: WrapperSessionReadyRequest): boolean {
   if (!isCodeReviewSession(request)) return false;
   const repo = request.repo;
-  // Only GitHub and GitLab: their session origin keeps working credentials via
-  // outbound injection, so blobs deferred by a partial clone can be fetched
-  // lazily during the review. Bitbucket's origin is credential-stripped, and
-  // other/unknown git remotes have no such guarantee, so they keep a full clone.
-  return repo?.kind === 'github' || (repo?.kind === 'git' && repo.platform === 'gitlab');
+  // GitHub/GitLab keep working credentials via outbound injection. Bitbucket
+  // keeps them only when the session uses an outbound capability (a raw-token
+  // origin is credential-stripped after bootstrap). Other/unknown git remotes
+  // have no such guarantee, so they keep a full clone.
+  if (repo?.kind === 'github') return true;
+  if (repo?.kind === 'git' && repo.platform === 'gitlab') return true;
+  return hasBitbucketReviewCapability(request);
 }
 
 async function cloneRepository(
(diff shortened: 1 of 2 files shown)


services/cloud-agent-next/wrapper/src/session-bootstrap.ts
```diff
@@ -539,12 +555,20 @@ async function sanitizeBitbucketCodeReviewRemote(
...
-  // origin credentials, which would break a partial clone's later blob fetches.
   if (!isBitbucketReviewSession(request)) {
     return false;
   }
+  // A capability origin stays authenticated through the outbound interceptor and
+  // is safe to expose (scoped to one repo, useless outside this container), so it
+  // must stay in place for a blobless clone's later lazy blob fetches. Only a raw
+  // workspace token needs stripping. Either way this is a handled code-review
+  // remote (return true), so callers do not refresh a token over it.
+  if (
```
pandemicsyn: sanitizeBitbucketCodeReviewRemote re-implements the capability-prefix check inline (`request.repo.token.startsWith(BITBUCKET_CAPABILITY_PREFIX)`) instead of reusing the newly added `hasBitbucketReviewCapability` helper, even though `isBitbucketReviewSession` has already narrowed `request` to the same shape the helper expects.

Suggested fix: Replace the inline `typeof request.repo.token === 'string' && request.repo.token.startsWith(BITBUCKET_CAPABILITY_PREFIX)` check with a call to `hasBitbucketReviewCapability(request)` to keep the single source of truth for what counts as a capability token.

services/cloud-agent-next/wrapper/src/session-bootstrap.ts
```diff
@@ -539,12 +555,20 @@ async function sanitizeBitbucketCodeReviewRemote(
...
+    typeof request.repo.token === 'string' &&
```
pandemicsyn: Related but separate from the DRY nit above: services/cloud-agent-next/src/session-service.ts has its own sanitizeBitbucketCodeReviewRemote (~lines 2610-2621) that isn't touched by this PR and has no equivalent capability check — it unconditionally calls updateGitRemoteUrl for any Bitbucket code-review session. That path runs whenever requiresPreparedDevcontainerRuntime is true (devcontainer-requested or dind- sandboxes), which is independent of createdOnPlatform === 'code-review', so a devcontainer/dind Bitbucket review session carrying a kbb1. capability token would have it stripped there instead of preserved. Suggest mirroring this capability check (or extracting a shared hasBitbucketReviewCapability-style helper) into session-service.ts as a follow-up so the optimization doesn't silently regress for that overlap.

pandemicsyn (approved):
one nit, and one potential issue (probably fine as follow up) inline.

PR #4997, opened 2026-08-04
feat(dev): give each worktree its own database

iscekic:
## Problem

Every worktree shared one Compose project (`dev`), so one PostgreSQL container, one volume, and one lifecycle. A sibling worktree's `pnpm dev:stop` ran `docker compose down` and dropped the connections of whatever else was running: a parallel section's check suite failed 713 times with refused connections, before any product assertion. The tmux-session guard in `dev:stop` cannot cover a caller that holds no dev session, such as `pnpm test:db` plus `pnpm validate`.

## Change

A worktree with a port offset runs its own Compose project on offset host ports. It owns its containers, its volume, and its teardown. The primary checkout keeps the default project and the default ports, so existing local databases and CI are unchanged.

`pnpm dev:infra-env` publishes the endpoints; `dev:start`, `dev:env`, and `test:db` call it:

- `dev/.env` carries `COMPOSE_PROJECT_NAME` and one `KILO_*_PORT` per container. Compose reads that file by default, because `dev/` is the project directory of `dev/docker-compose.yml` — so a raw `docker compose -f dev/docker-compose.yml …` call in a terminal agrees with the runner.
- `.env.local` gets this worktree's `POSTGRES_URL`, `REDIS_URL`, and `UPSTASH_REDIS_REST_URL`. That is what Next.js, drizzle-kit, `pnpm dev:seed`, and the app tests read. A value that names another host (a remote database, a tunnel) is left alone.
- Every worker command gets `CLOUDFLARE_HYPERDRIVE_LOCAL_CONNECTION_STRING_HYPERDRIVE`, because each `wrangler.jsonc` commits a `localConnectionString` on the default port.
- `scripts/dev.sh` now reads `REDIS_URL` and `UPSTASH_REDIS_REST_URL` from the env files before falling back to the default ports.
- `pnpm dev:stop` tears down its own containers whenever the worktree published `dev/.env`; without that file it keeps the sibling guard.
- `grafana` loses its fixed `container_name`, which would collide across projects.

A new worktree starts with an empty database. Run `pnpm test:db` for migrations, then seed what the task needs.

## Verification

- `pnpm test:dev-local`: 185 pass, including three new cases in `dev/local/infra-env.test.ts`.
- `dev/local/infra-env.ts` against a scratch root at offset 700: `dev/.env` and `.env.local` written as expected, and `docker compose config` in that root reports `name: kilo-dev-fakeroot` with published ports 6132, 7079, 8779.
- Two live stacks on different offsets reading their own data: running now, result to follow in a comment.


iscekic:
## Verification: two live stacks, two databases

Two worktrees of this branch, `dbtest-a` (offset 600) and `dbtest-b` (offset 700), each ran `pnpm test:db` and `pnpm dev:start app`.

Published endpoints:

| | dbtest-a | dbtest-b |
|---|---|---|
| Compose project | `kilo-dev-dbtest-a` | `kilo-dev-dbtest-b` |
| postgres | `0.0.0.0:6032->5432` | `0.0.0.0:6132->5432` |
| redis / redis-http | 6979 / 8679 | 7079 / 8779 |
| `POSTGRES_URL` in `.env.local` | `…@localhost:6032/postgres` | `…@localhost:6132/postgres` |
| nextjs | 3600 up | 3700 up |

Each `pnpm run test:db` applied the migrations to its own container. `pnpm dev:seed app:create-user` then wrote one distinct user per worktree, through the same `load-env` and `computeDatabaseUrl` path the app uses:

```
dbtest-a $ psql -Atc 'select normalized_email from kilocode_users'
ada-a@example.com
dbtest-b $ psql -Atc 'select normalized_email from kilocode_users'
bob-b@example.com
```

Each worktree's database holds only its own row.

## The reported failure no longer reproduces

`pnpm dev:stop` in `dbtest-a` removed only `kilo-dev-dbtest-a-*`:

```
$ docker ps            # after dbtest-a stopped
kilo-dev-dbtest-b-postgres-1   0.0.0.0:6132->5432/tcp
kilo-dev-dbtest-b-redis-1      0.0.0.0:7079->6379/tcp
kilo-dev-dbtest-b-redis-http-1 0.0.0.0:8779->80/tcp
```

`dbtest-b` kept serving (`nextjs 200`) and kept reading its own database. Before this change, that teardown dropped every worktree's connections.

jeanduplessis (approved):

iscekic:
## Review response

Three of the four findings are fixed in the amended head. One is rejected.

**Fixed — Compose project name collision (`dev/local/infra-env.ts`).** The name is now `kilo-dev-<slug>-<offset>`. The port offset lease already gives every live stack an offset of its own, so two worktrees whose basenames normalize to the same slug can no longer share one project. New case in `infra-env.test.ts`.

**Fixed — stale app env when the offset returns to 0 (`dev/local/infra-env.ts`).** The offset-0 branch now puts `.env.local` back on the default ports instead of only deleting `dev/.env`. `applyAppEnv` also compares host and port rather than the whole string, so `postgresql://…@localhost:5432/postgres` is left byte-for-byte alone. New case in `infra-env.test.ts`.

**Fixed — unused export (`dev/local/services.ts`).** `getInfraBasePort` has a caller now: the offset-0 reset needs the base ports, which are no longer reachable through `getService`.

**Rejected — `test:db` without `.env.local` (`package.json`).** The inline `POSTGRES_URL` default is the defect this PR removes: as a process env var it beat the worktree's own `.env.local`, so every offset worktree migrated the primary database. Restoring it restores that bug. Without `.env.local`, `pnpm test:db` now stops at a clear `POSTGRES_URL not configured`, and `.env.local` is a documented prerequisite: DEVELOPMENT.md creates it in step 3, before the database in step 4 and the migrations in step 5. A prepared worktree always has one, because `scripts/worktree-prepare.sh` copies it from the main worktree.

Also fixed: the `trufflehog` check flagged a Postgres connection string with credentials in a test fixture. The fixture never needed them; it now reads `postgres://db.example.com:5432/prod`. The fix is amended into the same commit, because a follow-up commit would leave the string in the branch's history.

PR #5015, opened 2026-08-04
feat(code reviewer) reap stale code reviews

St0rmz1:
## Summary

Reviews stranded in `pending`, `queued`, or `running` were never closed out.
Nothing on any schedule terminalizes them: the 90 minute staleness cutoff only
stops stale runs counting toward owner capacity, it never fails them. A stranded
review can leave an `in_progress` GitHub check run spinning on its pull request
indefinitely, and on any repo that requires the Kilo Code Review check, that
blocks merging.

This adds an hourly reaper cron that closes any review whose last sign of life
(`COALESCE(started_at, updated_at, created_at)`) is older than 48 hours:

- Marks the review failed with a new `abandoned` terminal reason and closes its
  own non-terminal attempts with it, the same way the supersede path closes both
  tables together.
- Completes the provider-side gate: GitHub check run as `timed_out`, GitLab
  commit status as `canceled`. Bitbucket has no status path and is skipped.
- The claim is an optimistic lock on the selected row's `updated_at`, so a
  re-dispatch or late callback arriving mid-batch wins the race instead of
  being overwritten.
- All provider work runs inside a per-row catch: suspended installations and
  malformed legacy rows are counted in the summary, not fatal to the batch.
- Dashboard-only manual reviews (`outputMode: 'kilo'`) publish nothing.
- Each run reports `remaining`, the count of still-stale rows using the same
  predicate as selection, so the backlog drain is observable from the log line.

Reaped rows are deliberately kept out of health metrics, in both directions:

- `abandoned` joins `CODE_REVIEW_BENIGN_TERMINAL_REASONS`, so the reaper's
  cleanup does not read as an error spike. The live queue health counters
  remain the signal for stuck work.
- The alerting detectors' shared CTE also excludes it. A reaped review that
  never started has no `started_at`, so its effective start collapsed to the
  reap timestamp and would have inflated `started_count`, diluting the
  error-spike and slow-review rates and masking a real incident.
- The admin `avg_duration_seconds` excludes it: a reap timestamp is a closure,
  not a duration measurement.

Supporting changes: `resolveGitLabAccessToken`/`getGitLabInstanceUrl` moved
out of the status-callback route into `lib/code-reviews/platform/gitlab-access`
(pure move, the reaper needs them too), and the repeated
`['pending','queued','running']` literal became `NON_TERMINAL_CODE_REVIEW_STATUSES`
in dispatch-constants. No migration: `terminal_reason` is an unconstrained text
column, and the db/worker-utils enum parity test covers the new value.

## Verification

- [ ] Not manually tested against a live pull request yet. The check-run close
      path is covered by unit tests only; the integration suite runs the
      selection, claim, and attempt-close SQL against a real database,
      including the microsecond `updated_at` round trip the optimistic lock
      depends on. Suggested first manual test: run the cron with a small batch
      and confirm a known stuck check run flips to `timed_out`.

## Visual Changes

N/A (the admin Error Analysis table gains an "Abandoned (reaped)" row label
once such rows exist; no layout changes).

## Reviewer Notes

- Deliberate scope cut: the reaper posts no PR comment. Closing the check run
  is visible on the PR already, and a notification can be layered on later
  without touching this design.
- The benign-list entry is an exception to the non-benign default documented in
  `schema-types.ts`; the reasoning is in a comment at the entry itself.
- During the initial drain expect nonzero `providerFailures`: part of the
  backlog references suspended or uninstalled integrations whose check runs
  cannot be closed. The row still terminalizes; the failure is counted, not
  retried.
- The drain runs oldest first at the capped batch rate; `remaining` in the
  summary log line tracks progress until it reaches zero.

### DIFF
--- a/apps/web/src/lib/code-reviews/reap-stale-reviews.ts
+++ b/apps/web/src/lib/code-reviews/reap-stale-reviews.ts
@@ -1,0 +1,243 @@
+import {
+  cloud_agent_code_review_attempts,
+  cloud_agent_code_reviews,
+  type CloudAgentCodeReview,
+} from '@kilocode/db/schema';
+import { and, asc, eq, inArray, sql } from 'drizzle-orm';
+import { captureException } from '@sentry/nextjs';
+
+import { db } from '@/lib/drizzle';
+import { APP_URL } from '@/lib/constants';
+import { CodeReviewPlatformSchema } from '@/lib/code-reviews/core/schemas';
+import { NON_TERMINAL_CODE_REVIEW_STATUSES } from '@/lib/code-reviews/dispatch/dispatch-constants';
+import { shouldPublishCodeReviewToProvider } from '@/lib/code-reviews/manual-config';
+import {
+  getGitLabInstanceUrl,
+  resolveGitLabAccessToken,
+} from '@/lib/code-reviews/platform/gitlab-access';
+import { getIntegrationById } from '@/lib/integrations/db/platform-integrations';
+import { PLATFORM } from '@/lib/integrations/core/constants';
+import { updateCheckRun } from '@/lib/integrations/platforms/github/adapter';
+import { setCommitStatus } from '@/lib/integrations/platforms/gitlab/adapter';
+import { logExceptInTest } from '@/lib/utils.server';
+
+/**
+ * How long a review may sit in a non-terminal state before the reaper closes it.
+ *
+ * Far past anything legitimate: capacity waits resolve in minutes to hours, and
+ * the cron drain already stops reconsidering pending work after 75 minutes.
+ */
+export const REAP_STALE_REVIEW_HOURS = 48;
+
+/**
+ * Upper bound on rows closed per invocation.
+ *
+ * Needed for two independent reasons. Each reaped review can cost a platform
+ * API call against a per-installation rate limit. And the alerting detector
+ * measures failure rate in a 30 minute window keyed on completion time, so an
+ * uncapped first drain would register thousands of failures at once and read
+ * as an incident.
+ */
+export const REAP_DEFAULT_BATCH_SIZE = 25;
+
+const REAP_TERMINAL_REASON = 'abandoned';
+const REAP_ERROR_MESSAGE = 'Stopped by Kilo after the review never reached a terminal state';
+const REAP_CHECK_TITLE = 'Kilo Code Review stopped';
+const REAP_CHECK_SUMMARY = 'This review did not finish and was closed by Kilo.';
+
+export type ReapStaleReviewsSummary = {
+  selected: number;
+  terminalized: number;
+  checksClosed: number;
+  providerFailures: number;
+};
+
+/**
+ * Reviews whose last sign of life is older than the reap threshold.
+ *
+ * The COALESCE mirrors the one the dispatcher uses for staleness, so a row that
+ * was updated recently is not reaped purely because it was created long ago.
+ *
+ * Oldest first: with no notification attached to a reap, there is no freshness
+ * window to protect, and the rows that have been spinning a provider-side check
+ * the longest are the most useful ones to close first.
+ */
+function selectStaleReviews(limit: number) {
+  return db
+    .select()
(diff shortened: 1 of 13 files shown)


apps/web/src/lib/code-reviews/reap-stale-reviews.ts
```diff
@@ -0,0 +1,264 @@
...
+    terminalized: 0,
+    checksClosed: 0,
+    providerFailures: 0,
+    remaining: 0,
+  };
+
+  for (const review of stale) {
+    // Claim first. Everything after this is best-effort cleanup, and a provider
+    // call that fails must not leave the row selectable forever.
+    if (!(await terminalizeReview(review))) continue;
```
jrf0110, line 242: Might better to terminalize all reviews in a single query, no? Why do them one-by-one?

apps/web/src/lib/code-reviews/reap-stale-reviews.ts
```diff
@@ -0,0 +1,264 @@
...
+ *
+ * Matching on the snapshot's `updated_at` makes the claim an optimistic lock:
+ * every write to this table refreshes that column, so any intervening dispatch,
+ * callback, or supersede invalidates the claim and the row is skipped this run.
+ *
+ * On a successful claim the review's own non-terminal attempts are closed with
+ * it, the same way the supersede path closes both tables together. Left open,
+ * they would count as in-progress work forever in every attempt-level query.
+ */
+async function terminalizeReview(review: CloudAgentCodeReview): Promise<boolean> {
```
jrf0110, line 119: The two updates in this function should probably occur in a transaction

apps/web/src/lib/code-reviews/reap-stale-reviews.ts
```diff
@@ -0,0 +1,264 @@
...
+  };
+
+  for (const review of stale) {
+    // Claim first. Everything after this is best-effort cleanup, and a provider
+    // call that fails must not leave the row selectable forever.
+    if (!(await terminalizeReview(review))) continue;
+    summary.terminalized += 1;
+
+    try {
+      if (await closeProviderGate(review)) summary.checksClosed += 1;
```
jrf0110, line 246: Should we try to close the provider gate first so that if it fails, the review gets picked up on the next cron tick?

jrf0110 (approved):

apps/web/src/lib/code-reviews/reap-stale-reviews.ts
St0rmz1: Deliberate: the claim is an optimistic lock against each row's own `updated_at` snapshot, so a bulk version needs a VALUES join to carry per-row timestamps, and per-row claims keep one row's provider failure isolated from its neighbors. At 25 rows/hour the saved round trips don't buy anything, so I kept the simpler shape. Happy to revisit if we ever raise the batch size materially.

apps/web/src/lib/code-reviews/reap-stale-reviews.ts
St0rmz1: Good question — considered it, but claim-first is load-bearing for two reasons. Part of the backlog references suspended installations whose gate close can never succeed; gate-first would keep those rows selectable forever, and since selection is oldest-first the batch would fill with unclosable rows and the drain would wedge. And the optimistic lock can only protect what happens after it: gate-first can stamp `timed_out` on a check run for a review a concurrent dispatch just legitimately revived. The cost of this ordering is that a transient gate failure orphans that one spinner (visible via `providerFailures` + Sentry); if that shows up in practice, a narrow sweep over `abandoned` rows with still-open check runs is the fix rather than reordering.

apps/web/src/lib/code-reviews/reap-stale-reviews.ts
St0rmz1: Agreed, fixed in 3f685b2c5 — the claim and the attempts close now run in one transaction. A crash between the two writes would have left the attempts stranded permanently, since the terminal parent is never reselected. The integration suite exercises the real transaction against the database.

PR #5016, opened 2026-08-04
fix(code reviewed) status dispatch after

St0rmz1:
## Summary

The code review status callback route triggered `tryDispatchPendingReviews()` as an unawaited promise right after processing a terminal status update, then continued running and eventually returned its response. Vercel can freeze a serverless invocation once nothing is holding it open, so the dispatch call's outbound fetches to the code review worker could stall mid-flight. This showed up in Sentry as spurious `Request timeout after 10000ms` errors on both the dispatch call and the follow-up status probe.

This PR wraps that dispatch promise in Next.js `after()`, the same pattern already used by the GitHub and bot webhook handlers, so the invocation stays alive until the dispatch settles. The dispatch is still kicked off immediately and not awaited inline, so it does not delay the route's own response or the GitHub/GitLab reaction and comment updates that follow it.

## Changes

- `route.ts`: import `after` from `next/server` and pass the already-started `tryDispatchPendingReviews(...).catch(...)` promise to it instead of leaving it as a bare unawaited promise.
- `route.test.ts`: mock `next/server`'s `after` so tests can run outside a real request scope (where the real `after()` throws), while preserving the existing fire-and-forget assertions on `tryDispatchPendingReviews`.

## Verification

- [ ] No manual testing performed. Verified via `NODE_ENV=test npx jest "src/app/api/internal/code-review-status"` (127/127 passing), `scripts/typecheck-all.sh --changes-only`, and `scripts/lint-all.sh --changes-only`, all clean.

## Visual Changes

N/A

## Reviewer Notes

Reviewer focus: confirm `after(dispatchPromise)` doesn't change response timing. It doesn't, since the promise is created and handed to `after()` before any subsequent `await` in the route, and `after()` itself only registers the promise with the platform's `waitUntil` rather than awaiting it.

### DIFF
--- a/apps/web/src/app/api/internal/code-review-status/[reviewId]/route.test.ts
+++ b/apps/web/src/app/api/internal/code-review-status/[reviewId]/route.test.ts
@@ -101,12 +101,15 @@ jest.mock('next/server', () => {
   const actual = jest.requireActual<typeof nextServerModule>('next/server');
   return {
     ...actual,
-    // Route handlers here fire the dispatch off in the background via
-    // `after()`. Outside a real request scope `after()` throws, so invoke
-    // the callback immediately (without awaiting) to preserve the existing
-    // fire-and-forget semantics under test.
-    after: (fn: () => Promise<void> | void) => {
-      void fn();
+    // The route hands `after()` an already-started promise (dispatch is
+    // kicked off immediately, not deferred behind it) so it can keep the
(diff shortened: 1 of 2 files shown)


eshurakov (approved):

RSO:
Merging because @St0rmz1 is out-of-office today, and I'd like to see if this fixes the issues we're seeing in Sentry!

PR #5017, opened 2026-08-05
feat(billing): settle container usage charges

pandemicsyn:
## Summary

- Adds the first paid container-billing rollout for Gastown. Every five-minute heartbeat records accepted container time, converts only the newly accepted time into credits at the price captured when that container started, and then returns `continue`, `warn`, or `stop` based on the wallet balance after the debit.
- The meter performs the debit inside the same PostgreSQL transaction that records the heartbeat. Retried heartbeats reuse a stable identity and do not create another charge; the container interval lock and its settled-time watermark are the primary exactly-once guard.
- Adds `compute_usage_charge`, an immutable financial ledger for metered infrastructure. A charge records the source that produced it, the payer, SKU, quantity, applied rate, final microdollar amount, and the source timestamp. It is intentionally generic enough for future Worker billing without making container settlement state generic.
- The ledger is partitioned by its immutable `created_at` month. The migration creates the initial partitions, and the existing partition-maintenance cron keeps the current and next two months ready. This avoids an ever-growing single ledger table while ensuring a retry is routed to the same monthly partition as the original source event.
- Container intervals keep their own `settled_billable_seconds` watermark because cumulative settlement is container-specific. Container ledger rows use `container_usage_segment` plus the segment idempotency key as their source identity; future compute producers can use their own stable source identities and quantities.
- Integrates compute charges into personal and organization balance recomputation. The personal or aggregate organization wallet is debited atomically; organization member usage and Kilo Pass accounting remain intentionally unchanged.
- Enables the approved personal Gastown canary and stop enforcement. User and organization payer types have independent allowlists: the current production configuration enables only the named user, while all other users and all organizations remain shadow-only. A `stop` verdict blocks new work, allows a 120-second save/drain period, then force-stops the container and suppresses automatic restart.

```mermaid
sequenceDiagram
    participant G as Gastown container
    participant M as Usage meter
    participant I as Container interval
    participant L as compute_usage_charge
    participant W as Kilo-credit wallet

    Note over G,W: Paid session runs for 7 minutes
    G->>M: recordStart at 00:00
    M->>I: Create paid interval and snapshot payer plus SKU rate
    M-->>G: Accepted

    G->>M: Heartbeat at 05:00 for 300 accepted seconds
    M->>I: Lock interval and persist segment
    M->>I: Advance settled watermark to 300 seconds
    M->>L: Insert charge 1 with heartbeat source key and rate snapshot
    M->>W: Atomically debit charge 1
    M-->>G: Return continue warn or stop from post-debit balance

    G->>M: recordStop at 07:00 for 120 final seconds
    M->>I: Lock interval and persist final segment
    M->>I: Advance watermark to 420 seconds and close interval
    M->>L: Insert charge 2 with final segment source key
    M->>W: Atomically debit charge 2
    M-->>G: Return final post-debit verdict
```

## Visual Changes

N/A

## Reviewer Notes

- Billing is fail-closed: an invalid service list, malformed payer list, no eligible payer, or invalid warning threshold leaves usage in shadow mode. An empty user or organization list disables only that payer type.
- The production configuration targets `gastown` and user `daef8451-f3f3-490e-93f1-21fafc2b005e`; organization billing is disabled by an empty organization allowlist.
- The ledger has real restrictive payer/SKU foreign keys, but deliberately has no polymorphic telemetry foreign key or delete cascade. Source telemetry can be retained or deleted independently without deleting financial charges.
- A rate is snapshotted on paid intervals. Historical and recovery-created intervals remain shadow-only, so changing allowlists or prices does not retroactively change existing billing behavior.
- The existing Exa-specific partition helper, test, API route, Vercel cron path, and logging are renamed to the agnostic usage-partition maintenance surface. The same monthly scheduler now provisions both Exa and compute partitions; this PR does not add a second cron.


apps/web/src/app/api/cron/exa-partition-maintenance/route.ts
```diff
@@ -2,7 +2,10 @@ import { NextResponse } from 'next/server';
 import { captureException } from '@sentry/nextjs';
 import { db } from '@/lib/drizzle';
 import { CRON_SECRET } from '@/lib/config.server';
-import { provisionExaUsageLogPartitions } from '@/lib/exa-usage-partitions';
+import {
+  provisionComputeUsageChargePartitions,
+  provisionExaUsageLogPartitions,
+} from '@/lib/exa-usage-partitions';
```
pandemicsyn: Why is apps/web/src/app/api/cron/exa-partition-maintenance/route.ts being modified. ...ComputeUsage is totally unrelated to exa. Either compute usage gets it own partition management cronjob or we should start naming/renaming so that its more agnostic.

apps/web/src/lib/user/recompute-balances.ts
```diff
@@ -31,7 +32,7 @@ export type MigrationResult = Result<UserBalanceUpdates, string>;
  * 6. Updates the user record and transaction baselines (unless dryRun is true).
  *
  * Postconditions:
- * - microdollars_used = sum(microdollar_usage) + sum(exa_usage_log where charged_to_balance, personal)
+ * - microdollars_used = all personal LLM, Exa, and container charge ledger records
```
pandemicsyn: The original comment directly explained how microdollars_used is derived and we should maintain that.

apps/web/src/lib/usage-partitions.ts
pandemicsyn, line 1: Same comment as above, having this in exa-usage-partitions.ts is confusing.

services/container-usage-meter/wrangler.jsonc
```diff
@@ -19,6 +19,12 @@
     "logs": { "enabled": true, "head_sampling_rate": 1 },
     "traces": { "enabled": true, "head_sampling_rate": 0.01 },
   },
+  "vars": {
+    "CONTAINER_BILLING_SERVICES": "gastown",
+    "CONTAINER_BILLING_USER_IDS": "daef8451-f3f3-490e-93f1-21fafc2b005e",
+    "CONTAINER_BILLING_ORG_IDS": "",
```
pandemicsyn, line 25: fyi - Just turning this on for me in prod for now. Will extend to our org after.

apps/web/src/app/api/cron/exa-partition-maintenance/route.ts
pandemicsyn: Fixed in `0215bf958`. Renamed the existing helper, route, cron path, and logging to agnostic usage partition maintenance; no second scheduler was added.

apps/web/src/lib/user/recompute-balances.ts
pandemicsyn: Fixed in `0215bf958`. Restored the explicit `microdollars_used` derivation across the personal LLM, Exa, and compute ledgers.

apps/web/src/lib/usage-partitions.ts
pandemicsyn: Fixed in `0215bf958`. Moved the shared partition functions to `usage-partitions.ts` and renamed its test accordingly.

services/container-usage-meter/src/postgres.ts
```diff
@@ -221,6 +393,18 @@ export async function applyStartWithDb(
       };
     }
 
+    const billingMode = billingModeFor(billingConfig, input.service, input.subject);
+    if (billingMode === 'paid') {
+      const remaining = await balanceForSubject(tx, input.subject, true);
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in `cf23ae434`. Paid supersede admission now locks the open interval before the payer, with a PostgreSQL concurrency test proving the payer remains lockable while start waits on the interval.

jrf0110 (approved):

PR #5024, opened 2026-08-05
chore(cost-insights): drop the cost_insight_* tables and schema

RSO:
Final PR of the Cost Insights shutdown, after #4993 (UI and API) and #4996 (all reads and writes). The twelve tables have been unused since #4996 merged.

## The migration

`0204_drop_cost_insight_tables.sql` — twelve `DROP TABLE ... CASCADE` statements and nothing else. Generated by `pnpm drizzle generate`; no hand-editing of DDL, snapshots, or the journal.

**Deploy ordering is a non-issue.** I verified against `0203_snapshot.json` that no non-Cost-Insights table has a foreign key to any of these tables, and no view references them. All 27 foreign keys involved originate *from* the dropped tables. Combined with #4996 removing every read and write, the migration is safe to run before, during, or after the code deploy.

`CASCADE` therefore only tears down the feature's own intra-table foreign keys, indexes, and constraints — it cannot reach outside the set.

One incidental improvement: `cost_insight_owner_hour_driver_buckets.actor_user_id` was an `ON DELETE NO ACTION` foreign key to `kilocode_users`, so it could previously block hard user deletion. That blocker is gone.

## Schema and code removed

- `packages/db/src/schema.ts` (−921): 12 table definitions, 16 inferred row types, the `CostInsightEventSnapshot` Zod schemas, the `schema-types` import, and the `SCHEMA_CHECK_ENUMS` registry entries.
- `packages/db/src/schema-types.ts` (−69): the 7 Cost Insights enums. I confirmed all 7 have zero consumers outside `packages/db`.
- `packages/db/src/schema.test.ts` (−19): the expected enum values. This **must** land with the registry change — removing one without the other fails the registry test with `Missing enums`.
- `packages/db/src/cost-insight-event-snapshot.test.ts`: deleted; its only subject was `CostInsightEventSnapshotSchema`.
- `.specs/cost-insights.md`: deleted. It was fully orphaned — #4996 removed the last `CONTEXT.md` reference, and contrary to what I assumed, it was never listed in the `specs` skill index, so no index entry needed updating.
- `CONTEXT.md`: the "scheduled to be dropped" language is now inaccurate.

## Deliberately kept

**The discontinued-notice routes** from #4993, so old bookmarks and previously sent alert emails still land on an explanation.

**Two untracked production indexes.** I told you in #4996 that these would be dropped in this migration. That was wrong, and worth correcting:

| Index | Table |
|---|---|
| `IDX_coding_plan_terms_credit_transaction` | `coding_plan_terms(credit_transaction_id)` |
| `idx_microdollar_usage_org_created_at` | `microdollar_usage(organization_id, created_at) WHERE organization_id IS NOT NULL` |

Neither is known to Drizzle — absent from `schema.ts`, all 204 prior migrations, and every snapshot. They exist only where the now-deleted maintenance script was run, so `drizzle generate` will not emit drops for them. `DROP INDEX CONCURRENTLY` cannot run inside Drizzle's transactional migration, and a plain `DROP INDEX` would take an `ACCESS EXCLUSIVE` lock on `microdollar_usage`, one of the hottest tables in the system. The second index also plausibly still helps organization usage queries over a time range. Left for a separate operational decision informed by real production index-usage stats rather than a guess from code.

## Verification

- `pnpm drizzle generate` produced only the intended drops; `pnpm drizzle check` reports `Everything's fine`.
- `pnpm drizzle migrate` applied locally, then confirmed via `psql`: **0** `cost_insight%` tables remain, 214 public tables remain, and `kilocode_users`, `organizations`, `microdollar_usage`, `coding_plan_terms`, `credit_transactions` are all intact.
- `pnpm drizzle:verify-bootstrap` passes, proving the full 0000 → 0204 chain replays on an empty database (0179 and 0183 create these tables, 0204 drops them).
- `scripts/typecheck-all.sh` and `scripts/lint-all.sh`: 0 errors across all workspaces. Repo-wide `pnpm format:check` clean.
- `schema.test.ts` + `schema-types.test.ts`: 42 tests pass, covering both the schema-versus-snapshot parity check and the enum registry.
- Full `apps/web` suite: 700 suites, 8,943 tests pass. `services/kiloclaw-billing`: 184 tests pass.

## Pre-existing gap found while verifying

`packages/db` tests do not run anywhere. `apps/web/jest.config.ts:38` lists `<rootDir>/../../packages/db/src/**/*.test.ts` in `testMatch`, but `roots` is unset so it defaults to `apps/web` and Jest never crawls that directory. `packages/db/package.json` has only `typecheck` and `lint`, and no other Jest config covers it.

So `schema.test.ts` — which exists specifically to fail with "Schema is out of date! Run 'pnpm drizzle generate'" — is never executed by CI. I ran it explicitly via `--roots` to validate this change. Not fixed here since it is unrelated to this PR and would newly enforce assertions across the whole schema, but it should be fixed separately.

Also note `packages/db/src/client.test.ts` fails under that ad-hoc invocation. I verified by stashing this branch that it fails identically on clean `origin/main`, so it is pre-existing and unrelated.


CONTEXT.md
```diff
@@ -15,7 +15,7 @@ Covered domains must use this contract's canonical terms in code, docs, task des
...
 | **Security Agent Email Delivery** | Dispatch-time revalidation, email rendering, owner-aware links, and Mailgun delivery | `apps/web/src/app/api/internal/security-agent/`, `apps/web/src/lib/email.ts`, `apps/web/src/emails/` | Accepts notification identity only and loads current data before sending |
 | **Shared Security Notification Policy** | Canonical config parsing, defaults, severity thresholds, and pure event eligibility rules | `packages/worker-utils/src/security-notification-policy.ts` | Web and Worker must use same policy contract |
-| **Cost Insights** | Retired. No owned behaviour remains | `apps/web/src/components/cost-insights/CostInsightsDiscontinuedNotice.tsx` | Only the discontinued-notice routes remain; the unused `cost_insight_*` tables are scheduled to be dropped |
+| **Cost Insights** | Removed. No owned behaviour, code, or storage remains | `apps/web/src/components/cost-insights/CostInsightsDiscontinuedNotice.tsx` | Only the discontinued-notice routes remain |
```
RSO:
```suggestion
```

CONTEXT.md
```diff
@@ -154,7 +154,7 @@ Covered domains must use this contract's canonical terms in code, docs, task des
...
 - Organization subscription page supports initial direct-child distribution before first issuance, then separates current immutable issuance snapshot from next **Kilo Pass allocation plan** and effective date; it shows purchased, parent-default, direct-child allocated, paid-through, pending-cancellation, blocked, and overallocated state.
-- **Cost Insights is retired.** No Cost Insights code runs: the UI, tRPC procedures, cron sweeps, spend capture, evaluation, alerting, and rollup maintenance are all removed, and none may be reintroduced. The `cost_insight_*` tables still exist but are no longer read or written, and are scheduled to be dropped.
+- **Cost Insights is fully removed.** The UI, tRPC procedures, cron sweeps, spend capture, evaluation, alerting, rollup maintenance, and the `cost_insight_*` tables are all gone, and none may be reintroduced.
```
RSO:
```suggestion
```

jeanduplessis (approved):

PR #5058, opened 2026-08-05
fix(web): filter AI usage-limit 402 errors out of Sentry

eshurakov:
## Summary

Filters `AI_APICallError` with `statusCode: 402` out of Sentry in `beforeSend`, alongside the existing tRPC 4xx filter.

## Why

The AI gateway returns 402 (`usage_limit_exceeded` — "Add credits to continue, or switch to a free model", `llm-proxy-helpers.ts`) when the **caller's own** credit balance is exhausted. Server-side callers acting on behalf of a user — e.g. Kilo Bot processing `/api/internal/bot-session-callback` — surface this as an unhandled `AI_APICallError` (statusCode 402), which `captureException` then reports. Each event just means "a user ran out of credits mid-workflow": expected per-user billing state, not an application bug, and it scales with free/low-credit usage.

This is safe to filter globally because upstream provider 402s (our provider account) never reach AI SDK callers: the gateway converts them to a 503 and reports them itself via `captureProxyError({ trackInSentry: true })` in `app/api/openrouter/[...path]/route.ts`, so that failure mode stays visible.

## Changes

- `apps/web/sentry.server.config.ts`: add exported `isAIUsageLimitError` predicate (duck-typed like the existing `isTRPC4xxError`, so the `ai` package stays out of the Sentry init bundle) and apply it in `beforeSend`.
- `apps/web/src/sentry.server.config.test.ts`: tests for the predicate, including one built from the real `APICallError` class to pin the duck-type contract.

## Verification

- `jest src/sentry.server.config.test.ts`: **24/24 pass** (run with a scratch config identical to `jest.config.ts` minus the DB-backed global setup — this sandbox has no Docker/Postgres; the standard config's setup requires it. The suite itself has no DB dependency, so CI should behave the same).
- `oxlint` on the changed test file: clean.
- `oxfmt` applied to both files.
- Scoped `tsgo` typecheck of both changed files: clean. Full `apps/web` typecheck not run (its `@kilocode/trpc` build prerequisite exceeded the sandbox time limit).

### DIFF
--- a/apps/web/sentry.server.config.ts
+++ b/apps/web/sentry.server.config.ts
@@ -79,6 +79,24 @@ function isTRPC4xxError(error: unknown): boolean {
   );
 }
 
+// The AI gateway returns 402 when the caller's own credit balance is exhausted
+// (`usage_limit_exceeded`, "Add credits to continue, or switch to a free
+// model"), and the AI SDK surfaces that as AI_APICallError with statusCode 402
+// (e.g. Kilo Bot calling the gateway on behalf of a user who is out of
+// credits). That is expected per-user billing state, not an application bug.
+// Upstream provider 402s (our provider account) never reach callers: the
+// gateway converts them to a 503 and reports them itself via
+// captureProxyError. Duck-typed like isTRPC4xxError to keep the `ai` package
+// out of the Sentry init bundle.
+export function isAIUsageLimitError(error: unknown): boolean {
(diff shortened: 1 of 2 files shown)


chrarnoldus:
Would it be better to not capture this class of errors in `app/api/openrouter/[...path]/route.ts`?

PR #5064, opened 2026-08-05
feat(kiloclaw): close new subscriptions

jeanduplessis:
## Summary

Close new KiloClaw instance provisioning in personal and organization contexts unless the requested context already has a current subscription.

### Why this change is needed

KiloClaw is no longer accepting new instance creation without an existing current subscription. The closure must apply consistently across personal and organization flows, web and mobile surfaces, provisioning services, checkout fulfillment, and navigation so stale or historical state cannot create replacement infrastructure.

Existing live instances and current subscriptions must continue to support normal access, billing management, and configuration updates.

### How this is addressed

- Require a current active, unsuspended KiloClaw subscription in the exact personal or organization context before provisioning replacement infrastructure.
- Prevent canceled, expired, suspended, transferred-out, or purely historical subscription rows from authorizing new instances.
- Enforce the same boundary in the web routers, KiloClaw billing service, KiloClaw provisioning service, and local service fallback.
- Prevent completed or stale Kilo Pass checkouts from creating a first KiloClaw subscription.
- Disable credit-funded reprovision recovery from canceled personal subscription history.
- Replace ineligible personal and organization onboarding actions with a no-action unavailable state.
- Hide KiloClaw navigation and promotional surfaces when the current context is ineligible.
- Preserve existing live-instance access, billing management, and configuration behavior.

## Human Verification

- Targeted mobile tests passed for closed-signup state derivation and Home behavior.
- KiloClaw platform provisioning tests passed: 26 tests.
- KiloClaw billing bootstrap tests passed: 19 tests.
- Targeted web tests passed for:
  - personal and organization provisioning rejection
  - current organization subscription navigation state
  - Kilo Pass post-checkout enforcement
  - disabled canceled-subscription reprovision recovery
  - unavailable-state copy and hosting recovery behavior

### Human Reviewer Flags

- `.specs/kiloclaw-billing.md` and `.specs/subscription-center.md` now define a current subscription in the requested ownership context as the provisioning boundary.
- Subscription history alone is intentionally insufficient: canceled and historical rows cannot authorize replacement infrastructure.
- Existing live instances remain manageable; the restriction applies to infrastructure creation rather than ordinary instance configuration and lifecycle management.

### Code Reviewer Agent

<details>

- The shared billing bootstrap guard is used by both the billing Worker RPC and the KiloClaw Worker's local fallback, preventing behavior drift when the service binding fails.
- Provision entitlement is checked before durable admission reservation or provider work.
- Organization eligibility is scoped by organization and assigned user; personal subscription state cannot authorize an organization instance.
- Explicit orphan-instance bootstrap attempts are checked before provisioning can mutate infrastructure.
- Kilo Pass fulfillment requires an existing subscription row on the checkout's personal instance before credit enrollment.
- Web and mobile navigation use server-backed context eligibility rather than the legacy trial-eligibility flag.

</details>

### DIFF
--- a/apps/web/src/app/(app)/claw/new/ClawNewClient.tsx
+++ b/apps/web/src/app/(app)/claw/new/ClawNewClient.tsx
@@ -127,7 +127,7 @@ function ClawNewLiveClient() {
 
   const billing = billingQuery.data;
 
-  if (billing && !billing.hasExistingPersonalSubscription) {
+  if (billing && !billing.hasCurrentPersonalSubscription) {
     return (
       <div
         className="container m-auto flex w-full max-w-[1140px] items-center justify-center p-4 md:p-6"
--- a/apps/web/src/app/(app)/claw/new/ClawNewClient.tsx
+++ b/apps/web/src/app/(app)/claw/new/ClawNewClient.tsx
@@ -13,7 +13,6 @@ import {
 } from '../components';
 import type { ClawOnboardingRenderStep } from '../components/ClawOnboardingFlow.state';
 import { ClawOnboardingFakeWalkthrough } from '../components/ClawOnboardingFakeWalkthrough';
-import { WelcomePage } from '../components/billing/WelcomePage';
 import { getClawNewStatusQueryForBoundary } from './ClawNewClient.state';
 
 const ClawOnboardingWithBoundary = withStatusQueryBoundary(ClawOnboardingFlow);
@@ -134,21 +133,6 @@ function ClawNewLiveClient() {
         style={{ minHeight: '50vh' }}
       >
         <KiloClawSignupUnavailable />
-      </div>
-    );
-  }
-
-  const isNewUser =
-    billing &&
-    !billing.hasAccess &&
-    billing.instance === null &&
-    !billing.earlybird &&
-    !billing.trial?.expired;
-
-  if (isNewUser && !billing.trialEligible) {
-    return (
-      <div className="container m-auto flex w-full max-w-[1140px] flex-col gap-6 p-4 md:p-6">
-        <WelcomePage />
       </div>
     );
   }
--- a/apps/web/src/app/(app)/claw/new/ClawNewClient.tsx
+++ b/apps/web/src/app/(app)/claw/new/ClawNewClient.tsx
@@ -5,6 +5,7 @@ import { Loader2 } from 'lucide-react';
 import { useQuery } from '@tanstack/react-query';
 import { useKiloClawStatus } from '@/hooks/useKiloClaw';
 import { useTRPC } from '@/lib/trpc/utils';
+import { KiloClawSignupUnavailable } from '@/components/subscriptions/kiloclaw/KiloClawSignupUnavailable';
 import {
   ClawOnboardingFlow,
   type ClawOnboardingMode,
@@ -125,6 +126,18 @@ function ClawNewLiveClient() {
   }
 
   const billing = billingQuery.data;
+
+  if (billing && !billing.hasExistingPersonalSubscription) {
+    return (
+      <div
+        className="container m-auto flex w-full max-w-[1140px] items-center justify-center p-4 md:p-6"
+        style={{ minHeight: '50vh' }}
+      >
+        <KiloClawSignupUnavailable />
+      </div>
+    );
+  }
+
   const isNewUser =
     billing &&
     !billing.hasAccess &&
--- a/apps/web/src/components/subscriptions/kiloclaw/KiloClawGroup.tsx
+++ b/apps/web/src/components/subscriptions/kiloclaw/KiloClawGroup.tsx
@@ -21,6 +21,7 @@ import { useInvalidateKiloClawBilling } from './useKiloClawBillingQueries';
 import KiloCrabIcon from '@/components/KiloCrabIcon';
 import { SubscriptionCard } from '@/components/subscriptions/SubscriptionCard';
 import { SubscriptionGroup } from '@/components/subscriptions/SubscriptionGroup';
+import { KiloClawSignupUnavailable } from './KiloClawSignupUnavailable';
 import { KiloClawSubscribeCard } from './KiloClawSubscribeCard';
 import {
   formatDateLabel,
@@ -195,6 +196,8 @@ export function KiloClawGroup({
             );
           })}
         </div>
+      ) : subscriptions.length === 0 ? (
+        <KiloClawSignupUnavailable />
       ) : nonTerminalSubscriptions.length === 0 ? (
         <KiloClawSubscribeCard
           standardCostMicrodollars={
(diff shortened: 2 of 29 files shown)


apps/mobile/src/components/home/home-screen.tsx
```diff
@@ -100,7 +102,7 @@ export function HomeScreen() {
     enabled: orgLoaded,
   });
 
-  const isLoading = instancesPending || sessionsLoading || !orgLoaded;
+  const isLoading = instancesPending || billingQuery.isPending || sessionsLoading || !orgLoaded;
```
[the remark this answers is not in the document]
jeanduplessis: Addressed by the rebase onto main. Home no longer queries KiloClaw billing or blocks its global skeleton on that endpoint; main's dedicated persisted KiloClaw tab-ownership flow is preserved.

apps/web/src/components/subscriptions/kiloclaw/KiloClawGroup.tsx
```diff
@@ -195,6 +196,8 @@ export function KiloClawGroup({
             );
           })}
         </div>
+      ) : subscriptions.length === 0 ? (
+        <KiloClawSignupUnavailable />
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 2583dc9fe. The personal billing summary now exposes the same server-derived canonical-history flag used by the other surfaces, and KiloClawGroup waits for and gates on that flag instead of inferring eligibility from the instance-backed subscription list. Added coverage for detached canonical history.

apps/web/src/routers/organizations/organization-kiloclaw-router.ts
```diff
@@ -172,6 +172,35 @@ async function requireOrgInstance(userId: string, orgId: string) {
...
+        eq(kiloclaw_instances.user_id, userId),
+        eq(kiloclaw_instances.organization_id, orgId)
+      )
+    );
+
+  return rows.some(
+    row =>
+      row.subscription.status === 'active' &&
+      row.subscription.suspended_at === null &&
+      row.subscription.payment_source !== 'stripe'
```
pandemicsyn: bot: hasCurrentOrganizationKiloClawSubscription excludes Stripe-funded organization subscriptions (`payment_source !== 'stripe'`) from counting as a current subscription for provisioning eligibility and nav-state visibility. If organizations can have Stripe-funded KiloClaw subscriptions (payment_source is either 'stripe' or 'credits' per the schema and org bootstrap previously supported org billing generically), an org whose only active row is Stripe-funded would be treated as ineligible: provisioning throws FORBIDDEN and the sidebar/new-instance surfaces hide KiloClaw, even though the org has a live paid subscription. This appears inconsistent with the PR's stated goal that 'existing live instances and current subscriptions must continue to support normal access.' Worth confirming whether org KiloClaw subscriptions can ever be Stripe-funded; if so this filter incorrectly locks out otherwise-eligible orgs.

Suggested fix: Either drop the `payment_source !== 'stripe'` condition (matching the personal-side eligibility logic, which does not exclude Stripe funding) or add an explicit comment/test proving organization KiloClaw subscriptions are never Stripe-funded so the exclusion is provably a no-op.

apps/web/src/app/(app)/claw/new/ClawNewClient.tsx
```diff
@@ -125,6 +126,18 @@ function ClawNewLiveClient() {
   }
 
   const billing = billingQuery.data;
```
pandemicsyn, line 127: bot: The new gate `if (billing && !billing.hasExistingPersonalSubscription)` renders KiloClawSignupUnavailable, but the pre-existing `isNewUser`/`trialEligible` welcome-page branch immediately below still references `billing.trialEligible`, a flag the PR's own commit message says navigation should stop relying on ('Web and mobile navigation use server-backed context eligibility rather than the legacy trial-eligibility flag'). Leaving this dead/legacy branch reachable when `hasExistingPersonalSubscription` is true but `trialEligible` is stale could produce inconsistent onboarding UI (e.g. still showing the old WelcomePage flow for users who technically have existing personal subscription history but `trialEligible` was never updated for the new semantics).

Suggested fix: Confirm whether `trialEligible`/`WelcomePage` gating downstream of the new unavailable-check still makes sense post-PR, or remove/replace it with the same `hasExistingPersonalSubscription`-driven logic for consistency.

apps/web/src/routers/kiloclaw-router.ts
```diff
@@ -2013,13 +1967,28 @@ async function getPersonalBillingStatus(user: {
     };
   }
```
pandemicsyn, line 1968: bot: getPersonalBillingStatus now runs an extra `anyPersonalSubscription` query (leftJoin + or/isNull) alongside the pre-existing `anySubscription` query, both scanning kiloclaw_subscriptions for the same user on every billing-status fetch. This adds a second round-trip/join on a hot path (used by web sidebar, mobile Home, BillingWrapper, etc.) where the two queries could plausibly be combined into one.

Suggested fix: Consider merging `anySubscription` and `anyPersonalSubscription` into a single query (e.g. select both flags with conditional aggregation) to avoid the added latency on a frequently polled endpoint.

pandemicsyn (changes requested):

apps/web/src/components/subscriptions/kiloclaw/KiloClawGroup.tsx
```diff
@@ -195,7 +203,9 @@ export function KiloClawGroup({
             );
           })}
         </div>
-      ) : nonTerminalSubscriptions.length === 0 ? (
+      ) : subscriptions.length === 0 && !hasExistingPersonalSubscription ? (
+        <KiloClawSignupUnavailable />
+      ) : subscriptions.length === 0 || nonTerminalSubscriptions.length === 0 ? (
```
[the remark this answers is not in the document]
jeanduplessis: Fixed in 13323d7c1. Billing status now distinguishes canonical history from a current, active, unsuspended personal subscription. Subscription Center, navigation, direct onboarding, billing, and mobile onboarding use current eligibility (while preserving management for a live instance), so canceled/historical-only users receive the unavailable state instead of dead-end enrollment actions. The spec and tests now encode that boundary.

apps/web/src/routers/kiloclaw-router.ts
jeanduplessis: Fixed in 813e247a9. Consolidated all-subscription, personal-history, and current-personal presence into one conditional-aggregation query, leaving only the separate instance-history query.

apps/web/src/routers/organizations/organization-kiloclaw-router.ts
jeanduplessis: Fixed in 813e247a9. Current organization eligibility no longer excludes Stripe-funded active rows, and getNavState now has explicit Stripe-funded organization coverage.

apps/web/src/app/(app)/claw/new/ClawNewClient.tsx
jeanduplessis: Fixed in 813e247a9. Removed the legacy trialEligible/WelcomePage branch from ClawNewClient; direct onboarding now uses the current-subscription gate consistently.

pandemicsyn (approved):

PR #5087, opened 2026-08-06
fix(cloud-agent-next): reconcile completed kilo turns on wrapper death

eshurakov:
## Problem

When the wrapper OOM-dies after emitting `wrapper_finalizing` but before sending `complete`, the DO liveness watchdog (`handleUnhealthyWrapper` via `wrapper_ping_timeout`) and the disconnect-grace expiry terminalize all accepted messages as `wrapper_failure` — even though the turn actually completed. Incident: 8 sessions posted their work but were reported as failed.

## Why the DO can settle this itself

Ingest is a single FIFO channel, and the wrapper only emits `wrapper_finalizing` after `message.updated(time.completed)` and `session.idle` have already been sent. Kilo events with an entity id are upserted last-wins into the DO's `events` table, so receiving `wrapper_finalizing` implies the DO already holds the terminal assistant state. (Verified via deadline arithmetic on the incident: 7 of 8 sessions provably reached `wrapper_finalizing`.)

## Change

Both wrapper-death paths (ping/no-output deadline, disconnect-grace expiry) now reconcile each accepted message against the DO's stored kilocode events before failing:

- assistant `info.error` → `failed/assistant_error` with the existing `classifyAssistantFailure` classification (extracted into a shared helper also used by `settleSealedBatch`)
- assistant `info.time.completed` present → `completed` with the assistant message id, `completionSource: 'idle_reconciliation'`
- no positive terminal evidence → previous `wrapper_failure` / `wrapper_disconnected` behavior, unchanged

Unlike the sealed-batch path, bare assistant-message presence is not sufficient here — no wrapper `complete` vouches for turn finality, so the same completion marker the wrapper itself uses (`time.completed` or terminal error) is required. Emits a warn log with reconciled/fallback counts for observability.

Residual gap: if the final `message.updated` never reached ingest at all, the store has no evidence and the outcome is `wrapper_failure` as today; closing that needs a live Kilo terminal query or an HTTP terminal assertion (separate work).

## Verification

- 4 new unit tests: ping-timeout reconcile to completed (incl. callback fired), no-completion-marker fallback, assistant-error reconcile (not masked as wrapper failure), disconnect-grace reconcile
- Full `cloud-agent-next` unit suite: 2451 passed; death-path integration tests in Workers runtime: 30 passed
- `tsgo --noEmit`, `oxlint`, `oxfmt` clean
- `services/cloud-agent-next/AGENTS.md` invariant bullet updated


eshurakov:
Ran manual generic cloud agent verification

jeanduplessis (approved):

PR #5098, opened 2026-08-06
fix(ai-gateway): stop the usage-record redelivery storm

RSO:
## Summary

Follow-up to #5034. That change fixed the hot-row convoy — the Frankfurt-local write is **5 ms at p50** and the primary is now idle (0 blocked sessions, oldest transaction 0.76 s, 17 `log_lock_waits` entries in 30 min with `log_lock_waits=on` and `deadlock_timeout=1s`, Supavisor `client_waiting=0` against `pool_size=320`). It replaced it with a self-inflicted retry storm.

`recordUsageInPrimaryRegion` used a **10 s** per-attempt timeout against an endpoint whose **p95 is ~50 s**. Every request slower than the timeout was abandoned and re-sent with the same `core.id`:

| 5-min bucket (UTC) | requests | p50 | p95 | duplicate-key errors |
|---|---|---|---|---|
| 12:20 | 10,331 | 5 ms | 43.7 s | 305 |
| 12:35 | 46,439 | 155 ms | 54.3 s | 4,320 |
| 12:45 | 46,362 | 155 ms | 50.2 s | 4,600 |

The violated constraint is `PK_a71b90d910e7882358c3defe8e2`, confirmed via `pg_constraint` to be `PRIMARY KEY (id)` on `microdollar_usage` — so ~20% of deliveries were redeliveries of a record already being written, and each retry added load that made the endpoint slower still. It stayed invisible because the client only logs after exhausting every attempt, which happened just **167** times.

### Breaking the loop

- **Attempt budget 10 s → 90 s**, above the observed tail and below the endpoint's `maxDuration` of 150 s.
- **A timeout is never retried.** The endpoint may still be mid-write, so re-sending reproduces exactly this collision. It falls back to the local write instead — slower, but the endpoint's conflict handling already recognises it.
- **Retryable statuses narrow from `>= 500 || 429` to 429/502/503** — refusals that prove the function did not run. A 500 means our handler threw, possibly after commit; 504 is equally ambiguous. Both used to be retried.
- **`insertUsageRecord` stops retrying a collision on this record's own id.** `id` is fixed for the delivery, so all three attempts collided, each rebuilding and re-logging the same ~30 KB statement before the outer handler recovered the committed identity. Deliberately narrow: `microdollar_usage` and `microdollar_usage_metadata` are keyed by the per-delivery id, while the twelve deduplicated lookup tables have a `WHERE NOT EXISTS` race that retrying genuinely does resolve. Checked against `pg_constraint` — lookup tables use `<table>_pkey`.

### Stops leaking prompts into logs

The failure log serialized the interpolated statement. A sample contained real end-user prompt text (`"Aun me siguen apareciendo los telefonos de un cliente…"`), `system_prompt_prefix`, client IP `134.82.68.167`, city and JA4 digest — ~30 KB per event, ~4,400 events per 5 minutes, into Axiom and Sentry. Both the retry warning and the final `captureException` now report a redacted description (`name`, `code`, `constraint`, `table`, `routine`). `detail` is excluded too, since PostgreSQL puts the conflicting key values there. The stack is preserved on the replacement error.

This predates #5034 — the same line fires on `kilocode-global-app` gateway paths — but that change amplified it about 45×.

### Instrumentation for the remaining tail

I could not determine *why* p95 is ~50 s. It is app-side (database and pooler both idle) and it is **not** memory: every route on `kilocode-app` sits at p95 920–1,055 MB and this one is typical at 1,034 MB, while its duration is the sole outlier at 57.7 s versus ≤ 2 s everywhere else. Two candidates remain and need different fixes, so they are separated by measurement:

- **In-process pg pool gauges** (`totalCount`/`idleCount`/`waitingCount`/`max`). `/api/cron/db-pool-metrics` reports Supavisor's counters, not this pool's, so its saturation is invisible today. `waitingCount` settles it.
- **Event-loop delay** via `monitorEventLoopDelay`, which would inflate every phase including ones doing no I/O.
- **Phase timings** — `validate`, `dedupe_check`, `write`.

One structured `{"type":"usage_record_timing"}` line, gated on ≥ 1 s or a 1% sample. Combined with removing the 30 KB dump, **net log volume falls.**

## Verification

Automated: `tsgo --noEmit -p apps/web/tsconfig.json` clean; `./scripts/lint-all.sh` 0 warnings 0 errors; 24 new diagnostics tests (including explicit assertions that no returned field contains the statement, params, prompt text or IP, and that lookup-table conflicts stay retryable); 16 client tests covering the never-retry-a-timeout regression and the ambiguous 500/504 cases; `src/lib/ai-gateway` + `src/lib/drizzle.test.ts` 699 passing across 59 suites. Targeted verification, not `pnpm validate`.

Manual: **none.** The behaviour only exists on a multi-region deployment — locally `VERCEL_REGION` is unset, so `isUSRegion()` is false and the remote path is never taken. I could not exercise the timeout, the collision path, or the pool gauges under load by hand.

- [ ] After deploy, confirm duplicate-key errors on `/api/internal/usage/record` fall to near zero.
- [ ] Read the new timing line and decide the next step (query below).
- [ ] Re-tune `ATTEMPT_TIMEOUT_MS` once the real p95 is known — see Reviewer Notes.

## Visual Changes

N/A

## Reviewer Notes

**90 s is calibrated against a p95 I expect this PR to remove.** If most of the tail was self-inflicted, the real p95 will drop a lot and 90 s becomes far more generous than needed. Treat it as provisional; the test asserts the invariant (above the tail, below `maxDuration`) rather than the number.

**This does not explain the tail, it measures it.** Please read the instrumentation as the deliverable it is, not as a fix. The query that decides the next step:

```
['vercel'] | where message startswith '{"type":"usage_record_timing"'
| summarize p95_total=percentile(total_ms,95), p95_write=percentile(['phases.write_ms'],95),
            max_waiting=max(pool_waiting_peak), p99_lag=percentile(['event_loop_lag.p99_ms'],99)
  by bin(_time, 1m)
```

`max_waiting > 0` → raise the per-instance pool `max`; there is headroom (141 of 480 Postgres connections, 100 of 3,000 pooler client slots). High `p99_lag` with `max_waiting` at 0 → the cost is CPU per request. Neither → I am wrong about both, and the phase split will say where the time went.

Other things worth a look:

- **Not retrying a timeout trades one failure mode for another.** A genuine transient timeout now goes straight to a local SFO write, re-adding the transatlantic lock hold #5034 removed. I judged that better than manufacturing collisions, but it is a deliberate trade.
- **`monitorEventLoopDelay` is never reset**, so values are process-lifetime. Per-request windowing would need a timer on the loop being measured. Sustained starvation is what we are looking for, so this is sufficient — but it will not show a single isolated stall.
- **`poolMax()` reads `pool.options.max`**, which is not in `pg`'s public types. Casted, and null-safe.
- I left the `pg_stat_statements` snapshot out of `/api/cron/db-pool-metrics` deliberately, to keep this diff reviewable while production is degraded. It would remove the cumulative-stats blind spot that made this hard to diagnose, and is worth a follow-up.

### DIFF
--- a/apps/web/src/app/api/internal/usage/record/route.ts
+++ b/apps/web/src/app/api/internal/usage/record/route.ts
@@ -11,6 +11,12 @@ import {
   UsageRecordRequestSchema,
   type UsageRecordResponse,
 } from '@/lib/ai-gateway/usage-record-contract';
+import {
+  createPhaseTimer,
+  emitUsageRecordTiming,
+  readPoolGauges,
+  shouldEmitUsageRecordTiming,
+} from '@/lib/ai-gateway/usage-record-diagnostics';
 
 /**
  * Frankfurt-local sink for AI-gateway usage writes.
@@ -37,8 +43,35 @@ export async function POST(request: NextRequest): Promise<NextResponse> {
     return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
   }
 
+  // Phase timings exist to locate the ~50s tail this endpoint exhibits while the
+  // database and Supavisor are both idle. Splitting validate / dedupe / write and
+  // pairing them with the in-process pool gauges distinguishes waiting for a pool
+  // connection from waiting for PostgreSQL from a stalled event loop.
+  const timer = createPhaseTimer();
+  const poolBefore = readPoolGauges();
+  let poolWaitingPeak = poolBefore.waiting;
+  const samplePool = () => {
+    const gauges = readPoolGauges();
+    poolWaitingPeak = Math.max(poolWaitingPeak, gauges.waiting);
+    return gauges;
+  };
+  const reportTiming = (usageId: string, outcome: UsageRecordResponse['status']) => {
+    const totalMs = timer.totalMs();
+    if (!shouldEmitUsageRecordTiming(totalMs)) return;
+    emitUsageRecordTiming({
+      usageId,
+      outcome,
+      totalMs,
+      phases: timer.phases(),
+      poolBefore,
+      poolAfter: samplePool(),
+      poolWaitingPeak,
+    });
+  };
+
   const rawBody: unknown = await request.json().catch(() => null);
   const parsed = UsageRecordRequestSchema.safeParse(rawBody);
+  timer.mark('validate');
   if (!parsed.success) {
     // Deliberately a 400: the client must not retry a payload we cannot accept,
     // and a schema failure here means a lost billing row that needs a human.
@@ -56,16 +89,17 @@ export async function POST(request: NextRequest): Promise<NextResponse> {
   // such a redelivery is detectable, and skipping the write avoids burning the
   // retry budget in `insertUsageRecord` and re-running its side effects.
   //
-  // This check cannot cover the redelivery that arrives while the first attempt
-  // is still open — the common case, since the client retries a 10s attempt
-  // timeout and this write can legitimately take longer. That window is closed by
-  // `insertUsageRecord`, which recovers the identity of an already-persisted row
-  // instead of reporting a primary-key collision as a failed write.
+  // This check cannot cover the redelivery that arrives while the first attempt is
+  // still open. That window is closed by `insertUsageRecord`, which recovers the
+  // identity of an already-persisted row instead of reporting a primary-key
+  // collision as a failed write.
   const existing = await db
     .select({ id: microdollar_usage.id })
     .from(microdollar_usage)
     .where(eq(microdollar_usage.id, core.id))
     .limit(1);
+  timer.mark('dedupe_check');
+  samplePool();
 
   if (existing.length > 0) {
     // Post-commit side effects ran on the first delivery, so do not re-run them.
@@ -75,6 +109,7 @@ export async function POST(request: NextRequest): Promise<NextResponse> {
       status: 'duplicate',
       result: { usageId: core.id, createdAt: core.created_at, newMicrodollarsUsed: null },
     };
+    reportTiming(core.id, 'duplicate');
     return NextResponse.json(duplicate);
   }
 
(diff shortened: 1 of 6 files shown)


eshurakov (approved):

RSO:
CI is green after a re-run. The `test` failure was flake, not a regression — I checked before re-running rather than assuming:

- **The failing suite cannot reach my code.** `src/lib/kiloclaw/provision-bootstrap-shared-org.test.ts` does not transitively import `usage-record-diagnostics` (walked 90 modules from its entry point). My diff is six files, all under `ai-gateway` plus the internal route — nothing in `kiloclaw`, `drizzle.ts`, or test setup.
- **The failure was a `beforeEach` hook timeout** on `cleanupDbForTest()` at 5000 ms, not an assertion — DB-cleanup contention under a fully parallel 736-suite run. 1 failure out of 9,555 tests.
- **The "active timers" hint was worth chasing** since I added `monitorEventLoopDelay` at module load, but the handle is unref-ed: a process that enables the histogram and does nothing else still exits on its own. So it cannot hold a worker open.
- The suite passes 3/3 locally, and `main` itself had a failed run about 20 minutes before this one.

I deliberately did **not** add a `NODE_ENV !== 'test'` guard around `enable()`. There is repo precedent for skipping process-level attachments in tests (`attachDatabasePool` in `drizzle.ts`), but that exists because pool attachment genuinely interferes with Jest teardown, and I have shown the histogram does not. Adding it would be a speculative change for a problem I have proven is not mine.

Current state: all checks SUCCESS, `reviewDecision: APPROVED`, `kilo-code-bot` at **No Issues Found / Recommendation: Merge** with all three earlier findings verified resolved at HEAD.

One thing for whoever merges: the branch is now 8 commits behind `main`. Given this is an incident mitigation I would rather not rebase without a reason — but if a required check needs it up to date, say so and I will.

PR #5107, opened 2026-08-06
diag(ai-gateway): measure whether pool slots are leaked

RSO:
## Summary

Item 2 of the follow-ups from #5098. **This is a measurement, not a fix** — it exists to decide between two explanations that call for opposite remedies.

#5098's instrumentation established that pool acquisition dominates `/api/internal/usage/record`: `idle == 0` on 96–98% of sampled requests with up to **245 queued against `max: 10`**, while PostgreSQL reports **zero active queries** and Supavisor reports `client_waiting: 0` with `pool_size` 320 against `server_active` 67. Connections are checked out without doing work. Either:

1. **Fluid compute packs high concurrency onto few instances** and ten connections is simply too few. Fix: more connections.
2. **Slots are leaked and never return.** drizzle-orm 0.45.2 issues `BEGIN` *outside* its `try`/`finally`, so a failed `BEGIN` never reaches the `release()` in the `finally` and burns that slot for the life of the process. Ten of those kill an instance permanently. Fix: release on a failed `BEGIN`.

Getting this wrong is expensive in both directions. If it's a leak, raising `max` postpones and conceals it — and `max` at `drizzle.ts:93` is a single module-level constant shared by every route in *both* Vercel projects, whose inline comment records that raising it previously exhausted Supabase's connection limit across ~2,200 instances.

The arithmetic is what makes this worth measuring rather than assuming: at ~47 req/s with Frankfurt-local statements of 2–10 ms, you need roughly **5** connections, not 245 queued.

### The discriminator

**The low-water mark of checked-out connections.** A healthy pool returns to zero checked out whenever the instance goes quiet, and `idleTimeoutMillis` of 5 s then closes the idle clients. A leaked client is never returned to `_idle`, so it counts as checked out forever and the low-water mark cannot fall below the number of leaks.

A low-water mark rather than an instantaneous reading, because this pool is shared with every other route on the instance — `/api/profile`, `/api/defaults` and the rest. One quiet moment resets it, so unrelated concurrent traffic cannot inflate it. An instantaneous `checked_out > 0` proves nothing; a *floor* that climbs does.

### What it adds

All read-only counters updated on `pg-pool` events. No queries, no timers.

- `acquires` / `releases` / `connects` / `removes`, plus `outstanding` as a cross-check that should track `pool.totalCount - pool.idleCount`. A persistent divergence means the accounting is wrong and nothing else should be trusted.
- `min_checked_out` and `max_checked_out`.
- **`begin_failures`** — counted where the retry loop already inspects the error. Under the drizzle bug each one is exactly one leaked slot, which makes it directly comparable to `min_checked_out`. Matching on the message is safe *only* for `BEGIN`: it takes no parameters, so unlike every other statement in this path the message carries no prompt text or client IP.
- `ms_since_last_request` and `checked_out_at_entry`, read before the request acquires anything.

Emission now also triggers on the first request after 5 s of instance quiet. Those requests are fast, so the duration threshold would never surface them — yet they are the only moment a leaked slot is distinguishable from a busy one.

Counters do not attach under `NODE_ENV=test`, matching how `drizzle.ts` already special-cases the test pool.

## Verification

Automated: `tsgo --noEmit -p apps/web/tsconfig.json` clean; `./scripts/lint-all.sh` 0 warnings 0 errors; 17 new probe tests and 6 new emission/clock tests; 923 passing across 70 suites in `src/lib/ai-gateway` + `src/lib/drizzle.test.ts` + `src/lib/db-pool-leak-probe.test.ts` + `src/app/api/internal`; `pnpm format` and `git diff --check` clean. Targeted verification, not `pnpm validate`.

Manual: **none, and it cannot be meaningful here.** The behaviour under investigation only appears under production concurrency; locally the pool never exceeds a couple of connections, so a local run would show `min_checked_out: 0` whether or not the bug exists. The unit tests cover the classifier and the arithmetic; only production can answer the question.

- [ ] After deploy, read the query in Reviewer Notes and reach a verdict.

## Visual Changes

N/A

## Reviewer Notes

**How to read the result.** The verdict is a comparison, not a threshold:

```
['vercel'] | where message startswith '{"type":"usage_record_timing"'
| extend quiet=toint(extract('"ms_since_last_request":([0-9]+)',1,message)),
         entry=toint(extract('"checked_out_at_entry":([0-9]+)',1,message)),
         minco=toint(extract('"min_checked_out":([0-9]+)',1,message)),
         beginf=toint(extract('"begin_failures":([0-9]+)',1,message)),
         uptime=toint(extract('"instance_uptime_ms":([0-9]+)',1,message))
| where quiet >= 5000
| summarize samples=count(), p50_entry=percentile(entry,50), max_entry=max(entry),
            p50_min=percentile(minco,50), max_min=max(minco),
            max_begin=max(beginf), p95_uptime=percentile(uptime,95)
  by bin(_time, 5m)
```

- `min_checked_out` at **0** and `checked_out_at_entry` near 0 on quiet instances → **no leak**. The pool is genuinely busy, and sizing (ideally a dedicated pool, to bound the blast radius) is the right next step.
- `min_checked_out` **above 0 and climbing with uptime**, tracking `begin_failures` → **leak confirmed**. The fix is wrapping `db.transaction` so a failed `BEGIN` releases its client, not more connections.
- `min_checked_out` above 0 but `begin_failures` at 0 → a leak from some *other* path, and I would want to know that before touching anything.

**Things worth challenging:**

- `min_checked_out` is process-lifetime and never reset, so on a long-lived instance a single early quiet moment pins it at 0 forever and could mask a leak that started later. Comparing it against `checked_out_at_entry` on quiet requests is the mitigation, but a windowed low-water mark would be strictly better if this proves inconclusive.
- The probe attaches four listeners to the shared primary pool at module load. They are counter increments, but it is a process-level side effect in a hot shared module.
- `begin_failures` is only incremented on the usage-write path. A `BEGIN` failure anywhere else in the process leaks a slot without being counted, which is exactly the third case above.
- This touches `route.ts` in the same import block as #5105, so expect a trivial conflict depending on merge order. Both changes are independent; either can land first.

### DIFF
--- a/apps/web/src/lib/db-pool-leak-probe.ts
+++ b/apps/web/src/lib/db-pool-leak-probe.ts
@@ -1,0 +1,174 @@
+import 'server-only';
+
+import { pool } from '@/lib/drizzle';
+
+/**
+ * Read-only probe that decides whether pooled connections are being leaked.
+ *
+ * `/api/internal/usage/record` shows the in-process pool pinned at `idle: 0` with
+ * up to 245 requests queued against `max: 10`, while PostgreSQL reports zero
+ * active queries and Supavisor reports zero waiting clients. Connections are
+ * therefore checked out without doing work, and there are two very different
+ * explanations:
+ *
+ * 1. Vercel Fluid compute packs high concurrency onto few instances, so ten
+ *    connections are genuinely in use and the cap is simply too low. Fix: more
+ *    connections.
+ * 2. Slots are leaked and never come back. drizzle-orm 0.45.2 issues `BEGIN`
+ *    *outside* its `try`/`finally`, so a `BEGIN` that fails never reaches the
+ *    `release()` in the `finally` and burns that slot for the life of the
+ *    process. Ten of those kill an instance permanently. Fix: release on a failed
+ *    `BEGIN`. Raising `max` would only postpone and hide it.
+ *
+ * Guessing wrong is expensive in opposite directions, so measure instead. Note
+ * `max` in `drizzle.ts` is a single module-level constant shared by every route in
+ * both Vercel projects, and its comment records that raising it previously
+ * exhausted Supabase's connection limit across ~2,200 instances.
+ *
+ * The discriminator is the **low-water mark** of checked-out connections. A
+ * healthy pool returns to zero checked out whenever the instance goes quiet, and
+ * `idleTimeoutMillis` of 5s then closes the idle clients. A leaked client is never
+ * returned to `_idle`, so it is counted as checked out forever and the low-water
+ * mark can never fall back below the number of leaks.
+ *
+ * A low-water mark is deliberately chosen over an instantaneous reading because
+ * this pool is shared with every other route on the instance: a single quiet
+ * moment resets it, so concurrent unrelated traffic cannot inflate it. If it
+ * climbs over an instance's lifetime and tracks `beginFailures`, explanation 2 is
+ * confirmed with a number.
+ *
+ * Everything here is counters updated on pool events. No queries, no timers.
+ */
+
+type PoolLeakCounters = {
+  acquires: number;
+  releases: number;
+  connects: number;
+  removes: number;
+  /**
+   * Failures of the `BEGIN` statement itself. Under the drizzle bug each one is
+   * exactly one permanently leaked slot, which is what makes this directly
+   * comparable to `minCheckedOut`.
+   */
+  beginFailures: number;
+  maxCheckedOut: number;
+  /** Low-water mark of checked-out connections. The leak discriminator. */
+  minCheckedOut: number;
+  startedAtMs: number;
+};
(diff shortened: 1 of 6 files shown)


apps/web/src/lib/db-pool-leak-probe.ts
```diff
@@ -0,0 +1,174 @@
...
+    sampleCheckedOut();
+  });
+  pool.on('connect', () => {
+    counters.connects++;
+  });
+  pool.on('remove', () => {
+    counters.removes++;
+    sampleCheckedOut();
+  });
+}
```
jrf0110, line 100: Probably not, but do we need something to teardown these listeners? Perhaps if drizzle ever recreates the pool?

jrf0110 (approved):

PR #5120, opened 2026-08-07
fix(auth): audit admin logins safely

pandemicsyn:
## Summary
- emit a minimal structured event after a successful platform-admin login
- record only the opaque Kilo user ID, provider, and super-admin flag for Axiom monitoring
- disable NextAuth debug output to prevent OAuth credentials from reaching application logs

## Validation
- pnpm --filter web lint
- pnpm --filter web typecheck
- git diff --check


pandemicsyn:
Closing this PR: the proposed general-log audit event and blanket NextAuth debug disable were unsafe. The branch has been reverted; a future audit design should use a dedicated, access-controlled sink and preserve test/local debug behavior.

PR #5156, opened 2026-08-09
feat(exports): add queued user data exports

pandemicsyn:
## Summary

Adds the Phase 1 foundation for Kilo-admin user data exports:

- Adds a `/data-exports` account page for requesting exports, monitoring status, and downloading completed artifacts.
- Adds a dedicated Cloudflare Worker using Queues, Hyperdrive, and private R2 multipart uploads.
- Reads approved source data through a separate read-replica binding and enforces user ownership in every source query.
- Exports the matched Kilo account's approved identity, profile, usage-total, preference, safety-identifier, attribution, and signup fields with native JSON value types.
- Uses a fixed data cutoff of `2026-08-03T00:00:00.000Z`; records created after that timestamp are excluded.
- Streams pages through native gzip compression and writes resumable, uniform 5 MiB R2 multipart checkpoints.
- Adds PostgreSQL job, part, and transactional outbox state with leases, generation fencing, retries, expiry, and email-delivery state.
- Reuses the existing shared production Hyperdrive for primary state; only the read-replica Hyperdrive is a new rollout resource.
- Preserves account-deletion cleanup through durable R2 deletion tombstones, including in-flight multipart aborts.
- Sends export-ready notifications through the existing Mailgun integration.
- Uses the existing `x-internal-api-key` / `INTERNAL_API_SECRET` contract for web-to-Worker and Worker-to-web communication.
- Implements five-minute, attachment-only R2 download signing with independent ownership/readiness validation.
- Restricts navigation, the server page, and all export procedures to existing Kilo admins for the initial release.
- Registers `pnpm dev:start data-export` with worktree-aware ports, generated local URLs, both Hyperdrive overrides, and the shared internal API key.
- Matches the usage-meter Worker compatibility date and compact `wrangler types --include-runtime=false` generation setup.
- Leaves only Cloudflare resource, database-role, credential-binding, and production-origin setup for rollout.

```mermaid
flowchart TB
    Admin["Kilo admin browser"]

    subgraph Web["Kilo Web trust boundary"]
        Page["/data-exports<br/>server page: adminOnly"]
        TRPC["userExports router<br/>adminProcedure + ctx.user.id"]
        Notify["Ready-email endpoint<br/>timing-safe internal key check"]
    end

    subgraph Worker["Export Worker trust boundary"]
        Internal["user-data-export.kilosessions.ai<br/>timing-safe x-internal-api-key"]
        Consumer["Queue consumer<br/>lease + generation fencing"]
        Signer["Download signer<br/>re-checks exportId + kiloUserId<br/>ready + unexpired"]
    end

    subgraph Data["Private data plane"]
        Primary[("PostgreSQL primary<br/>existing shared Hyperdrive<br/>job state, outbox, object key")]
        Replica[("Read-only replica role<br/>matched Kilo user row<br/>and approved source tables only")]
        Queue["Cloudflare Queue + DLQ<br/>IDs/generation only; no prompts"]
        Bucket[("Private R2 bucket<br/>no public domain")]
        Tombstone[("Deletion tombstones<br/>survive user/export cascade")]
    end

    Mailgun["Mailgun transactional email"]
    Signed["5-minute presigned GET<br/>attachment; private, no-store"]

    Admin -->|"authenticated session"| Page --> TRPC
    TRPC -->|"rows scoped by ctx.user.id"| Primary
    TRPC -->|"x-internal-api-key"| Internal
    Internal --> Queue
    Queue --> Consumer
    Consumer -->|"SELECT only; user predicate;<br/>cutoff <= 2026-08-03 UTC"| Replica
    Consumer -->|"state/checkpoints"| Primary
    Consumer -->|"multipart .jsonl.gz"| Bucket
    Consumer -->|"x-internal-api-key; exportId only"| Notify --> Mailgun
    TRPC -->|"exportId + ctx.user.id"| Signer
    Signer -->|"ownership/readiness lookup"| Primary
    Signer -->|"sign exact stored key"| Signed --> Admin
    Primary -->|"account deletion copies key + upload ID"| Tombstone
    Consumer -->|"abort multipart + delete object"| Tombstone

    classDef secure fill:#103b2d,stroke:#55d68b,color:#fff
    classDef private fill:#17233d,stroke:#73a7ff,color:#fff
    class TRPC,Notify,Internal,Consumer,Signer secure
    class Primary,Replica,Queue,Bucket private
```

```mermaid
flowchart LR
    Request["Admin request"] --> Admit["Single DB transaction<br/>create queued job and generation 0 outbox"]
    Admit --> Dispatch["Immediate Queue send<br/>plus scheduled outbox recovery"]
    Dispatch --> Delivery["At-least-once delivery<br/>exportId and generation only"]
    Delivery --> Claim{"Lease claim matches<br/>current generation?"}

    Claim -->|"No: stale, duplicate, or terminal"| Ack["Acknowledge safe no-op"]
    Claim -->|"Yes"| Read["Read bounded keyset pages<br/>for one user and fixed cutoff"]
    Claim --> Attach["Create multipart and CAS-persist upload ID<br/>abort immediately if row deleted or lease lost"]
    Attach --> Read
    Read --> Upload["Stream gzip to R2<br/>uniform multipart parts"]
    Upload --> Checkpoint["One DB transaction<br/>persist all ETags and cursor<br/>advance generation and outbox"]
    Checkpoint --> More{"Sources complete?"}
    More -->|"No"| Dispatch
    More -->|"Yes"| Final["Complete multipart<br/>HEAD exact object"]
    Final --> Ready["Lease-fenced ready transition<br/>store key, size, expiry"]

    Claim -.->|"expired lease"| Recover["Reconciler clears lease<br/>re-arms current outbox"]
    Recover --> Dispatch
    Delivery -.->|"configured retry limit"| Failed["DLQ consumer marks failed<br/>reconciler aborts multipart"]
    Claim -.->|"five consecutive claims expire"| Failed

    classDef durable fill:#17233d,stroke:#73a7ff,color:#fff
    classDef guarded fill:#103b2d,stroke:#55d68b,color:#fff
    class Admit,Dispatch,Checkpoint,Ready,Recover durable
    class Claim,Attach,Upload,Final guarded
```

## Verification

- [x] Reset and migrated the local development database through the repository-supported workflow.
- [x] Signed in as a fake Kilo admin, requested an export, and verified queued status plus duplicate-request disablement.
- [x] Verified the page at desktop and 375px mobile widths.
- [x] Signed in as a non-admin and verified `/data-exports` returns 404 and the navigation item is absent.

## Visual Changes

| Before | After |
|---|---|
| No account data-export surface | Desktop queued export |
| No mobile export workflow | Mobile queued export |

## Reviewer Notes

- This PR intentionally contains development-phase code only. `PRIMARY_STATE_DB` reuses the existing shared production Hyperdrive; it does not provision or deploy Queues, R2, the read-replica Hyperdrive, standard Worker secrets, environment values, or production migrations.
- Focus review on cross-user ownership predicates, Queue retry/idempotency behavior, multipart gzip checkpoints, database state transitions, and account-deletion cleanup.
- Poisoned Queue work terminates through both generation-fenced DLQ handling and a five-consecutive-claim database bound for isolate deaths.
- The primary Queue consumer starts at concurrency five with a five-minute CPU allowance; Cloudflare's fixed 15-minute consumer wall-clock limit still applies. The DLQ consumer remains at concurrency one.
- Five-minute private R2 signing is implemented; rollout only needs bucket-scoped read credentials, account/bucket values, and Worker bindings.
- Sensitive Worker values use vanilla `wrangler secret put` secrets; this Worker intentionally does not use Secrets Store.
- The initial release is Kilo-admin-only instead of feature-flagged.
- Exports deliberately stop at `2026-08-03T00:00:00.000Z`. No WAL replay gate is needed because the workload is intentionally historical and runs against four large read replicas.
- The detailed deviations and rollout register is maintained outside the repository at `~/fd-plans/research/data-export.md`.


services/user-data-export/src/databases.ts
```diff
@@ -0,0 +1,326 @@
...
+    async attachMultipartUpload(input: {
+      exportId: string;
+      generation: number;
+      leaseToken: string;
+      multipartUploadId: string;
+    }): Promise<'attached' | 'deleted' | 'lost_lease'> {
+      const rows = await db.execute<{ attached: boolean; export_exists: boolean }>(sql`
+        WITH attached AS (
+          UPDATE user_data_exports
+          SET multipart_upload_id = ${input.multipartUploadId}, updated_at = now()
```
St0rmz1, line 64: **Critical, Functional:** This UPDATE always violates the `user_data_exports_multipart_checkpoint_shape` CHECK constraint, so every export fails on its first upload attach.

The constraint (migration `0209_curved_mordo.sql`) is `multipart_upload_id IS NULL OR next_part_number > 1`. Attach only runs on the fresh path in `processGenerateMessage` (when `job.multipart_upload_id` is null), and on this state machine that always means `next_part_number` is still at its default of 1: only `checkpoint()` advances it, and checkpoint always persists an upload ID at the same time. So this statement leaves the row with an upload ID and `next_part_number = 1`, Postgres rejects it with a 23514, and the export retries into the DLQ, where `markFailed` marks it failed. Users would see "The export could not be completed after multiple attempts" on 100 percent of requests. The abort in `attachNewMultipartUpload` does fire, so no R2 uploads leak, but the pipeline cannot progress.

The constraint appears to be written backwards relative to the intended invariant, which is "you cannot have checkpointed parts without an upload." Suggest changing it to `CHECK (next_part_number = 1 OR multipart_upload_id IS NOT NULL)`, or dropping it. Existing tests miss this because the worker tests mock the state layer, and the web tests only insert rows that already satisfy the constraint (upload ID together with `next_part_number: 2`); an integration test over the real claim, attach, checkpoint, complete sequence would have caught it (separate comment on that).

services/user-data-export/src/worker.ts
```diff
@@ -0,0 +1,410 @@
...
+      cursor,
+      nextSource,
+      nextGeneration: job.dispatch_generation + 1,
+      multipartUploadId: upload.uploadId,
+    });
+    if (!checkpointed) return;
+
+    // Finalization runs from the checkpointed continuation so a crash after the
+    // R2 upload cannot complete a multipart object whose part is not durable.
+    if (nextSource) return;
```
St0rmz1, line 260: **Moderate, Performance:** After a successful `checkpoint()`, nothing sends the next generation's queue message. The checkpoint creates the outbox row, but the only sender is `processScheduledExportWork`, which runs on the `*/5 * * * *` cron. So every continuation chunk, the finalization step, and (indirectly) the ready email each wait up to 5 minutes idle. The PR's flow diagram says checkpoint loops back to "Immediate Queue send plus scheduled outbox recovery," and the admit path does send immediately via the dispatch endpoint, but this path only implements the recovery half.

Each invocation is capped at 32 MiB uncompressed or 200 pages, so a large account needing N invocations accumulates up to 5 idle minutes per chunk; an export whose real work takes 10 minutes could take hours of wall clock. The crash safety comment above justifies not finalizing in the same invocation, but it does not require deferring the send: delivery is at least once and claims are generation fenced, so an inline send after a successful checkpoint is safe.

Suggest: when `checkpoint()` returns true, send the next generation message to `EXPORT_QUEUE` and mark that outbox row sent (mirroring the dispatch endpoint), keeping the cron as recovery only.

services/user-data-export/src/databases.ts
```diff
@@ -0,0 +1,326 @@
+import { getWorkerDb, pg } from '@kilocode/db/client';
```
St0rmz1, line 1: **Moderate, Testing:** This file contains the most intricate SQL in the PR (claim with lease and generation fencing, the attach CTE with the tombstone fallback, the multi CTE checkpoint, reconcile), and none of it runs against real Postgres in any test. `worker.test.ts` mocks the state object entirely, and the `vitest-pool-workers` test only covers HTTP auth and body limits. Meanwhile the web side does have a real database harness (the router and `softDeleteUser` tests use it).

These queries interact with CHECK constraints, partial unique indexes, and cascade rules defined in this same PR, and mocks cannot validate that interaction; the attach constraint violation flagged above is exactly the kind of failure this gap hides. The state machine here (queued to processing to finalizing to ready, lease takeover, generation fencing) is the correctness core of the feature.

Suggest an integration test that drives claim, attach, checkpoint, complete, and the reconcile sweeps against the real migrated schema, plus the deletion race (soft delete between attach and checkpoint). Reusing the web test database harness for these functions would be enough even if the R2 side stays mocked.

services/user-data-export/src/worker.ts
```diff
@@ -0,0 +1,410 @@
...
+    }
+  }
+}
+
+async function dispatchReadyNotifications(
+  env: ExportEnv,
+  state: ReturnType<typeof createStateDb>
+): Promise<void> {
+  if (!env.USER_DATA_EXPORT_WEB_URL) return;
+  const baseUrl = new URL(env.USER_DATA_EXPORT_WEB_URL);
```
St0rmz1, line 433: **Low, Functional:** `dispatchReadyNotifications` calls `new URL(env.USER_DATA_EXPORT_WEB_URL)` with no try/catch, and `isAllowedWebCallbackUrl` also throws on an unparseable value. Every other branch of `reconcile` is defensive, but a malformed env value here throws and fails the whole scheduled invocation. Since notifications run last, the practical symptom would be "ready emails never send" with only an uncaught cron exception as the signal.

Suggest parsing inside a try/catch (or making `isAllowedWebCallbackUrl` return false on parse failure) and logging a structured warning, matching the surrounding style.

apps/web/src/app/api/internal/user-data-exports/ready/route.ts
```diff
@@ -0,0 +1,99 @@
...
+type ClaimedExport = {
+  id: string;
+  email: string | null;
+  expires_at: string;
+  email_lease_token: string;
+};
+
+export async function POST(request: NextRequest): Promise<NextResponse> {
+  const secret = request.headers.get('x-internal-api-key');
+  if (!INTERNAL_API_SECRET || !secret || !timingSafeEqual(secret, INTERNAL_API_SECRET)) {
```
St0rmz1, line 24: **Low, Testing:** This auth guard has no test coverage. The Worker's equivalent `x-internal-api-key` guard is tested in `test/worker.test.ts`, but on the web side only `delivery-state.ts` is covered; nothing asserts this route returns 401 for a missing key, a wrong key, or an unset `INTERNAL_API_SECRET`. Since this endpoint triggers outbound email to users, it should get rejection and acceptance coverage in the same PR, consistent with how the other authorization guards in this feature are tested.

St0rmz1 (approved):

services/user-data-export/src/databases.ts
pandemicsyn: Fixed in 7bc695032. Removed the invalid multipart checkpoint constraint and regenerated migration 0209 against latest main.

services/user-data-export/src/databases.ts
pandemicsyn: Not implemented per project direction: test-only feedback was explicitly excluded from this feedback pass. Existing state-machine unit, runtime, migration bootstrap, and database-backed web tests remain in place.

services/user-data-export/src/worker.ts
pandemicsyn: Fixed in 7bc695032. Malformed callback URLs now fail closed and emit a structured warning instead of throwing from the scheduled invocation.

services/user-data-export/src/worker.ts
pandemicsyn: Fixed in 7bc695032. Continuations now dispatch immediately after a durable checkpoint; the transactional outbox remains the recovery path.

services/user-data-export/src/source-adapters.ts
```diff
@@ -0,0 +1,236 @@
...
+  next_credit_expiration_at, auto_top_up_enabled, default_model, completed_welcome_form,
+  linkedin_url, github_url, discord_server_membership_verified_at,
+  openrouter_upstream_safety_identifier, openrouter_downstream_safety_identifier,
+  vercel_downstream_safety_identifier, customer_source, signup_ip, normalized_email, email_domain
+FROM kilocode_users
+WHERE id = $1
+LIMIT 1`;
+
+function cursorFor(row: Record<string, unknown>): ExportCursor {
+  return { createdAt: new Date(String(row.created_at)).toISOString(), id: String(row.id) };
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in b5593b2bc. Paginated queries now return six-digit UTC timestamp text from PostgreSQL and persist it without a lossy Date round-trip.

apps/web/src/lib/user-data-export-worker-client.ts
```diff
@@ -0,0 +1,93 @@
+import 'server-only';
+
+import { z } from 'zod';
+import { INTERNAL_API_SECRET, USER_DATA_EXPORT_WORKER_URL } from '@/lib/config.server';
+
+const LOCAL_EXPORT_WORKER_HOSTS = new Set(['localhost', '127.0.0.1', '::1']);
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in b5593b2bc. The loopback allowlist now matches WHATWG URL.hostname as `[::1]`.

packages/db/src/migrations/0209_nasty_paladin.sql
```diff
@@ -0,0 +1,97 @@
...
+	CONSTRAINT "user_data_exports_email_lease_shape" CHECK (("user_data_exports"."email_status" = 'sending') = ("user_data_exports"."email_lease_token" IS NOT NULL AND "user_data_exports"."email_lease_expires_at" IS NOT NULL)),
+	CONSTRAINT "user_data_exports_email_sent_shape" CHECK (("user_data_exports"."email_status" = 'sent') = ("user_data_exports"."email_sent_at" IS NOT NULL))
+);
+--> statement-breakpoint
+ALTER TABLE "user_data_export_outbox" ADD CONSTRAINT "user_data_export_outbox_export_id_user_data_exports_id_fk" FOREIGN KEY ("export_id") REFERENCES "public"."user_data_exports"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "user_data_export_parts" ADD CONSTRAINT "user_data_export_parts_export_id_user_data_exports_id_fk" FOREIGN KEY ("export_id") REFERENCES "public"."user_data_exports"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "user_data_exports" ADD CONSTRAINT "user_data_exports_kilo_user_id_kilocode_users_id_fk" FOREIGN KEY ("kilo_user_id") REFERENCES "public"."kilocode_users"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in b5593b2bc. Changed `user_data_exports.kilo_user_id` to `ON DELETE RESTRICT` so hard deletion cannot bypass R2 tombstones.

packages/db/src/migrations/0209_windy_mindworm.sql
```diff
@@ -0,0 +1,97 @@
...
+CREATE INDEX "IDX_user_data_export_outbox_pending" ON "user_data_export_outbox" USING btree ("available_at","created_at","id") WHERE "user_data_export_outbox"."sent_at" IS NULL;--> statement-breakpoint
+CREATE UNIQUE INDEX "UQ_user_data_exports_single_active" ON "user_data_exports" USING btree ("kilo_user_id") WHERE "user_data_exports"."status" IN ('queued', 'processing', 'finalizing');--> statement-breakpoint
+CREATE INDEX "IDX_user_data_exports_user_created" ON "user_data_exports" USING btree ("kilo_user_id","created_at","id");--> statement-breakpoint
+CREATE INDEX "IDX_user_data_exports_lease_expiry" ON "user_data_exports" USING btree ("lease_expires_at","id") WHERE "user_data_exports"."status" IN ('processing', 'finalizing');--> statement-breakpoint
+CREATE INDEX "IDX_user_data_exports_ready_expiry" ON "user_data_exports" USING btree ("expires_at","id") WHERE "user_data_exports"."status" = 'ready';--> statement-breakpoint
+CREATE INDEX "IDX_user_data_exports_email_lease_expiry" ON "user_data_exports" USING btree ("email_lease_expires_at","id") WHERE "user_data_exports"."email_status" = 'sending';
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in b5593b2bc. Added the partial failed-multipart cleanup index on `(updated_at, id)`.

packages/db/src/migrations/0209_windy_mindworm.sql
```diff
@@ -0,0 +1,97 @@
+CREATE TABLE "user_data_export_object_deletions" (
+	"object_key" text PRIMARY KEY NOT NULL,
+	"multipart_upload_id" text,
+	"reason" text DEFAULT 'account_deletion' NOT NULL,
+	"attempt_count" integer DEFAULT 0 NOT NULL,
+	"available_at" timestamp with time zone DEFAULT now() NOT NULL,
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in b5593b2bc. Added an index on `(available_at, created_at, object_key)` for the tombstone drain query.

packages/db/src/schema.ts
```diff
@@ -495,6 +495,216 @@ export const kilocode_users = pgTable(
...
+    next_part_number: integer().notNull().default(1),
+    dispatch_generation: integer().notNull().default(0),
+    lease_token: uuid(),
+    lease_expires_at: timestamp({ withTimezone: true, mode: 'string' }),
+    attempt_count: integer().notNull().default(0),
+    row_count: bigint({ mode: 'number' }).notNull().default(0),
+    size_bytes: bigint({ mode: 'number' }),
+    r2_object_key: text(),
+    r2_etag: text(),
+    sha256: text(),
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in b5593b2bc. Removed the unused `sha256` column and format constraint from the unshipped migration.

services/user-data-export/wrangler.jsonc
```diff
@@ -0,0 +1,57 @@
...
+  "observability": { "enabled": true, "logs": { "enabled": true, "head_sampling_rate": 1 } },
+  "routes": [
+    {
+      "pattern": "user-data-export.kilosessions.ai",
+      "zone_name": "kilosessions.ai",
+      "custom_domain": true,
+    },
+  ],
+  "vars": {
+    "R2_ACCOUNT_ID": "e115e769bcdd4c3d66af59d3332cb394",
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in c7d3cfa3d. `R2_BUCKET_NAME` is now explicitly configured as `user-data-export`.

services/user-data-export/wrangler.jsonc
```diff
@@ -0,0 +1,57 @@
...
+      },
+      {
+        "queue": "user-data-export-dlq",
+        "max_batch_size": 1,
+        "max_batch_timeout": 5,
+        "max_concurrency": 1,
+      },
+    ],
+  },
+  "r2_buckets": [{ "binding": "EXPORT_BUCKET", "bucket_name": "user-data-export-local" }],
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in c7d3cfa3d. `EXPORT_BUCKET` now binds to the production `user-data-export` bucket.

apps/web/src/app/api/internal/user-data-exports/ready/route.ts
pandemicsyn: Not implemented per project direction: test-only feedback was explicitly excluded from this feedback pass. The route uses the same timing-safe internal API-key contract as the tested Worker boundary.

packages/db/src/schema.ts
```diff
@@ -495,6 +495,216 @@ export const kilocode_users = pgTable(
...
+      .default(sql`pg_catalog.gen_random_uuid()`)
+      .primaryKey()
+      .notNull(),
+    export_id: uuid()
+      .notNull()
+      .references(() => user_data_exports.id, { onDelete: 'cascade' }),
+    generation: integer().notNull(),
+    operation: text().$type<'generate'>().notNull().default('generate'),
+    available_at: timestamp({ withTimezone: true, mode: 'string' }).defaultNow().notNull(),
+    attempt_count: integer().notNull().default(0),
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in b5593b2bc. Failed outbox sends now increment `attempt_count` and apply capped exponential backoff without blocking later rows.

PR #5171, opened 2026-08-10
feat(admin): add data export recovery console

pandemicsyn:
## Summary
- add an admin workload-health console at `/admin/data-exports` with summary metrics, exact filters, paginated workload state, and detailed execution/outbox/artifact diagnostics
- allow any current Kilo admin to redispatch, cancel and purge, or cancel and retry exports through generation-checked, primary-database transactions
- preserve PostgreSQL as the control plane: durable outbox publication precedes immediate dispatch, cleanup tombstones precede row deletion, and stale Queue deliveries are fenced by generation
- extend cleanup reasons for admin cancellation and replacement and harden Worker dispatch/outbox selection against obsolete generations

### Redispatch
- requires an active export without a live lease or completed artifact
- advances the generation to fence stale Queue/DLQ messages
- restarts the one-shot export from the beginning
- clears retired cursor/part/progress/failure/email state
- retains an interrupted multipart ID so the Worker aborts it before starting
- preserves the export ID and original snapshot

### Cancel and purge
- requires exact export-ID confirmation
- persists deferred R2/multipart cleanup before deleting the export control-plane row
- cascades parts and outbox history
- lets the user request another export immediately

### Cancel and retry
- requires exact export-ID confirmation
- persists cleanup for the old export
- creates a fresh export UUID and generation-zero outbox task in the same transaction
- preserves the logical snapshot, schema version, owner, and original request time
- bypasses the normal user request cooldown

## One-shot generator compatibility
This branch is based on #5167. It does not restore cross-invocation checkpoints or multipart continuation. Dispatch generation is used only for stale-message fencing. The 12-minute source cutoff, 13-minute hard abort, 16-minute lease, per-record compression writes, 5 MiB multipart buffer, abort-before-transient-retry behavior, and terminal deadline handling remain unchanged.

## Authorization and safety
- health/list/detail and all recovery mutations use `adminProcedure`
- any Kilo admin may operate the console; superadmin is not required
- mutations revalidate admin status, expected generation, owner state, lease, artifact shape, and active-export constraints under advisory and row locks
- the browser never supplies object keys, multipart IDs, Queue identifiers, or R2 credentials
- destructive actions require typing the full export UUID

## Migration
`0210_user_data_export_admin_recovery.sql` expands cleanup reasons to:
- `account_deletion`
- `admin_cancel`
- `admin_replace`

The migration was generated with Drizzle, applied locally, and verified against an empty bootstrap database.

## Verification
- web typecheck passed
- 55 targeted admin/router/UI tests passed
- 56 Worker unit tests passed
- 5 Workers-runtime tests passed
- 39 shared schema tests passed
- DB, Worker, and targeted web lint passed
- migration apply and empty-database bootstrap passed
- `git diff --check` passed
- browser-tested responsive health/list/detail views and all three recovery actions
- browser-verified Redispatch as a normal admin with `is_admin=true` and `is_super_admin=false`
- independent race/security and #5167 compatibility reviews found no medium or high-severity defects


apps/web/src/routers/admin/user-data-exports-router.test.ts
```diff
@@ -0,0 +1,556 @@
...
+  it('summarizes persisted failure, lease, dispatch, cleanup, and email health', async () => {
+    const result = await (await createCallerForUser(admin.id)).admin.userDataExports.summary();
+
+    expect(result).toMatchObject({
+      active: 2,
+      needsAttention: 4,
+      staleLeases: 1,
+      pendingDispatches: 1,
+      failed: 1,
+      cleanupDue: 1,
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in bd9af1a9d. The assertion now includes a captured baseline for pre-existing due cleanup rows.

apps/web/src/routers/admin/user-data-exports-router.test.ts
```diff
@@ -0,0 +1,556 @@
...
+    });
+
+    expect(result.pagination).toMatchObject({ page: 1, limit: 2, total: 4, totalPages: 2 });
+    expect(result.rows).toHaveLength(2);
+  });
+
+  it('returns legacy part aggregates, outbox history, and redacted failure details', async () => {
+    const caller = await createCallerForUser(admin.id);
+    const processing = await db.query.user_data_exports.findFirst({
+      where: eq(user_data_exports.status, 'processing'),
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in bd9af1a9d. The test now retains and uses the leased processing fixture ID directly.

apps/web/src/app/admin/data-exports/[id]/DataExportRecoveryCard.tsx
```diff
@@ -0,0 +1,335 @@
...
+        queryClient.removeQueries(
+          trpc.admin.userDataExports.detail.queryFilter({ exportId: detail.id })
+        );
+        await Promise.all([
+          queryClient.invalidateQueries(trpc.admin.userDataExports.list.queryFilter()),
+          queryClient.invalidateQueries(trpc.admin.userDataExports.summary.queryFilter()),
+        ]);
+        const copy = retryToastCopy(result);
+        toast[copy.kind](copy.title, { description: copy.description });
+        router.replace(`/admin/data-exports/${encodeURIComponent(result.replacementExportId)}`);
```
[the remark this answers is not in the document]
Comment (pandemicsyn) [reply apps/web/src/app/admin/data-exports/[id]/DataExportRecoveryCard.tsx:112]: Fixed in bd9af1a9d. Retry success now closes the dialog and clears the typed confirmation before navigation.

apps/web/src/routers/admin/user-data-exports-router.ts
```diff
@@ -0,0 +1,692 @@
...
+      return { exportId: target.id, kiloUserId: target.kilo_user_id, generation: nextGeneration };
+    });
+    return {
+      exportId: result.exportId,
+      generation: result.generation,
+      ...(await dispatchAfterCommit(result)),
+    };
+  }),
+
+  cancelAndPurge: adminProcedure.input(RecoveryInputSchema).mutation(async ({ input }) => {
```
St0rmz1: **Moderate, Security:** `cancelAndPurge` and `cancelAndRetry` hard-delete the export row, but nothing records which export was destroyed or whose data it was.

The only audit signal is the `admin_access` event emitted by `adminProcedure`. That carries the acting admin's ID and email plus `route: "admin.userDataExports.cancelAndPurge"`, but not the procedure input, and `sentryMiddleware` is configured with `attachRpcInput: false`. Grepping this file for `ctx` or any log call returns nothing, so the mutation itself writes no trace.

After the delete, `kilo_user_id` is gone with the row and the parts and outbox rows cascade away. The `user_data_export_object_deletions` tombstone preserves the object key, which embeds the export UUID, and a `reason` of `admin_cancel` or `admin_replace`, so you can recover which export and roughly when. You cannot recover who did it or whose data it was, because those two halves live in different systems with no join key. Any Kilo admin can run this, not just a superadmin, so the set of people who could have done it is large.

Suggestion: emit a structured log line inside each mutation before returning, with `exportId`, `kiloUserId`, the acting admin's ID, the action, and the generation. `lockRecoveryTarget` already returns the owner ID, so the data is in hand. The codebase has `logExportEvent` on the Worker side and `recordKiloAdminElevation` in `lib/admin/admin-access-log.ts` for this kind of attribution.

apps/web/src/routers/admin/user-data-exports-router.test.ts
```diff
@@ -0,0 +1,564 @@
...
+
+  it('returns not found for an unknown export', async () => {
+    await expect(
+      (await createCallerForUser(admin.id)).admin.userDataExports.detail({
+        exportId: crypto.randomUUID(),
+      })
+    ).rejects.toMatchObject({ code: 'NOT_FOUND' });
+  });
+
+  it('redispatches an active export as a clean one-shot run with a fenced generation', async () => {
```
St0rmz1, line 325: **Moderate, Testing:** The recovery safety interlocks have no coverage.

This suite covers the authorization boundary on all five procedures, generation staleness, cascade behavior, and the replacement blocker, which is the hard part and it is well done. Four guard branches in the router are untested:

- soft-deleted or blocked owner returns `PRECONDITION_FAILED` (`user-data-exports-router.ts`:156)
- unexpected `r2_object_key` returns `PRECONDITION_FAILED` (:175)
- a live worker lease blocks `redispatch` (:579)
- an existing artifact blocks `redispatch` (:585)

These are the checks that stop an admin from restarting an export while a Worker is mid-flight, or from purging an artifact whose key does not match the expected shape, which the code itself treats as needing investigation. They are also the branches most likely to be silently dropped by a future rewrite of the `lockRecoveryTarget` SQL, and losing any of them would not fail a single existing test.

Suggestion: four fixtures and four `rejects.toThrow` assertions, following the shape of the existing "rejects recovery when the expected generation is stale" test.

apps/web/src/app/admin/data-exports/DataExportsContent.tsx
```diff
@@ -0,0 +1,382 @@
...
+  const pendingQueryRef = useRef<string | null>(null);
+  const [isNavigationLocked, setIsNavigationLocked] = useState(false);
+  const [isNavigationPending, startNavigation] = useTransition();
+
+  const [searchDraft, setSearchDraft] = useState(filters.search ?? '');
+  useEffect(() => {
+    desiredFiltersRef.current = filters;
+  }, [filters.emailStatus, filters.health, filters.page, filters.search, filters.status]);
+  useEffect(() => {
+    if (pendingQueryRef.current === null || pendingQueryRef.current !== searchParams.toString())
```
St0rmz1: **Low, Functional:** This effect is the only path that clears `isNavigationLocked`, and it requires an exact match.

```ts
if (pendingQueryRef.current === null || pendingQueryRef.current !== searchParams.toString())
  return;
```

If the user clicks Next, which sets `pendingQueryRef.current` to `page=2` and locks at line 107, and then hits browser Back before the push settles, `searchParams` resolves to something other than `page=2`. The effect returns early, `pendingQueryRef.current` is never nulled, and the lock is never released. `isNavigationPending` clears on its own when the transition ends, but `isNavigationLocked` does not.

`controlsDisabled` at line 175 then stays true for the rest of the session, which disables all three filter selects, the search input, the Search and Clear buttons, and both pagination buttons. Recovering takes a full page reload. Same story for any history navigation that lands on a different query string while a push is in flight.

Suggestion: clear the lock whenever `searchParams` changes at all rather than only on an exact match, or additionally clear it when `isNavigationPending` goes false.

apps/web/src/routers/admin/user-data-exports-router.ts
```diff
@@ -0,0 +1,692 @@
...
+        ),
+        available_at = GREATEST(
+          user_data_export_object_deletions.available_at,
+          EXCLUDED.available_at
+        ),
+        updated_at = now()
+  `);
+}
+
+async function dispatchAfterCommit(input: {
```
St0rmz1, line 211: **Low, Functional:** `dispatchAfterCommit` runs after the transaction commits, so a throw here reports a succeeded action as failed.

`dispatchUserDataExport` is well behaved for network and HTTP failures: `postExportWorker` catches and returns null, and the result degrades to `pending`. But `generateInternalServiceToken` is called outside that try block, so a signing misconfiguration throws out of the mutation.

For `cancelAndRetry` that is the worse case. The original export is already deleted and the replacement already exists when this runs. The client shows "Retry failed", the dialog stays open with the typed confirmation preserved, and the admin is left on a detail page for an export that no longer exists. Retrying is fail-safe because it hits `NOT_FOUND` or `CONFLICT`, but there is no signal that the first attempt actually worked.

Narrow trigger, so treat this as optional. Suggestion: wrap the call in try/catch and return `dispatch: 'pending'` on throw. That matches the existing semantics, since the toast already says the worker could not be reached and the outbox will dispatch it automatically.

apps/web/src/routers/admin/user-data-exports-router.ts
```diff
@@ -0,0 +1,692 @@
...
+  'queued',
+  'processing',
+  'finalizing',
+  'ready',
+  'failed',
+  'expired',
+]);
+const EmailStatusSchema = z.enum(['pending', 'sending', 'sent', 'failed']);
+const HealthFilterSchema = z.enum(['needs_attention', 'active', 'terminal', 'all']);
+const MAX_PAGE = 10_000;
```
St0rmz1: **Low, Performance:** `MAX_PAGE` of 10,000 with a `limit` up to 100 allows an `OFFSET` of one million.

The default `needs_attention` filter is a disjunction of fourteen predicates including a correlated `EXISTS` against `user_data_export_parts`, which no index can serve. Every list page is therefore a sequential scan of `user_data_exports` plus a per-row subquery, and the count query at line 424 repeats the same work on every page load.

Nothing is wrong today, since the table is small and this is admin-only. It will degrade quietly as export volume grows, and the doubled cost is paid on every request rather than only on deep pages.

Not blocking. If you want cheap insurance, lowering `MAX_PAGE` to a few hundred bounds the worst case at no cost, since no operator is paging to 10,000 by hand. Keyset pagination on `(created_at, id)` would be the real fix and `IDX_user_data_exports_user_created` partially supports it.

apps/web/src/routers/admin/user-data-exports-router.ts
```diff
@@ -0,0 +1,692 @@
...
+  if (row.r2_object_key !== null && row.r2_object_key !== objectKey(row.id)) {
+    throw new TRPCError({
+      code: 'PRECONDITION_FAILED',
+      message: 'The export has an unexpected artifact key and requires investigation',
+    });
+  }
+  return row;
+}
+
+async function scheduleCleanup(
```
St0rmz1, line 181: **Nit, Design:** `scheduleCleanup` inserts a tombstone even when there is nothing to clean.

Purging a queued export that never wrote an artifact and never started a multipart upload still creates a row in `user_data_export_object_deletions` for an object key that never existed. The sweeper then issues a delete for a nonexistent key, and the row counts toward `summary.cleanupDue`, which is the number an admin reads to judge whether cleanup is backed up.

Skipping the insert when both `r2_object_key` and `multipart_upload_id` are null looks safe: the one window where a Worker creates the R2 multipart just before the purge is already covered by the `cleanup` CTE in `attachMultipartUpload`, which writes the tombstone when it finds the export row gone.

That said, the account deletion path in `lib/user/index.ts`:1088 tombstones unconditionally too, so this is consistent with existing behavior. Leaving it alone is a fine answer.

apps/web/src/routers/admin/user-data-exports-router.ts
```diff
@@ -0,0 +1,692 @@
...
+    leaseExpiresAt: nullableIso(row.lease_expires_at),
+    failureCode: row.failure_code,
+    failureMessage: row.last_error_redacted,
+    emailStatus: row.email_status,
+    emailAttemptCount: row.email_attempt_count,
+    health: classifyExportHealth(healthInput(row), asOf),
+  };
+}
+
+const needsAttentionSql = sql`
```
St0rmz1, line 274: **Nit, Design:** Two of the fourteen predicates are already subsumed by earlier ones.

`e.status = 'failed'` on the first line covers `e.status = 'failed' AND e.multipart_upload_id IS NOT NULL` five lines down. Likewise `e.status NOT IN ('queued', 'processing', 'finalizing') AND e.lease_token IS NOT NULL` covers the `e.status = 'queued' AND e.lease_token IS NOT NULL` case that follows it.

No correctness impact, since the clause is a disjunction. Worth trimming anyway because this fragment is the hardest thing in the file to hold in your head, and it is interpolated into the `summary` query as well, so any confusion about it is paid for twice.

apps/web/src/routers/admin/user-data-exports-router.ts
pandemicsyn: Fixed in 423ca5229. Recovery mutations now emit non-throwing structured audit events with actor, owner, export, action, generation, replacement, and cleanup attribution.

apps/web/src/routers/admin/user-data-exports-router.test.ts
pandemicsyn: Fixed in 423ca5229. Added DB-backed coverage for blocked owners, noncanonical artifact keys, live leases, and existing artifacts.

apps/web/src/app/admin/data-exports/DataExportsContent.tsx
pandemicsyn: Fixed in 423ca5229. Any committed search-param change now releases the navigation lock, including browser history navigation.

apps/web/src/routers/admin/user-data-exports-router.ts
pandemicsyn: Fixed in 423ca5229. Post-commit dispatch exceptions now degrade to pending so durable outbox success is reported correctly.

apps/web/src/routers/admin/user-data-exports-router.ts
pandemicsyn: Fixed in 423ca5229. The page cap is now 400, bounding the maximum offset to 39,900 at the API limit.

apps/web/src/routers/admin/user-data-exports-router.ts
pandemicsyn: Fixed in 423ca5229. Pristine lease-free exports now skip tombstones; live-lease exports still queue deferred cleanup for late Worker writes.

apps/web/src/routers/admin/user-data-exports-router.ts
pandemicsyn: Partially addressed in 423ca5229: removed the redundant failed-with-multipart clause. Kept queued-with-lease because the preceding terminal-with-lease predicate explicitly excludes queued, processing, and finalizing rows.

RSO (approved):

PR #5180, opened 2026-08-10
feat(exports): enable user data exports for all users

pandemicsyn:
## Summary
- remove the temporary Kilo admin gate from user data export tRPC procedures
- allow all authenticated users to open the data exports page
- show the Data exports account-menu entry to every user
- retain web-session-only access and per-user export ownership checks
- allow asking for one data export every hour


jrf0110 (approved):
Probably double check all of the authorization code here and make sure there aren't ways to request downloads that don't actually belong to you

PR #5183, opened 2026-08-10
chore(web): stop scheduling weekly recommendations digest email

jobrietbergen:
## Summary

Stops the weekly enterprise recommendations digest email from being sent.

The email was triggered exclusively by a Vercel cron entry (`0 9 * * 1` — Mondays 09:00 UTC) hitting `/api/cron/enterprise-recommendations-digest`. Removing that schedule is sufficient to stop all delivery, because nothing else in the codebase calls `dispatchEnterpriseRecommendationsDigests()`.

No architectural changes. No database changes.

Deliberately left in place, for a reviewer to decide on as a possible follow-up:

- The cron route, `recommendations-digest.ts`, its tests, and `recommendationsDigest.html` remain as unused code.
- The "Weekly recommendations email" toggle still appears in organization settings for enterprise orgs, and still writes `recommendations_digest_enabled`. It no longer causes any email to be sent, so the UI now overpromises.
- Existing `recommendations_digest_enabled` values and historical `transactional_email_log` rows are untouched. `OrganizationSettingsSchema` is a non-strict `z.object`, so no read path breaks.

The in-app feature-adoption recommendations dashboard is unaffected: `getOrganizationRecommendations` is shared with `organization-usage-details-router.ts` and was not modified.

## Verification

Manual verification of the send path was not performed, because reproducing it end-to-end requires an enterprise org with the digest flag enabled, an owner with a valid email, and at least one open recommendation — none of which exist in a local database, and the Vercel cron does not fire in local dev.

- [ ] Confirmed the removed entry was the only trigger: `dispatchEnterpriseRecommendationsDigests` has no other callers in the repo
- [ ] Confirmed `apps/web/vercel.json` still parses and retains its other 26 cron entries
- [ ] Confirmed no test or repo script validates cron paths against route files, so nothing asserts this entry's presence
- [ ]

## Visual Changes

N/A

## Reviewer Notes

Single-file, four-line deletion; revert by restoring the entry.

Worth a decision from you: should a follow-up PR remove the now-dead code and the settings toggle? I scoped this PR to stopping delivery only, keeping it minimal and easily reversible.

### DIFF
--- a/apps/web/src/components/organizations/OrganizationEmailPreferencesCard.tsx
+++ b/apps/web/src/components/organizations/OrganizationEmailPreferencesCard.tsx
@@ -3,14 +3,9 @@
 import { useState } from 'react';
 import { Card, CardContent, CardDescription, CardHeader, CardTitle } from '@/components/ui/card';
 import { Button } from '@/components/ui/button';
-import { Switch } from '@/components/ui/switch';
-import { Bell, ChartLine, Mail } from 'lucide-react';
+import { Bell, Mail } from 'lucide-react';
 import type { LucideIcon } from 'lucide-react';
-import { toast } from 'sonner';
-import {
-  useOrganizationWithMembers,
-  useUpdateRecommendationsDigest,
-} from '@/app/api/organizations/hooks';
+import { useOrganizationWithMembers } from '@/app/api/organizations/hooks';
 import { SpendingAlertsModal } from './SpendingAlertsModal';
 
 type Props = {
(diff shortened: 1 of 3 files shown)


jrf0110 (approved):
Looks good. I added b77e9f0da to hide the weekly recommendations email UI now that the digest is no longer scheduled.

jrf0110:
Looks good. I added a fix in b77e9f0da to hide the weekly recommendations email toggle and the onboarding summary status so the UI no longer promises an email that will not be sent.

PR #5190, opened 2026-08-10
feat(extension): use-case task benchmark and read-depth harness fixes

iscekic:
# Use-case benchmark and read-depth harness fixes

**Status: ready for human review.** CI green; adversarial reviewer HAPPY;
Kilobot "No Issues Found — Merge".

## Summary

PR #5154 made workflow creation reliable. This change extends the same
data-driven method to the rest of the popular browser-agent use cases: it
ranks them from online data, pins seven task scenarios that measure them,
and fixes the harness root causes the baseline surfaced. The headline fix
gives safe mode full read access to long pages: snapshot text paging plus
full-page text search.

## The use cases, data-driven

Ranked from install counts, review themes, and platform feature bets
(Sider 5M installs, Monica 3M, Merlin 900k, MaxAI 700k, Immersive
Translate 20M+, Edge Copilot, Gemini in Chrome, Perplexity Comet, HARPA):

1. Summarize the page — the lead feature of every top extension.
2. Page Q&A — the core sidebar function.
3. Write and reply — draft text from page content.
4. Translate — full-page translation.
5. AI search / research — now served by the `web_search` tool.
6. Multi-tab compare — needs cross-tab reads. Documented gap.
7. Agentic actions — log in, fill forms, operate the page.
8. Repeatable automation — the twenty workflow scenarios of PR #5154.

Reading tasks (1, 2, 4) carry roughly 80% of real usage. Use case 6 stays
a documented gap (deliberately deferred); use case 5 is closed by this
change.

## Web search through the existing Exa proxy

The backend already runs an Exa proxy at `/api/exa/*` with auth, a path
allowlist, the monthly free allowance, balance checks, and per-request
billing. The extension therefore needs no key and no new backend: the new
`web_search` tool posts the query to `/api/exa/search` with the user's
bearer token and returns up to five title/url/date/snippet records.
Allowance and balance errors pass through as tool errors so the model can
tell the user why a search failed. The tool is available in both modes,
and the `web-research` scenario pins it: the answer must carry the fact
and a source URL, with tool evidence, from a page that cannot answer it.

`web-research` passes **14/14** across the model set — every paid model
and every free model tried, including `inclusionai/ling-3.0-tiny:free`,
which scores 4/7 on the other scenarios. Regression on the same head:
`qa-deep-fact` and `hn` still pass, and the model correctly keeps using
page tools for page questions rather than reaching for search.

## Task scenarios

Seven pinned scenarios in `agent-task-bench-scenarios.ts` cover use cases
1-4 and 7 (see `workflow-create-benchmark.md` for the table). The driver
runs them with `--scenario <id>`; correctness scores the final assistant
answer: pinned content checks, a minimum length, and an ok `eval`
exchange for action scenarios. Task batches gate on the turn total: every
answer must land inside 120 s.

**Anti-hallucination rule.** A check may set `requireToolEvidence`: the
same pattern must appear in an ok tool result. This exists because the
truncation-era baseline produced a dishonest pass — the model summarized a
famous essay "from familiarity" with the full text while presenting it as
page content. With the rule, a pass proves the page was actually read.

## Root causes found and fixed

The baseline on `kilo-auto/efficient` (raw transcripts) surfaced these:

- **Safe mode could not read past 8000 characters.** The snapshot capped
  visible text with no continuation, and `find_in_page` searched only that
  window. On the deep-fact scenario the model probed `find_in_page` eleven
  times, correctly concluded the harness could not see deeper content, and
  gave up after 163 s. Fixes:
  - `get_page_snapshot` takes optional `textStart`; the result reports
    `textStart`/`textTotalChars`/`textTruncated` plus a continuation note.
  - `find_in_page` searches the full visible page text inside the same
    snapshot injection (one DOM walk); matches carry excerpts with
    character offsets that compose with `textStart`.
  - Result: the deep-fact scenario went from 163 s / 14 tool calls /
    honest failure to 31 s / 4 tool calls / correct grounded answer; the
    summarize scenario pages through the whole essay (9 windows, 87 s).
- **The model presented training knowledge as page content.** New system
  prompt rule: answer from tool results, never from training knowledge of
  the site or document; the evidence gate in scoring enforces it.
- **The driver could not run dangerous-mode scenarios.** The mode toggle
  label reads "Danger mode", not "Dangerous mode"; the driver now switches
  and waits on the real label.

## Results

On `kilo-auto/efficient` (head 398c50885, safe/dangerous per scenario,
production gateway, one attempt per cell):

| Scenario | Baseline (pre-fix) | After fixes | Turn time |
|---|---|---|---|
| qa-deep-fact | fail (could not reach deep content, 163 s) | pass | 31 s |
| summarize-article | dishonest pass (training data) | honest pass (9 paged windows) | 87 s |
| extract-table | pass | pass | 16 s |
| translate-page | pass | pass | 60 s |
| draft-reply | pass (ungrounded) | pass (grounded, evidence-gated) | ~40 s |
| action-login | driver blocker | pass | — |
| action-cart | driver blocker | pass | — |

Workflow-scenario regression check on the same head: hn, weather, npr all
pass (the full 20-scenario suite ran in PR #5154; the snapshot shape
change is exercised by these three plus the 1197 unit tests).

### Multi-model sweep

27 models (kilo-auto/efficient plus the top OpenRouter paid and free set
from PR #5154), 7 task scenarios each, one attempt per cell, production
gateway. 155/189 cells pass. Paid models: 130/133 (97.7%); 16 of 19 paid
models score 7/7, including `kilo-auto/efficient`. `qa-deep-fact` — the
scenario the read-depth fixes exist for — passes 27/27, including every
capability-limited free model.

| Model | Passed |
|---|---|
| kilo-auto/efficient | 7/7 |
| anthropic/claude-opus-4.7, opus-4.8, sonnet-4.6, sonnet-5 | 7/7 |
| anthropic/claude-opus-5 | 6/7 |
| deepseek/deepseek-v4-flash, v4-flash-0731, v4-pro | 7/7 |
| google/gemini-3.1-pro-preview, gemini-3.5-flash | 7/7 |
| google/gemini-3-flash-preview, gemini-3.6-flash | 6/7 |
| openai/gpt-5.5, gpt-5.6-luna, gpt-5.6-sol | 7/7 |
| tencent/hy3, xiaomi/mimo-v2.5, z-ai/glm-5.2 | 7/7 |
| poolside/laguna-s-2.1:free | 6/7 |
| poolside/laguna-xs-2.1:free, stepfun/step-3.7-flash:free | 5/7 |
| cohere/north-mini-code:free, inclusionai/ling-3.0-tiny:free | 4/7 |
| nvidia/nemotron-3-ultra:free, nemotron-3-nano-omni:free | 3/7 |
| nvidia/nemotron-3-super-120b:free | 2/7 |

Residuals are model behavior, not harness faults:

- `claude-opus-5` stops and asks before the credentialed cart action even
  though the message carries explicit approval (over-caution; twice).
- `gemini-3-flash-preview`/`gemini-3.6-flash` answer the summarize task
  after reading 2-3 of 9 text windows, or with a one-line stub, despite
  the continuation note saying how to read on (under-reading).
- The weak free models fail eval-driven actions and deep reads — the same
  capability-limited set PR #5154 documents.

### Scoring evolution during the sweep (recorded honestly)

Two scoring rules changed after the first full pass, and only the
affected cells were re-run on the new head:

- Theme checks accept honest paraphrases and a 2-of-3 quorum for the
  summary — eight capable models had read the entire essay (evidence
  green, up to nine paged windows) yet failed on the literal words
  "procrastination"/"originality". Evidence gates were never relaxed.
- The action predicate accepts a real `run_workflow` as well as `eval` —
  one model completed the cart task by saving and running a workflow,
  which is a legitimate harness action path.

Widening can flip fails to passes but never passes to fails, so passing
cells were not re-run.

## A/B on the tools the failing scenarios use

Both levers were measured with the repo's A/B protocol, one variable per
step, on the models that actually failed.

**Snapshot text window 8000 to 24000 characters — KEPT.** Three attempts
per cell on `summarize-article`:

| Model | pass 8k → 24k | LLM requests | snapshot bytes | median turn |
|---|---|---|---|---|
| gemini-3.6-flash | 2/3 → 2/3 | 10 → 4 | 113k → 82k | 30s → 21s |
| ling-3.0-tiny:free | 2/3 → 3/3 | 14 → 6 | 103k → 97k | 33s → 19s |
| nemotron-ultra:free | 0/3 → 1/3 | 9 → 6 | 52k → 82k | — → 207s |
| kilo-auto/efficient | 3/3 → 3/3 | 4 → 4 | 113k → 82k | 42s → 48s |

7/12 to 9/12 passes with roughly half the round-trips, and *fewer* total
bytes: reading a long article in three calls instead of nine avoids
re-sending a growing conversation eight times. Under-reading was never
stubbornness — nine sequential round-trips give a weak model nine chances
to stop early. Regression on the same head: `qa-deep-fact` (36 s → 20 s),
`extract-table`, `web-research`, and the `hn` workflow scenario all pass.

The one task shape where a wider window could cost more — workflow
creation, which is prompted to take *at most one* snapshot and needs form
fields rather than prose — was measured too: the flights page snapshot is
13,207 bytes at 8k and 13,208 at 24k, because that page's visible text
sits far below either cap. No byte cost on that path.

**Eval navigation recovery — MEASURED AND REJECTED.** Raw transcripts
showed weak models spending 23-60 requests on the two-step cart task,
retrying a click that had already navigated and destroyed the JS
execution context. Returning the landed page as a success instead of that
error did not help: 3/8 → 2/8 passes, and `laguna-xs-2.1:free` went from
46 to 77 requests. Removing the error appears to remove the signal that
something went wrong. Reverted; the error text stays.

## Adversarial review round 1

The ponytail adversarial reviewer found four real defects; all are fixed:

- Harness metadata (the paging note's character counts, the base36
  snapshotId) could satisfy an evidence pattern. Evidence collection now
  skips `note` and `snapshotId` strings, with a regression test.
- `draft-reply` was passable from training memory; it now pins a
  page-only fact with tool evidence.
- The full-text walk called getComputedStyle per ancestor per text node;
  a huge DOM could blow the injection deadline and fail the snapshot
  entirely. Visibility is memoized per element and a mid-walk deadline
  keeps the collected text instead of throwing.
- `find_in_page` paid two DOM walks and its `truncated` flag lied when
  the page had more matches than returned excerpts. The search now runs
  inside the snapshot injection (one walk, aligned offsets) and the flag
  compares total matches against returned matches.

Declined from the same review: `useCase` stays on the scenario spec (it
is registry documentation), and the evidence string collector stays
separate from the workflow scorer's (their semantics now differ — the
evidence collector must skip metadata keys).

## Honesty notes

- Scenario pins were corrected once during authoring (books.toscrape
  truncates long titles server-side), before any A/B claim.
- The workflow scenarios were re-checked on the new head (hn, weather,
  npr) because the snapshot shape changed; the full 20x30 workflow sweep
  from PR #5154 was not re-run.
- Raw transcripts stay local (`KILO_BENCH_RAW=1`); committed artifacts are
  redacted as before, and `find_in_page` results stay byte-count-only.
- The flights scenario returned 0/2, 1/2 and 2/2 across three batches of
  effectively identical code, which is the clearest available measure of
  how noisy that particular scenario is.
- Single-attempt cells carry real run-to-run variance: two models that
  failed `summarize-article` in the one-attempt sweep passed it on a
  re-run. That is why both A/B arms use multiple attempts per cell, and
  why the sweep table above should be read as a coarse capability signal
  rather than a precise per-model score.


iscekic:
@kilo-code-bot review

iscekic:
@kilo-code-bot /review

PR #5196, opened 2026-08-11
fix(extension): stop retrying context-window overflows

### LINKED ISSUE #5191: Extension: a context-window overflow costs three identical billed attempts
Found while reviewing PR #5190. Pre-existing, not introduced there.

`model_context_window_exceeded` is in `TRUNCATED_FINISH_REASONS`
(`apps/extension/src/shared/agent-llm-turn-runner-core.ts`), so the turn
runner retries it up to `MAX_STREAM_ATTEMPTS` (3) with the **same**
messages array. A prompt that overflowed the context cannot fit on retry,
so both retries are guaranteed to fail, cost a billed request each, and
add a delay before the runner degrades to the partial text.

Auto-compaction runs after a turn ends, so a small-context model that
takes several large snapshots inside one turn hits this mid-turn.

Reproduction: a model with a small context window plus three
`get_page_snapshot` calls in one turn.

Fix sketch: split that one finish reason out of the retriable set and
return the degraded completion immediately.

Deliberately not fixed in PR #5190: it lives in the shared turn runner,
affects every long conversation, and changing retry classification
deserves its own change and its own measurement.

iscekic:
Fixes #5191.

## Problem

`model_context_window_exceeded` sat in `TRUNCATED_FINISH_REASONS`, so the turn runner retried it up to 3 times with the same messages. A prompt that overflowed the context cannot fit on a retry: both retries were guaranteed failures, each one billed, plus 5 s of retry delay before the runner degraded to the partial text.

## Fix

Drop that one finish reason from the retriable set (`agent-llm-turn-runner-core.ts`). The first attempt now returns the completion at once, which is the same degrade path the third attempt took before — minus two billed requests and the delay.

## Verification

- Unit test: an overflow completion makes exactly 1 gateway call and keeps the partial text. The test fails on the pre-fix code (3 calls) and passes on the fix.
- E2E (`context-overflow-no-retry.test.ts`): the side panel shows the partial text, and the mocked gateway sees exactly 1 chat request after the old retry window has passed. Ran locally against the built extension: pass.
- `pnpm --filter kilo-extension verify`: 1210 unit tests pass.
- Full Chrome e2e suite: 141 pass, 13 skipped (live-gated), 0 fail.

## Base-branch e2e repair included

PR #5190 added `web_search` to the tool list but not to the e2e fixture's expected tool names. Every message-sending e2e spec failed on `ext-harness-usecases` because of it. This PR adds `web_search` to `safeToolNames`, `dangerousToolNames`, and the four inline lists in `safe-mode`, `conversation-spacing`, and `remote-mcp-settings` specs.

## Known residual (pre-existing, out of scope)

When the partial overflow text looks like a work announcement (or the overflow lands in a thinking-only completion), the continue nudge still sends one follow-up request with a strictly larger prompt, which overflows again. `continueNudgeSent` caps it at one request. This path predates this change; suppressing the nudge on this finish reason deserves its own measurement.


iscekic:
@kilo-code-bot review

PR #5202, opened 2026-08-11
fix(user-data-export): show expired downloads as disabled with a tooltip

RSO:
## Summary

On the Data exports page, an export whose download window had lapsed rendered no download control at all — the row's action area was simply empty, which reads as a missing/broken UI rather than an intentional state.

The download button now renders for `expired` as well as `ready`, disabled, with a tooltip on hover/focus: `Download expired. Request a new export.`

- Extracted the button into `DownloadExportButton` so the ready and expired variants share one definition.
- The disabled button is wrapped in `<span tabIndex={0}>` inside `TooltipTrigger asChild`, because a disabled button emits no pointer events and would never open the tooltip. Same pattern as `ModelExperimentsContent.tsx` and `NotConnectedShell.tsx`.
- `aria-label` becomes `Download expired for export requested <date>`, since Radix attaches its tooltip description to the span wrapper rather than the button.
- Deleted `page.test.tsx`. `apps/web`'s jest `testMatch` is `**/src/**/*.test.ts`, so this file — the only `.test.tsx` in `apps/web` — was never collected.

Note that `expired` is derived, not just stored: `getDisplayStatus` treats a `ready` export past `expiresAt` as expired, and the list query already schedules a refetch at the next expiry, so a row flips to the disabled state without a reload.

## Verification

- [ ] Not manually tested. Reproducing the state needs a `user_exports` row with `status = 'ready'` and a past `expires_at`, and I did not have local services up for this worktree. Automated checks (`lint`, `tsgo --noEmit`, `oxfmt`) pass.

## Visual Changes

No screenshots — see the Verification note above; I did not render the expired state locally.

| Before | After |
|---|---|
| Expired row: no download button | Expired row: disabled Download button, tooltip `Download expired. Request a new export.` |

## Reviewer Notes

- Two copy decisions worth a look: the tooltip says `Download expired. Request a new export.` rather than a bare `Expired`, to satisfy the design guidance that disabled states explain blocked progress. The row body already carries the `Expired` badge and `Expired <timestamp>` detail, so there is some repetition.
- Deleting `page.test.tsx` drops the only assertion that the page calls `getUserFromAuth({ adminOnly: true })`. The file contains no JSX, so renaming it to `page.test.ts` would have made it run instead of removing it — happy to do that if you'd rather keep the admin-gating coverage.
- Only `ready` and `expired` render a button; `queued`, `processing`, and `failed` are unchanged.


RSO:
The `page.test.tsx` looked really cool, but was never run because we don't support tsx tests, so the net result of me removing the file is that we're actually clearer about the fact that the test doesn't do anything ;)

eshurakov (approved):

PR #5215, opened 2026-08-12
fix(ai-gateway): apply org access policy to Auto Efficient routing

chrarnoldus:
## Summary

Auto Efficient only sent the enterprise `model_deny_list` to `/decide`. Models forbidden by disabled providers (org `provider_allow_list` / group provider grants) or by a group model allow-list could still be chosen, then 403 after resolution.

This evaluates `getEffectiveModelDecision` against routing-table candidates plus coding-plan defaults (`minimax/minimax-m3`, `byteplus-coding/bytedance-seed-code`) and forwards denied IDs as `deniedModelIds`. The worker already filters that set, so Efficient picks an allowed model instead of a forbidden one.

## Test plan

- [ ] `pnpm --filter web exec jest src/lib/ai-gateway/auto-routing-denied-models.test.ts 'src/app/api/openrouter/[...path]/route.test.ts'`
- [ ] Enterprise org with a provider disabled: Auto Efficient should not route to a model that only exists on that provider
- [ ] Enterprise deny list still excludes those model IDs
- [ ] Group `model_access` selected mode: Efficient should not pick a model the member cannot use
- [ ] Personal / non-enterprise requests are unchanged


apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -309,42 +311,49 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
...
+          // for unauthenticated requests: auto-routed models resolve to a
+          // paid model, so an unauthenticated caller is rejected downstream
+          // regardless, and a null decision simply falls back to balanced.
+          // This stops anonymous or abusive traffic from repeatedly spending
+          // Kilo-funded classification with no user to attribute it to.
+          if (!user || authFailedResponse) return null;
+          const { settings, plan } = await balanceAndSettingsPromise;
+          const groupPolicy = await organizationGroupPolicyPromise;
+          const deniedFromPolicy = groupPolicy
+            ? await collectDeniedAutoRoutingModelIds(groupPolicy)
```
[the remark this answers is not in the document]
Comment (chrarnoldus) [reply apps/web/src/app/api/openrouter/[...path]/route.ts]: Valid. Policy load/evaluation is now try/caught in the efficient-decision thunk and falls back to the enterprise deny list so resolution still degrades to balanced. Added a route test that a rejected collectDenied still returns 200.

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -309,42 +311,57 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
   let classifierCostUsd = 0;
   if (isKiloAutoModel(requestedModelLowerCased)) {
     autoModel = requestedModelLowerCased;
-    const efficientDecision =
+    const usesEfficientDecision =
```
Comment (chrarnoldus) [inline apps/web/src/app/api/openrouter/[...path]/route.ts]: rename to isAutoEfficientId

apps/web/src/app/api/openrouter/[...path]/route.ts
```diff
@@ -309,42 +311,57 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
...
+          // paid model, so an unauthenticated caller is rejected downstream
+          // regardless, and a null decision simply falls back to balanced.
+          // This stops anonymous or abusive traffic from repeatedly spending
+          // Kilo-funded classification with no user to attribute it to.
+          if (!user || authFailedResponse) return null;
+          const { settings, plan } = await balanceAndSettingsPromise;
+          const deniedFromSettings =
+            plan === 'enterprise' ? (settings?.model_deny_list?.map(normalizeModelId) ?? []) : [];
+          let deniedFromPolicy: string[] = [];
+          try {
```
Comment (chrarnoldus) [inline apps/web/src/app/api/openrouter/[...path]/route.ts]: remove try/catch

apps/web/src/lib/ai-gateway/auto-routing-denied-models.ts
```diff
@@ -0,0 +1,79 @@
...
+import { MINIMAX_CURRENT_MODEL_ID } from '@/lib/ai-gateway/providers/minimax';
+import {
+  getEffectiveModelDecision,
+  type EffectiveOrganizationModelPolicy,
+} from '@/lib/organizations/effective-model-access.server';
+
+// Keep in sync with services/auto-routing/src/coding-plan-preference.ts.
+const BYTEPLUS_CODING_PLAN_DEFAULT_MODEL_ID = 'byteplus-coding/bytedance-seed-code';
+
+const ALWAYS_CONSIDERED_MODEL_IDS = [
```
chrarnoldus: rename to indicate purpose better and remove unnecessary entries

apps/web/src/app/api/openrouter/[...path]/route.ts
Comment (chrarnoldus) [reply apps/web/src/app/api/openrouter/[...path]/route.ts]: Renamed to `isAutoEfficientId`.

apps/web/src/app/api/openrouter/[...path]/route.ts
Comment (chrarnoldus) [reply apps/web/src/app/api/openrouter/[...path]/route.ts]: Removed. Policy evaluation is awaited directly again.

apps/web/src/lib/ai-gateway/auto-routing-denied-models.ts
chrarnoldus: Renamed to `CODING_PLAN_DEFAULT_MODEL_IDS` and dropped Qwen. Only MiniMax and BytePlus remain — `/decide` can short-circuit to those without the routing table.

apps/web/src/lib/ai-gateway/auto-routing-denied-models.ts
```diff
@@ -0,0 +1,76 @@
...
+  const denied = new Set(
+    (policy.organizationModelDenyList ?? []).map(modelId => normalizeModelId(modelId))
+  );
+  if (!policyNeedsCandidateEvaluation(policy) && options.candidateModelIds === undefined) {
+    return [...denied];
+  }
+
+  const candidateIds =
+    options.candidateModelIds ??
+    (await (options.loadCandidateModelIds ?? loadAutoRoutingCandidateModelIds)());
```
chrarnoldus: **[P1] Evaluate the effective custom pool, not only the platform table.** `/decide` replaces the platform table with `loadCustomRoutingTable(...)` whenever the effective org/personal settings contain a pool, but this loader only returns platform-table candidates plus coding-plan defaults. An organization pool is validated against default org access rather than the requesting member’s group grant, so a custom-pool-only model excluded by that member’s selected grant is absent from `deniedModelIds`, can be selected here, and is then rejected by the gateway’s final policy check. Please project the policy over the actual effective pool (or enforce it in `/decide` against the table being used) and cover a selected member grant with a custom-pool-only model.

apps/web/src/lib/ai-gateway/auto-routing-denied-models.ts
```diff
@@ -0,0 +1,76 @@
...
+  const decide = options.decideModel ?? getEffectiveModelDecision;
+  const decisions = await Promise.all(
+    [...new Set(candidateIds.filter(id => !isVirtualAutoModelId(id)))].map(async modelId => ({
+      modelId: normalizeModelId(modelId),
```
chrarnoldus: **[P1] Preserve the exact candidate ID sent to `/decide`.** The policy is evaluated against the original ID, but a denied candidate such as `deepseek/deepseek-v4-pro:discounted` is recorded as `deepseek/deepseek-v4-pro`. The worker filters with exact `deniedModelIds.has(c.model)`, so the suffixed candidate remains eligible and can still produce the late 403 this PR is intended to prevent. This also matters because suffix-specific exclusive-provider policy can deny the discounted route while allowing the base catalog model; emitting the normalized ID can filter the wrong candidate. Keep evaluated candidate denials exact, and separately expand normalized organization deny rules to matching candidate IDs as needed.

chrarnoldus (commented):
Two blocking policy-projection gaps remain. The worker can select models from a configured custom pool that this helper never evaluates, and exact suffixed candidate IDs are lost before they reach the worker.

apps/web/src/lib/ai-gateway/auto-routing-denied-models.ts
chrarnoldus: Valid. Candidate collection now uses the effective org/personal pool when one is configured (org pool wins), otherwise the platform table. Added a selected-grant + custom-pool-only model test.

apps/web/src/lib/ai-gateway/auto-routing-denied-models.ts
chrarnoldus: Valid. Evaluated denials now keep the exact candidate ID (`:discounted` stays). Normalized deny-list entries are still recorded and also expanded onto matching suffixed candidates.

apps/web/src/lib/ai-gateway/auto-routing-denied-models.ts
```diff
@@ -0,0 +1,133 @@
...
+  policy: EffectiveOrganizationModelPolicy,
+  options: CollectDeniedAutoRoutingModelIdsOptions = {}
+): Promise<string[]> {
+  const denyList = policy.organizationModelDenyList ?? [];
+  const normalizedDeny = new Set(denyList.map(normalizeModelId));
+  const denied = new Set(normalizedDeny);
+  const shouldEvaluateCandidates =
+    policyNeedsCandidateEvaluation(policy) ||
+    options.candidateModelIds !== undefined ||
+    options.owner !== undefined;
```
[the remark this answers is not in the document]
chrarnoldus: Valid. `owner` now only triggers a candidate/pool load when there is a deny list to expand. No-op policies (non-enterprise / unrestricted, empty deny list) return `[]` without extra worker fetches.

jrf0110 (approved):

PR #5226, opened 2026-08-12
fix(web): disable R2 checksums that break cloud agent image uploads

chrarnoldus:
## Summary

Cloud agent image uploads fail in the browser with `Failed to upload file: Network error during upload`.

The composer PUTs directly to a Cloudflare R2 presigned URL. `@aws-sdk/client-s3` 3.729+ defaults to `requestChecksumCalculation: WHEN_SUPPORTED`, so `getSignedUrl` embeds `x-amz-checksum-crc32=AAAAAA==` (CRC32 of an empty body — the `PutObjectCommand` has no `Body` at sign time). The browser then uploads the real file; R2 rejects the checksum. That shows up as `xhr.onerror` rather than a readable HTTP status.

This disables flexible checksums on the shared R2 S3 client (`WHEN_REQUIRED`), matching the working Kilo Chat `aws4fetch` path.

## Test plan

- [x] `pnpm exec jest src/lib/r2/create-r2-s3-client.test.ts src/lib/r2/cloud-agent-attachments.test.ts` in `apps/web`
- [ ] Attach a PNG/JPEG in the cloud agent composer and confirm the thumbnail completes without the network-error toast
- [ ] Confirm send still includes the attachment

### DIFF
--- a/apps/web/src/lib/r2/client.ts
+++ b/apps/web/src/lib/r2/client.ts
@@ -1,5 +1,5 @@
-import { S3Client } from '@aws-sdk/client-s3';
 import { getEnvVariable } from '@/lib/dotenvx';
+import { createR2S3Client } from './create-r2-s3-client';
 
 // R2 configuration from environment variables
 const R2_ACCOUNT_ID = getEnvVariable('R2_ACCOUNT_ID');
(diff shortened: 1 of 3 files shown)


chrarnoldus:
**Could this be an expired credential instead?**

Possibly, but it would be a different 403 that the browser then hides as `xhr.onerror`.

Presigning is local HMAC — `getAttachmentUploadUrl` still returns 200 with a `signedUrl` even if `R2_ACCESS_KEY_ID` / `R2_SECRET_ACCESS_KEY` are revoked or rotated. The browser PUT is what R2 rejects (`InvalidAccessKeyId` / `SignatureDoesNotMatch`). If that 403 has no CORS headers, XHR reports status 0 → `Network error during upload`, same toast.

Same shape for a stale presigned URL (`X-Amz-Expires` is 900s). Unlikely here: the composer PUTs immediately after picking the file.

Why I still think checksums are the primary bug:

- Default AWS SDK v3 signing injects `x-amz-checksum-crc32=AAAAAA==` (CRC32 of an empty body) into every PutObject URL. That fails **every** browser upload, not only after a rotation.
- Expired/revoked keys would also break server-side R2 (CLI session blobs, experiment prompts, app-builder assets). This report is isolated to the cloud-agent composer PUT.
- R2 S3 API tokens do not expire on their own unless they were created with a TTL or revoked.

To rule credentials out in DevTools: if `getAttachmentUploadUrl` 200s and the following PUT is 403/`InvalidAccessKeyId`, rotate keys. If the PUT URL contains `x-amz-checksum-crc32=AAAAAA==`, this PR is the fix.

PR #5248, opened 2026-08-13
fix(ai-gateway): filter suffixed models by endpoint providers

chrarnoldus:
## Summary
- index suffixed OpenRouter models by their exact endpoint provider tags
- preserve unsuffixed provider associations and suffix-insensitive model deny lists
- filter organization model catalogs using exact suffixed provider availability
- add coverage for free and non-free suffixes, including `nvidia/nemotron-3.5-lightning:free`

## Test plan
- CI only, as requested
- focused lint passed locally
- `git diff --check` passed locally


apps/web/src/lib/organizations/group-policies/model-access/model-access.server.ts
```diff
@@ -140,7 +141,7 @@ export async function getEffectiveModelDecision(
         ? { allowed: true, eligibleProviderRoutes: organizationRoutes }
         : { allowed: false, denialSource: 'organization_provider' };
     }
-    const modelProviders = currentModelProviders ?? (await providerLookup(normalizedModelId));
+    const modelProviders = currentModelProviders ?? (await providerLookup(providerLookupModelId));
```
chrarnoldus, line 144: The third lookup in this function (line 167, the member-grant fallback) still uses `providerLookup(normalizedModelId)`, while this line and line 127 were switched to `providerLookupModelId`. It appears unreachable today — a `selected` member grant always comes with `requireModelInCurrentSnapshot: true`, so `currentModelProviders` is defined except for latest aliases, which take the other branch — but any policy constructed with `requireModelInCurrentSnapshot: false` plus a selected grant would evaluate suffixed models against base-model providers, reintroducing the bug this PR fixes. Suggest using `providerLookupModelId` there too for consistency.

chrarnoldus: Fixed in e34e39095. The member-provider fallback now uses `providerLookupModelId`, matching the other provider lookups. I also added a regression test that constructs the defensive `requireModelInCurrentSnapshot: false` plus selected-provider policy and verifies a `:free` model uses its exact endpoint providers.

apps/web/src/lib/organizations/effective-model-access.server.test.ts
```diff
@@ -166,6 +166,62 @@ describe('effective organization model access', () => {
...
+    const policy = {
+      requireModelInCurrentSnapshot: false,
+      organizationModelDenyList: [],
+      memberGrant: {
+        mode: 'selected',
+        modelAllowList: [],
+        providerAllowList: ['nvidia'],
+      },
+      policyRevision: 1,
+    } as const;
```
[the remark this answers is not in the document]
chrarnoldus: Fixed in 24abef0f5. The fixture now uses `satisfies EffectiveOrganizationModelPolicy` instead of `as const`, preserving literal discrimination without producing readonly arrays. Focused formatting and lint pass; refreshed CI is running on that commit.

PR #5301, opened 2026-08-17
fix(ai-gateway): remove unsupported legacy aliases

chrarnoldus:
## Summary
- remove legacy OpenRouter aliases absent from the approved model ID list
- retain all existing aliases present in the approved list
- add no new aliases

## Testing
- Not run locally; CI will handle tests as requested.


chrarnoldus:
The OpenRouter/Vercel asymmetry is intentional here. This change is scoped to `legacyOpenRouterAliases`, and the supplied approved OpenRouter model ID list excludes these aliases. The Vercel mapping has a separate routing scope, so it is left unchanged; restoring the removed OpenRouter aliases would contradict the requested allowlist.

PR #5316, opened 2026-08-18
fix(mobile): show empty message when pasting an empty clipboard

iscekic:
## Summary

Tapping paste with nothing copied shows a file-read error ("Couldn't read this file"). The paste hook reported empty, denied, and unsupported-type through one `onUnreadable` callback, so callers could not distinguish an empty clipboard from a denied or unreadable one.

The change classifies the paste failure. `useClipboardPaste` now reports `onFailure(reason)` where `reason` is `'empty'` or `'unreadable'`. The three paste call sites toast `Nothing to paste` only for the empty case; denied permission and unreadable image keep the current unreadable copy.

**User**: Tapping paste with nothing copied now shows "Nothing to paste" instead of a file-read error. The user can copy and tap again.

**Product manager**: The empty-clipboard paste case now has a clear, retryable message, distinct from denied permission and unreadable image.

**Maintainer**: `useClipboardPaste` classifies paste failure as `'empty'` vs `'unreadable'`. A new `hasClipboardUrl` helper inspects content type only (no iOS paste prompt) and returns `false` on error. Classification runs after the existing reads; `readClipboardText` is not gated on a content-type probe, so URL-only clipboards still paste.

## Verification

- Unit tests cover the new classification paths: empty, denied text, denied URL, unreadable image with and without text.
- `pnpm typecheck`, `pnpm lint`, `pnpm check:unused` pass in `apps/mobile`.
- `git diff --check` is clean.
- E2E (bot-e2e, iOS): new-session empty-clipboard paste toasts `Nothing to paste`.

## Visual Changes

The empty-clipboard paste toast copy changes from "Couldn't read this file" to "Nothing to paste". Denied and unreadable toasts are unchanged. This is a text-only change to a transient toast; no layout or styling changed. The rendered copy was verified by E2E (Appium hierarchy).

## Reviewer Notes

- The `paste` no-file branch order is the core logic; review it against the plan's step 2.
- `hasClipboardUrl` catches the Android `UnavailabilityError` and returns `false`.

## Human steps

No step is needed. No secret, migration, flag, or deploy-order change.


apps/mobile/src/lib/agent-attachments/use-clipboard-paste.ts
```diff
@@ -21,10 +24,13 @@ type UseClipboardPasteOptions = {
    *  paste with text on the clipboard, and an unreadable-image toast would be
    *  wrong there. Omit it to keep the image-only behavior. */
   addText?: (text: string) => void;
-  /** Called when neither an image nor text could be read (empty, denied, or
-   *  unsupported type). The caller supplies its own unreadable-toast copy to
-   *  match that composer's existing pick-path message. */
-  onUnreadable: () => void;
+  /** Called when neither an image nor text could be used. `'empty'` means no
```
pandemicsyn, line 27: bot: The PR description claims two new content-type-only helpers, `hasClipboardText` and `hasClipboardUrl`, but only `hasClipboardUrl` was added; there is no `hasClipboardText` anywhere in the diff or repository. The updated doc comment here (and the implementation) is internally consistent, so this is a description/code mismatch rather than a functional bug.

Suggested fix: Update the PR description to drop the `hasClipboardText` mention, or confirm whether a follow-up was meant to add it.

pandemicsyn (approved):
lgtm. just flagging that the robot pointed there might be a missing helper.

apps/mobile/src/lib/agent-attachments/use-clipboard-paste.ts
iscekic: (bot) Fixed the PR description: dropped the `hasClipboardText` mention. Only `hasClipboardUrl` was added; `hasClipboardText` was removed during E2E defect fixing because `hasStringAsync` returns true for empty-string items and cannot distinguish empty from denied.

PR #5349, opened 2026-08-18
fix(ai-gateway): remove Laguna from Auto Free

chrarnoldus:
## Summary
- remove Laguna S 2.1 from Auto Free candidate routing
- keep Laguna S 2.1 in the preferred model list
- update model-list coverage for the new behavior

## Testing
- CI only, as requested


iscekic (approved):

chrarnoldus:
seem stepfun already has trouble with current load

PR #5351, opened 2026-08-18
fix(extension): session context, cost, and model parity with mobile

iscekic:
## What

Six agents-tab fixes, for parity with the mobile app and consistency with the extension's own browser tab.

| Reported | Fix |
|---|---|
| No context or cost for a session | The open session header shows the browser tab's `ContextDonut`: context usage, percent when the model's window is known, and session cost. |
| A running session has no model picker | The composer now carries a model picker. Cloud-agent and legacy remote sessions use the gateway catalog; a v1 remote CLI session uses the catalog the CLI itself reports. |
| A new remote CLI session has no model picker | The new-session form offers the picker for a CLI target too. The pick rides `create_session`; `CLI default` leaves the CLI's own model alone. |
| Opening a session does not show the latest messages | The conversation list pins to the newest message on open, settling on virtualizer measurement instead of a fixed frame budget. |
| Weird spacing in the session list | One vertical rhythm and one empty-state treatment across Active and History, one text column for titles, sub-lines, and empty text. |
| `save_memory` should have a confirmation option | New `Auto-approve memory saves` setting in the memories section, **off by default**, so the confirmation card stays the default. |

## Notes

- **Reuse over new code.** Context usage and cost come from the existing `@kilocode/cloud-agent-sdk` atoms (`contextUsage`, `totalCost`) and `fetchedSessionData.totalCostMicrodollars`; the indicator is the browser tab's existing `ContextDonut`; the picker is the existing `ModelPicker`. Mobile's behavior is mirrored, including its rule that a mid-session model change applies to the next message and is not gated on streaming.
- **Honest states.** An unknown context window renders tokens only, never a fake percent. A CLI session with no catalog shows a disabled picker with a reason instead of a gateway model it cannot run.
- **Favorites stay clean.** `CLI default` and CLI-catalog rows are not gateway ids, so they are never sent to `modelPreferences.addFavorite` and show no favorite control.
- **Scroll fix root cause.** `virtual-core` issues corrective `scrollTo` writes on row measurement, from an offset that lags the DOM by a frame. The old code could not tell those from a user scroll, so one corrective write latched auto-scroll off for the whole mount and nothing re-pinned. The pin now compares against the live offset and drops stale corrections; a reader who scrolls up stays put.
- The dead permanently-disabled `Compact now` button is gone where compaction is impossible, and `ContextDonut` takes an explicit `placement` instead of a child-selector hack.
- `AGENTS.md` updated: `save_memory` is no longer unconditionally card-gated.

## Verification

- `pnpm --filter kilo-extension typecheck` / `lint` / `test` — clean; unit tests 89 files.
- `pnpm --filter kilo-extension e2e:chrome` — full mocked suite green.
- `pnpm --filter kilo-extension e2e:firefox` — green; one assertion narrowed, because the Memories section legitimately holds a settings toggle now.
- `pnpm --filter kilo-extension e2e:local` — run against a local stack with a real connected `kilo` CLI instance and a real cloud-agent session. The live suite has 7 failures that reproduce identically on `main` at the same commit (frontier timing, dangerous-mode eval, manual compaction, CLI reopen). Each was baselined on `main` before being ruled pre-existing; `local-backend-live.test.ts:608` was run twice per side to confirm it is flaky on both.

### DIFF
--- a/apps/extension/AGENTS.md
+++ b/apps/extension/AGENTS.md
@@ -59,7 +59,7 @@ Before committing extension changes, run `pnpm format`. Prefer `pnpm --filter ki
 
 - Safe mode exposes read tools (`get_page_snapshot`, `find_in_page`, `get_element_details`, `search_memories`, `get_memory`, and when the model supports images `get_viewport_screenshot`), workflow read tools (`search_workflows`, `get_workflow`), and card-gated tools (`save_workflow`, `save_memory`).
 - `search_workflows` without a query lists workflows scoped to the selected tab. With a query it searches every site, ranks in-scope matches first, and reports `inScope` and `startUrl` per result.
-- `save_workflow` and `save_memory` are card-gated — the executor blocks the tool turn until the user approves or rejects on the approval card, unless auto-approve workflow changes is on, which stores a workflow save with no card; `save_memory` is always card-gated.
+- `save_workflow` and `save_memory` are card-gated — the executor blocks the tool turn until the user approves or rejects on the approval card. Auto-approve workflow changes stores a workflow save with no card. Auto-approve memory saves does the same for `save_memory`; it is off by default, so the confirmation card is the default behavior. A settings read failure always falls back to showing the card.
 - `run_workflow` is gated behind the "Allow workflows in safe mode" toggle. In dangerous mode the toggle is bypassed and `run_workflow` is always available.
 - `delete_workflow` and `eval` are dangerous-mode only and never exposed in safe mode.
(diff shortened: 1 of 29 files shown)


iscekic:
@kilo-code-bot the review run failed with "Assistant request failed". Please re-run the review on the latest commit.

eshurakov (approved):
Session context/cost, model picker, scroll pin, and memory auto-approve look sound. Serial save chain addresses the overlapping-toggle case. CI green including e2e-chrome.

PR #5359, opened 2026-08-19
fix(mobile): keep transcript row gaps uniform around blank text parts

iscekic:
Follow-up to #5318, which did not fix the reported gap.

## What changed

A whitespace-only assistant text part no longer counts as transcript content, so the gap between two neighboring transcript rows is always one gap.

## User note

The gap above and below a collapsed `THOUGHT` row now matches the gap between two tool rows. Nothing else in the transcript moves.

## Product note

Row rhythm is uniform again: every adjacent row pair in a session transcript sits 8pt apart, whatever the hidden parts between them are.

## Maintainer note

Root cause is not the dashed box that #5318 changed. `partRendersContent` accepted any text part with `text !== ''`, so a whitespace-only text part (models emit `"\n\n"` between tool calls) passed as content:

- inside one message it rendered a zero-height markdown row, which consumed one of the `gap-2` gaps in `MessageBubble` — the visible gap doubled to 16pt;
- alone in a message it kept `messageRendersContent` true, so `mergeSessionTranscript` kept the message and its `px-4 py-1` wrapper added the same 8pt.

`partRendersContent` now trims, matching the reasoning-part rule next to it, and `TextPartRenderer` trims its own guard so the two stay in step.

## Verification

Local e2e on an iPhone 17 Pro simulator (real dev build, local stack, CLI session seeded through `session-ingest`), pixel-measured from `simctl` screenshots:

| boundary | before | after |
|---|---|---|
| tool row → tool row | 21px | 21px |
| tool row → `THOUGHT` row | 21px | 21px |
| across a blank text part | 42px | 21px |
| across a blank-text-only message | 42px | 21px |

`pnpm format && pnpm typecheck && pnpm lint && pnpm check:unused` and `pnpm test` (4916 tests) pass in `apps/mobile`.

## Human steps

No human step is needed.

### DIFF
--- a/apps/mobile/src/components/agents/message-visibility.test.ts
+++ b/apps/mobile/src/components/agents/message-visibility.test.ts
@@ -88,6 +88,11 @@ describe('partRendersContent', () => {
     expect(partRendersContent(textPart({ text: '' }))).toBe(false);
   });
 
+  it('returns false for a whitespace-only text part', () => {
+    expect(partRendersContent(textPart({ text: '\n\n' }))).toBe(false);
+    expect(partRendersContent(textPart({ text: '   ' }))).toBe(false);
+  });
+
   it('returns true for a bash tool part', () => {
     expect(partRendersContent(toolPart('bash'))).toBe(true);
   });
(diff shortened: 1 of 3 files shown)


eshurakov (approved):
Looks correct. Whitespace-only text parts no longer count as content, matching the renderer. Tests cover the cases.

PR #5361, opened 2026-08-19
fix(mobile): match linked-account rows to other profile cards

iscekic:
## What
Linked-account rows on the mobile profile screen now use the same `ConfigureRow` card as Agents, Reviews, Organization, and App.

## Why
The WorkOS (and every other provider) row used a custom layout: a bare key icon and the raw provider id (`workos`). That broke the visual rhythm of the rest of the screen.

## How
- Render each linked provider with `ConfigureRow` (tinted icon tile, title, email subtitle).
- Map known provider ids to the same display names as web (`Enterprise SSO`, `GitHub`, `Google`, …).
- Leave the row inert (no chevron): there is no linked-account detail screen on mobile.

## Visual
Before: bare icon + capitalized slug, no tile.
After: boxed tinted icon + proper name + email, same card chrome as the other rows.

### DIFF
--- a/apps/mobile/src/components/profile-screen.tsx
+++ b/apps/mobile/src/components/profile-screen.tsx
@@ -43,8 +43,21 @@ import {
 import { getSecurityAgentPath } from '@/lib/security-agent';
 import { useTRPC } from '@/lib/trpc';
 
-function providerIcon(_provider: string) {
-  return KeyRound;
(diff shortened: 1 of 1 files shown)


eshurakov (approved):
Linked-account rows now use ConfigureRow and the same provider labels as web. Inert (no chevron) is the right call.

PR #5385, opened 2026-08-19
feat(dev): prompt for a base URL in setup-env, defaulting to unset

IamCoder18:
## Summary

`pnpm dev:setup-env` now prompts for a `BASE_URL` ("the URL you use to access the dev server") at the start, before the `NEXTAUTH_URL` prompt. When provided, it becomes the default for `NEXTAUTH_URL` (which the user can still override) and `APP_URL_OVERRIDE` is also written so auth and server-side redirects resolve at the same public origin — useful for LAN testing (`http://192.168.1.82:3100`) and dev URLs behind Cloudflare Access (`https://clouddev.example.com`). Pressing enter leaves both unset, so `NEXTAUTH_URL` falls back to the `http://localhost:3000` example default. CI mode (`--ci`) skips the `BASE_URL` prompt and preserves the existing CI placeholders.

Both prompts share a URL validator with a non-fatal recovery menu (no fatal exit):
- Rejects whitespace/quotes/newlines/`#` (env-reserved), unparseable input, non-http/https protocol, missing hostname.
- Suggests a fix when the input is missing a protocol (e.g. `192.168.1.82:3100` → `http://192.168.1.82:3100`).
- On invalid input: prints the error, offers `[r] Re-enter` / `[s] Use suggested` (when available) / `[d] Use default or clear`. Enter defaults to `d`.

`APP_URL_OVERRIDE` is appended to `.env.local` since it isn't in `.env.local.example`; the file-write path already handles that.

## Verification

- `pnpm exec oxlint dev/local/setup-env.ts` — 0 warnings, 0 errors.
- `pnpm exec tsc --noEmit --target es2022 --module nodenext --moduleResolution nodenext --strict --skipLibCheck dev/local/setup-env.ts` — clean.
- Manual smoke test of the `validateUrl` helper via `node -e` against 10 inputs covering valid LAN/Cloudflare-Access/localhost URLs, protocol-less host:port (suggests `http://` prefix), wrong protocol (`ftp:`), garbage strings, whitespace, and `#`-fragment inputs. All produce the expected `ok`/`error`/`suggestion` result.
- No automated tests exist for `dev/local/setup-env.ts` in the repo, so the prompt flow itself was not exercised end-to-end.
- Pre-push hook (`pnpm format:check`, `pnpm lint`, `pnpm typecheck --changes-only`) ran clean on push.

## Visual Changes

N/A

## Reviewer Notes

- `validateUrl` returns the user's exact input (not `parsed.origin`) on success, so what you type is what gets written to `.env.local`. The runtime already derives `.origin` when it reads `APP_URL_OVERRIDE` / `NEXTAUTH_URL` via `new URL(...).origin`, so a trailing path is harmless — but I chose not to silently rewrite the value. Happy to normalize to origin (reject paths/queries) if you'd rather enforce the strict origin form.
- The "use default / clear" recovery option always returns `''`. The caller maps `''` to the appropriate fallback: empty for `BASE_URL` (unset), the shown default for `NEXTAUTH_URL` (base URL or `http://localhost:3000`). So one recovery path serves both prompts.
- The validator's "suggest `http://`" path only fires when the input contains no `://`. It does not try to fix typos like `htttp://` or missing single slashes — those surface as a parse error with a re-enter prompt.

### DIFF
--- a/dev/local/setup-env.ts
+++ b/dev/local/setup-env.ts
@@ -216,6 +216,92 @@ async function promptForValue(
   });
 }
 
+// Prompts for a URL and validates it. On invalid input, offers a non-fatal
+// recovery menu: re-enter, use a suggested fix (when available), or use the
+// default / clear. Returns the validated URL string, or '' to mean
+// "use the default / clear" — callers map '' to the appropriate fallback.
+async function promptForUrl(args: {
+  key: string;
+  defaultValue: string;
+  description?: string;
+}): Promise<string> {
+  while (true) {
+    const raw = await promptForValue(args.key, args.defaultValue, args.description, false);
+    if (raw === '') return '';
+    const result = validateUrl(raw);
+    if (result.ok) return result.value;
+    if (result.suggestion) {
+      // Make sure the suggestion itself is valid before offering it.
+      const suggestionCheck = validateUrl(result.suggestion);
+      if (!suggestionCheck.ok) delete result.suggestion;
+    }
+    console.log(`  ${RED}✗ ${args.key}: ${result.error}${RESET}`);
+    if (result.suggestion) {
+      console.log(`  ${YELLOW}Suggested: ${result.suggestion}${RESET}`);
+    }
+    console.log('  Options:');
+    console.log('    [r] Re-enter');
+    if (result.suggestion) console.log('    [s] Use suggested');
+    console.log('    [d] Use default / clear');
+    const choice = await promptForUrlRecoveryChoice(Boolean(result.suggestion));
+    if (choice === 'r') continue;
+    if (choice === 's' && result.suggestion) return result.suggestion;
+    return '';
+  }
+}
+
+async function promptForUrlRecoveryChoice(hasSuggestion: boolean): Promise<'r' | 's' | 'd'> {
+  const hint = hasSuggestion ? '[r/s/d]' : '[r/d]';
+  const rl = readline.createInterface({ input: process.stdin, output: process.stdout });
+  return new Promise(resolve => {
+    rl.question(`  Choice ${hint} (default d) > `, answer => {
+      rl.close();
+      const a = answer.trim().toLowerCase();
+      if (a === 'r' || a === 're-enter') return resolve('r');
+      if (hasSuggestion && (a === 's' || a === 'suggested')) return resolve('s');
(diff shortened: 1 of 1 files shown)


IamCoder18:
Ready for review!

dev/local/setup-env.ts
```diff
@@ -216,6 +216,92 @@ async function promptForValue(
...
+      suggestion,
+    };
+  }
+  if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
+    return { ok: false, error: `protocol must be http or https (got "${parsed.protocol}")` };
+  }
+  if (!parsed.hostname) {
+    return { ok: false, error: 'URL is missing a hostname' };
+  }
+  return { ok: true, value: raw };
```
pandemicsyn: bot: Normalize or reject non-origin URL components before persisting the value.

Suggested fix: `validateUrl` currently returns `raw`, so an accepted base URL such as `https://dev.example.test?tenant=a` is written verbatim to `NEXTAUTH_URL`. `APP_URL_OVERRIDE` is normalized by `resolveAppUrl` to `https://dev.example.test`, but direct consumers such as `apps/web/src/lib/stripe/index.ts:1656` append a path to raw `process.env.NEXTAUTH_URL`, yielding a malformed callback URL (the appended path becomes part of the query). Return `parsed.origin` on success, or reject credentials, pathname, and search explicitly, and use that normalized value for both persisted variables.

pandemicsyn: bot: Reject or escape JavaScript replacement tokens in accepted URLs.

Suggested fix: The validator permits `$`, and the URL is later inserted into the string replacement passed to `finalContent.replace()` at lines 429–431. For example, an accepted `http://localhost:3000/$&` makes `$&` expand to the matched `NEXTAUTH_URL=...` line, corrupting the generated setting. Prefer the function form of `replace`, e.g. `finalContent.replace(regex, (_match, prefix) => `${prefix}${formatValue(value)}`)`, so user values are treated literally.

pandemicsyn (changes requested):
couple  of relative minor things inline

IamCoder18:
Addressed both review issues and verified the fix.

IamCoder18:
Addressed the Copliot review too.

PR #5388, opened 2026-08-20
fix(web): return to Cloud Agent setup from integration detail

IamCoder18:
## Summary

When Cloud Agent onboarding shows the "Connect a repository provider" panel (e.g. after a user lands on /cloud without a connected repo), picking GitHub/GitLab/Bitbucket drops the user on the integration detail page whose Back link is hard-coded to /integrations. The user lands somewhere they did not intend instead of returning to the Cloud Agent setup flow.

Two pieces conspired:

- `RepositoryProviderOnboardingPanel` built provider links as plain `/integrations/<platform>` with no `returnTo` query param.
- `UserIntegrationDetailPage` in `IntegrationDetailPage.tsx` ignored `returnTo` entirely; only the organization variant of the same page already honored it.

This change passes a `returnTo` path through the onboarding panel (`/cloud` for personal, `/organizations/<id>/cloud` for org) and updates the user variant of the integration detail page to validate and use it, mirroring the organization variant. The validated `returnTo` is also threaded into the rendered detail component's search params so downstream branches (e.g. `GitHubIntegrationDetails`' `appReturnPath`) keep working for the OAuth round-trip.

Files touched:
- `apps/web/src/components/cloud-agent-next/RepositoryProviderOnboardingPanel.tsx`
- `apps/web/src/components/cloud-agent-next/NewSessionPanel.tsx`
- `apps/web/src/components/integrations/IntegrationDetailPage.tsx`

## Verification

- [x] Manually walk through: visit /cloud without a connected repo -> pick GitHub -> click "Back". Confirm it now returns to /cloud instead of /integrations. Repeat for `/organizations/<id>/cloud`.

Local checks run:
- `oxlint` on the three changed files: 0 warnings, 0 errors.
- `tsgo --noEmit` (apps/web): clean.

## Visual Changes

image

## Reviewer Notes

- `withReturnTo` is a small inline helper to append `returnTo` without breaking callers that already had query strings; today only the generated integration hrefs use it, but the helper is defensive.
- `validateReturnPath` already rejects non-`/` paths, unsafe characters, and double-slash tricks, so the Back href is constrained the same way as the organization variant.
- The pre-existing `GitHubIntegrationDetails` `appReturnPath` plumbing was already `returnTo`-aware but the page never set it; that gap is closed here.

### DIFF
--- a/apps/web/src/components/cloud-agent-next/NewSessionPanel.tsx
+++ b/apps/web/src/components/cloud-agent-next/NewSessionPanel.tsx
@@ -1201,6 +1201,7 @@ export function NewSessionPanel({ organizationId, isDevcontainerAvailable }: New
             organizationId={organizationId}
             isCheckingConnection={isCheckingConnection}
             onCheckConnection={() => void checkSourceControlConnection()}
+            returnTo={organizationId ? `/organizations/${organizationId}/cloud` : '/cloud'}
           />
         </div>
       </div>
--- a/apps/web/src/components/cloud-agent-next/RepositoryProviderOnboardingPanel.tsx
+++ b/apps/web/src/components/cloud-agent-next/RepositoryProviderOnboardingPanel.tsx
@@ -32,6 +32,7 @@ type RepositoryProviderOnboardingPanelProps = {
   organizationId?: string;
   isCheckingConnection?: boolean;
   onCheckConnection: () => void;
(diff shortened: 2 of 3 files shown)


eshurakov (approved):

IamCoder18:
All checks pass

PR #5391, opened 2026-08-20
fix(ai-gateway): forward LongCat user ID header

chrarnoldus:
## Summary
- forward the gateway hashed user ID to LongCat via `Mt-User-Id`
- add focused coverage for the provider request transform

## Verification
- `pnpm --filter web test -- --runInBand src/lib/ai-gateway/providers/provider-definitions.test.ts`
- `pnpm --filter web lint`
- `git diff --check`


WebReflection (approved):
simple fix and well tested, not sure why last CI failed but looks good to me.

marius-kilocode (approved):

Issue #1001, opened 2026-03-10
Git commit co-authorship Ã¢ human as Author, agent as Committer


jrf0110:
## Parent

Part of #204 (Phase 4: Hardening)

## Problem

Agent commits currently show `Toast (gastown) <toast@gastown.local>` as both author and committer (`agent-runner.ts:84-87`). There's no trace of the human who initiated the work. On GitHub/GitLab, these commits look like they came from a bot Ã¢ the human gets no contribution credit, and there's no provenance trail for compliance ("who authorized this code?").

## Solution

Dual-attribute every agent commit: the **human** who initiated the work is the `Author`, the **agent** is the `Committer`. This way GitHub contribution graphs credit the human, while the agent identity remains traceable.

### Git environment variables

Currently in `agent-runner.ts:84-87`:
```ts
GIT_AUTHOR_NAME: `${request.name} (gastown)`,
GIT_AUTHOR_EMAIL: `${request.name}@gastown.local`,
GIT_COMMITTER_NAME: `${request.name} (gastown)`,
GIT_COMMITTER_EMAIL: `${request.name}@gastown.local`,
```

Change to:
```ts
GIT_AUTHOR_NAME: request.userName,          // "John Smith"
GIT_AUTHOR_EMAIL: request.userEmail,        // "john@company.com"
GIT_COMMITTER_NAME: `${request.name} (Polecat)`,
GIT_COMMITTER_EMAIL: `${request.name}@gastown.kilo.ai`,
```

### Data flow

The user's name and email need to flow from the initiating action through to the agent:

1. **Mayor message or `gt_sling`**: The `userId` of the human is already on the request. Resolve their name/email from the Kilo user record.
2. **Bead creation**: Store `initiated_by_user_id` on the bead (or use the existing `created_by` field if it stores the userId).
3. **Agent dispatch**: `startAgentInContainer()` passes `userName` and `userEmail` to the container alongside the existing `agentName`.
4. **Container**: `agent-runner.ts` sets `GIT_AUTHOR_*` to the human, `GIT_COMMITTER_*` to the agent.

### Co-authored-by trailers

In addition to the Author/Committer split, append trailers to commit messages:

```
Co-authored-by: Toast (Polecat) <toast@gastown.kilo.ai>
Requested-by: John Smith <john@company.com>
Bead-ID: gt-abc-123
```

This requires either:
- Instructing the polecat's system prompt to include trailers in commit messages
- Or post-processing commits via a git hook in the worktree

[...]

PR #5397, opened 2026-08-20
feat(admin): finish Cloud GDPR deletion catalog v2

eshurakov:
## Summary

Move remaining CSA-owned GDPR deletion work onto the Cloud admin queue.

- Freeze catalog v1; new requests use catalog v2 (`pylon_finalize`, `csa_support_db`).
- Ticket-only intake: customer email resolves at preflight from the Pylon ticket.
- Pylon reply, tag `delete-complete`, close the ticket, then delete the contact.
- Cloud calls CSA to scrub the support DB after the Pylon contact is gone.
- PostHog uses environment APIs with a 3-attempt verify cap, then succeeds unconfirmed.
- Local `deletion-mock` covers PostHog, Pylon, Substack, Customer.io, and CSA.

## Verification

- `pnpm format:changed` and `git diff --check`
- oxlint on changed GDPR/deletion-mock files: clean
- `pnpm --filter web exec tsgo --noEmit` and `pnpm --filter @kilocode/db exec tsgo --noEmit`
- Jest (161/161) for changed deletion-queue, router, and GDPR-removal tests
- deletion-mock + services tests: 46/46
- Manual local E2E against deletion-mock: happy path `#1001` completed 12/12; 429/5xx/expired-Substack/missing-user/`delete_ready_missing` parked as designed

## Visual Changes

Admin deletion queue intake is visible. New v2 steps: Pylon finalize and CSA support DB.


eshurakov:
Addressed the Pylon subdomain relay warning.

`isInternalOrRelayEmail` now uses `includes('service.usepylon.com')`, matching `pylon-reply`. Ticket-only preflight skips `user@mail.service.usepylon.com` as a requester and walks the thread for a customer email.

`93eebb407`

apps/web/src/app/api/internal/support/users/gdpr-removal/route.ts
```diff
@@ -104,6 +104,7 @@ export async function POST(request: NextRequest): Promise<NextResponse> {
     const [result] = await enqueueUserDeletionTargets({
       actor: { kiloUserId: null, email: actorEmail },
       targets: [{ email: user.google_user_email, trustedUserId: user.id }],
+      catalogVersion: 1,
```
intentionally-left-nil, line 107: What do catalogVersions mean in this context? Should this be 1 or 2?

apps/web/src/lib/user/deletion-queue/handlers/csa-support-db.ts
```diff
@@ -0,0 +1,57 @@
...
+  const secret = getEnvVariable('SUPPORT_API_SECRET').trim();
+  if (!secret) return configurationMissing();
+
+  const baseUrl = getEnvVariable('CSA_APP_BASE_URL').trim().replace(/\/$/, '');
+  if (!baseUrl) {
+    return { kind: 'retry', errorCode: 'csa_base_url_missing', httpStatusClass: 'error' };
+  }
+
+  const headers: Record<string, string> = {
+    Authorization: `Bearer ${secret}`,
```
intentionally-left-nil, line 29: I checked, and if deletionFetch fails, there are no errors thrown and the bearer secret ought not to be in logfiles

apps/web/src/lib/user/deletion-queue/handlers/csa-support-db.ts
```diff
@@ -0,0 +1,57 @@
...
+  requireTargetEmail,
+  type DeletionHandler,
+} from '@/lib/user/deletion-queue/handlers/common';
+
+export const handleCsaSupportDb: DeletionHandler = async ({ request, context }) => {
+  const stop = continueIfLowTime(context);
+  if (stop) return stop;
+
+  const emailOrOutcome = requireTargetEmail(request);
+  if (typeof emailOrOutcome !== 'string') return emailOrOutcome;
```
intentionally-left-nil, line 18: Without trying to get in and suggest a whole new way of determining codeflows and error handlings - this type of checking is the result of overloading the return type, as opposed to e.g. chaining promises, or other exception types. 

My first instinct would have been to check for a generic Outcome type, but since these have no hierarchy, you can't do e.g. an instanceof check.

Anyways, it's not worth redoing this whole code around how requireTargetEmail (and all of these outcome patterns work), but it's one of the first things I noticed

apps/web/src/lib/user/deletion-queue/handlers/csa-support-db.ts
```diff
@@ -0,0 +1,57 @@
...
+  configurationMissing,
+  continueIfLowTime,
+  deletionFetch,
+  isRecord,
+  readJsonUnknown,
+  requireTargetEmail,
+  type DeletionHandler,
+} from '@/lib/user/deletion-queue/handlers/common';
+
+export const handleCsaSupportDb: DeletionHandler = async ({ request, context }) => {
```
intentionally-left-nil, line 13: Given that this function returns bare objects, can we specify a return type to catch any future tying errors?

apps/web/src/lib/user/deletion-queue/handlers/posthog.ts
```diff
@@ -1,82 +1,129 @@
...
+  return `${host}/api/environments/${encodeURIComponent(environmentId)}`;
+}
+
 function collectEmails(person: Record<string, unknown>): string[] {
   const emails = new Set<string>();
   const properties = isRecord(person.properties) ? person.properties : null;
-  const propertyEmail = properties ? properties.email : undefined;
-  if (typeof propertyEmail === 'string' && propertyEmail) emails.add(propertyEmail);
-  if (typeof person.name === 'string' && person.name.includes('@')) emails.add(person.name);
+  for (const value of [properties?.email, properties?.$email]) {
```
intentionally-left-nil, line 65: Oof why is there a `.email` and a `.$email`

apps/web/src/lib/user/deletion-queue/handlers/posthog.ts
```diff
@@ -1,82 +1,129 @@
...
     if (typeof value === 'string' && value.includes('@')) emails.add(value);
   }
   return [...emails];
 }
 
-function parsePersons(payload: unknown): PosthogPerson[] | null {
-  if (!isRecord(payload) || !Array.isArray(payload.results)) return null;
-  const persons: PosthogPerson[] = [];
+function parsePersons(payload: unknown): ParsePersonsResult {
+  if (!isRecord(payload) || !Array.isArray(payload.results)) return { kind: 'incomplete' };
```
intentionally-left-nil, line 76: Is this expected? Would we want any kind of error logging to determine what happened?

apps/web/src/lib/user/deletion-queue/handlers/posthog.ts
```diff
@@ -1,82 +1,129 @@
...
-      : [];
-    persons.push({
-      id,
-      emails: collectEmails(entry),
-      distinctIds: distinctIds.length > 0 ? distinctIds : [id],
-    });
+    if (!isRecord(entry)) continue;
+    const uuid = typeof entry.uuid === 'string' && entry.uuid.length > 0 ? entry.uuid : null;
+    if (!uuid) {
+      missingUuid = true;
```
intentionally-left-nil, line 83: This is another case we could not delete emails if the uuid is a different type. Would it be better to error out here instead?

apps/web/src/lib/user/deletion-queue/handlers/posthog.ts
```diff
@@ -91,6 +138,186 @@ async function saveProgress(requestId: string, progress: UserDeletionTaskProgres
...
+      )
+    ),
+  ];
+  const ids = new Set<string>();
+  for (const email of extraEmails) {
+    const users = (
+      await db
+        .select({ id: kilocode_users.id, blocked_reason: kilocode_users.blocked_reason })
+        .from(kilocode_users)
+        .where(eq(sql`lower(${kilocode_users.google_user_email})`, normalizeDeletionEmail(email)))
```
intentionally-left-nil, line 174: Do we normalize the email in any other way besides lowercasing it?

apps/web/src/lib/user/deletion-queue/handlers/posthog.ts
```diff
@@ -91,6 +138,186 @@ async function saveProgress(requestId: string, progress: UserDeletionTaskProgres
...
+
+    const statusUrl = new URL(`${base}/persons/deletion_status/`);
+    statusUrl.searchParams.set('person_uuid', personId);
+    statusUrl.searchParams.set('status', 'all');
+    const statusPoll = await deletionFetch(context, statusUrl.toString(), { headers });
+    if ('outcome' in statusPoll) return statusPoll;
+    if (!statusPoll.response.ok) return { outcome: classifyPosthogResponse(statusPoll.response) };
+    const body = await readJsonUnknown(statusPoll.response);
+    if (!isRecord(body) || !Array.isArray(body.results)) {
+      return { pending: true };
```
intentionally-left-nil, line 272: Why is this case pending, but not an error?

intentionally-left-nil (commented):
I only got as far along as pylon-client

apps/web/src/app/api/internal/support/users/gdpr-removal/route.ts
eshurakov: v1 is the frozen original catalog. v2 is current (`USER_DELETION_CATALOG_VERSION`) and adds Posthog, Substack, pylon_finalize, and csa_support_db. This CSA support route stays on 1 so CSA-initiated deletes do not run Cloud-owned v2 steps, including calling CSA back. Admin enqueue uses 2.

So currently this flow is split into two parts with one being on CSA (customer support website) and the other in Cloud admin. And I'm moving it all to Cloud admin. In the end after everything is done there will be no catalog version needed.

apps/web/src/lib/user/deletion-queue/handlers/csa-support-db.ts
eshurakov: Yes. `deletionFetch` classifies the failure and does not log request headers, so the bearer should not land in logs.

apps/web/src/lib/user/deletion-queue/handlers/csa-support-db.ts
eshurakov: Same union as the other handlers (`string | DeletionHandlerOutcome`). Not changing that pattern in this PR.

apps/web/src/lib/user/deletion-queue/handlers/csa-support-db.ts
eshurakov: Already typed as `DeletionHandler`, which is `Promise<DeletionHandlerOutcome>`. The bare objects have to match that.

apps/web/src/lib/user/deletion-queue/handlers/posthog.ts
eshurakov: PostHog reserved person property is `$email`; some payloads also set `email`. We collect both.

apps/web/src/lib/user/deletion-queue/handlers/posthog.ts
eshurakov: `parsePersons` only classifies. `lookupPersons` maps `incomplete` to `needs_attention` / `posthog_lookup_incomplete`, so it does surface.

apps/web/src/lib/user/deletion-queue/handlers/posthog.ts
eshurakov: Non-string or missing `uuid` already fails the whole parse as `missing_uuid` → `manual_action_required`. The `continue` is so we still scan the rest of the page before returning that.

apps/web/src/lib/user/deletion-queue/handlers/posthog.ts
eshurakov: Trim and lowercase only, same as `normalizeDeletionEmail` everywhere else. No plus-tag or Gmail-dot folding.

apps/web/src/lib/user/deletion-queue/handlers/posthog.ts
eshurakov: Malformed or empty `deletion_status` while PostHog is still processing is treated as not-yet-verified. We retry, then after `USER_DELETION_POSTHOG_MAX_VERIFY_ATTEMPTS` succeed unconfirmed instead of blocking the queue.

The most important part is that we make sure that deletion request was accepted. Posthog deletion is an async process and we don't want to wait forever (it really takes a while) till they actually delete it, so we try a few times and then mark it as success.

pandemicsyn (approved):

PR #5412, opened 2026-08-21
feat(mobile): preview first-message cloud-agent attachments

iscekic:
## Summary

- Tapping a Markdown file that a cloud agent sends now opens the preview with the file's content.
- Tapping an image file that a cloud agent sends now shows the image inline and opens the full-screen viewer.
- Both previews still work after leaving and reopening the session.
- A file that fails to load shows an error with a Retry button that fetches a fresh link.
- An empty file shows "This file is empty."
- Small inline images sent as file parts now render in the live stream.

---

The live stream no longer blanks every file-part URL. It now keeps `file://` sandbox paths, `http(s)` URLs, and `data:` URLs up to 512 KiB on top-level file parts, and it still strips larger `data:` URLs and the source text. The new `MAX_INLINE_FILE_URL_LENGTH` limit is half the ingest-frame byte budget, so one inline file cannot alone overflow a frame.

<details>
<summary>Files</summary>

- `services/cloud-agent-next/src/shared/trim-payload.ts` — adds `MAX_INLINE_FILE_URL_LENGTH`; `stripFilePartFields` keeps non-`data:` URLs and small top-level `data:` URLs, strips large `data:` URLs, and still blanks `source.text.value`.

</details>

A new parser turns the wrapper's sandbox attachment URL into a `CloudAgentAttachmentRef` with the message UUID and filename. The pattern accepts only the exact sandbox path shape, one UUID segment followed by one filename segment, so unrelated file URLs do not match. History keeps this URL form forever, so the parser must handle both live events and replays.

<details>
<summary>Files</summary>

- `apps/mobile/src/components/agents/file-part-preview.ts` — adds the `CloudAgentAttachmentRef` type and `parseCloudAgentAttachmentUrl`.

</details>

The cache entry contract changes: `url` becomes optional, and `attachmentRef` plus `resolveFailed` join the shape. A sandbox attachment URL now stores a reference-only entry with no URL key, and three new helpers overwrite the URL, mark a failed presign, and clear that mark. First write still wins, and `data:` URLs still write to disk.

<details>
<summary>Files</summary>

- `apps/mobile/src/components/agents/file-part-cache.ts` — makes `url` optional, adds `attachmentRef` and `resolveFailed`, adds `overwriteFilePartCacheEntry`, `markFilePartResolveFailed`, and `clearFilePartResolveFailed`, and changes `useFilePartCache` to snapshot the entry object.

</details>

The new `useResolvedFilePartUrl` hook owns URL resolution for a file part. It returns a captured `http(s)` or `data:` URL as-is and presigns a sandbox attachment through the existing `getAttachmentDownloadUrl` endpoint on demand. A module-level set dedupes in-flight presigns by part id, and `refreshFilePartUrl` re-presigns and returns false instead of throwing.

<details>
<summary>Files</summary>

- `apps/mobile/src/components/agents/file-part-url-resolver.ts` — adds `useResolvedFilePartUrl` and `refreshFilePartUrl`; exposes `ready`, `resolving`, `unavailable`, and `error` states.

</details>

The renderer now consumes the resolver's state instead of reading the cache directly. Image parts get loading, error, and retry rows, the preview modal gets an `onRetry` prop that re-presigns before re-reading, and the chip shows a busy spinner while resolution runs.

<details>
<summary>Files</summary>

- `apps/mobile/src/components/agents/file-part-renderer.tsx` — drops the `resolveUsableUrl` helper, uses `useResolvedFilePartUrl`, adds the image states, and passes `onRetry` to the modal.

</details>

Tests: 5 files changed — apps/mobile/src/components/agents/file-part-cache.test.ts, apps/mobile/src/components/agents/file-part-preview.test.ts, apps/mobile/src/components/agents/file-part-renderer.mounted.test.tsx, services/cloud-agent-next/src/shared/ingest-frame.test.ts, services/cloud-agent-next/src/shared/trim-payload.test.ts.

Generated: none.

---

## Verification

On iOS, four scenarios ran across three verify rounds and the g4 takeover round.

|---|---|---|---|
| E1 | The live markdown tap opens the preview modal and renders the file text. | ios | passed after the fix in round g4 takeover |
| E2 | The live image tap opens the full-screen viewer. | ios | passed after the fix in round g4 takeover |
| E3 | After re-entering the session, the markdown tap still opens the preview modal. | ios | passed after the fix in round g4 takeover |
| E4 | After re-entering the session, the image tap still opens the full-screen viewer. | ios | passed after the fix in round g4 takeover |

Defects reproduced on the unfixed build:
- Live markdown tap showed `Preview unavailable`, no modal (repro R1). Fixed; no longer reproduces.
- Live image tap showed `Image unavailable`, no viewer (repro R2). Fixed; no longer reproduces.
- After reopen, markdown showed `Preview unavailable` (repro R3). Fixed; no longer reproduces.

Recordings: repro-r0-ios.mp4, repro-r1-ios.mp4.

## Visual Changes

**Markdown preview modal.** Tapping the markdown chip opens a modal that renders the file text, not the `Preview unavailable` toast. The `Repro brief` heading and body line sit under the title bar; the Done button is top right.

**Full-screen image viewer.** Tapping the inline image opens a full-screen viewer that shows the PNG, not the `Image unavailable` row. The red PNG spans the width on a black backdrop; the close button is top left.

## Reviewer Notes

Human steps: none known.

Notes: none.

### DIFF
--- a/apps/mobile/src/components/agents/file-part-url-resolver.ts
+++ b/apps/mobile/src/components/agents/file-part-url-resolver.ts
@@ -1,0 +1,122 @@
+import { type FilePart } from '@kilocode/cloud-agent-sdk';
+import { useEffect, useState } from 'react';
+
+import { trpcClient } from '@/lib/trpc';
+
+import {
+  clearFilePartResolveFailed,
+  getFilePartCacheEntry,
+  isUsableFilePartUrl,
+  markFilePartResolveFailed,
+  overwriteFilePartCacheEntry,
+  useFilePartCache,
+} from './file-part-cache';
+import { type CloudAgentAttachmentRef, parseCloudAgentAttachmentUrl } from './file-part-preview';
+
+/** Part IDs with an on-demand presign in flight. Dedupes a StrictMode
+ *  double-mount and a leave/reopen during the mutate. */
+const inFlight = new Set<string>();
+
+export type ResolvedFilePartUrl = {
+  status: 'ready' | 'resolving' | 'unavailable' | 'error';
+  // set only when status === 'ready'
+  url?: string;
+  // set whenever a ref is known
+  attachmentRef?: CloudAgentAttachmentRef;
+  // set only when status === 'error'
+  retry?: () => void;
+};
+
+/**
+ * Resolve a usable URL for a FilePart. A captured `http(s)`/`data:` URL (or a
+ * cached one) is used directly. A cloud-agent sandbox `file://` attachment is
+ * presigned on demand via `getAttachmentDownloadUrl`. Failure state lives in
+ * the cache store so a remounted instance sees a failed presign instead of a
+ * stuck `resolving`.
+ */
+export function useResolvedFilePartUrl(part: FilePart): ResolvedFilePartUrl {
+  const cached = useFilePartCache(part.id);
+  const [attempt, setAttempt] = useState(0);
+
+  const url = cached?.url ?? (isUsableFilePartUrl(part.url) ? part.url : undefined);
+  const ref = cached?.attachmentRef ?? parseCloudAgentAttachmentUrl(part.url);
+  const failed = cached?.resolveFailed === true;
+
+  useEffect(() => {
+    const entry = getFilePartCacheEntry(part.id);
+    if (entry?.url || entry?.resolveFailed) {
+      return;
+    }
+    const attachmentRef = entry?.attachmentRef ?? parseCloudAgentAttachmentUrl(part.url);
+    if (!attachmentRef || inFlight.has(part.id)) {
+      return;
+    }
+    inFlight.add(part.id);
+    void (async () => {
+      try {
+        const result = await trpcClient.cloudAgentNext.getAttachmentDownloadUrl.mutate({
+          messageUuid: attachmentRef.messageUuid,
+          filename: attachmentRef.filename,
+        });
+        overwriteFilePartCacheEntry(part.id, {
+          url: result.signedUrl,
+          mime: part.mime,
+          filename: part.filename,
+        });
+      } catch {
+        markFilePartResolveFailed(part.id);
+      } finally {
+        inFlight.delete(part.id);
+      }
+    })();
+  }, [part.id, part.url, part.mime, part.filename, cached, attempt]);
+
+  if (url !== undefined) {
+    return { status: 'ready', url, ...(ref ? { attachmentRef: ref } : {}) };
+  }
+  if (!ref) {
+    return { status: 'unavailable' };
+  }
+  if (failed) {
+    return {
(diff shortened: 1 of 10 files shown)


apps/mobile/src/components/agents/file-part-url-resolver.ts
```diff
@@ -0,0 +1,120 @@
...
+      return;
+    }
+    inFlight.add(part.id);
+    void (async () => {
+      try {
+        const result = await trpcClient.cloudAgentNext.getAttachmentDownloadUrl.mutate({
+          messageUuid: attachmentRef.messageUuid,
+          filename: attachmentRef.filename,
+        });
+        overwriteFilePartCacheEntry(part.id, {
```
pandemicsyn, line 60: bot: Presigning an attachment from an uncached `part.url` drops the reference needed for later retries.

Suggested fix: Carry `attachmentRef` into the cache write (for example, allow `overwriteFilePartCacheEntry` to accept it, or seed a ref-only entry before the mutation). Today the overwrite preserves a reference only if an entry already existed. On the uncached fallback path, the first image/text preview works, but after its signed URL expires or an image/text load fails, `refreshFilePartUrl(part.id)` sees an entry without `attachmentRef` and returns `false`. Add a regression test that mounts a sandbox `part.url` without calling `cacheFilePart`, resolves it, then verifies a later refresh re-presigns successfully.

services/cloud-agent-next/src/shared/trim-payload.ts
```diff
@@ -13,8 +22,25 @@ function truncate(s: string, max: number): string {
...
+  // the live stream must keep both. Large `data:` URLs carry inline base64
+  // bytes and are stripped for size.
+  const url = part.url;
+  const stripDataUrl =
+    typeof url === 'string' &&
+    url.startsWith('data:') &&
+    (!options?.preserveSmallDataUrls || url.length > MAX_INLINE_FILE_URL_LENGTH);
+  const out: Record<string, unknown> = {
+    ...part,
+    ...(stripDataUrl ? { url: '' } : {}),
```
pandemicsyn, line 42: bot: The new preservation rule forwards every non-`data:` URL scheme, not just the intended sandbox `file://` and HTTP(S) URLs.

Suggested fix: Replace the data-only stripping condition with an explicit allowlist: preserve only the validated sandbox attachment `file://` shape, `http://`/`https://`, and eligible small `data:` URLs; blank all other schemes. Add tests for unsupported values such as `ftp:` and case-variant/other schemes. As written, arbitrary non-data URLs are included in the ingest payload, widening the prior behavior that always removed file-part URLs and exceeding the stated retention contract.

pandemicsyn (approved):

PR #5413, opened 2026-08-21
fix(mobile): scope child session error to the child sheet

iscekic:
## Summary

- A subagent failure now appears only in the failed subagent's sheet, not on the main session screen.
- The main session's composer stays clear of a subagent's error.
- The main composer stays usable and shows Stop while the main session works.
- A failed subagent with no transcript shows a "Subagent session failed" error screen.
- A failed subagent with a transcript shows the error as a banner above its messages.

---

A `session.error` event with a non-root `sessionId` now routes to a new `onChildSessionError` callback and returns before the shared root status or `onError` path. An event without a `sessionId` keeps the legacy root behavior. This keeps a child failure from touching the parent status indicator.

<details>
<summary>Files</summary>

- `packages/cloud-agent-sdk/src/service-state.ts` — adds `onChildSessionError` to `ServiceStateConfig`; `processSessionError` routes non-root `sessionId` events to it and returns before the root error path.
- `packages/cloud-agent-sdk/src/session.ts` — adds `onChildSessionError` to `CloudAgentSessionConfig` and forwards it to `createServiceState`.

</details>

The session manager records each child's failure message in a new `childSessionErrorsAtom` map and exposes a read-only `childSessionError` atom that returns the message for a child id, or null. The map clears with the other atoms on switch and destroy, so a stale child error cannot leak into the next session.

<details>
<summary>Files</summary>

- `packages/cloud-agent-sdk/src/session-manager.ts` — adds the `childSessionError` atom to `SessionManagerAtoms`, the backing map, the derived reader, the `onChildSessionError` handler, and the clear/reset wiring.

</details>

The child session sheet renders a runtime failure only in the failed child: a child with no messages shows a "Subagent session failed" error state, and a child with messages shows an error banner above the transcript. The parent detail screen reads the per-child error atom and passes the open child's message into the sheet.

<details>
<summary>Files</summary>

- `apps/mobile/src/components/agents/child-session-sheet-state.ts` — `getChildSessionSheetState` takes a `sessionError` argument and returns `error` for it when the child has no messages.
- `apps/mobile/src/components/agents/child-session-sheet.tsx` — accepts `sessionError`; renders the no-message error state or the above-transcript banner.
- `apps/mobile/src/components/agents/session-detail-content.tsx` — reads `manager.atoms.childSessionError` and passes the open child's message to the sheet.

</details>

Tests: 3 test files updated (`child-session-sheet-state.test.ts`, `service-state.test.ts`, `session-manager.test.ts`).
Generated: none.

---

## Verification

One iOS device round ran two cases.

|---|---|---|---|
| S1 — parent happy path | The parent session shows the Stop control while the agent works, then returns the composer to a usable state with no error text above it. | iOS | passed |
| S2 — cold-start reopen | Reopening the parent session shows a usable composer and no error text above it. | iOS | passed |

No defect was reproduced on an unfixed build: the fix landed before round 1, so no round ran on an unfixed build.

## Visual Changes

**Parent session detail screen (mobile).** The user sees a usable composer and no error banner above it after a message round trip. In the picture, the Preparation complete row sits above the Message... composer, where an error banner would appear, and no error text is present.

## Reviewer Notes

Human steps: none needed.

Notes: none.

### DIFF
--- a/packages/cloud-agent-sdk/src/service-state.ts
+++ b/packages/cloud-agent-sdk/src/service-state.ts
@@ -109,6 +109,7 @@ function upsertByRequestId<T extends { requestId: string }>(
 }
 
 function createServiceState(config: ServiceStateConfig): ServiceState {
+  let rootSessionId = config.rootSessionId;
   let activity: SessionActivity = INITIAL_ACTIVITY;
   let status: AgentStatus = IDLE_STATUS;
   let cloudStatus: CloudStatus | null = null;
@@ -135,7 +136,7 @@ function createServiceState(config: ServiceStateConfig): ServiceState {
   }
 
   function isRootSession(sessionId: string): boolean {
-    return sessionId === config.rootSessionId;
+    return sessionId === rootSessionId;
   }
 
   function processSessionStatus(event: Extract<ServiceEvent, { type: 'session.status' }>): void {
@@ -246,6 +247,9 @@ function createServiceState(config: ServiceStateConfig): ServiceState {
   }
 
   function processSessionCreated(event: Extract<ServiceEvent, { type: 'session.created' }>): void {
+    if (event.info.parentID == null) {
+      rootSessionId = event.info.id;
+    }
     // Only track root session info
     if (isRootSession(event.info.id)) {
       sessionInfo = event.info;
--- a/packages/cloud-agent-sdk/src/service-state.ts
+++ b/packages/cloud-agent-sdk/src/service-state.ts
@@ -26,6 +26,7 @@ type ServiceStateConfig = {
   /** The root session ID we're tracking (to detect child sessions). */
   rootSessionId: string;
   onError?: ((message: string) => void) | undefined;
+  onChildSessionError?: ((sessionId: string, message: string) => void) | undefined;
   onQuestionAsked?: ((requestId: string, questions?: QuestionInfo[]) => void) | undefined;
   onQuestionResolved?: ((requestId: string) => void) | undefined;
   onPermissionAsked?:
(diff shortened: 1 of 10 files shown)


packages/cloud-agent-sdk/src/service-state.ts
```diff
@@ -230,6 +231,14 @@ function createServiceState(config: ServiceStateConfig): ServiceState {
   function processSessionError(event: Extract<ServiceEvent, { type: 'session.error' }>): void {
     if (terminated) return;
 
+    // Child session errors are scoped to the child. They must not touch the
+    // shared root status or onError, which drive the parent status indicator.
+    // Events without a sessionId keep the legacy root behavior.
+    if (event.sessionId !== undefined && !isRootSession(event.sessionId)) {
```
pandemicsyn, line 238: bot: A server-reported root ID can be routed as a child error and disappear from the parent error path.

Suggested fix: Keep the root ID used by `isRootSession` synchronized with the server-reported root session ID (the manager already adopts `session.created` IDs with `parentID == null`), or provide service state an authoritative root-ID predicate. Add a test where the configured `kiloSessionId` differs from the root `session.created` ID, then verify a subsequent `session.error` for that server ID invokes `onError` and sets root status rather than `onChildSessionError`.

pandemicsyn (approved):
lgtm one in line

jrf0110 (approved):

packages/cloud-agent-sdk/src/service-state.ts
iscekic: Valid. `ServiceState` kept the configured ID after `SessionManager` adopted the server root ID. Commit cb0aaff52 now adopts each root `session.created` ID inside `ServiceState`. The new regression test verifies that a later root `session.error` calls `onError`, updates root status, and does not call `onChildSessionError`.

PR #5416, opened 2026-08-21
fix(mobile): harden offline and query integrity flows (w6c)

iscekic:
## Summary

A live session that stops reconnecting now shows "Connection lost" with a Retry button, so the user can recover instead of watching "Reconnecting…" forever.

The Security Agent settings screen now shows a load error with a Retry button when it cannot reach the service, instead of telling an owner that only owners can manage the agent.

---

The SDK connection gains an additive reconnect-exhaustion contract: an optional `maxReconnectAttempts` cap, a two-edge `onReconnectExhaustionChange` callback, and a `retryReconnect` method. Exhaustion sets when the retry cap is reached or a terminal auth failure stops retries, and clears on every recovery path; `refreshAndConnect` now threads the attempt through so a live session that refreshes auth on reconnect still reaches the cap. All additions are optional, so web and extension consumers compile and behave unchanged.

<details>
<summary>Files</summary>

- `packages/cloud-agent-sdk/src/base-connection.ts` — adds `maxReconnectAttempts`, `onReconnectExhaustionChange`, and `retryReconnect`; sets exhaustion at the retry cap and on a terminal auth failure, clears it on every recovery path, and threads the reconnect attempt through `refreshAndConnect`.
- `packages/cloud-agent-sdk/src/user-web-connection.ts` — forwards `maxReconnectAttempts`, mirrors the exhaustion snapshot, exposes `isReconnectExhausted`, `onReconnectExhaustionChange`, and `retryConnection`, and resets the snapshot when the connection stops.

</details>

The session connection indicator gains an explicit `exhausted` state: a remote session whose reconnects are exhausted shows "Connection lost" with a Retry action instead of "Reconnecting…" forever. The state resolver consumes the exhaustion signal, and a new `useUserWebConnectionHealth` hook binds readiness and exhaustion reactively through the shared connection.

<details>
<summary>Files</summary>

- `apps/mobile/src/components/agents/session-connection-indicator-state.ts` — adds the `exhausted` state and the `reconnectExhausted` input.
- `apps/mobile/src/components/agents/session-connection-indicator.tsx` — renders the "Connection lost" label and a Retry pressable that calls `retryConnection`, with separate accessibility elements.
- `apps/mobile/src/lib/hooks/use-user-web-connection-state.ts` — adds `useUserWebConnectionHealth` binding both readiness and exhaustion.

</details>

`useSecurityAgentCapability` returns a discriminated `status` (`loading | error | denied | allowed`) instead of a bare canManage boolean, so callers can tell a failed or pending permission load from a resolved denial. The settings overview screen renders distinct loading, error, offline, and denied states, so an owner never sees permission-denied copy on a network error or an offline launch. A settled role stays authoritative: a failed background refetch does not demote an already-resolved capability to error.

<details>
<summary>Files</summary>

- `apps/mobile/src/lib/hooks/use-security-agent.ts` — adds the discriminated capability status and exposes `hasData` and `isPending` from the role query.
- `apps/mobile/src/components/security-agent/settings-overview-screen.tsx` — branches on capability status and committed connectivity for the offline and permission-error screens.

</details>

The offline banner store now exposes a committed `BannerState` tri-state (online, offline, unknown) and notifies on every committed change, so screens can gate on a settled connectivity state. The hook layer keeps one module-level store for the whole app, so every consumer shares a single NetInfo subscription, and adds a `useCommittedConnectivityStatus` hook.

<details>
<summary>Files</summary>

- `apps/mobile/src/lib/offline-banner-state.ts` — exports `BannerState`, adds the `state` snapshot, and notifies on every committed change.
- `apps/mobile/src/lib/hooks/use-offline-banner-state.ts` — replaces the per-mount store lifecycle with one lazy shared store and adds `useCommittedConnectivityStatus`.

</details>

Every optimistic mutation now stamps its write with a per-cache generation from `nextMutationGeneration` and rolls back only when `isLatestMutationGeneration` confirms it is still the latest, so an older failure cannot stomp cache state a newer mutation owns. Mutations that share one cache serialize their network calls through a `scope.id` or a `chainSave`, and each changed mutation declares its onError policy in a comment. The accepted residual is that when two writes both fail, the older optimistic value can show until the settle invalidation refetches.

<details>
<summary>Files</summary>

- `apps/mobile/src/lib/hooks/mutation-generations.ts` — new pure per-key generation counter and latest-generation check.
- `apps/mobile/src/lib/hooks/use-code-reviewer.ts` — stamps toggle and save rollbacks with generations and chains both onto the shared scope+platform save key.
- `apps/mobile/src/lib/hooks/use-model-preferences.ts` — stamps favorite rollbacks with generations and chains favorite writes on one key.
- `apps/mobile/src/lib/hooks/use-organization-mutations.ts` — stamps member, rename, and balance rollbacks with generations and serializes org writes with a scope id.
- `apps/mobile/src/lib/hooks/use-security-agent-mutations.ts` — stamps config save and toggle rollbacks with generations and serializes them with a scope id.
- `apps/mobile/src/lib/hooks/use-security-remediation.ts` — stamps remediation-cancel rollbacks with generations.
- `apps/mobile/src/lib/hooks/use-session-mutations.ts` — stamps session delete and rename rollbacks with generations after the cancel, and chains both per session.
- `apps/mobile/src/lib/pr-review/discussion/use-review-discussion-mutations.ts` — stamps thread resolve, unresolve, and reaction rollbacks with a shared generation key, chains resolve and unresolve per thread, and scopes reactions per thread.
- `apps/mobile/src/components/notifications-screen.tsx` — stamps push-token and preference rollbacks with generations and serializes both with static scope ids.

</details>

The global query client sets `refetchOnWindowFocus` to false, and a new `useRouteForegroundRefresh` hook invalidates only the active route's query prefixes on foreground and focus regain. Owner mounts across layouts refresh each route's domain, and the root layout keeps app-wide freshness for the signed-in user, organizations, and kilo-chat. The Agents list reads focus live via `navigation.isFocused()` so a frozen unfocused tab does not refetch on foreground.

<details>
<summary>Files</summary>

- `apps/mobile/src/lib/query-client.ts` — sets `refetchOnWindowFocus: false`.
- `apps/mobile/src/lib/hooks/use-route-foreground-refresh.ts` — new hook invalidating path-prefix keys on app-active and focus regain.
- `apps/mobile/src/app/(app)/_layout.tsx` — adds the app-wide freshness mount.
- `apps/mobile/src/app/(app)/(tabs)/(0_home)/_layout.tsx` — refreshes the active-sessions domain.
- `apps/mobile/src/app/(app)/(tabs)/(3_profile)/_layout.tsx` — refreshes the user, organizations, reviewer, security-agent, and kilo-pass domains.
- `apps/mobile/src/app/(app)/pr-review/[owner]/[repo]/[number]/_layout.tsx` — refreshes the PR-review domain.
- `apps/mobile/src/app/(app)/pr-review/index.tsx` — refreshes the PR-review domain.
- `apps/mobile/src/app/(app)/agent-chat/[session-id].tsx` — refreshes the session and model-preference domains.
- `apps/mobile/src/app/(app)/agent-chat/model-picker.tsx` — refreshes the model-preference domain.
- `apps/mobile/src/app/(app)/device-sessions.tsx` — refreshes the user domain.
- `apps/mobile/src/app/(app)/kilo-pass.tsx` — refreshes the kilo-pass domain.
- `apps/mobile/src/components/agents/session-list-screen.tsx` — gates the foreground refetch on tab focus and invalidates the active-sessions tray.

</details>

The credits card's own focus-based refetch is removed because the profile route mount now owns that refresh.

<details>
<summary>Files</summary>

- `apps/mobile/src/components/profile-credits-card.tsx` — removes the focus-based refetch.

</details>

Tests: 25 test files updated — 20 in the mobile app and 5 in the SDK, adding chaos scenarios for offline flapping, process kill with persisted state, cold-launch offline, hibernation zombie-socket replacement, reconnect exhaustion, and mutation interleave.
Generated: none.

---

## Verification

iOS ran three cases in one round; a fourth case was not run.

|---|---|---|---|
| cold-launch-restore | A relaunch renders the cached Agents session list without an error screen or a stuck skeleton. | iOS | passed |
| foreground-freshness | The app refetches the session list on foreground return and shows no connection-lost label. | iOS | passed |
| security-capability-happy-path | A personal-scope owner sees the Security Agent switch and settings rows, never denial copy. | iOS | passed |
| mutation-error-surface | A failed Security Agent change rolls back and shows an error toast. | iOS | not run — the fresh account has no repositories, so the switch is disabled; the sandbox lacks a GitHub installation to seed one; the vitest/jest suite holds the rollback and toast evidence |

The round reproduced no defect on the unfixed build; all three cases passed on the first run.
No recording exists.

Human steps: none.

### Agents tab (session list) after cold-launch restore

After a cold relaunch, the user sees the Agents tab with its session list restored from the cache as a settled empty state. The `Agents` header at the top and the `No sessions yet` block in the middle of the screen show the restored list.

### Agents tab after foreground return

When the app returns to the foreground, the Agents tab refreshes and shows no `Connection lost` or `Reconnecting…` label. The empty session list fills the app window, and no connection label appears on the screen. The capture catches the foreground transition, but the screen content stays readable.

### Security Agent settings (personal scope)

The owner sees a Settings screen with the Security Agent row and its switch, and no owner-denial copy. The `Security Agent` row and its switch sit under the `Settings` header, above the repository hint text.

## Reviewer Notes

Notes: none.


packages/cloud-agent-sdk/src/base-connection.ts
```diff
@@ -160,7 +177,11 @@ export function createBaseConnection<T>(config: BaseConnectionConfig<T>): Connec
   function scheduleReconnect(attempt: number, expectedGeneration: number) {
     if (destroyed || intentionalDisconnect || expectedGeneration !== generation) return;
 
-    if (attempt >= MAX_RECONNECT_ATTEMPTS) {
+    if (attempt >= maxReconnectAttempts) {
```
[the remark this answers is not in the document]
iscekic: (bot) Fixed in 165fe8c24: refreshAndConnect now threads the reconnect attempt through connectInternal, so live sessions whose reconnects refresh auth reach the exhaustion cap and fire the two-edge signal. Added a jest test reproducing the live-session case.

apps/mobile/src/lib/hooks/use-session-mutations.ts
```diff
@@ -61,31 +65,42 @@ export function useSessionMutations() {
...
 
+  // onError policy: roll back the onMutate snapshot (latest generation only)
+  // and toast error.message.
   const deleteSessionMutation = useMutation(
     trpc.cliSessionsV2.delete.mutationOptions({
-      // eslint-disable-next-line typescript-eslint/promise-function-async -- conflicting require-await rule
-      onMutate: ({ session_id }) =>
-        snapshotAndUpdate(data => removeStoredSession(data, session_id)),
+      onMutate: async ({ session_id }) => {
+        const generation = nextMutationGeneration(hashKey(listKey));
```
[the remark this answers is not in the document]
iscekic: (bot) Fixed in 55ac7098f: the generation is now stamped inside snapshotAndUpdate after cancelQueries resolves, in the same synchronous block as the snapshot and write, for both delete and rename.

apps/mobile/src/components/agents/session-list-screen.tsx
```diff
@@ -102,17 +109,20 @@ export function AgentSessionListScreen() {
...
   // through the same wrapped `refetch` as navigation focus — to keep every
   // stored refetch serialized by the shared operation coordinator (backfill
-  // and departure never overlap a refetch).
+  // and departure never overlap a refetch). A frozen (unfocused) Agents tab
+  // must NOT refetch on foreground: only the focused tab refreshes the stored
+  // list and invalidates the active-sessions tray.
   useEffect(() => {
     const subscription = AppState.addEventListener('change', nextState => {
-      if (nextState === 'active') {
+      if (nextState === 'active' && focusedRef.current) {
```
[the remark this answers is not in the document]
iscekic: (bot) Fixed in 56b335e23: the AppState callback now reads navigation.isFocused() live instead of a focusedRef a frozen tab never updates. The mounted test now blurs without re-rendering to cover the freeze scenario.

packages/cloud-agent-sdk/src/base-connection.ts
```diff
@@ -234,6 +255,7 @@ export function createBaseConnection<T>(config: BaseConnectionConfig<T>): Connec
       // Reset auth refresh flag on successful message
       authRefreshAttempted = false;
       reconnectAttempt = 0;
+      clearExhausted();
```
pandemicsyn, line 265: bot: A late event from a replaced socket can falsely clear reconnect exhaustion and mark the transport recovered.

Suggested fix: At the start of `newWs.onmessage`, return unless `ws === newWs`, before changing staleness timing, parsing the message, clearing exhaustion, or dispatching callbacks. This matches the existing ownership guard in `onclose`. Add a regression test that replaces/exhausts a connection, then emits a valid late message from the old socket and asserts that exhaustion, readiness, and events remain unchanged.

apps/mobile/src/components/notifications-screen.tsx
```diff
@@ -360,19 +372,26 @@ export function NotificationsScreen() {
         if (next === undefined) {
           return undefined;
         }
-        return applyAgentPushOptimistic({
+        const generation = nextMutationGeneration(hashKey(preferencesQueryKey));
```
pandemicsyn: bot: The notification-preference generation is stamped before the awaited cancellation and optimistic write, so it can identify the wrong cache write.

Suggested fix: Move `nextMutationGeneration(hashKey(preferencesQueryKey))` into `applyAgentPushOptimistic` immediately after its `cancelQueries` await and alongside its snapshot/setQueryData operations (or otherwise serialize the whole onMutate critical section). Return that generation with the snapshot. Add a test that starts two preference mutations, resolves their cancellations in reverse order, and fails one to prove no rollback or optimistic write clobbers the other.

apps/mobile/src/lib/hooks/use-route-foreground-refresh.ts
```diff
@@ -0,0 +1,59 @@
...
+  const wasActiveRef = useRef(isActive);
+  const focusedRef = useRef(isFocused);
+  const firstFocusRef = useRef(true);
+
+  useEffect(() => {
+    focusedRef.current = isFocused;
+  }, [isFocused]);
+
+  useEffect(() => {
+    if (!wasActiveRef.current && isActive && focusedRef.current) {
```
pandemicsyn: bot: Foreground refresh can invalidate an unfocused frozen tab because this guard reads a focus ref that only updates after the route renders.

Suggested fix: Use a live navigation focus check in the active-transition callback, as `session-list-screen` does, or subscribe to a focus source that updates while the tree is frozen. Add a mounted test using a layout that calls this hook: blur it without rerendering, foreground the app, and assert its query prefixes are not invalidated.

pandemicsyn (approved):
lgtm, couple from the bot.

packages/cloud-agent-sdk/src/base-connection.ts
iscekic: Fixed in 5eeb233ce.

`newWs.onmessage` now returns unless `ws === newWs`, before the staleness timing, the parse, the exhaustion clear, and the callbacks. This matches the `onclose` ownership guard.

Test: `base-connection.test.ts` — "ignores a late message from a replaced socket". It exhausts the cap, emits a valid message from socket 0, and asserts `onEvent`, `onReconnected`, exhaustion, and the socket count stay unchanged. The test fails on the unfixed build.

Also rewrote "successful message after exhaustion fires false": exhaustion only ever sets while `ws` is null, so that scenario was impossible. It now recovers with `retryReconnect()` and asserts the recovery socket keeps exhaustion cleared.

apps/mobile/src/components/notifications-screen.tsx
iscekic: Fixed in 5eeb233ce.

`applyAgentPushOptimistic` now stamps the generation itself, after its `cancelQueries` await and in the same synchronous block as `setQueryData`, and returns it in the context. The call site no longer stamps.

Test: `agent-push-preference.optimistic.test.ts` — "gives the latest generation to the write that lands last". It starts two flips, releases the cancellations in reverse order, and asserts the last write holds the latest generation. The test fails when the stamp moves before the await.

apps/mobile/src/lib/hooks/use-route-foreground-refresh.ts
iscekic: Fixed in 5eeb233ce.

The hook now subscribes to `AppState` directly and reads focus live with `navigation.isFocused()`, the same pattern as `session-list-screen`. It holds no React state, so an AppState change cannot rerender a frozen route.

Test: `use-route-foreground-refresh.mounted.test.tsx` — "reads focus live without rerendering on the foreground transition". It blurs the probe without a rerender, foregrounds the app, and asserts no invalidation and no extra render. The test fails on the unfixed hook.

PR #5421, opened 2026-08-21
feat(mobile): harden bootstrap, routing, and release gates (w6b)

iscekic:
## Summary

The Home screen now shows an Explore area with three shortcuts: Code Reviewer, Security Agent, and PR Review. The area stays visible when the session list is loading, empty, filled, or failed to load.

The Profile screen now paints first and loads account and organization details only after the screen transition settles, showing placeholders while it waits.

Opening a link with a broken review, session, or organization identifier now shows a not-found screen with a back action instead of a blank or broken screen.

---

The production host allowlist is now complete. It holds all seven production hosts from the committed environment file, including the three `kiloapps.io` endpoints, and a unit test reads that file and asserts every URL value passes the scheme and host checks, so a missing host fails before any build. A production build still fails on a missing required value, a wrong scheme, an off-allowlist host, or a missing Sentry source-map token.

<details>
<summary>Files</summary>

- `apps/mobile/src/lib/url-contract.js` — new; holds `URL_SCHEMES`, `PRODUCTION_HOSTS`, `assertUrlScheme`, and `assertProductionHost`; the allowlist now lists all seven production hosts.
- `apps/mobile/app.config.ts` — applies the scheme and host checks; makes missing env fatal in production; adds the Sentry token gate; bakes `extra.isProductionBuild`.
- `apps/mobile/src/lib/config.ts` — applies the scheme check at runtime and the production host check when `extra.isProductionBuild` is true.

</details>

The root layout now derives its bootstrap routing and its render gating from one pure resolver instead of two hand-kept copies, so the routing effect and the visibility flags can no longer disagree. The resolver returns a closed union of routing tags plus the error, redirect, and hidden flags, and the routing effect switches over the tag. The old inline if-chain and the old inline gating derivations are deleted; behavior stays the same.

<details>
<summary>Files</summary>

- `apps/mobile/src/lib/bootstrap-decision.ts` — new; exports the input, decision, and tag types and the pure resolver.
- `apps/mobile/src/app/_layout.tsx` — calls the resolver in the routing effect and the render gate; replaces the inline if-chain with a switch.

</details>

The release workflow pins EAS CLI to 21.8.0 and splits the old build-and-submit step into a preflight job, a build job, an artifact download, an inspection step, and separate iOS and Android submissions. The preflight fails closed when the EAS production environment misses a required value or the config contract fails. The build no longer uses `--auto-submit`; submission runs from downloaded artifacts after inspection.

<details>
<summary>Files</summary>

- `.github/workflows/kilo-app-release.yml` — pins the CLI, adds the preflight job, splits build and submit, adds download and inspection steps.
- `apps/mobile/eas.json` — pins `cli.version` to 21.8.0.
- `apps/mobile/package.json` — pins every eas-cli script to 21.8.0 and adds the `assert:config` script.

</details>

A new preflight script runs the fully evaluated Expo config and fails when the bundle id, package, scheme, associated domain, blocked permissions, Sentry plugin, or any required extra value violates the contract. The preflight invokes it inside the EAS production environment, so config drift fails before any build.

<details>
<summary>Files</summary>

- `apps/mobile/scripts/assert-expo-config.mjs` — new; runs `expo config --json` and checks the contract values.

</details>

A new inspection script validates the EAS build result and the signed IPA and AAB before submission, and submission stops when an artifact violates a bundle, permission, endpoint, or debug-symbol contract. The IPA is checked for its bundle id, privacy manifest, usage descriptions, and SKAdNetwork endpoints; the AAB is checked for its package, blocked permissions, cleartext traffic, intent-filter host, and debug symbols.

<details>
<summary>Files</summary>

- `scripts/inspect-mobile-artifacts.mjs` — new; validates build.json, inspects the IPA and AAB, and checks debug symbols.

</details>

The home screen gains an Explore section with Code Reviewer, Security Agent, and PR Review rows, and the PR Review row stays behind the existing PR Review feature flag. The section renders in the empty, populated, and error states of the session list, so the shortcuts remain available when sessions fail to load.

<details>
<summary>Files</summary>

- `apps/mobile/src/components/home/product-choices.tsx` — new; three rows, PR Review gated by the flag.
- `apps/mobile/src/components/home/home-screen.tsx` — renders ProductChoices after the sessions/promo branch.

</details>

The profile screen defers its two mount-time queries until the navigation transition settles, using a new `useAfterInteractions` hook, so the screen paints before the account and organization fetches start. The skeleton, stale, and error states stay, and the linked-accounts section now shows its skeleton while the interactions are pending.

<details>
<summary>Files</summary>

- `apps/mobile/src/lib/hooks/use-after-interactions.ts` — new; returns false on mount, then true after `InteractionManager.runAfterInteractions`.
- `apps/mobile/src/components/profile-screen.tsx` — gates both queries on the hook and keeps the skeleton and error states.

</details>

Six routes that read a required path parameter now parse it with `parseParam` and render `InvalidRouteState` with a bounded back target when the parameter is missing or malformed. Optional parameters keep their existing first-element unwrapping, and the session route no longer fires its query for a null session id.

<details>
<summary>Files</summary>

- `apps/mobile/src/app/(app)/(tabs)/(3_profile)/code-reviewer/[scope]/index.tsx` — parses scope; renders InvalidRouteState.
- `apps/mobile/src/app/(app)/(tabs)/(3_profile)/code-reviewer/[scope]/manual-review.tsx` — parses scope; renders InvalidRouteState.
- `apps/mobile/src/app/(app)/(tabs)/(3_profile)/code-reviewer/[scope]/reviews/[id].tsx` — parses scope and id; back target depends on scope.
- `apps/mobile/src/app/(app)/(tabs)/(3_profile)/code-reviewer/[scope]/reviews/index.tsx` — parses scope; renders InvalidRouteState.
- `apps/mobile/src/app/(app)/(tabs)/(3_profile)/organization/member-limit.tsx` — parses memberId; renders InvalidRouteState.
- `apps/mobile/src/app/(app)/agent-chat/[session-id].tsx` — parses session-id; disables the query and renders InvalidRouteState when null.

</details>

Tests: 13 test files added or updated.
Generated: none.

---

## Verification

Two cases ran on iOS across two verification rounds.

|---|---|---|---|
| EV-03 cold start to interactive Home | A relaunch from force-quit lands on the interactive Home; no local sink observes the startup event in dev builds. | iOS | passed |
| EV-08 invalid route boundary | A malformed deep link renders a bounded not-found state on the device; the route tests prove the InvalidRouteState back target. | iOS | passed |

No round reproduced a defect on the unfixed build.
Recording: `ev03-coldstart.mp4` shows the EV-03 cold launch from force-quit to interactive Home.

## Visual Changes

**Home screen.** The user now sees an Explore section on Home with rows for Code Reviewer, Security Agent, and PR Review. In the picture, the EXPLORE header and the three rows sit below the Kilo Agents card and above the tab bar.

**Invalid-route screen.** The user now gets a bounded not-found state with a back action when a route param is malformed. No picture shows this change. The single candidate shows the session not-found state after a malformed deep link; no deep link yields a malformed param on the device, so the picture fails to show the new InvalidRouteState with Page not found and Go back. The co-located route tests prove that state instead.

## Reviewer Notes

Human steps:

1. Confirm the EAS production environment URL values match the committed `apps/mobile/.env` hosts before the first real production build; the release preflight fails closed on any mismatch. (after merge)
2. Store `SENTRY_AUTH_TOKEN` in the EAS production environment with non-secret visibility (`plain` or `sensitive`). (after merge)
3. Set `EXPO_PUBLIC_SENTRY_ENVIRONMENT=production` in the EAS production environment; the release preflight asserts it. The EAS preview environment no longer needs `EXPO_PUBLIC_SENTRY_ENVIRONMENT=preview` to avoid a crash, because the runtime host check now keys off the baked build intent. (after merge)
4. Watch the first real release run: the signed IPA may lack `PrivacyInfo.xcprivacy` or the AAB may lack debug-symbols metadata; the follow-up fix is named in the plan. (after merge)
5. The signed IPA's binary `Info.plist` is parsed with `python3 plistlib`; the release runner must provide `python3`. (after merge)

Notes: E2E: the follow-up commit 13f4d4b5a adds three production hosts to the build-time allowlist and a unit test; the host check only runs in production builds, so it has no runtime behavior to verify. The existing iOS verification still holds.


apps/mobile/src/lib/config.ts
```diff
@@ -37,6 +44,20 @@ export const PLAY_INTEGRITY_PROJECT_NUMBER: string | undefined = optional(
 );
 export const SENTRY_ENVIRONMENT: string | undefined = optional('sentryEnvironment');
 
+// URL contract at module evaluation. The `required` presence check above is
+// the old presence-only check; remove it when every build passes through the
+// config boundary in app.config.ts, which already throws on missing values.
+const sentryEnvironment = resolveSentryEnvironment(SENTRY_ENVIRONMENT, __DEV__);
+const runProductionHostCheck =
+  !__DEV__ && sentryEnvironment !== 'development' && sentryEnvironment !== 'preview';
```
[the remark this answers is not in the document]
iscekic: (bot) Fixed in d7e4503a0. The runtime production-host check now keys off the baked EAS build intent (extra.isProductionBuild from EAS_BUILD_PROFILE === "production"), not the Sentry environment. A preview release build no longer fails closed on preview hosts; the EAS production build still runs the check at config time and runtime.

.github/workflows/kilo-app-release.yml
```diff
@@ -52,8 +52,64 @@ jobs:
...
+
+      - name: Verify EAS production environment
+        working-directory: apps/mobile
+        env:
+          EXPO_TOKEN: ${{ secrets.EXPO_TOKEN }}
+        run: |
+          ENV_OUTPUT=$(pnpx eas-cli@21.8.0 env:list --environment production --format long)
+          MISSING=""
+          for NAME in API_BASE_URL WEB_BASE_URL CLOUD_AGENT_WS_URL SESSION_INGEST_WS_URL APPSFLYER_DEV_KEY APPSFLYER_APP_ID KILO_CHAT_URL EVENT_SERVICE_URL NOTIFICATIONS_URL POSTHOG_API_KEY SENTRY_AUTH_TOKEN EXPO_PUBLIC_SENTRY_ENVIRONMENT; do
+            if ! printf '%s\n' "$ENV_OUTPUT" | grep -qE "^Name[[:space:]]+${NAME}[[:space:]]*$"; then
```
pandemicsyn, line 95: bot: The EAS environment presence check does not match `env:list --format long` data rows, so it reports configured variables as missing and blocks every production preflight.

Suggested fix: Parse a machine-readable EAS response if the CLI offers one, or change the check to match the actual data-row format (where the variable name is the first column), rather than a line shaped `Name <VARIABLE>`. Add a workflow-level fixture/test covering populated and missing production variables so a CLI output-format change cannot silently block releases.

apps/mobile/src/lib/url-contract.js
```diff
@@ -0,0 +1,86 @@
...
+ *  production environment
+ *  (`pnpx eas-cli@21.8.0 env:list --environment production --format long`)
+ *  by extracting the host of each URL value. Otherwise preflight is the
+ *  completeness safety net: the release preflight runs assertProductionHost
+ *  against the real production values, so an incomplete allowlist fails
+ *  preflight before any build, never at runtime in a store build.
+ *  cloud-agent-next.kilosessions.ai and ingest.kilosessions.ai come from the
+ *  repo-documented extension defaults
+ *  (apps/extension/src/shared/cloud-agent-config.ts:10-11). */
+export const PRODUCTION_HOSTS = [
```
pandemicsyn, line 30: bot: The new production host allowlist rejects three URL hosts used by the committed mobile configuration, blocking production config assertion/builds when production uses those defaults.

Suggested fix: Add `chat.kiloapps.io`, `events.kiloapps.io`, and `notifications.kiloapps.io` to `PRODUCTION_HOSTS` after confirming they are the intended production endpoints, and add positive host-contract tests for all configured production URL values. If the EAS production values are intentionally different, update the checked-in configuration and the allowlist together before enabling the release gate.

pandemicsyn (approved):
pre-approving but think robot's inline comment is right and deploys via this new workflow will probably fail.

.github/workflows/kilo-app-release.yml
iscekic: (bot) Rejected: the check matches the actual `env:list --format long` output. eas-cli v21.8.0 renders `--format long` as a field list, not a table: `formatVariable` emits each field as a padded label plus two spaces plus the value, so the name line is `Name           API_BASE_URL`. The preflight grep `^Name[[:space:]]+${NAME}[[:space:]]*$` matches that line, so a configured variable is not reported missing. Verified against the pinned CLI source at the v21.8.0 tag (expo/eas-cli `src/commands/env/list.ts`, `src/utils/variableUtils.ts`, `src/utils/formatFields.ts`) and by simulating the field-list output through the exact grep.

apps/mobile/src/lib/url-contract.js
iscekic: (bot) Fixed in 13f4d4b5a. Added chat.kiloapps.io, events.kiloapps.io, and notifications.kiloapps.io to PRODUCTION_HOSTS, and added a host-contract test that reads the committed apps/mobile/.env and asserts every URL value passes assertUrlScheme and assertProductionHost. chat.kiloapps.io is confirmed as the production Kilo Chat endpoint in apps/web/src/lib/kiloclaw/kilo-chat-internal-client.ts.

PR #5425, opened 2026-08-21
fix(mobile): align markdown list markers with item text

iscekic:
## Summary

Bullets and numbers in chat message lists now sit on the same line as the first line of each item's text. The alignment holds for lists with and without blank lines between items.

---

The renderer adds a `listItem` override that flattens the leading paragraph View's vertical margin in loose lists. Loose lists wrap each item's text in a paragraph View whose top margin pushed the first line below the marker, so the override finds that View by its `paddingVertical` style key, sets `marginTop` to 0, and keeps the bottom margin. It clones the View in place, which preserves the renderer's key sequence.

<details>
<summary>Files</summary>

- `apps/mobile/src/components/agents/markdown-renderer.ts` — adds the `listItem` override; it detects the leading paragraph View, rewrites its vertical margins, and clones it in place.

</details>

The `list` palette style changes from `{ marginVertical: 2 }` to `{ marginBottom: 4 }`. react-native-marked applies `list` to each item's marker box, not the list container, so the old top margin pushed the marker below the first text line. The new style moves that space to the bottom, so the marker box keeps its 28px height and the marker sits at the row top.

<details>
<summary>Files</summary>

- `apps/mobile/src/components/agents/markdown-palette.ts` — changes the `list` style to `{ marginBottom: 4 }` and documents the marker-box mapping.

</details>

Tests: 2 files updated — `apps/mobile/src/components/agents/markdown-renderer.test.ts`, `apps/mobile/src/components/agents/markdown-palette.test.ts`.
Generated: none.

---

## Verification

Verification ran 2 cases on the iOS platform.

|---|---|---|---|
| ML-1 | Tight bullet and numbered list markers sit on the same line as the first line of the item text. | iOS | passed |
| ML-2 | Loose list markers stay on the item text line and the rows keep an even vertical gap. | iOS | passed |

No defect reproduced on the unfixed build.

## Visual Changes

Agent chat screen, Markdown list rendering. The user sees each bullet and number on the same line as the first line of that item's text. In the lower half, the bullets and numbers each sit level with their item word.

## Reviewer Notes

Human steps: none known.
Notes: none.

### DIFF
--- a/apps/mobile/src/components/agents/markdown-renderer.ts
+++ b/apps/mobile/src/components/agents/markdown-renderer.ts
@@ -11,7 +11,6 @@ import {
   type GestureResponderEvent,
   type ImageStyle,
   Pressable,
-  StyleSheet,
   Text,
   type TextStyle,
   View,
@@ -201,12 +200,10 @@ export class MarkdownRenderer extends Renderer {
   // in markdown-palette.ts). Rewrite the leading block's vertical margin so
   // loose and tight items start their first line at the row top alike.
   override listItem(children: ReactNode[], styles?: ViewStyle): ReactNode {
-    const first = Array.isArray(children) ? children[0] : undefined;
+    const first = children[0];
     if (isValidElement(first) && first.type === View) {
-      const styleProp = (first.props as { style?: ViewStyle | ViewStyle[] }).style;
-      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- StyleSheet.flatten(undefined) returns undefined at runtime; the RN type says non-null
-      const flat = StyleSheet.flatten(styleProp) ?? {};
-      const { marginVertical, marginBottom, ...rest } = flat;
+      const styleProp = (first.props as { style?: ViewStyle }).style;
+      const { marginVertical, marginBottom, ...rest } = styleProp ?? {};
       const adjusted: ViewStyle = {
         ...rest,
         marginTop: 0,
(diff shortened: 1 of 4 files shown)


apps/mobile/src/components/agents/markdown-renderer.ts
```diff
@@ -187,6 +194,38 @@ export class MarkdownRenderer extends Renderer {
...
 
+  // Loose Markdown lists — any blank line between items — wrap every item's
+  // text in a paragraph View (marked rewrites item `text` tokens to
+  // `paragraph` tokens). That paragraph's top margin pushes the first line
+  // below the marker, which sits at the row top (see the `list` style note
+  // in markdown-palette.ts). Rewrite the leading block's vertical margin so
+  // loose and tight items start their first line at the row top alike.
+  override listItem(children: ReactNode[], styles?: ViewStyle): ReactNode {
+    const first = Array.isArray(children) ? children[0] : undefined;
+    if (isValidElement(first) && first.type === View) {
```
pandemicsyn, line 206: bot: Avoid removing top spacing from every leading block View in a list item.

Suggested fix: `first.type === View` also matches block-level content, not only the loose-list paragraph the override intends to adjust. For example, `- > quoted text` renders a leading blockquote View with the palette's intentional `marginVertical: 4`; this branch flattens it and forces `marginTop: 0`, changing its spacing. Narrow the rewrite to the known paragraph-view shape/style (rather than any View), preserving other block styles, and add a mounted regression test for a list item that begins with a blockquote.

pandemicsyn (approved):
lgtm one nit inline from the bot

apps/mobile/src/components/agents/markdown-renderer.ts
iscekic: (bot) Narrowed the listItem rewrite to the paragraph View shape (style carries paddingVertical); a leading blockquote or hr now keeps its own vertical margin. Added a mounted regression test for `- > quoted text` asserting marginVertical: 4 with no forced marginTop. Pushed in f0e38d03e.

PR #5437, opened 2026-08-21
feat(organizations): add verified domain auto-join

jrf0110:
## Summary

image

- add WorkOS-backed verified domain claims with canonical validation, ownership constraints, lifecycle auditing, and organization settings UI
- automatically add matching authenticated users as ordinary organization members across browser, WorkOS SSO, and native authentication
- prefer the verified-domain organization after sign-in while preserving personal accounts, unrelated memberships, explicit destinations, and removal tombstones

## Validation
- 11 focused Jest suites, 269 tests passed
- `scripts/typecheck-all.sh --changes-only`
- `pnpm --filter web lint`
- `pnpm --filter @kilocode/db exec drizzle-kit check`
- `pnpm format:check`
- `pnpm drizzle generate` reports no schema changes
- `git diff --check`

## Remaining validation gaps
- `pnpm drizzle:verify-bootstrap` exits before PostgreSQL receives SQL with a non-diagnostic Drizzle `undefined` error; migration consistency and database-backed feature tests pass
- WorkOS behavior is covered with provider mocks rather than a live verification flow
- popup and responsive behavior have component coverage but no browser smoke test


apps/web/src/lib/user/server.ts
```diff
@@ -943,6 +946,10 @@ export const authOptions: NextAuthOptions = {
           return redirectUrlForCode(`BLOCKED`);
         }
 
+        if (!isAccountLinking) {
+          await ensureVerifiedDomainOrganizationMembership(result.user.id);
```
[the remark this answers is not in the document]
jrf0110: Fixed in bf86a30a0. Ordinary browser auto-join now uses the same proven-email gate as safe account linking: magic links, fake login, and profiles asserting `email_verified` remain eligible, while GitHub/GitLab/Discord-style profiles without proof settle normally but do not trigger verified-domain membership. The GitHub regression test now explicitly verifies that admission is not called.

apps/web/src/components/organizations/VerifiedDomainsCard.tsx
```diff
@@ -0,0 +1,343 @@
...
+
+  const invalidateClaims = () =>
+    queryClient.invalidateQueries({
+      queryKey: trpc.organizations.verifiedDomains.list.queryKey({ organizationId }),
+    });
+
+  const createClaim = useMutation(
+    trpc.organizations.verifiedDomains.create.mutationOptions({
+      onSuccess: async data => {
+        if (verificationPortal.current) {
```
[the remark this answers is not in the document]
jrf0110: Fixed in bf86a30a0. Popup reservation/navigation/cleanup are now non-throwing, closed windows fall back safely, the pending guard resets in `finally`, and unmount cleanup closes any reserved tab with a mounted guard preventing late mutation callbacks from reopening it. Added coverage for closed tabs, navigation failure, blocked popup APIs, opener assignment failure, and cleanup failure.

apps/web/src/lib/organizations/verified-domain-service.ts
```diff
@@ -0,0 +1,402 @@
...
+  if (!organization) {
+    throw new TRPCError({ code: 'NOT_FOUND', message: 'Organization not found' });
+  }
+  let refreshed: OrganizationDomain;
+  if (claim.workos_domain_id) {
+    try {
+      refreshed = await provider.organizationDomains.get(claim.workos_domain_id);
+    } catch (error) {
+      if (providerStatus(error) !== 404) throw providerError(error);
+      refreshed = await ensureProviderDomain(provider, claim, organization);
```
pandemicsyn: bot: A refresh can recreate a WorkOS domain after a concurrent removal has deleted the local claim.

Suggested fix: Serialize the whole refresh reconciliation with removal using the same claim/domain/organization lock before any provider read/create/delete (or use a durable reconciliation/outbox design). In particular, re-read and lock the claim before `ensureProviderDomain`, keep removal from proceeding until synchronization finishes, and add a test where a 404 refresh and removal interleave. Without this, refresh can create a new provider domain, removal can delete the old ID and local row, and `synchronizeProviderState` then fails with NOT_FOUND, leaving the newly created WorkOS domain orphaned.

apps/web/src/routers/organizations/organization-verified-domains-router.ts
```diff
@@ -0,0 +1,34 @@
...
+    .mutation(({ ctx, input }) =>
+      createVerifiedDomainClaim(input.organizationId, input.domain, ctx.user)
+    ),
+  refresh: organizationAdminProcedure
+    .input(ClaimInputSchema)
+    .mutation(({ ctx, input }) =>
+      refreshVerifiedDomainClaim(input.organizationId, input.claimId, ctx.user)
+    ),
+  remove: organizationAdminProcedure.input(ClaimInputSchema).mutation(async ({ ctx, input }) => {
+    await removeVerifiedDomainClaim(input.organizationId, input.claimId, ctx.user);
```
pandemicsyn, line 31: bot: Deleted organizations can still list and remove domain claims through these endpoints.

Suggested fix: Require `organizations.deleted_at IS NULL` for all verified-domain operations, either in a dedicated active-organization procedure or in `listVerifiedDomainClaims` and `removeVerifiedDomainClaim`, and add deleted-organization tests for list and remove. `organizationAdminProcedure` only checks membership and `ensureOrganizationAccess` does not filter soft-deleted organizations; unlike create and refresh, list/remove therefore allow a retained owner or admin membership to inspect claims or delete the provider domain after organization deletion.

pandemicsyn (approved):
lgtm, that refresh/removal race my bot flagged seems very edge case. Not sure this would trigger in the real world. If it takes excessive locking or something to box out that case it might not be worth addressing.

apps/web/src/routers/organizations/organization-verified-domains-router.ts
jrf0110: Fixed in 928064c32. Listing now requires an active non-deleted organization, and removal verifies the organization is active inside its transaction before taking provider locks or calling WorkOS. Added coverage showing list/remove return NOT_FOUND for a soft-deleted organization and provider deletion is not called.

apps/web/src/lib/organizations/verified-domain-service.ts
jrf0110: Fixed in 928064c32. Refresh now holds the claim row lock plus the existing domain and WorkOS-organization advisory locks across provider lookup/recreation and local synchronization, using the same lock order as removal. A concurrent removal waits, then reads and deletes the recreated provider domain ID. Added a regression test for the exact provider-404, recreate, and concurrent-remove interleave.

PR #5445, opened 2026-08-24
fix(cloud-agent-next): enforce git credential helper config

eshurakov:
## Summary

Install the read-only Git credential helper in all Cloud Agent sandbox images and force Git to use it for setup, submodules, and LFS. Remove profile-supplied GIT_CONFIG_* overrides so the helper cannot be replaced at runtime.

## Verification

- [ ] No manual UI path; this is container/runtime behavior.

## Visual Changes

N/A

## Reviewer Notes

- Focused tests: 117 passed in session-service.test.ts and kilo-git-credential.test.ts.
- format:check, lint, and typecheck passed.
- Full unit suite: 2,689 passed and 3 skipped. The unrelated wrapper hook test for chaining existing hooks and suppressing co-authorship timed out at its 5-second default but passed with a 15-second timeout.

### DIFF
--- a/services/cloud-agent-next/scripts/kilo-git-credential
+++ b/services/cloud-agent-next/scripts/kilo-git-credential
@@ -18,7 +18,6 @@ done
 
 [ "$protocol" = https ] || exit 0
 
-host="${host%%:*}"
 username=
 password=
 
@@ -36,7 +35,6 @@ bitbucket.org)
 	gitlab_host="${gitlab_host#https://}"
 	gitlab_host="${gitlab_host#http://}"
 	gitlab_host="${gitlab_host%%/*}"
-	gitlab_host="${gitlab_host%%:*}"
 	if [ "$host" = "$gitlab_host" ]; then
 		username=oauth2
 		password="${GITLAB_TOKEN:-}"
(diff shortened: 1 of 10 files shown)


services/cloud-agent-next/scripts/kilo-git-credential
```diff
@@ -0,0 +1,49 @@
...
+	[ -z "$line" ] && break
+	case "$line" in
+	protocol=*) protocol="${line#protocol=}" ;;
+	host=*) host="${line#host=}" ;;
+	esac
+done
+
+[ "$protocol" = https ] || exit 0
+
+host="${host%%:*}"
```
pandemicsyn: bot: Do not discard the requested port before deciding whether to return a credential.

Suggested fix: Git's credential protocol includes an explicitly specified port in `host`, so match the requested authority exactly rather than applying `host="${host%%:*}"`. Normalize the configured GitLab authority in the same way without removing its port, and add negative tests showing that `github.com:8443` and a configured GitLab host on a different port receive no token. This keeps a bearer token scoped to the actual HTTPS authority Git will contact.

pandemicsyn (approved):
lgtm, one line from the robot

services/cloud-agent-next/scripts/kilo-git-credential
eshurakov: Fixed in c6327c3a7. The credential helper now preserves the requested port and configured GitLab port, so tokens are returned only for the exact HTTPS authority. Added GitHub and GitLab port-mismatch regression coverage.

PR #5446, opened 2026-08-24
fix(gdpr): use connect.sid for Substack credentials

eshurakov:
## Summary

- use the Substack `connect.sid` cookie when normalizing stored credentials
- default the publication URL to `https://blog.kilo.ai` and share that behavior with the deletion handler
- update admin guidance, environment documentation, and focused tests

## Verification

- `oxfmt --list-different` passed
- type-aware `oxlint` passed with 0 warnings and 0 errors
- `git diff --check` passed
- targeted Jest could not start because this worktree has no installed dependencies and the external Jest setup is missing `ts-node`

### DIFF
--- a/apps/web/src/lib/user/deletion-queue/deletion-substack-credential.ts
+++ b/apps/web/src/lib/user/deletion-queue/deletion-substack-credential.ts
@@ -3,6 +3,7 @@ import { user_deletion_provider_credentials } from '@kilocode/db/schema';
 import { UserDeletionProviderScope } from '@kilocode/db/schema-types';
 import { getEnvVariable } from '@/lib/dotenvx';
 import { db } from '@/lib/drizzle';
+import { USER_DELETION_DEFAULT_SUBSTACK_PUBLICATION_URL } from '@/lib/user/deletion-queue/deletion-constants';
 import {
   decryptDeletionCredential,
   DeletionCryptoError,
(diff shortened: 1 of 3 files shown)


apps/web/src/lib/user/deletion-queue/deletion-substack-credential.ts
```diff
@@ -34,16 +35,20 @@ export function cookieFromCredential(material: string): string | null {
     }
     return null;
   }
-  return trimmed.includes('=') ? trimmed : `substack.sid=${trimmed}`;
+  return trimmed.includes('=') ? trimmed : `connect.sid=${trimmed}`;
```
pandemicsyn, line 38: bot: Legacy stored `substack.sid` cookies are not migrated to `connect.sid`.

Suggested fix: Recognize and rewrite the legacy cookie name in raw cookie material before returning it (including when it appears in a multi-cookie header), then add a regression test for `substack.sid=...`. Add or update a handler-level test that stores the legacy value and asserts the outbound deletion request sends `connect.sid=...`. This is needed because the deletion handler calls this normalizer for persisted credentials, while its existing test setup still stores `substack.sid=test-cookie`.

pandemicsyn (approved):
Bot flagged one, but i think as long as theres no deletions in flight when this rolls out its kind of a non issue.

PR #5444, opened 2026-08-24
feat(mobile): add language picker, RTL, and localized push copy

iscekic:
## Summary

- The mobile app now renders in 20 languages. It follows the device language by default and falls back to English.
- A language picker on the login screen and in Profile lets a user override the device language.
- Choosing a language that flips the layout direction restarts the app into a mirrored, right-to-left layout; a same-direction choice applies in place.
- Money, dates, and relative times now follow the active language.
- Push notifications arrive in the language the device had when it registered; each device gets its own language.
- Android notification channel names update to the active language.
- A language change that restarts the app keeps the typed login email and the SSO banner.

---

The mobile app gains a static 20-language i18n runtime. Device locale resolves through a supported-tag list with Chinese-script handling, and an RTL bootstrap forces the native direction before the first layout pass. The root layout holds the splash until the catalog is ready, and a failed cold-start direction reload surfaces a Retry/Continue screen instead of painting in the wrong direction.

<details>
<summary>Files</summary>

- `apps/mobile/src/i18n/index.ts` — creates the i18next instance with all 20 bundled catalogs, English default and fallback.
- `apps/mobile/src/i18n/languages.ts` — the 20 supported tags, the Arabic/Hebrew RTL set, and each language's endonym.
- `apps/mobile/src/i18n/resolve-language.ts` — maps a device locale list to a supported tag, with `zh` script subtag handling.
- `apps/mobile/src/i18n/rtl.ts` — enables RTL and exposes `syncRtl`, which forces the direction and reports whether a reload is needed.
- `apps/mobile/src/i18n/apply-language.ts` — persists and applies a preference; reloads for direction changes, switches in place otherwise, and rolls back on a failed persist.
- `apps/mobile/src/i18n/return-target.ts` — one-shot SecureStore target so an RTL reload reopens login or profile.
- `apps/mobile/src/i18n/i18next.d.ts` — module augmentation typing the catalog as the default namespace.
- `apps/mobile/src/app/_layout.tsx` — resolves and applies the language before first paint, handles the cold-start RTL reload and its failure screen, and renames Android channels on apply.
- `apps/mobile/src/lib/bootstrap-decision.ts` — adds the `settle-language-error` bootstrap tag and the language-reload-error flag.
- `apps/mobile/src/lib/startup-timing.ts` — adds the `language-error` startup outcome.
- `packages/app-shared/src/analytics/event-map.ts` — adds `language-error` to `STARTUP_OUTCOMES`.
[...]

PR #5451, opened 2026-08-24
feat(mobile): ship 87 languages, lazy catalogs, and mechanical i18n checks

iscekic:
Stacked on #5444. Base branch is `mobile-i18n-61ad`, so review this diff on its own.

## Summary

- The app ships in 87 languages instead of 20. Every language on the product list has a mobile catalog and a push notification catalog.
- A cold start parses one catalog, not 87.
- The language picker gains search, reads each language in its own script, and moves into Preferences with Device sessions.
- Two CI checks make the i18n rules mechanical, and they caught real untranslated copy the 20-language PR was already shipping.
- Both review comments on #5444 are resolved.

## Languages

`SUPPORTED_LANGUAGES` is the single source of truth: the catalog loader, the picker, the push catalog, and the iOS bundle all read it.

Portuguese ships twice. The words a phone app uses most are exactly the ones that differ, so `pt` reads *ficheiro, utilizador, ecrã, aplicação* and `pt-BR` keeps *arquivo, usuário, tela, aplicativo*. A `pt-BR` device matches exactly; every other Portuguese tag falls back to `pt`. Six scripts lay out right to left: `ar he fa ur ckb ps`.

`CFBundleLocalizations` now lists every locale. Without it iOS treats the app as English-only, so OS-drawn text we cannot translate — the native Sign in with Apple button above all — stays English on a localized device. That button's label comes from the OS and follows the *device* language; there is no locale prop, so an in-app override that differs from the device still shows the device's wording.

## Performance

`src/i18n/catalogs.ts` holds one loader per language, and the i18next backend calls only the active one. Metro keeps every catalog in the bundle, but a start parses one instead of 87.

`src/lib/intl-cache.ts` caches `Intl` formatters by locale and options. A session list was building a currency and a relative-time formatter per row per render, and constructing one is the expensive part. Three formatters also read the wrong locale: session day headings were pinned to `en-US`, and message time labels took the device locale instead of the active language.

The picker renders 87 rows through a `FlatList`.

## The picker

Rows lead with the endonym, because someone lost in a language they cannot read scans for their own name. The English name is the dimmed second line, so the row is findable either way. No flags: flags mark countries, and Arabic, Portuguese, Spanish, Swahili, and Kurdish have no correct one. Search matches the endonym, the English name, and the tag, and folds diacritics, so `espanol` finds `Español`. The active language pins to the top, each row reads in its own direction, and an RTL reload now returns to Preferences.

## The checks

`check:i18n` runs both, in CI:

- **`no-literal-copy`** (hardened). It now reports a copy prop on an object literal and a `SCREAMING_SNAKE` constant in a copy position, and its old test-ID pattern no longer waves through an ordinary lowercase word — `title="settings"` and `accessibilityLabel="hello"` were both silently exempt. Test files are out of scope. It runs over `packages/app-shared` too, where an English Kilo Pass string had been shipping to every locale.
- **`tools/i18n/check-catalogs.mjs`** (new). Proves each locale is a key-for-key copy of English: missing, extra, duplicate, and empty keys, matching `{{token}}` sets, every used key defined, and every defined key used. The dead-key direction reads every string literal, not just `t()` arguments, so a key named in a lookup table counts as used.

What they caught, all now in the catalog: the auth error map (15 strings), the default and device-admission errors, the Kilo Pass legal disclosure and its link labels, the PR link toasts, the free-model labels, the three remote-spawn toasts, the session-exited toast, and the voice input labels. Fifteen dead keys deleted. Four constants are allowlisted by name — a brand, an acronym, a number, and a URL — so the exemption list is the review record.

## Review comments on #5444

- **Push locale cache.** It was keyed by locale alone, so it could answer `already-registered` for another account's row — one device keeps one Expo token across a sign-out and sign-in. `getMyPushTokens` now returns `locale` and the server row decides; the client keeps only a hint that can bypass the spacing skip, never suppress a write. The Notifications screen also registered without a locale, so a device enrolled there received English pushes.
- **`TEST_ID_PATTERN`.** Removed from the shared ignore, as described above.

## Review comments on this PR

- **`toSorted`.** Hermes does not implement it, so the picker crashed on open. The rows now sort a mapped copy with `.sort()`, the pattern the repository already uses for this reason.
- **The search field.** It controlled its text with `value` plus state, which drops keystrokes on iOS. The field is uncontrolled now; the state only drives the filtered list.

## Diffs

Translated catalogs are marked `linguist-generated`, so a review reads the English change instead of 86 copies of it.

## Verification

`pnpm --filter kilo-app run lint` 0, `typecheck` clean, `check:i18n` clean, 5800 mobile tests and 22 notification tests pass.


pandemicsyn:
Little misscomm but lgtm.

PR #5453, opened 2026-08-24
feat(auth): add email-first sign-in

pandemicsyn:
## Summary

- Changes normal web sign-in from provider-first to email-first while preserving returning-user, invite, explicit signup, and get-started behavior.
- Discovers required SSO or linked authentication methods for the submitted email, automatically continues single-method flows, and shows filtered Google-first choices for multi-provider accounts.
- Shows the complete eligible account-creation methods for unknown emails and hardens discovery with Turnstile, fail-closed IP/email limits, normalized identity ambiguity handling, and an indexed case-insensitive provider-email lookup.
- Adds cancellation-safe Turnstile and magic-link transitions, accessible form/error states, and focused coverage across the route, hook, components, navigation, eligibility, and user lookup.

## Verification

- [x] Exercised the initial email-first sign-in screen on the local Next.js app at desktop width.
- [x] Confirmed a valid email enables Continue.
- [x] Exercised an unknown email against the real local discovery endpoint and confirmed the complete signup method list with Google first.
- [x] Exercised the existing multi-provider selection state in the local app using a deterministic mocked discovery response.

## Visual Changes

| Before | After |
|---|---|

First sign in

01-email-first-sign-in-desktop
02-existing-account-methods-google-first
03-unknown-email-account-creation

Returning (with github and gmail linked)

Screenshot 2026-08-25 at 9 29 02 AM

05-valid-email-continue-enabled

## Reviewer Notes

- Adds Drizzle migration `0224_medical_george_stacy`, concurrently creating `IDX_user_auth_provider_lower_email` on `lower(user_auth_provider.email)`.
- Discovery intentionally fails closed if either infrastructure-managed Vercel Firewall policy is unavailable.
- Callback-side SSO enforcement and provider identity checks remain authoritative; the submitted email is routing input, not authentication.
- Cancellation suppresses stale client transitions but cannot revoke an already-started magic-link request or external redirect.
- Independent static review/remediation completed with no remaining findings.


apps/web/src/components/auth/SignInForm.tsx
```diff
@@ -68,8 +63,11 @@ export function SignInForm({
     );
   }
 
-  // Show error notification at the top level for all states that can display errors
-  const errorNotification = flow.error ? <AuthErrorNotification error={flow.error} /> : null;
+  // Syntax errors belong to the email field. Discovery and provider errors need
+  // one global live region rather than being announced twice.
+  const fieldError = !flow.emailValidation.isValid ? flow.error : undefined;
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in ca3f3a33d. Callback auth codes now remain global notifications; only schema validation errors render beside the email field.

apps/web/src/hooks/useSignInFlow.ts
```diff
@@ -566,53 +774,73 @@ export function useSignInFlow({
...
     newParams.delete('email');
     newParams.delete('org');
-    window.history.replaceState({}, '', `${window.location.pathname}?${newParams.toString()}`);
+    const query = newParams.toString();
+    window.history.replaceState(
+      {},
+      '',
+      query ? `${window.location.pathname}?${query}` : window.location.pathname
+    );
     setEmailState('');
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in ca3f3a33d. Clearing an invite now suppresses stale query prefill and clears the email before URL replacement.

apps/web/src/hooks/useSignInFlow.ts
```diff
@@ -234,118 +351,184 @@ export function useSignInFlow({
...
       }
     }
-  }, [params.email, storybookInitialState, initialError]);
+  }, [
+    params.email,
+    storybookInitialState,
+    initialError,
+    isSignUp,
+    isInviteCleared,
+    tier,
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in ca3f3a33d. Query-prefill verification now runs only while the flow remains on the landing state.

apps/web/src/hooks/useSignInFlow.ts
```diff
@@ -234,118 +352,186 @@ export function useSignInFlow({
...
       }
     }
-  }, [params.email, storybookInitialState, initialError]);
+  }, [
+    params.email,
+    storybookInitialState,
+    initialError,
+    isSignUp,
+    isInviteCleared,
+    flowState,
```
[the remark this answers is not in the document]
pandemicsyn: Fixed in e4d88c85c. Query-prefilled emails now auto-verify at most once per distinct email during the mounted flow. Returning from provider selection to email entry no longer reopens Turnstile or repeats discovery, while a genuinely changed query email still receives one automatic attempt. Added mounted hook regressions for all three paths.

apps/web/src/app/api/sso/organizations/route.ts
```diff
@@ -32,117 +51,133 @@ export async function POST(request: Request): Promise<NextResponse> {
...
+      checkRateLimit(DISCOVERY_IP_RATE_LIMIT_ID, { request }),
+      checkRateLimit(DISCOVERY_EMAIL_RATE_LIMIT_ID, {
+        request,
+        rateLimitKey: discoveryEmailRateLimitKey(email),
+      }),
+    ]);
+    if (ipLimit.rateLimited || emailLimit.rateLimited) {
+      return NextResponse.json({ error: 'Please try again later.' }, { status: 429 });
+    }
+    if (ipLimit.error || emailLimit.error) {
```
iscekic, line 71: **Check the deploy order.** `checkRateLimit` returns `error: 'not-found'` when the firewall rule ID does not exist. This branch turns that into a 503, so a missing or renamed `sign-in-discovery-ip` / `sign-in-discovery-email` rule takes down every sign-in, not just the rate limit.

The existing pattern in `apps/web/src/lib/auth/data-export-download-code-rate-limit.ts:30` logs `not-found` and continues. Either provision both rules before this merges, or split the handling: fail closed on `'blocked'`, log and continue on `'not-found'`.

apps/web/src/app/api/sso/organizations/route.ts
```diff
@@ -32,117 +51,133 @@ export async function POST(request: Request): Promise<NextResponse> {
...
+        extra: {
+          ipLimiterError: ipLimit.error ?? null,
+          emailLimiterError: emailLimit.error ?? null,
+        },
+      });
+      return NextResponse.json({ error: 'Please try again later.' }, { status: 503 });
+    }
+
+    const providerLookup = await getAllUserProviders(email);
+    if (providerLookup.kind === 'ambiguous') {
```
iscekic, line 84: The `ambiguous` result returns 503 with no Sentry event. `kilocode_users.normalized_email` carries only an index (`packages/db/src/schema.ts:507`), not a unique constraint, so duplicate normalized emails can already exist in production.

An affected user then sees "Unable to find sign-in methods. Please try again." forever, and nobody gets a signal. Add a `warnInSentry` here with both candidate user IDs.

apps/web/src/components/auth/SignInForm.tsx
```diff
@@ -68,8 +63,11 @@ export function SignInForm({
     );
   }
 
-  // Show error notification at the top level for all states that can display errors
-  const errorNotification = flow.error ? <AuthErrorNotification error={flow.error} /> : null;
+  // Syntax errors belong to the email field. Discovery and provider errors need
+  // one global live region rather than being announced twice.
+  const fieldError = flow.emailValidation.error ?? undefined;
+  const errorNotification =
+    flow.error && !fieldError ? <AuthErrorNotification error={flow.error} /> : null;
```
iscekic: `AuthErrorNotification` matches a fixed list of error codes. Anything else falls through to "Oops! Something went wrong trying to log in."

Every new discovery message the hook sets is free text, so all four collapse into that one generic box:
- `Too many attempts. Please try again later.`
- `Unable to find sign-in methods. Please try again.`
- `No supported sign-in method is available for this account. Use a different email.`
- `Failed to send magic link. Please try again.`

The 429 case reads as a bug to the user. Render `flow.error` directly here, or add codes for these states.

apps/web/src/components/auth/sign-in/EmailInputForm.tsx
```diff
@@ -27,32 +29,51 @@ export function EmailInputForm({
   placeholder = 'you@example.com',
   autoFocus = false,
   disabled = false,
+  isLoading = false,
 }: EmailInputFormProps) {
+  const [hasBlurred, setHasBlurred] = useState(false);
+  const validationError =
+    hasBlurred && email && !emailValidation.isValid ? emailValidation.error : null;
+  const visibleError = validationError ?? error;
```
iscekic: The `hasBlurred` gate never defers anything. `SignInForm.tsx:68` passes `error={flow.emailValidation.error}`, which is the same string as `emailValidation.error`. Before the first blur, `validationError` is `null` and `visibleError` falls through to `error`, so the message still appears on the first keystroke.

Pick one source: keep `hasBlurred` and stop passing the validation error as `error`, or delete `hasBlurred` and its state.

apps/web/src/app/api/sso/organizations/route.ts
```diff
@@ -32,117 +51,133 @@ export async function POST(request: Request): Promise<NextResponse> {
...
-      if (userProviderInfo.providers.length > 0) {
-        const response: SSOOrganizationsResponse = {
-          providers: userProviderInfo.providers,
-          newUser: false,
-        };
-        return NextResponse.json(response);
-      }
+      // Return the matched account even when its methods are unsupported. The
+      // client presents a recoverable error rather than broadening to signup.
+      return discoveryResponse({ kind: 'existing', providers: userProviderInfo.providers });
```
iscekic, line 117: Trace an account whose only linked provider is `workos` while its domain resolves `not_required`. Both SSO branches above return null, so this line returns `{kind: 'existing', providers: ['workos']}`. `resolveSignInMethods` filters `workos` out of `ProdNonSSOAuthProviders`, so the client reaches `no-supported-method` and shows "Use a different email" (`useSignInFlow.ts:508`).

The only way back is the Enterprise SSO footer link, and the message does not name it. Point that error at Enterprise SSO.

apps/web/src/app/api/sso/organizations/route.ts
```diff
@@ -32,117 +51,133 @@ export async function POST(request: Request): Promise<NextResponse> {
...
-    return NextResponse.json(response);
+    // No organization or provider found → new user signup flow. Resolve the
+    // email option on the server so account-creation rules remain authoritative
+    // without spending the magic-link rate-limit allowance during discovery.
+    const providers = [
+      ...((await isNewAccountEligibleForMagicLink(email))
+        ? ProdNonSSOAuthProviders
+        : ProdNonSSOAuthProviders.filter(provider => provider !== 'email')),
+    ];
+    return discoveryResponse({ kind: 'new', providers });
```
iscekic, line 138: Note for the record: `existing` against `new` is now an explicit account-existence oracle, and the `new` provider list also reveals whether the address passed the blacklist and TLD checks. Turnstile and both rate limits bound it, and email-first sign-in cannot avoid it. I flag it only so the trade-off is a recorded decision.

packages/db/src/migrations/0225_tiny_reavers.sql
```diff
@@ -0,0 +1,3 @@
+COMMIT;--> statement-breakpoint
+CREATE INDEX CONCURRENTLY "IDX_user_auth_provider_lower_email" ON "user_auth_provider" USING btree (lower("email"));--> statement-breakpoint
```
iscekic: `CREATE INDEX CONCURRENTLY` without `IF NOT EXISTS`. If the build fails part way, PostgreSQL leaves an INVALID index behind and the retry fails with "already exists".

`user_auth_provider` is large, so this outcome is realistic. Migrations 0132 and 0168 use `IF NOT EXISTS` for the same reason. Per `packages/db/AGENTS.md` the fix belongs in the schema, not in the generated DDL.

apps/web/src/components/auth/SignInForm.tsx
```diff
@@ -11,9 +11,11 @@ import { FakeLoginForm } from '@/components/auth/FakeLoginForm';
 import { AuthErrorNotification } from '@/components/auth/AuthErrorNotification';
 import { AnimatedLogoMark } from '@/components/AnimatedLogoMark';
 import Link from 'next/link';
-import { Mail, SquareUserRound } from 'lucide-react';
+import { SquareUserRound } from 'lucide-react';
+import React from 'react';
```
iscekic, line 15: Unused import. This file has no `React.` reference. The same applies to `apps/web/src/app/users/sign_in/page.tsx:5` and `apps/web/src/components/auth/sign-in/ProviderSelectView.tsx:5`, which can keep the `{ useState }` part. `TurnstileView.tsx` and `AuthErrorNotification.tsx` do use the namespace, so leave those.

iscekic (approved):
Approved, bot comments below

apps/web/src/app/api/sso/organizations/route.ts
pandemicsyn: Validated as an intentional fail-closed deployment prerequisite. This endpoint is an unauthenticated account/provider-routing oracle after Turnstile, so a missing discovery limiter must not silently remove that control. The PR reviewer notes explicitly require both `sign-in-discovery-ip` and `sign-in-discovery-email` policies to be provisioned before deployment; `not-found` continues to return 503 by design.

apps/web/src/components/auth/sign-in/EmailInputForm.tsx
pandemicsyn: Fixed in rebased commit ab1d8580d. EmailInputForm now owns syntax-error presentation through its blur state; the duplicate parent error channel was removed. Rendered tests cover invalid pre-blur, invalid post-blur with ARIA association, and clearing after valid input.

apps/web/src/components/auth/SignInForm.tsx
pandemicsyn: Fixed in rebased commit ab1d8580d. Discovery throttling/failure, unsupported methods, and magic-link delivery now use a closed set of internal error codes with actionable presentation. Arbitrary callback query errors remain normalized to the generic fallback rather than reflected.

apps/web/src/app/api/sso/organizations/route.ts
pandemicsyn: Validated the trace, but no code change: a stored WorkOS credential is not evidence of a currently active SSO authority. If both linked and primary domains resolve `not_required`, sending the user to Enterprise SSO would retry the same inactive path and may target deprovisioned configuration. The route intentionally fails closed without broadening to signup; this legacy availability case requires explicit account/admin remediation.

apps/web/src/app/api/sso/organizations/route.ts
pandemicsyn: Acknowledged and accepted as an explicit product/security trade-off. Email-first provider routing necessarily distinguishes existing and new accounts. Exposure is bounded by IP-bound Turnstile proof plus per-IP and normalized-email HMAC Firewall limits, while provider callbacks and magic-link authorization independently re-evaluate identity and active SSO policy.

apps/web/src/app/api/sso/organizations/route.ts
pandemicsyn: Fixed in rebased commit ab1d8580d. The ambiguous lookup path now emits a fixed Sentry warning before the generic 503. It intentionally omits email, provider data, and candidate user IDs to comply with the repository PII logging policy; route coverage asserts the signal and response.

packages/db/src/migrations/0225_tiny_reavers.sql
pandemicsyn: Validated, but `IF NOT EXISTS` is not safe recovery for this case: a failed concurrent build can leave a same-name INVALID index, and IF NOT EXISTS would skip it and allow the migration to appear successful without a usable index. Drizzle also has no schema-level `.ifNotExists()` API. The rebased `0225_tiny_reavers` follows recent repo precedent with COMMIT / CREATE INDEX CONCURRENTLY / BEGIN and no IF NOT EXISTS; recovery requires inspecting and dropping/reindexing an invalid index before retry.

apps/web/src/components/auth/SignInForm.tsx
pandemicsyn: No change after validation. Production uses the automatic JSX runtime, but the current `@swc/jest` transform for these suites emits classic `React.createElement` calls. These imports are therefore required at Jest runtime; removing them causes `React is not defined` in the affected component/page tests.

PR #5464, opened 2026-08-25
fix(mobile): clear auth image caches and temp copies

iscekic:
## Summary

- When you sign out or switch accounts, the app now deletes the previous account's cached images and files, so nothing carries over to the next account.
- Image views now hold images only in memory while they are shown, so a viewed image no longer leaves its own saved copy on the device.
- Photos and documents you attach are now deleted automatically after they are sent, and unused copies are deleted after 24 hours.
- A copy made for the share sheet is an exception: the receiving app still reads it after the sheet closes, so it stays in the app cache until the next reap deletes it, at most 24 hours after it was written. Organization invoice PDFs follow the same path.

---

A new temp-file registry tracks every app-owned Picker and share copy so the copies can be deleted
later. `registerTempFile` is idempotent per URI and persists a cache-directory JSON registry that
starts empty when missing or corrupt; `reapTempFiles()` deletes only entries older than 24 hours,
while `reapTempFiles({ all: true })` deletes every entry. The root layout runs the TTL reap at cold
start and whenever the app returns to the foreground, deferred past the interaction frame; the reap
skips the registry write when nothing expired.

<details>
<summary>Files</summary>

- `apps/mobile/src/lib/temp-file-registry.ts` — new module: registerTempFile, reapTempFiles, TEMP_FILE_TTL_MS, and a JSON registry persisted under Paths.cache.
- `apps/mobile/src/app/_layout.tsx` — adds a cold-start reap and an AppState foreground listener, both deferred through scheduleCacheMaintenance, and gates the share ingest on the shell being ready.
- `apps/mobile/src/lib/pending-share-navigation.ts` — adds the `shouldIngestShareIntent` guard and `SHARE_INTENT_OPTIONS` (`resetOnBackground: false`), so a sign-in that leaves the app cannot drop a deferred share payload.
- `apps/mobile/src/components/agents/attachment-picker.ts` — registers each camera, library, and document candidate URI after the picker returns.
- `apps/mobile/src/lib/share-payload.ts` — registers the cache destination after each incoming share file is copied.

</details>

Sign-out and direct account switch now delete the three authenticated image caches plus every
registered temp file, so one account's media never leaks to the next. Three new best-effort clears
— `clearToolCardImageCache`, `clearFilePartCache`, and `clearClipboardImages` — each delete their
cache directory and reset their in-memory map, and `reapTempFiles({ all: true })` deletes every
registered copy; no call throws. The sign-out path clears after remote cleanup, and the sign-in
path clears after credential publication so a switch cannot resurrect the previous account's files.

<details>
<summary>Files</summary>

- `apps/mobile/src/lib/auth/auth-context.tsx` — emits the three clears plus reapTempFiles all in both the sign-in and the sign-out teardown.
- `apps/mobile/src/components/agents/tool-card-image-cache.ts` — adds clearToolCardImageCache, which deletes the tool-card-images directory and clears the URI map and write dedupe set.
- `apps/mobile/src/components/agents/file-part-cache.ts` — adds clearFilePartCache, which deletes the session-file-parts directory and clears the entry map.
- `apps/mobile/src/lib/agent-attachments/clipboard-image.ts` — adds clearClipboardImages, which deletes the clipboard-images directory.

</details>

Every authenticated expo-image render now sets `cachePolicy="memory"`, so the viewed bytes never
enter the image disk cache. The change covers the file-part image, the tool-card attachment image,
the full-screen viewer, the PR review author and comment avatars, and the share-payload preview
thumbnails. The cost is a fresh decode on each mount instead of a reusable disk cache.

<details>
<summary>Files</summary>

- `apps/mobile/src/components/agents/file-part-renderer.tsx` — renders the file-part image with the memory policy.
- `apps/mobile/src/components/agents/tool-card-image-attachments.tsx` — renders the tool-card attachment image with the memory policy.
- `apps/mobile/src/components/image-viewer-modal.tsx` — renders the full-screen viewer image with the memory policy.
- `apps/mobile/src/components/pr-review/discussion/comment-row.tsx` — renders the comment author avatar with the memory policy.
- `apps/mobile/src/components/pr-review/pr-review-overview-parts.tsx` — renders the overview author avatar with the memory policy.
- `apps/mobile/src/components/share/share-payload-preview.tsx` — renders the share preview thumbnails with the memory policy.

</details>

Temp copies are released as soon as their consumer is done with them. The new-session creator resets
the composer's local attachment copies once the session handoff succeeds. `shareMaterializedRemoteFile`
takes the opposite route on purpose: the receiving app reads the file after the share sheet returns, so
a successful share registers the copy for the reap instead of deleting it, and only the failure path —
where nothing consumed the copy — deletes immediately. A shared copy therefore lives in the app cache
for up to 24 hours, organization invoice PDFs included, since they reach the share sheet through
`organization-invoice-download.ts`. The trade is deliberate: a truncated read in the receiving app is
worse than a cache copy the reap removes within a day.

<details>
<summary>Files</summary>

- `apps/mobile/src/lib/share-remote-file.ts` — registers the materialized file for the reap after a successful Android share, deletes it on the failure path, and drops the now-unused Platform import.
- `apps/mobile/src/components/agents/use-new-session-creator.ts` — calls attachments.reset after a successful prepareSession, before navigation.

</details>

Tests: 5 test files updated — `temp-file-registry.test.ts` (new) and `tool-card-image-cache.test.ts`, `file-part-cache.test.ts`, `share-remote-file.test.ts`, `pending-share-navigation.test.ts`.
Generated: none.

---

## Verification

- [x] No manual test paths for this level: automated tests cover the unit logic, and runtime verification runs on the tip PR, not on this level.

## Visual Changes

Visual Changes: N/A

## Reviewer Notes

Human steps: none needed.

Notes:
- E2E: bot-e2e — runtime verification runs on the tip PR, not this level.

Full verification (E2E, user advocacy, simplify, bot review) runs on the tip PR over every level.
A review comment on a lower PR is answered on the tip. A red check on a level is repaired on that level.

1. `audit-w7c-privacy-security-9cac` — #5458
2. `audit-w7c-privacy-security-9cac-s2` — #5461
3. `audit-w7c-privacy-security-9cac-s3` — #5464 ← this PR
4. `audit-w7c-privacy-security-9cac-s4` — #5466
5. `audit-w7c-privacy-security-9cac-s5` — #5469
6. `audit-w7c-privacy-security-9cac-s6` — #5470
7. `audit-w7c-privacy-security-9cac-s7` — #5471 (tip)


iscekic:
All three warnings were valid. Fixed in 61c9692:

- `share-remote-file.ts:82` (Android delete race): VALID — the success path now calls `registerTempFile(file.uri)` instead of `file.delete()`, so the TTL reap owns the copy after the receiver finishes reading (`apps/mobile/src/lib/share-remote-file.ts:80-90`). The failure path still deletes at once.
- `_layout.tsx:897` (cold start never reaps): VALID — `AppState` is already `active` at launch, so no change event fires. The effect now also runs one `reapTempFiles()` on mount (`apps/mobile/src/app/_layout.tsx:894-899`).
- `auth-context.tsx:234` (sign-in reap kills an in-flight share): VALID, fixed at the root. The share ingest now waits for `isShellReady`, so the copy never happens before the sign-in reap runs (`apps/mobile/src/app/_layout.tsx:537-543`). The gate already opened only under `isShellReady`, so nothing is lost by deferring the copy.

Checks: format, typecheck, lint, check:unused, and 6096 mobile tests pass.

iscekic:
@kilocode-bot review the latest head

iscekic:
Residual items from the last review, addressed in 5017ba9.

- **W1** — already fixed in b5ddc87: `registerTempFile` refreshes `createdAt` on re-registration (`apps/mobile/src/lib/temp-file-registry.ts:69-79`), covered by `temp-file-registry.test.ts:73`.
- **W2** — fixed: the provider now takes `SHARE_INTENT_OPTIONS` (`resetOnBackground: false`) at `apps/mobile/src/app/_layout.tsx:920`. Confirmed against `expo-share-intent@8.0.1` (`build/useShareIntent.js:11-15,70`), where the default resets the intent on `active → background`. We chose the option over an in-flight pin because it removes the payload loss without touching the reap, so the sign-in reap gate stays exactly as it is.
- **S1** — fixed: the PR body no longer claims a successful Android share deletes the copy, and now states that a shared copy lives in the app cache for up to 24 hours, organization invoice PDFs included.
- **S2** — fixed: the ingest guard is now `shouldIngestShareIntent` in `apps/mobile/src/lib/pending-share-navigation.ts:27`, with a test that asserts it tracks `isShellReadyForShare` for every flipped guard. `reapTempFiles` returns before `persistEntries` when nothing expired (`temp-file-registry.ts:97-101`), so a cold start and every foreground stop rewriting the registry unchanged.

Checks from `apps/mobile`: format, typecheck, lint, check:unused, and 6106 tests pass.

RSO (approved):

PR #5471, opened 2026-08-25
fix(mobile): union Android backup rules and enable shrinking

iscekic:
Android cloud backups and device-to-device transfers no longer copy the secure key store or the AppsFlyer attribution data. All other app data still backs up as before.

The Android app download is now smaller, because the build no longer bundles resources the app never uses.

Linked images in chat now offer their link options the way text links do: a long-press opens them on a loaded or failed image, and screen readers can reach them on a blocked, failed, or unconfirmed image.

---

The media proxy now classifies addresses with the shared `@kilocode/mcp-gateway` helpers and enforces the 10 MiB body cap by streaming. `isIpAddress` and `isPublicIp` replace the `ipaddr.js` range check, so SSRF blocking uses one classifier shared with the gateway. The body reader cancels the upstream stream the moment it exceeds the cap instead of buffering the whole response, and the `ipaddr.js` dependency is removed from the web app.

<details>
<summary>Files</summary>

- `apps/web/src/lib/media-proxy.ts` — swaps the classifier for `isIpAddress`/`isPublicIp` and adds the streaming capped body reader that cancels at the limit.
- `apps/web/package.json` — drops `ipaddr.js` from dependencies.

</details>

Attachment ledger linking is now server-authoritative, and the web client no longer links before send. The finalize step only builds the payload, so a failed or abandoned send leaves rows pending for the 24-hour reaper instead of linked rows the reaper never deletes. Both routers already link after a successful send, and their comments now state that contract.

<details>
<summary>Files</summary>

- `apps/web/src/hooks/useCloudAgentAttachmentUpload.ts` — drops the two link mutations and the pre-send link in the finalize step.
- `apps/web/src/routers/cloud-agent-next-router.ts` — updates the send-path comment to describe server-authoritative linking.
- `apps/web/src/routers/organizations/organization-cloud-agent-next-router.ts` — updates the send-path comment to describe server-authoritative linking.

</details>

Releasing abandoned pending uploads now deletes their private R2 objects before deleting the ledger rows. Object deletion runs first, so a failed delete leaves the row pending for the 24-hour reaper instead of orphaning the object, and a later reap of an already-absent key is a no-op. Removing a composer file still frees the per-message quota immediately.

<details>
<summary>Files</summary>

- `apps/web/src/lib/r2/cloud-agent-pending-uploads.ts` — selects the caller's pending rows and deletes each R2 object before deleting the rows.

</details>

Android backup rules are unified into one rule set that both manifest attributes point at. The expo-secure-store and AppsFlyer SDK conflict over the backup attributes is resolved by writing one union resource and setting both attributes through `tools:replace`. Shared preferences still back up, while the SecureStore file and the AppsFlyer capture files are excluded from cloud backup and device transfer.

<details>
<summary>Files</summary>

- `apps/mobile/plugins/backup/kilo_backup_rules.xml` — new full-backup rules for Android 11 and lower; includes shared preferences and excludes the SecureStore and AppsFlyer files.
- `apps/mobile/plugins/backup/kilo_data_extraction_rules.xml` — new data-extraction rules for Android 12 and higher, with the same exclusions in the cloud-backup and device-transfer blocks.
- `apps/mobile/plugins/withAndroidManifestFix.js` — copies both XML resources and an unused raw sentinel into the generated res directory and points both manifest attributes at the union rules.

</details>

Release builds now enable resource shrinking, and the artifact inspector fails when the shrink sentinel survives. The plugin writes an unreferenced raw resource that shrinking strips, so its presence in the AAB proves the shrink step was skipped. The inspector also guards its main path behind a direct-run check so tests can import it.

<details>
<summary>Files</summary>

- `apps/mobile/app.config.ts` — turns on `enableShrinkResourcesInReleaseBuilds` next to the existing minify flag.
- `scripts/inspect-mobile-artifacts.mjs` — adds `checkResourceShrinking`, calls it as a failure condition, and guards `main()` behind a direct-run check.

</details>

The full-screen markdown image viewer now fetches through the authenticated media proxy, and image chips expose their link actions to screen readers and long-press. The viewer receives the proxy-resolved URI and bearer auth headers, so the device never requests the source host directly at full size. A long-press on a failed or loaded linked image opens the link actions, and blocked, retry, and confirmed chips register a `showLinkActions` accessibility action.

<details>
<summary>Files</summary>

- `apps/mobile/src/components/agents/markdown-image.tsx` — passes the proxy URI and auth headers to the viewer; adds the link action and long-press to the blocked, retry, and confirmed chips.
- `apps/mobile/src/components/image-viewer-modal.tsx` — adds the headers prop and forwards it to the image source.

</details>

Stripped image copies now join the temp-file registry. The upload hook registers the new cache file produced by metadata stripping, so sign-out reaping deletes it instead of leaking the stripped copy across accounts. The original picker URI stays registered separately.

<details>
<summary>Files</summary>

- `apps/mobile/src/lib/agent-attachments/use-agent-attachment-upload.ts` — registers the stripped URI after the metadata strip.

</details>

The sign-out cleanup comment is corrected; the synchronous preference clears are now described as best-effort only. The old note claiming a synchronous throw falls through to the state reset is removed. No code behavior changes in this file.

<details>
<summary>Files</summary>

- `apps/mobile/src/lib/auth/auth-context.tsx` — rewords the sign-out preference-clear comment.

</details>

Test: 4 test files changed — `scripts/inspect-mobile-artifacts.test.mjs` added; `apps/mobile/src/components/agents/markdown-image.test.ts`, `apps/web/src/lib/media-proxy.test.ts`, and `apps/web/src/lib/r2/cloud-agent-pending-uploads.test.ts` updated.
Generated: 1 lockfile updated — `pnpm-lock.yaml`.

---

## Visual Changes

**Markdown chat, external link tap (iOS).** A link tap now shows a confirm Alert before the browser opens. The picture shows the Alert at screen center: the title, the example.com body, and the Open, Trust this host, Cancel buttons.

**Markdown chat, image blocks (iOS).** An HTTPS image now waits behind a Load control, and an http image shows a static host chip. No picture shows this screen: the handoff lists two candidates, but both are full black frames; they show neither the Load control nor the chip.

**In-app browser after Trust this host (iOS).** Trust this host now opens the link in an in-app browser and trusts the host. The picture shows the loaded Example Domain page under the example.com URL bar, over the chat.

**Trusted hosts screen (iOS).** Revoke now removes the host row and leaves the empty state. The picture shows the Trusted hosts header and the No trusted hosts empty state after the revoke tap.

**Markdown chat after sign-out and sign-in (iOS).** Sign-out now clears trusted hosts, so the Alert asks again after sign-in. The picture shows the Open external link? Alert over the chat after re-login.

**Android Recents, agent chat card.** The Recents card for agent chat now shows an opaque cover, not message text. The picture shows the card as a uniform blank panel with no readable text.

**Android Recents, profile card.** The Recents card for profile now shows an opaque cover, not the email or balance. The picture shows the card as a uniform blank panel with no readable text.

**Android Recents, home card.** The home card keeps its content in Recents; no cover is forced. The picture shows the Welcome to Kilo home content inside the task card.

## Verification

Four rounds on iOS and Android ran 9 scenarios; all 9 pass.

|---|---|---|---|
| ios-inert-https | An HTTPS image shows a Load control and renders no pixels before the tap. | iOS | passed |
| ios-http-chip | An http image shows a host chip, and a tap on it fetches nothing. | iOS | passed |
| ios-host-alert | A markdown link tap shows the host confirm Alert before the browser opens. | iOS | passed after the fix in round 3 |
| ios-trust-skip | Trust this host opens the link, and a later link to the host skips the Alert. | iOS | passed after the fix in round 3 |
| ios-revoke | The Trusted hosts screen removes the host row on revoke. | iOS | passed after the fix in round 3 |
| ios-logout-clears | Sign-out clears trusted hosts, and the Alert returns after sign-in. | iOS | passed after the fix in round 3 |
| android-chat-cover | The Recents card for agent chat hides message text under an opaque cover. | Android | passed |
| android-profile-cover | The Recents card for profile hides the email and balance under an opaque cover. | Android | passed |
| android-home-no-cover | The Recents card for home shows the home screen; no cover is forced. | Android | passed |

Host-confirm Alert: absent within 15 s on the unfixed build (rounds 1 and 2); the fixture re-seed fixed it; round 3 does not reproduce it.

## Human Steps

No human steps are needed before or after merge.

## Notes

None.

Full verification (E2E, user advocacy, simplify, bot review) runs on the tip PR over every level.

1. `audit-w7c-privacy-security-9cac` — #5458
2. `audit-w7c-privacy-security-9cac-s2` — #5461
3. `audit-w7c-privacy-security-9cac-s3` — #5464
4. `audit-w7c-privacy-security-9cac-s4` — #5466
5. `audit-w7c-privacy-security-9cac-s5` — #5469
6. `audit-w7c-privacy-security-9cac-s6` — #5470
7. `audit-w7c-privacy-security-9cac-s7` — #5471 ← this PR (tip)


iscekic:
CRITICAL (cloud-agent-pending-uploads.ts:120) — valid, fixed in 18e4561b0.

`releasePendingUploads` now claims with a single `DELETE ... RETURNING` and deletes the R2 objects only for the claimed keys (apps/web/src/lib/r2/cloud-agent-pending-uploads.ts:120). This is the same atomic-claim idiom `reapAbandonedUploads` already uses, so a `linkPendingUploads` that lands during a release updates no row and its sent object is never deleted. The R2-failure test was replaced with one asserting the row is already claimed when the object delete runs; it fails against the previous select-then-delete order.

apps/web/src/lib/media-proxy.ts
```diff
@@ -1,5 +1,5 @@
 import { lookup } from 'dns/promises';
-import ipaddr from 'ipaddr.js';
+import { isIpAddress, isPublicIp } from '@kilocode/mcp-gateway';
```
RSO: Does it actually make sense that these helpers are in the mcp-gateway package?

RSO (approved):
It's not entirely clear to me why these different changes were grouped together in a single PR

apps/web/src/lib/media-proxy.ts
iscekic: (bot) Yes — these helpers are the repo's canonical SSRF classifier: `apps/web/src/lib/mcp-gateway/discovery-service.ts:45-78` already imports `isIpAddress`/`isPublicIp` from `@kilocode/mcp-gateway` to reject non-public IP literals and DNS results, and `apps/web` depends on the package in 24 files. Reusing them here (decision: replace the new `ipaddr.js` third-party dependency with the repo-owned classifier) is deliberate — it removes supply-chain surface without copying a range list. A neutral-package extraction would churn the shared package's public API and every consumer, which is outside this section's scope.

PR #5476, opened 2026-08-25
feat(web): enable self-service account deletion

eshurakov:
## Summary

- Add a profile Danger Zone flow with email challenge confirmation and sign-out.
- Prevent self-service deletion for organization-managed accounts using the primary eligibility flag.
- Add a completion email queue step with durable checkpoints, ambiguity handling, pre-deployment compatibility, and admin visibility.
- Add the transactional email template, migration, and focused regression coverage.

Screenshot 2026-08-25 at 14 58 16
Screenshot 2026-08-25 at 14 58 26
Screenshot 2026-08-25 at 15 08 02

## Verification

- Local account-deletion flow tested; screenshot to follow.
- `pnpm --filter web exec jest --watchman=false --runInBand --runTestsByPath src/app/admin/deletion-queue/deletion-queue-format.test.ts src/components/profile/DeleteAccountDialog.test.ts` — 2 suites, 28 tests passed
- Changed-file Oxlint — passed
- Changed-file Oxfmt — passed


apps/web/src/lib/user/deletion-queue/deletion-completion.ts
```diff
@@ -89,7 +89,7 @@ export async function tryCompleteRequest(requestId: string): Promise<boolean> {
       .where(eq(user_deletion_steps.request_id, requestId))
       .for('update');
 
-    const required = catalogForVersion(request.catalog_version).map(entry => entry.stepKey);
+    const required = steps.map(step => step.step_key);
```
pandemicsyn, line 92: bot: Completion now treats every currently materialized row as the required set, so a request missing any other catalog finalization step can be scrubbed as completed.

Suggested fix: Derive required keys from `catalogForVersion(request.catalog_version)` and explicitly allow only the documented legacy exception for an absent `CompletionEmail` row. Apply the same narrow exception in task dependency selection rather than filtering all absent dependencies. Add tests that deleting a non-email finalization row (for example `PylonReply`, `PylonFinalize`, or `CsaSupportDb`) neither permits dependent work to run nor permits request completion; retain the existing test that an old request without `CompletionEmail` can complete.

pandemicsyn (approved):

apps/web/src/lib/user/deletion-queue/deletion-completion.ts
eshurakov: Fixed in 8ce5c776a. Shared catalog validation now requires every versioned deletion step except the legacy v2 CompletionEmail exception, so both task selection and completion reject missing mandatory rows. Regression coverage includes missing PylonReply, PylonFinalize, and CsaSupportDb steps, blocked dependent work and completion, and preserved legacy compatibility.

PR #5508, opened 2026-08-26
feat(cloud-agent-next): add interactive control-plane terminals

eshurakov:
## Summary

Add interactive terminals for `workspace_*` sessions on the existing call-home control plane. Shared control messages negotiate terminal lifecycle; authenticated single-use reverse WebSockets relay raw PTY frames through `SandboxSession`. Legacy `agent_*` terminals are unchanged.

PTYs are bound to owner, session, root Kilo session, workspace directory, sandbox, wrapper runtime, and bridge generation. Credentials are scrubbed from the attach environment. Vercel billing stays fail-closed.

### DIFF
--- a/services/cloud-agent-next/src/sandbox-session/terminal-lifecycle.ts
+++ b/services/cloud-agent-next/src/sandbox-session/terminal-lifecycle.ts
@@ -323,7 +323,7 @@ export function createSandboxTerminalLifecycle(deps: TerminalLifecycleDeps) {
     control: TerminalControl,
     session: SessionRequestIdentity,
     ptyId: string
-  ): Promise<void> {
+  ): Promise<boolean> {
     const payload = sessionTerminalClosePayloadSchema.parse({ ptyId });
     try {
       const response = await control.request({
@@ -331,9 +331,11 @@ export function createSandboxTerminalLifecycle(deps: TerminalLifecycleDeps) {
         session,
         payload,
       });
-      if (response.ok) sessionTerminalCloseResultSchema.safeParse(response.result);
+      if (!response.ok) return false;
+      const result = sessionTerminalCloseResultSchema.safeParse(response.result);
+      return result.success && result.data.success;
     } catch {
-      return;
+      return false;
     }
   }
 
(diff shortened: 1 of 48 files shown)


services/cloud-agent-next/src/sandbox-session/terminal-lifecycle.ts
```diff
@@ -0,0 +1,849 @@
...
+    const payload = sessionTerminalClosePayloadSchema.parse({ ptyId });
+    try {
+      const response = await control.request({
+        operation: 'session.terminal.close',
+        session,
+        payload,
+      });
+      if (response.ok) sessionTerminalCloseResultSchema.safeParse(response.result);
+    } catch {
+      return;
```
pandemicsyn: bot: Terminal close is reported as successful even when the wrapper did not delete the PTY.

Suggested fix: Have `closeWrapperTerminal` return a validated close result and require `response.ok`, a successful `sessionTerminalCloseResultSchema` parse, and `success === true` before calling `markEnded`. If the request fails or deletion returns false, keep a durable pending-cleanup state (or leave the record running), return a retryable failure, and reconcile/retry cleanup. Add coverage for both a control-request failure and `{ success: false }` from the wrapper.

pandemicsyn (approved):
lgtm, one from the bot inline

services/cloud-agent-next/src/sandbox-session/terminal-lifecycle.ts
eshurakov: Fixed in e29ccdc70. Explicit close now requires `response.ok`, a schema-valid close result, and `success === true` before marking the terminal ended or closing its bridge. Failures return retryable `SERVICE_UNAVAILABLE` and preserve durable running/operation state so the same close can be retried, including after lifecycle recreation. Existing deletion/revocation fencing is unchanged.

Added regression coverage for rejected requests, `ok: false`, `{ success: false }`, invalid results, live sockets after a failed close, successful retry, and deletion races. Passed: 177 targeted Node tests, 61 Workers integration tests, 16 wrapper tests, package lint/typecheck, focused integration-test lint/typecheck, and formatting. The supplemental all-tests `test/tsconfig.json` check still reports unrelated existing test typing errors.

PR #5533, opened 2026-08-27
fix(kilo-pass): prevent duplicate checkout sessions

pandemicsyn:
## Summary

Prevent concurrent Kilo Pass purchase requests from creating multiple Stripe Checkout Sessions and subscriptions.

- coordinate checkout creation per user across the generic Kilo Pass and KiloClaw upsell entrypoints
- reuse an existing open Kilo Pass Checkout Session and reject live Stripe subscriptions not yet persisted by webhooks
- paginate Stripe session and subscription checks and preserve legitimate resubscription after ended subscriptions

## Verification

- [ ] No manual Stripe checkout was run because this change targets a provider concurrency race and local checkout uses mocked Stripe APIs.

## Visual Changes

N/A

## Reviewer Notes

Automated verification completed:

- `pnpm --filter web lint`
- targeted Jest coverage for `createCheckoutSession` and `createKiloPassUpsellCheckout` (24 passed)
- independent static review repeated after remediation until clean

The first request to acquire checkout coordination wins. A competing request reuses that open Checkout Session, including its selected tier, cadence, and optional hosting intent.

### DIFF
--- a/apps/web/src/lib/kilo-pass/checkout-session.ts
+++ b/apps/web/src/lib/kilo-pass/checkout-session.ts
@@ -1,0 +1,79 @@
+import { db } from '@/lib/drizzle';
+import { getKiloPassStateForUser } from '@/lib/kilo-pass/state';
+import { isStripeSubscriptionEnded } from '@/lib/kilo-pass/stripe-subscription-status';
+import { client as stripe } from '@/lib/stripe-client';
+import { TRPCError } from '@trpc/server';
+import { sql } from 'drizzle-orm';
+import type Stripe from 'stripe';
+
+type CreateOrReuseKiloPassCheckoutSessionParams = {
+  userId: string;
+  stripeCustomerId: string;
+  createSession: () => Promise<Stripe.Checkout.Session>;
+};
+
+export async function createOrReuseKiloPassCheckoutSession(
+  params: CreateOrReuseKiloPassCheckoutSessionParams
+): Promise<{ url: string | null }> {
+  return db.transaction(async tx => {
+    await tx.execute(
+      sql`SELECT pg_advisory_xact_lock(hashtextextended(${`kilo-pass:checkout:${params.userId}`}, 0))`
+    );
+
+    const existing = await getKiloPassStateForUser(tx, params.userId);
+    if (existing && !isStripeSubscriptionEnded(existing.status)) {
+      throw new TRPCError({
+        code: 'BAD_REQUEST',
+        message: 'You already have an active Kilo Pass subscription.',
+      });
+    }
+
+    const isUserKiloPassSession = (session: Stripe.Checkout.Session) =>
+      session.metadata?.type === 'kilo-pass' && session.metadata.kiloUserId === params.userId;
+    let checkoutCursor: string | undefined;
+    do {
+      const openSessions = await stripe.checkout.sessions.list({
+        customer: params.stripeCustomerId,
+        status: 'open',
+        limit: 100,
+        ...(checkoutCursor ? { starting_after: checkoutCursor } : {}),
+      });
+      const openKiloPassSession = openSessions.data.find(isUserKiloPassSession);
+      if (typeof openKiloPassSession?.url === 'string') {
+        return { url: openKiloPassSession.url };
+      }
+      checkoutCursor = openSessions.has_more ? openSessions.data.at(-1)?.id : undefined;
+    } while (checkoutCursor);
+
+    let subscriptionCursor: string | undefined;
+    let hasLiveStripeSubscription = false;
+    do {
+      const subscriptions = await stripe.subscriptions.list({
+        customer: params.stripeCustomerId,
+        status: 'all',
+        limit: 100,
+        ...(subscriptionCursor ? { starting_after: subscriptionCursor } : {}),
+      });
+      hasLiveStripeSubscription = subscriptions.data.some(
+        subscription =>
+          subscription.metadata.type === 'kilo-pass' &&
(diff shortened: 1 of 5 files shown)


pandemicsyn:
## Hyper Review Audit — PR #5533

Reviewed statically against merge-base of `main` @ `75f6f14`. Scope: `apps/web/src/lib/kilo-pass/checkout-session.ts`, both router call sites, and new tests. No tests/typecheck were run; this is a static review.

---

### Must-fix

**1. Open-session reuse is request-shape-blind: a stale session for another tier/cadence/hosting plan is handed back**
`apps/web/src/lib/kilo-pass/checkout-session.ts:31-32`

The reuse predicate matches only `metadata.type === 'kilo-pass' && metadata.kiloUserId === userId` and ignores the `tier`, `cadence`, `affiliateTrackingId`, and (upsell path) `kiloclawHostingPlan`/`kiloclawInstanceId`/`kiloclawPriceVersion` metadata the routers write into each session.

- Wrong-product charge: a user who abandons an open Tier19/Monthly session and later clicks Tier49/Yearly receives the old session URL; Stripe charges the old price while the UI showed the new selection. Downstream the webhook derives tier/cadence from subscription metadata, so the user is persisted under the wrong plan.
- KiloClaw upsell: a reused session lacking `kiloclawHostingPlan` skips hosting activation after purchase (`kilo-pass-router.ts:1672` reads that key), so the user pays but gets no hosted instance despite passing eligibility checks in `createKiloPassUpsellCheckout`.
- Test note: `kilo-pass-router.test.ts` ("reuses one open Stripe session for concurrent initial checkout requests") codifies returning one Tier19/Monthly URL to a Tier49/Yearly request — that behavior should not be asserted as correct.

Remediation: pass the product-defining metadata into `createOrReuseKiloPassCheckoutSession` and require exact equality on those keys before reuse (compare `tier`, `cadence`, and for upsell also the `kiloclaw*` fields; deliberately exclude mutable attribution like `affiliateTrackingId`). Prefer `stripe.checkout.sessions.expire()` on non-matching open Kilo Pass sessions so dedupe stays strict. Update/remove the concurrent-reuse test accordingly.

**2. Pooled DB connection + advisory lock held across unbounded Stripe I/O, with no timeout configured on the Stripe client**
`apps/web/src/lib/kilo-pass/checkout-session.ts:18-76`, `apps/web/src/lib/stripe-client.ts:13`

The helper's entire body runs inside `db.transaction(...)`: two potentially multi-page Stripe list scans plus `params.createSession()` all execute while pinning one of only 10 primary pool connections (`drizzle.ts`: `max = 10`) and holding `pg_advisory_xact_lock`. stripe-node defaults to no client-level timeout here, so during a Stripe slowdown a single attempt can hold a connection for minutes; concurrent checkouts across distinct users can exhaust the entire primary write pool app-wide.

Remediation (any combination): configure `timeout`/`maxNetworkRetries` on the shared client or per-request options on these calls; cap the pagination loops (e.g., 2–3 pages); ideally restructure so Stripe scans run outside the transaction and only the advisory-lock + cheap local re-check + `createSession()` happen inside it — no DB writes occur in this transaction, so a short lock window preserves correctness.

---

### Suggested

**3. Matching open session with `url === null` is neither reused nor expired** — `apps/web/src/lib/kilo-pass/checkout-session.ts:41-46`
If a matching open session has no URL (Stripe clears it on expiry/completion), the code falls through and creates a new session while the stale one lingers open — recreating the duplicate this PR exists to prevent. Suggested: best-effort `stripe.checkout.sessions.expire(session.id)` on a matched-but-unusable session (catch + log), and add a test for this branch.

**4. Test gaps tied to the introduced behavior** — `kilo-pass-router.test.ts`, `kiloclaw-billing-router.test.ts`
Missing cases: (a) mismatched-tier reuse (see must-fix #1), (b) upsell path reusing a session whose `kiloclaw*` metadata differs from the current request, (c) matched-but-null-URL session, (d) non-Stripe (App Store/Play) active subscription blocking checkout via the helper's DB check.

---

### Nits

**5. `'kilo-pass'` metadata tag string duplicated across three modules** (`checkout-session.ts:32,59`, `kilo-pass-router.ts:2655`, `kiloclaw-router.ts:5285`) — if producer and matcher drift apart, reuse silently disables with no type error. Export a shared constant/factory co-located with the matcher.

**6. Two near-identical paginated do/while loops** (`checkout-session.ts:33-46` vs `48-67`) — extract a small page-walk helper so timeout caps and cursor fixes land once.

**7. Optional-chaining inconsistency** (`checkout-session.ts:59-60`) — `subscription.metadata.type` omits `?.` used by the session matcher above; use `?.` consistently so a null-metadata surface yields BAD_REQUEST instead of TypeError.

---

### Clean areas
Tenant isolation (scans filter by caller's own Stripe customer *and* own `kiloUserId`), the raw-sql advisory lock call (userId is a bound parameter), Zod validation/auth boundaries upstream, and error-message leakage were reviewed and found sound.

pandemicsyn:
Addressed the reported checkout coordination findings in `a91aa63fe`:

- reuse now requires matching tier, cadence, and KiloClaw hosting metadata
- mismatched and null-URL open sessions are expired before replacement
- Stripe list/create/expire calls use 10s request timeouts and scans are capped at three pages while retaining the advisory lock for race safety
- added coverage for tier mismatch, hosting-intent mismatch, null URLs, and active store subscriptions

Kept the advisory-lock design intentionally: moving provider scans outside it would reopen the cross-instance race, while a durable checkout-intent model would add disproportionate schema and lifecycle complexity.

iscekic (approved):

PR #5539, opened 2026-08-27
feat(remote): pick a remote session folder

iscekic:
## Summary

When starting a session on a connected remote instance, a new Folder field lets the user pick a child folder of the instance's launch directory. The user can drill down any number of levels, and Start then opens the session in that folder.

An older connected instance that cannot list folders shows that message, and Start still opens the session in the instance's launch folder.

---

The kilocode CLI adds two launch-directory-scoped contracts: a `list_directories` command that lists one directory level, and an optional `create_session.directory` field that starts the session in a chosen child folder. Both resolve a client-supplied relative path under the launch directory through one helper that rejects absolute paths, escapes, and symlinks that resolve outside it, so the CLI owns path safety. A listing is capped at 256 entries, and old clients that omit `directory` keep the legacy sessionId or launch-directory fallback.

<details>
<summary>Files</summary>

- `packages/opencode/src/kilo-sessions/remote-command.ts` — adds the `list_directories` request, entry, and response schemas and the `MAX_DIRECTORIES` cap.
- `packages/opencode/src/kilo-sessions/remote-sender.ts` — dispatches `list_directories`, resolves `create_session.directory`, and adds the `resolveUnderLaunch` helper plus the one-level `readdir` loop.

</details>

A new SDK helper `listDirectoriesOnConnection` sends a connection-scoped `list_directories` command and classifies every outcome into `ok`, `unsupported`, `invalid`, or `transport`. It never throws and never logs, so the mobile hook maps outcomes to UI states without error handling. The strict `listDirectoriesV1Schema` fails closed on protocol drift, over-limit entries, and malformed entry strings.

<details>
<summary>Files</summary>

- `packages/cloud-agent-sdk/src/list-directories.ts` — new helper with the outcome classification and the unsupported relay-code set; imports the shared retryable set.
- `packages/cloud-agent-sdk/src/instance-model-catalog.ts` — exports the `RETRYABLE_RELAY_CODES` set that the helper now imports.
- `packages/cloud-agent-sdk/src/schemas.ts` — adds `listDirectoriesV1Schema` with the 256-entry and 2000-character caps; the caps stop being exported.
- `packages/cloud-agent-sdk/package.json` — exports the `./list-directories` subpath.

</details>

`createRemoteSessionOnConnection` now forwards an optional `directory` field, and the `:ext`/`:bare` old-CLI retry treats `directory` as an extended field. An old CLI that rejects the extended body still gets the bare `{protocolVersion: 1}` retry, which omits `directory`, so a folder choice degrades to the launch directory instead of failing the spawn. The two retries keep distinct durable mutation identities, so the relay does not replay the extended attempt's error to the bare retry.

<details>
<summary>Files</summary>

- `packages/cloud-agent-sdk/src/create-session.ts` — includes `directory` in the wire data and the extended-field check.
- `packages/cloud-agent-sdk/src/transport.ts` — adds `directory` to `CreateRemoteSessionInput`.

</details>

The user-web relay now admits `list_directories` as a viewer command and maps an old CLI's `unknown command: list_directories` to the structured `CLI_UPGRADE_REQUIRED` error. That mapping is temporary and must be removed once every supported CLI ships the command.

<details>
<summary>Files</summary>

- `services/session-ingest/src/dos/UserConnectionDO.ts` — adds `list_directories` to `ALLOWED_VIEWER_COMMANDS` and `CLI_UPGRADE_REQUIRED_COMMANDS`.

</details>

A new formSheet route lists one directory level at a time beneath the chosen instance's launch directory, with drill-down, in-sheet Back, per-path caching, and skeleton, empty, retryable, and unsupported states. The leading header control is relabeled Back when nested, and Done keeps the launch directory when the CLI cannot list folders. New copy covers the Folder field, the list hint, the empty state, and the retryable and unsupported messages, and the same strings are translated into every other catalog.

<details>
<summary>Files</summary>

- `apps/mobile/src/app/(app)/agent-chat/folder-picker.tsx` — the picker screen with the drill stack and per-phase rendering.
- `apps/mobile/src/lib/hooks/use-list-directories.ts` — the one-level listing client with a generation counter and per-path cache.
- `apps/mobile/src/app/(app)/_layout.tsx` — registers the `agent-chat/folder-picker` formSheet route.
- `apps/mobile/src/components/picker-sheet.tsx` — threads the `cancelLabel` override to the header.
- `apps/mobile/src/components/sheet-header.tsx` — resolves the leading-control label override so Back is announced as Back.
- `apps/mobile/src/i18n/locales/en.json` — adds the `folderPicker` strings.
- `apps/mobile/src/i18n/locales/*.json` — the translation pass fills the `folderPicker` keys in the 86 other catalogs.

</details>

The new-session configure form shows a Folder field for remote targets, whose confirmed relative path is held per instance and cleared when the connection changes. The field hands the connection, project, current path, and callback to the picker through a new folder-picker registry slot, and the confirmed path is threaded into the spawn dispatch as `create_session.directory`. The field shows the project name at launch and the selected folder path once the user has drilled in.

<details>
<summary>Files</summary>

- `apps/mobile/src/components/agents/folder-selector.tsx` — the Folder field and its handoff to the picker route; the component export is pruned.
- `apps/mobile/src/lib/hooks/use-launch-folder.ts` — per-instance folder state that resets when the connection changes.
- `apps/mobile/src/components/agents/new-session-configure-form.tsx` — renders the Folder field for remote targets.
- `apps/mobile/src/components/agents/new-session-screen-body.tsx` — owns the folder state and passes it to the form and the spawn dispatch.
- `apps/mobile/src/components/agents/use-remote-spawn-dispatch.ts` — accepts `directory`, reads it at press time, and keys the operation on it.
- `apps/mobile/src/lib/use-new-session-share-remote.ts` — forwards `folderPath` as `directory` to the dispatch.
- `apps/mobile/src/lib/hooks/remote-instance-spawn-classifier.ts` — includes `directory` in the built `create_session` input.
- `apps/mobile/src/lib/picker-bridge.ts` — adds the `FolderPickerBridge` type.
- `apps/mobile/src/lib/route-registry.ts` — adds the `folderPicker` slot kind.

</details>

Tests: 11 test files changed (9 in `Kilo-Org/cloud`, 2 in `Kilo-Org/kilocode`).
Generated: none.

---

## Verification

On iOS, three cases ran across the six folder rounds.

|---|---|---|---|
| fp-confirm-and-start | Done confirms the child path, and Start opens a session that runs in that child folder. | ios | passed |
| fp-happy-open-and-drill | The sheet lists child folders with the tap-to-list hint and chevron rows, and drill-in and Back navigate. | ios | passed |
| fp-dismiss-and-instance-reset | A swipe-down dismiss keeps the confirmed folder, the re-shown sheet keeps Done tappable, and an instance change resets the folder. | ios | passed after the fix in round folder-r5 |

The folder-r2 round report is missing: the handoff names `e2e-folder-r2-ODY7BP/e2e-round.md`, but that directory is empty. The folder-r3 carry-forward that cites it is unverified; the fp-happy-open-and-drill result rests on folder-r6.

Defects:

- Re-presentation defect (Done and Back unresponsive on the re-shown sheet after a swipe-down dismiss): round folder-r4 reproduced it twice on the unfixed build; commit 48922812a fixed it; rounds folder-r5 and folder-r6 no longer reproduce it.
- ExpoScreenCapture native-module redbox on cold load: repro-r1 reproduced it; it is an environment residual, not a product defect; out of scope, not fixed.

Recordings exist: `repro-r0-ios.mp4` and `repro-r1-ios.mp4` in the scratch folder.

### Folder picker sheet

The user now sees a sheet that lists the remote folders as rows, with a tap-to-list hint and a Done button.
In the sheet, the hint sits above the rows `.config`, `node_modules`, and `xdg`; each row ends in a chevron, not a radio mark.

### New-session form after folder confirmation

After the user picks a child folder and taps Done, the Folder field shows the child path.
Under the Run on row, the Folder field reads `.config`, and the Start session button sits below it.

### Started session in the chosen folder

The session runs in the confirmed child folder, and the pwd reply shows that path.
The agent reply path ends in `remote-session-ux-8f16/.config`, below the `Print working directory` tool row.

## Reviewer Notes

Human steps: none known. The code adds no environment value, secret, migration, or flag to flip, and old CLIs degrade gracefully through the existing retry and unsupported paths.

Notes:
- Folder picking lets the user start a remote session in a chosen child folder.
- Live-list latency is out of scope for this PR (the latency work was not landed; a reproduction did not name the cause).
- This change spans `Kilo-Org/cloud` and `Kilo-Org/kilocode`; the `Kilo-Org/kilocode` PR is this PR's sibling. Both PRs share one body and cross-link.


packages/cloud-agent-sdk/src/transport.ts
```diff
@@ -28,6 +28,8 @@ type CreateRemoteSessionInput = {
     variant?: string;
   };
   orgId?: string;
+  /** Launch directory for the new remote session (`create_session` wire data). */
+  directory?: string;
```
pandemicsyn, line 32: bot: `directory` is added to the shared create-session input contract, but the standard session-manager/live-transport route silently drops it.

Suggested fix: Propagate `input.directory` through `session-manager.ts` when it constructs `merged`, add it to `cli-live-transport.ts`'s `buildCreateSessionWireData`, and include it in `hasExtendedFields` so an old CLI gets the bare fallback. Add tests that invoke `manager.createRemoteSession({ directory: 'child' })` through the live transport and assert both the initial wire payload and the old-CLI fallback behavior. The mobile connection-scoped helper does forward this field, but SDK consumers using the manager/live-transport route currently create the session in the launch directory instead.

pandemicsyn (approved):

PR #5545, opened 2026-08-27
fix(mobile): avoid choosing the wrong GitHub install from a repo link

RSO:
## Why this PR exists

Organizations can now connect more than one GitHub App installation. A pasted or deep-linked `owner/repo` can therefore match multiple mobile repository options, especially while installations overlap during a repository move. The previous first-match behavior could silently choose the wrong installation before the user even sees the new-session form.

This is a small defensive PR: when repository provenance is ambiguous, mobile leaves the selection empty and asks the user to choose rather than guessing.

## Place in the rollout

This is an independent client-safety fix in the broader multiple-GitHub-organizations rollout. It does not require the full mobile installation selector in #5549, and it is useful even before that PR lands.

## What changes

- Resolve pasted and deep-linked repositories case-insensitively only when exactly one loaded option matches.
- Preserve GitHub canonical casing for a unique match.
- Return no preselection for zero or multiple matches.

## What stays unchanged

- Repository links still prefill normally when the match is unique.
- This PR does not change repository rendering, session submission, or continuation behavior. Those are intentionally isolated in #5549 and #5546.

## Review guide

The production change is six lines in `new-session-prefill.ts`; the rest is focused ambiguity coverage. The key question is whether mobile ever selects a repository when provenance is not unique.

## Validation

- Focused mobile prefill tests
- Focused lint and format check
- `git diff --check`

### DIFF
--- a/apps/mobile/src/components/agents/new-session-prefill.test.ts
+++ b/apps/mobile/src/components/agents/new-session-prefill.test.ts
@@ -243,6 +243,14 @@ describe('resolvePrefillRepo', () => {
     expect(result).toBe('Kilo-Org/cloud');
   });
 
+  it('returns null when multiple options match case-insensitively', () => {
+    const result = resolvePrefillRepo(
+      [{ fullName: 'Kilo-Org/cloud' }, { fullName: 'kilo-org/CLOUD' }],
+      { mode: 'code', repo: 'KILO-ORG/cloud' }
+    );
+    expect(result).toBeNull();
+  });
+
   it.each([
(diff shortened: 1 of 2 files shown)


RSO:
Superseded by #5601. That PR now includes this ambiguity-safe prefill behavior and adapts it to the composite picker identity: unique personal/legacy rows resolve to their exact unqualified key, unique organization rows retain integration provenance, and duplicate case-insensitive GitHub rows remain unselected.

PR #5546, opened 2026-08-27
feat(mobile): keep continued sessions on their GitHub installation

RSO:
## Why this PR exists

A continued Cloud Agent session must use the same GitHub App installation as the source session. With multiple installations, repository name alone is no longer sufficient: two installations can expose the same `owner/repo`, and choosing a different one can change both authorization and credentials.

Without this PR, mobile continuation can silently move work to whichever matching installation happens to be returned first. This PR makes continuation preserve exact installation identity and fail closed when that identity cannot be determined uniquely.

## Place in the rollout

This is the continuation-specific mobile slice of the multiple-GitHub-organizations rollout. It is independent from the new-session selector in #5549 so reviewers can evaluate session cloning and idempotency separately from picker UI.

## What changes

- Resolve the source repository against the current repository DTOs and require a unique match.
- Carry `platformIntegrationId` into continuation as `githubIntegrationId`.
- Include installation identity in the operation fingerprint so retries for different installations cannot collide.
- Keep legacy behavior for old DTOs that genuinely have no provenance, while still rejecting ambiguity.

## What stays unchanged

- Continuations remain clone-only and do not create a synthetic prompt or initial message.
- New-session repository selection is not changed here.
- Server-side token authorization remains the fence already merged in #5485.

## Review guide

Start with `continuation-seed.ts`, then verify `use-continue-cloud-create.ts` forwards the resolved identity and fingerprints it. The tests cover unique, ambiguous, legacy, and retry cases.

## Validation

- Focused continuation tests
- Mobile typecheck, lint, unused-code check, and format check
- `git diff --check`

### DIFF
--- a/apps/mobile/src/components/agents/continuation-seed.test.ts
+++ b/apps/mobile/src/components/agents/continuation-seed.test.ts
@@ -88,4 +88,52 @@ describe('resolveContinuationResolution', () => {
 
     expect(result).toEqual({ kind: 'unresolved-model' });
   });
+
+  it('preserves unique repository integration provenance when continuing', () => {
+    const result = resolveContinuationResolution({
+      gitUrl: GIT_URL,
+      mode: 'code',
+      model: 'test-model',
+      variant: 'default',
+      repositories: [{ fullName: 'owner/repo', platformIntegrationId: 'integration-1' }],
+      models: MODELS,
+    });
+
+    expect(result).toEqual({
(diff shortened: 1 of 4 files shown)


RSO:
Replaced by #5587. Cloud Agent now resolves and persists GitHub installation identity centrally and inherits it from source session metadata during continuation, so mobile no longer reverse-maps repository DTOs.

PR #5548, opened 2026-08-27
feat(extension): let cloud sessions target any connected GitHub org

RSO:
## Why this PR exists

The extension currently treats repository full name as option identity. After an organization connects multiple GitHub installations, duplicate `owner/repo` entries can represent different authorization boundaries. Rendering or submitting only the name makes those entries indistinguishable and can start a Cloud Agent session with the wrong installation.

This PR makes the extension selection explicit and sends the chosen installation to the already-merged server authorization fence.

## Place in the rollout

This is the extension client slice of the multiple-GitHub-organizations rollout. It is independent of mobile and product-specific automation work. The server DTO and session input already support `platformIntegrationId` from #5485 and #5488.

## What changes

- Preserve integration ID and GitHub account login in repository options.
- Use an installation-qualified internal option key while continuing to display and submit canonical `owner/repo`.
- Group and search repositories by GitHub organization.
- Send optional `githubIntegrationId` for cloud sessions.
- Treat stale selections as invalid instead of falling back to another row.

## What stays unchanged

- Connected-CLI session creation does not use this cloud installation field and remains unchanged.
- Legacy API responses without provenance continue to work.
- This PR does not alter server token resolution or any webhook behavior.

## Review guide

There are only two changed files. Review the stable-key helpers and cloud submit payload first, then the grouped picker rendering. Tests exercise duplicate names, account search, stale values, and legacy fallback.

## Validation

- Full extension `verify` suite
- Extension production build
- `git diff --check`

### DIFF
--- a/apps/extension/entrypoints/sidepanel/agents-new-session.test.ts
+++ b/apps/extension/entrypoints/sidepanel/agents-new-session.test.ts
@@ -4,6 +4,9 @@ import { describe, expect, it, vi } from 'vitest';
 import {
   buildPrepareSessionInput,
   buildSubmitInput,
+  filterRepoOptions,
+  getRepoOptionKey,
+  groupRepoOptions,
   isModelPreferencesGetResult,
   MODE,
   PROMPT_MAX_LENGTH,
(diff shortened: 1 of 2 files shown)


RSO:
Replaced by the client-selection/central-authorization version: https://github.com/Kilo-Org/cloud/pull/5602

PR #5549, opened 2026-08-27
feat(mobile): let new sessions target any connected GitHub org

RSO:
## Why this PR exists

Mobile can receive repositories from several GitHub App installations, but its picker and retry logic currently identify a repository only by provider and full name. When two installations expose the same repository, one row can overwrite or impersonate the other, and the selected installation is lost before session creation.

The server already accepts and enforces exact `githubIntegrationId`. This PR makes mobile preserve that identity from API response through visible selection, submission, and safe retry.

## Place in the rollout

This is the new-session mobile slice of the broader rollout. Prefill ambiguity and continuation are separate PRs (#5545 and #5546) so this review can focus on the picker and new-session submission path.

## What changes

- Preserve GitHub integration ID and account login in the mobile repository model.
- Use installation-qualified option keys so duplicate names remain separate.
- Group and search GitHub repositories by organization account.
- Submit canonical `githubRepo` plus the selected `githubIntegrationId`.
- Include installation identity in retry fingerprints and reject stale option keys.

## What stays unchanged

- GitLab and Bitbucket selection semantics remain unchanged.
- Legacy GitHub responses without provenance remain supported.
- This PR does not change continuation or deep-link ambiguity behavior.

## Why this touches several mobile files

Selection identity crosses the API adapter, repository state, native picker bridge, picker UI, submit payload, and retry fingerprint. Splitting those pieces would create an intermediate app that displays duplicate rows but cannot submit them safely, so they are intentionally one deployable client change.

## Review guide

Follow one repository through `toNewSessionGitHubRepository`, `getRepoOptionKey`, picker state, and `setRepositoryField`. The key invariant is that the composite key is UI-only; `githubRepo` remains canonical `owner/repo`, and `githubIntegrationId` carries authorization provenance.

## Validation

- Focused mobile selector tests
- Mobile typecheck, lint, unused-code check, and format check
- `git diff --check`

### DIFF
--- a/apps/mobile/src/app/(app)/agent-chat/repo-picker.tsx
+++ b/apps/mobile/src/app/(app)/agent-chat/repo-picker.tsx
@@ -10,12 +10,12 @@ import { EmptyState } from '@/components/empty-state';
 import { PickerSheet } from '@/components/picker-sheet';
 import { Text } from '@/components/ui/text';
 import { useThemeColors } from '@/lib/hooks/use-theme-colors';
-import { REPO_PLATFORM_LABEL_KEYS, type RepoOption } from '@/lib/picker-bridge';
+import { getRepoOptionKey, REPO_PLATFORM_LABEL_KEYS, type RepoOption } from '@/lib/picker-bridge';
 import { repoPickerSlot, UNFENCED_ROUTE_KEY, useRouteRegistry } from '@/lib/route-registry';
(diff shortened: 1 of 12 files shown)


RSO:
Replaced by the client-selection/central-authorization version: https://github.com/Kilo-Org/cloud/pull/5601

PR #5554, opened 2026-08-27
feat(deployments): rebuild only from the GitHub install that pushed

RSO:
## Why this PR exists

Secondary GitHub installations are currently prevented from delivering `push` behavior. Simply allowing those webhooks would be unsafe: deployment matching historically used repository and branch, so a push from one installation could redeploy a record associated with another installation that exposes the same repository.

This PR enables the deployment use case only after making the handler installation-aware. The webhook can rebuild deployments and App Builder previews, but only records pinned to the installation that delivered the push.

## Place in the rollout

This is the deployment-specific webhook slice of the multiple-GitHub-organizations rollout. Other secondary event types remain suppressed until their owning product PRs establish equivalent provenance.

## What changes

- Admit `push` events from secondary organization installations.
- Match deployments by repository, branch, source type, and exact `platform_integration_id`.
- Keep App Builder preview matching tied to the delivering integration.
- Mint deployment credentials with the actual Standard or Lite app type from the integration row.

## What stays unchanged

- Unrelated secondary webhook events remain suppressed.
- Deployment records without the matching installation are not touched.
- App Builder migration identity persistence and destination selection are handled separately in #5597 and #5600.

## Review guide

Review the admission predicate in `webhook-handler.ts`, then the exact deployment query in `push-handler.ts`. The critical negative test is that a sibling installation with the same repository and branch is not redeployed.

## Validation

- 3 focused Jest suites, 19 tests
- Web typecheck and lint
- Changed-file format check
- `git diff --check`

### DIFF
--- a/apps/web/src/lib/integrations/platforms/github/adapter.installation-token.test.ts
+++ b/apps/web/src/lib/integrations/platforms/github/adapter.installation-token.test.ts
@@ -1,0 +1,37 @@
+const mockAuth = jest.fn();
+
+jest.mock('@octokit/auth-app', () => ({
+  createAppAuth: jest.fn(() => mockAuth),
+}));
+
+jest.mock('@/lib/integrations/platforms/github/app-selector', () => ({
+  getGitHubAppCredentials: jest.fn(() => ({
+    appId: 'app-id',
+    privateKey: 'private-key',
+    webhookSecret: 'webhook-secret',
+  })),
+}));
+
+import { generateGitHubInstallationToken } from './adapter';
+
+describe('generateGitHubInstallationToken', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockAuth.mockResolvedValue({ token: 'token', expiresAt: '2099-01-01T00:00:00.000Z' });
+  });
+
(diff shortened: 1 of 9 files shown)


RSO:
Audit follow-up rationale:

- Removed `platform_integrations.repositories` / `repository_access` from the deployment authorization decision. That data remains useful for UI discovery, but it is a synchronized cache and is not authoritative for direct token minting.
- Deployment creation and redeployment now use the exact persisted `platform_integration_id`, validate exact owner, healthy lifecycle state, `platform = github`, `integration_type = app`, installation ID, and the row's Standard/Lite app type.
- The GitHub installation token request is scoped to the exact repository name. A GitHub mint failure is therefore the authoritative proof that the selected installation cannot access the repository; no broad resolver or HTTP bridge was added.
- Added stale-negative coverage (newly granted repository absent from cache still reaches GitHub mint) and stale-positive coverage (revoked repository present in cache is rejected by GitHub before the builder runs), plus adapter coverage for scoped and existing unscoped token minting.

Validation after `1b51f0b7a`:
- 4 focused Jest suites, 23 tests
- Web typecheck
- Web lint
- Repository format pass and changed-file format check
- `git diff --check`

PR #5547, opened 2026-08-27
superseded: cached web repository installation resolver

RSO:
Superseded by #5584.

This PR resolves installation identity from cached web/database metadata. The replacement moves resolution to Git Token Service and validates each candidate against live GitHub repository access.

Do not merge this PR. It remains open temporarily only because several existing product branches were based on it; those branches are being rebuilt or simplified against #5584/#5587.

PR #5555, opened 2026-08-27
feat(code-review): keep review follow-ups on the originating GitHub install

RSO:
## Why this PR exists

Code Reviewer already records the GitHub installation that produced a review, but several follow-up paths discard it. `Fix with Cloud Agent`, incremental review continuation, and Review Memory change requests can then resolve a different organization installation by repository name. That is both a correctness problem and a credential-boundary problem when installations overlap.

This PR makes every review follow-up continue through the installation that owns the review, and makes legacy repository-only Review Memory changes fail closed when more than one installation could serve them.

## Place in the rollout

This is the follow-up/continuation slice for Code Reviewer. Initial multi-install review dispatch was piloted in merged #5486. Installation health and action-required isolation are intentionally separate in #5579.

## Dependency

Depends on #5547 for shared fail-closed repository resolution. After #5547 merges, retarget this PR to `main`.

## What changes

- Pass `review.platform_integration_id` into `Fix with Cloud Agent`.
- Pin incremental review sessions to the exact integration.
- Refresh moving GitHub pull refs when a warm review session is reused.
- Resolve Review Memory change requests uniquely instead of selecting the first installation.
- Preserve the resolved Standard or Lite app type for follow-up token issuance.

## What stays unchanged

- Initial webhook dispatch and reviewer configuration are not redesigned here.
- GitLab continuation behavior is preserved.
- This PR does not change organization-wide status or action-required UI.

## Why this crosses web and Cloud Agent services

The integration ID is chosen in the web review flow but must survive the worker contract and warm-session preparation. Changing only one side would create a partial pin that is lost during continuation.

## Review guide

Trace `platform_integration_id` from `prepare-review-payload.ts` and the fork route into Cloud Agent session state, then inspect the warm pull-ref refresh. Review Memory resolution is a separate, small use of the same safety contract.

## Validation

- 53 focused web tests
- 111 Cloud Agent session-service tests
- 57 code-review worker tests
- 355 worker-utils tests
- Affected-package typechecks, lint, and formatting
- `git diff --check`

### DIFF
--- a/apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.test.ts
+++ b/apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.test.ts
@@ -35,6 +35,7 @@ type ReviewResult =
 
 type PrepareSessionInput = {
   githubRepo: string;
+  githubIntegrationId?: string;
   prompt: string;
   mode: string;
   model: string;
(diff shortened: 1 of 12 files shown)


RSO:
Replaced by the simplified central-resolution version: https://github.com/Kilo-Org/cloud/pull/5592

PR #5558, opened 2026-08-27
feat(chat-bots): let Slack and Discord target any connected GitHub org

RSO:
## Why this PR exists

Slack and Discord bot sessions currently build repository context from one organization GitHub installation and prefetch its token. Once an organization has several installations, repositories on secondary GitHub organizations are invisible; manually naming one can also select an arbitrary installation when access overlaps.

This PR gives the model explicit repository provenance and validates its tool call before Cloud Agent starts. The bot passes the selected installation ID instead of choosing a primary token itself.

## Place in the rollout

This is the user-selected chat-bot slice of the multiple-GitHub-organizations rollout. Incoming GitHub bot mentions use webhook identity and are handled separately in #5556.

## Dependency

Depends on #5547 for owner-scoped, fail-closed repository resolution. After #5547 merges, retarget this PR to `main`.

## What changes

- Aggregate repository context across all healthy organization GitHub installations.
- Include account and `platformIntegrationId` beside each repository in model context.
- Require the bot tool call to preserve that identity.
- Validate repository, account, and integration before spawning Cloud Agent.
- Pass `githubIntegrationId` and let Cloud Agent mint its short-lived token.

## What stays unchanged

- Personal bot sessions keep their existing token path.
- Slack/Discord workspace ownership remains independent from GitHub ownership.
- An unlisted or ambiguous repository produces an error rather than a guessed installation.

## Why this PR is larger

The shared bot runner and the older Discord-specific runner are separate execution paths. Both must enforce the same provenance rule or one platform would remain unsafe. Most added lines are focused tests for context formatting and session input.

## Review guide

Start with `github-repository-context.ts`: it defines what the model sees and validates what comes back. Then check the two callers only forward validated `githubIntegrationId` and no longer fetch organization-primary GitHub tokens.

## Validation

- 3 focused Jest suites, 9 tests
- Web typecheck and lint
- Changed-file format check
- `git diff --check`

### DIFF
--- a/apps/web/src/lib/bot/agent-runner.ts
+++ b/apps/web/src/lib/bot/agent-runner.ts
@@ -19,6 +19,7 @@ import type { CloudAgentAttachments } from '@/lib/cloud-agent/constants';
 import {
   formatGitHubRepositoriesForPrompt,
   getGitHubRepositoryContext,
+  resolveGitHubRepositorySelection,
 } from '@/lib/slack-bot/github-repository-context';
 import {
   formatGitLabRepositoriesForPrompt,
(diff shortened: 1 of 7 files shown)


RSO:
Replaced by the simplified central-resolution version: https://github.com/Kilo-Org/cloud/pull/5591

PR #5561, opened 2026-08-27
feat(gastown): keep each rig on the GitHub install it was created from

RSO:
## Why this PR exists

A Gastown can contain several rigs, and an organization can now connect several GitHub installations. Gastown currently keeps repository and credentials largely at town scope, so adding or refreshing one rig can cause another rig to use the wrong installation. This is especially dangerous after a restart, when in-memory context is gone and only persisted rig state remains.

This PR makes GitHub installation identity a property of each rig and uses it for every Git operation.

## Place in the rollout

This is the Gastown backend-safety layer. It must land before the repository selector in #5570 exposes secondary-installation repositories to users.

## Dependency

Depends on #5547 for fail-closed legacy repository resolution. After #5547 merges, retarget this PR to `main`.

## What changes

- Persist Kilo `platformIntegrationId` per rig in owner and Town Durable Object state.
- Use the rig installation for clone, push, PR polling, feedback, merge, branch creation, and token refresh.
- Keep credentials rig-scoped instead of mutating process-global GitHub token state.
- Preserve old rigs without identity, but resolve them only when one installation matches.

## What stays unchanged

- Existing rigs remain readable.
- This PR does not add or redesign repository selection UI.
- Town-level fallback exists only for legacy unpinned rigs and fails closed on ambiguity.

## Why this crosses Gastown, container, and token-service code

A rig moves through persisted Durable Object state, container startup, runtime Git commands, and credential refresh. Pinning only creation would be lost at restart; pinning only token refresh would leave clone and PR operations unsafe. The changes follow one identity across those existing boundaries.

## Review guide

Start with `user-rigs.table.ts` and Town config normalization, then trace `platformIntegrationId` into `town-scm.ts` and refresh handling. Finally verify the container no longer relies on a process-global token. Tests cover persistence, two-rig isolation, legacy ambiguity, and refresh.

## Validation

- Gastown, Git Token Service, and container typechecks/lint
- 166 shared resolver tests
- 32 Gastown identity tests
- Focused container and Durable Object identity tests
- `git diff --check`

### DIFF
--- a/services/gastown/container/src/control-server.test.ts
+++ b/services/gastown/container/src/control-server.test.ts
@@ -1,0 +1,28 @@
+import { afterEach, describe, expect, it } from 'vitest';
+import { app } from './control-server';
+
+const originalGitToken = process.env.GIT_TOKEN;
+
+afterEach(() => {
+  if (originalGitToken === undefined) delete process.env.GIT_TOKEN;
+  else process.env.GIT_TOKEN = originalGitToken;
+});
+
+describe('town config git identity', () => {
+  it('clears process-global GitHub credentials for rig-scoped identity', async () => {
+    process.env.GIT_TOKEN = 'other-rig-token';
+
+    const response = await app.request('/sync-config', {
+      method: 'POST',
+      headers: {
+        'X-Town-Config': JSON.stringify({
+          rig_scoped_git_identity: true,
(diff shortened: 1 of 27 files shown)


RSO:
Replaced by the authoritative Git Token Service version: https://github.com/Kilo-Org/cloud/pull/5595

PR #5562, opened 2026-08-27
feat(security): sync repositories from every connected GitHub org

RSO:
## Why this PR exists

Security Agent configuration is organization-wide, but its sync worker currently behaves as though the organization has one GitHub installation. Repositories on secondary GitHub organizations are skipped, and an unordered or primary-only lookup can mint the wrong app token. A failure in one installation can also incorrectly interfere with healthy siblings.

This PR makes the sync plan installation-aware before any UI exposes secondary repositories.

## Place in the rollout

This is the Security Agent backend foundation. The repository/status UI in #5568 and finding operations in #5571 depend on authoritative sync provenance established here.

## Dependency

Depends on #5547 for shared fail-closed repository resolution. After #5547 merges, retarget this PR to `main`.

## What changes

- Load every healthy organization GitHub installation with the required vulnerability-alert permissions.
- Build an explicit repository-to-integration sync plan.
- Mint each token with that integration's provider installation ID and Standard/Lite app type.
- Isolate authorization failures to repositories served by the affected installation.
- Fail closed when the same configured repository is visible through multiple installations.
- Update a finding's integration ID when authoritative sync observes a legitimate repository move.

## What stays unchanged

- Security Agent remains one organization-wide configuration.
- UI selection and reauthorization presentation are handled in #5568.
- Analysis, remediation, dismissal, and writeback are handled in #5571.

## Why the worker change is substantial

The old sync function assumes one token and one repository set throughout. Supporting several installations safely requires partitioning work and errors by integration rather than wrapping the old loop in another loop. The production change is confined to the Security Sync service; roughly half the diff is scenario coverage.

## Review guide

Review the sync-plan construction first, then token creation per plan entry, then finding provenance updates. The important negative cases are duplicate visibility and an unhealthy sibling not blocking healthy work.

## Validation

- 66 Security Sync tests
- Package typecheck and lint
- Formatting and `git diff --check`

### DIFF
--- a/services/security-sync/src/index.ts
+++ b/services/security-sync/src/index.ts
@@ -449,6 +449,9 @@ function syncCommandTerminalState(result: Awaited<ReturnType<typeof syncOwner>>)
   if (result.commandResultCode === 'CONFIG_DISABLED') {
     return { status: 'no_op', resultCode: 'CONFIG_DISABLED' };
   }
+  if (result.commandResultCode === 'AMBIGUOUS_GITHUB_INTEGRATION') {
+    return { status: 'failed', resultCode: 'AMBIGUOUS_GITHUB_INTEGRATION' };
+  }
   if (result.commandResultCode === 'REPOSITORY_UNAVAILABLE' || result.staleRepos.length > 0) {
     return { status: 'failed', resultCode: 'REPOSITORY_UNAVAILABLE' };
   }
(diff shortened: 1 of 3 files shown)


RSO:
Replaced by the authoritative Git Token Service version: https://github.com/Kilo-Org/cloud/pull/5593

PR #5563, opened 2026-08-27
feat(auto-triage): handle issues from every connected GitHub org

RSO:
## Why this PR exists

Auto Triage currently derives organization repositories and tokens from one GitHub installation. Issues opened on a secondary connected organization are suppressed, while manually submitted issue URLs can be fetched through an arbitrary matching installation. That means connecting another GitHub organization does not actually make Auto Triage work there, and overlapping access can cross credential boundaries.

This PR enables only the Auto Triage issue lifecycle after making ticket creation and all downstream GitHub calls carry exact installation identity.

## Place in the rollout

This is the Auto Triage product slice. Auto Fix uses different webhook actions and ticket behavior and remains isolated in #5572.

## Dependency

Depends on #5547 for unique repository-to-installation resolution. After #5547 merges, retarget this PR to `main`.

## What changes

- Aggregate Auto Triage repository configuration across healthy installations with account and integration provenance.
- Resolve manual issue URLs to exactly one installation before fetching or creating a ticket.
- Persist `platform_integration_id` on the ticket and use its app type for classification, comments, labels, and retries.
- Admit secondary `issues.opened` and `issues.reopened` events.

## What stays unchanged

- Secondary label events remain suppressed for Auto Fix.
- Ambiguous manual URLs fail with a conflict instead of choosing an installation.
- Personal Auto Triage behavior is unchanged.

## Why routes and internal callbacks change too

A triage ticket is asynchronous: webhook/manual creation, classification, comment, label, and retry can occur in different requests. Every stage must recover credentials from the persisted integration, or the initial correct choice is lost later.

## Review guide

Trace `platform_integration_id` from repository listing/manual resolution or webhook delivery into ticket creation, then inspect `installation-token.ts` use in internal routes. Finally review the narrow secondary-event predicate.

## Validation

- 5 focused suites, 34 tests
- Web typecheck and lint
- `git diff --check`

The standard DB-backed Jest harness was unavailable; focused tests used isolated mocks.

### DIFF
--- a/apps/web/src/app/api/internal/triage/add-label/route.ts
+++ b/apps/web/src/app/api/internal/triage/add-label/route.ts
@@ -20,8 +20,8 @@ import { logExceptInTest, errorExceptInTest } from '@/lib/utils.server';
 import { captureException } from '@sentry/nextjs';
 import { INTERNAL_API_SECRET } from '@/lib/config.server';
 import { addIssueLabel } from '@/lib/auto-triage/github/add-label';
-import { generateGitHubInstallationToken } from '@/lib/integrations/platforms/github/adapter';
 import { getIntegrationById } from '@/lib/integrations/db/platform-integrations';
+import { generateAutoTriageInstallationToken } from '@/lib/auto-triage/github/installation-token';
 
 const addLabelRequestSchema = z.object({
   ticketId: z.string().uuid(),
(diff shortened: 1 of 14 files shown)


RSO:
Replaced by the smaller exact-provenance version: https://github.com/Kilo-Org/cloud/pull/5596

PR #5565, opened 2026-08-27
feat(cli-sessions): remember which GitHub install owns cached PR data

RSO:
## Why this PR exists

CLI session PR badges and review decisions are backed by a cache keyed around repository and branch. With multiple GitHub installations, identical repository/branch values can arrive through different authorization boundaries. If the cache does not record the delivering integration, later refreshes can apply one installation's PR state to sessions using another.

This PR adds provenance at the write boundary before #5569 enables secondary review events and installation-aware refresh.

## Place in the rollout

This is the schema/write foundation for CLI PR metadata. It deliberately does not change refresh or review-decision fetching; that behavior is isolated in #5569.

## What changes

- Add nullable `platform_integration_id` to cached GitHub branch/PR associations.
- Record the exact delivering integration on pull-request webhook writes.
- Preserve an existing cache pin when a legacy caller has no authoritative integration ID.
- Leave old rows null rather than inventing a backfill.

## What stays unchanged

- Existing cache rows remain readable.
- Review-decision batching and manual refresh are unchanged until #5569.
- No secondary `pull_request_review` event is enabled here.

## Why a separate PR

The schema can be deployed backward-compatibly before readers depend on it. Separating storage from behavior keeps this review focused on data ownership and lets #5569 assume the field exists. The large addition count is almost entirely the generated Drizzle snapshot.

## Review guide

Review the nullable foreign key and conflict-update behavior in `upsert-cli-session-pull-requests.ts`. The key invariant is “write exact provenance when known; never infer or erase it when unknown.”

## Migration note

This PR and #5564 currently use the same generated migration sequence. Whichever merges second must be rebased and regenerated; generated SQL, snapshot, and journal files must not be hand-merged.

## Validation

- DB and web typechecks
- Targeted lint
- Schema and migration tests
- Migration regeneration check
- `git diff --check`

DB-backed webhook tests were unavailable because their harness requires runtime environment configuration.


RSO:
Superseded by #5611, which replays the same two schema/cache commits as an explicit child of #5590 and preserves the generated migration artifacts and documented 0233 collision policy. Behavior continues in #5612.

PR #5568, opened 2026-08-27
feat(security): let users configure repositories from every GitHub org

RSO:
## Why this PR exists

#5562 teaches Security Agent sync to process every healthy GitHub installation, but the current web/mobile configuration still projects one installation. Users cannot select secondary repositories or tell which connected GitHub organization lacks required permissions.

This PR exposes only repositories the backend can now sync, preserves their installation identity, and makes recovery installation-specific.

## Place in the rollout

This is the Security Agent configuration/UI child of #5562. Finding execution is separate in #5571 so this review can focus on repository discovery, selection, and installation health.

## Dependency

Depends on #5562. Merge and deploy sync support first, then retarget this PR to `main`.

## What changes

- Aggregate repository APIs across healthy GitHub installations.
- Include account and integration ID on each repository.
- Group repositories by GitHub organization in web and mobile selectors.
- Expose permission, authentication, and reauthorization state per installation.
- Keep healthy repositories selectable when a sibling installation is unhealthy.

## What stays unchanged

- Security Agent remains one organization-wide configuration.
- The sync semantics are defined in #5562.
- Analysis, remediation, and writeback use finding provenance in #5571.

## Why this touches shared selector code

Security Agent reuses `RepositoryMultiSelect`, which previously assumed one label space. The change adds optional grouping/provenance support while preserving existing consumers; the Security-specific components supply installation health and recovery behavior.

## Review guide

Start with `shared-handlers.ts` output: verify every repository carries its integration. Then review selector grouping and the per-installation status rendering. The key negative case is one unhealthy installation not hiding healthy siblings.

## Validation

- Workspace typecheck and lint
- Full mobile suite: 6,244 tests
- 4 database-free web grouping tests
- `git diff --check`

DB-backed web and browser checks were unavailable because no test database/server was running.

### DIFF
--- a/apps/mobile/src/components/security-agent/repository-settings-screen.tsx
+++ b/apps/mobile/src/components/security-agent/repository-settings-screen.tsx
@@ -21,6 +21,7 @@ import { Skeleton } from '@/components/ui/skeleton';
 import { Text } from '@/components/ui/text';
 import { TabScreenScrollView } from '@/components/tab-screen';
 import { getGitHubIntegrationUrl } from '@/lib/agent-github-integration';
+import { i18n } from '@/i18n';
 import { WEB_BASE_URL } from '@/lib/config';
 import { openExternalUrl } from '@/lib/external-link';
 import { trpcClient } from '@/lib/trpc';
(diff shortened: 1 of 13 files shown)


RSO:
Replaced on the authoritative Security Sync parent: https://github.com/Kilo-Org/cloud/pull/5604

PR #5569, opened 2026-08-27
feat(cli-sessions): refresh PR status with the session's GitHub install

RSO:
## Why this PR exists

#5565 records which GitHub installation produced cached PR data. Readers still need to honor that provenance. Today manual refresh and batched review-decision fetches can use a primary installation, and secondary review webhooks are suppressed because their cache side effects are not isolated.

This PR makes refresh, batching, and webhook updates operate on the exact installation associated with the session/cache row, then safely admits those secondary events.

## Place in the rollout

This is the behavior child of #5565. The parent is a backward-compatible schema/write PR; this PR is where CLI sessions gain correct secondary-installation PR badges and review decisions.

## Dependency

Depends on #5565. Merge its schema first, then retarget this PR to `main`.

## What changes

- Use a Cloud Agent session's pinned `githubIntegrationId` for manual PR refresh.
- Resolve old unpinned sessions only when repository/account access identifies one installation.
- Group review-decision requests by integration, provider installation, and app type.
- Scope compare-and-set batching and cache updates by integration.
- Admit secondary `pull_request` and `pull_request_review` cache events only after those writes are isolated.

## What stays unchanged

- Ambiguous legacy sessions fail closed.
- Cache schema/write ownership remains in #5565.
- Code Reviewer dispatch is not changed by this PR.

## Why this crosses web and Cloud Agent contracts

The web refresh endpoint needs the session's persisted installation identity, but that metadata is owned by Cloud Agent. A small contract addition exposes it; the remaining changes keep cache and GitHub API work grouped under the same identity.

## Review guide

Review `cli-session-integration.ts` and manual refresh first, then the batch grouping key, then webhook admission and integration-scoped writes. Verify old unpinned sessions never select the first overlapping installation.

## Validation

- Web and Cloud Agent typechecks
- 148 focused Cloud Agent tests
- Changed-file lint and formatting
- `git diff --check`

DB-backed web tests were typechecked but not executed because their setup requires environment files.

### DIFF
--- a/apps/web/src/lib/cloud-agent-next/cloud-agent-client.ts
+++ b/apps/web/src/lib/cloud-agent-next/cloud-agent-client.ts
@@ -282,6 +282,7 @@ export type GetSessionOutput = {
 
   // Repository info (no tokens)
   githubRepo?: string;
+  githubIntegrationId?: string;
   gitUrl?: string;
   platform?: 'github' | 'gitlab' | 'bitbucket';
 
--- a/apps/web/src/lib/integrations/platforms/github/batch-review-decisions.test.ts
+++ b/apps/web/src/lib/integrations/platforms/github/batch-review-decisions.test.ts
@@ -1,5 +1,9 @@
 import { db } from '@/lib/drizzle';
-import { cli_sessions_v2, github_branch_pull_requests } from '@kilocode/db/schema';
+import {
+  cli_sessions_v2,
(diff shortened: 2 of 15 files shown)


RSO:
Replaced by the canonical-session/exact-cache version: https://github.com/Kilo-Org/cloud/pull/5606

PR #5570, opened 2026-08-27
feat(gastown): let each rig choose its GitHub organization

RSO:
## Why this PR exists

#5561 makes installation identity durable per Gastown rig, but onboarding and later rig creation still expose repositories as if there were one GitHub installation. Duplicate repository names cannot be distinguished, and the backend pin cannot be populated without a client selection.

This PR lets users choose the GitHub organization for each rig and submits the exact integration that #5561 persists.

## Place in the rollout

This is the UI child of the Gastown backend-safety PR #5561. It is intentionally separate so the persistence/runtime contract can be reviewed and deployed before secondary repositories become selectable.

## Dependency

Depends on #5561. Merge the per-rig identity backend first, then retarget this PR to `main`.

## What changes

- Use one installation-aware repository selector in onboarding and Create Rig.
- Group repositories by GitHub organization/account.
- Keep duplicate names distinct by platform integration ID.
- Submit `platformIntegrationId` and repository default branch with rig creation.
- Invalidate stale selections instead of switching to another installation.
- Truncate long labels visually while retaining full accessible/title text.

## What stays unchanged

- Rig token, clone, PR, and restart behavior remains in #5561.
- Existing rigs are not modified by opening the selector.
- The selector does not choose a replacement when its selected repository disappears.

## Review guide

Review `GastownRepositorySelector` as the identity boundary, then verify both onboarding and dialog callers submit its complete selection. Most added lines are component tests.

## Validation

- 53 focused tests
- Web and affected-package typechecks
- Web lint
- `git diff --check`

### DIFF
--- a/apps/web/src/app/(app)/gastown/onboarding/OnboardingContext.tsx
+++ b/apps/web/src/app/(app)/gastown/onboarding/OnboardingContext.tsx
@@ -11,6 +11,7 @@ import { presetToConfig } from './onboarding.domain';
 
 type OnboardingRepo = {
   platform: 'github' | 'gitlab' | 'manual';
+  selectionKey: string;
   fullName: string;
   gitUrl: string;
   defaultBranch: string;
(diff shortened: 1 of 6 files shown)


RSO:
Replaced on the authoritative Gastown parent: https://github.com/Kilo-Org/cloud/pull/5605

PR #5571, opened 2026-08-27
feat(security): keep each finding on its GitHub installation

RSO:
## Why this PR exists

#5562 records authoritative installation provenance while syncing Security findings. The later operations on those findings still include paths that fetch a primary token or trust a command-supplied provider installation ID. With multiple installations, analysis, remediation, dismissal, or Dependabot writeback can therefore act through a sibling installation.

This PR makes the finding's persisted `platform_integration_id` the authorization source for every operation.

## Place in the rollout

This is the Security Agent execution child of #5562. Repository configuration and status UI are separate in #5568.

## Dependency

Depends on #5562 because sync must establish and repair finding provenance first. Retarget to `main` after #5562 merges.

## What changes

- Pass expected integration identity through manual and automatic analysis.
- Pin remediation sessions and callback PR verification.
- Resolve manual/automatic Dependabot dismissal tokens from the finding, not owner primary or command installation ID.
- Support exact personal and organization integration resolution.
- Preserve Standard/Lite app type in writeback.
- Allow legacy unpinned findings only when one repository installation matches.

## What stays unchanged

- Finding schema already contains the integration ID; this PR adds no migration.
- Repository selection and installation health UI are not changed here.
- Stale, suspended, moved, mismatched, or ambiguous pins fail closed rather than being repaired implicitly during an operation.

## Why several services change

A finding moves through web requests, Security Auto Analysis, Security Sync dismissal, remediation, callbacks, and Git Token Service. Every credential boundary must enforce the same expected integration or the pin can be bypassed downstream. Much of the diff is cross-service negative coverage.

## Review guide

Trace `finding.platform_integration_id` through manual analysis, automatic consumer, remediation, and dismissal. At each Git Token Service call, verify it becomes an expected integration fence and app type comes from the resolved row.

## Validation

- Web, Security Analysis, Security Sync, and Git Token Service typechecks/lint
- 181 Security Analysis tests
- 68 Security Sync tests
- 535 Git Token Service tests
- `git diff --check`

Web DB-backed tests could not run because PostgreSQL was unavailable.


RSO:
Replaced on the authoritative Security Sync parent: https://github.com/Kilo-Org/cloud/pull/5599

PR #5572, opened 2026-08-27
feat(auto-fix): fix issues from every connected GitHub org safely

RSO:
## Why this PR exists

Auto Fix currently assumes one organization GitHub installation. Secondary label and review-comment events are suppressed, repository configuration omits secondary installations, and retry/callback paths can re-resolve credentials by repository name. With overlapping installations, that can either make Auto Fix unavailable or continue a fix through the wrong GitHub App.

This PR enables only the Auto Fix event lifecycle after pinning each ticket and callback to one exact installation.

## Place in the rollout

This is the Auto Fix product slice. Auto Triage uses different issue actions and is separate in #5563. Review Memory feedback shares the review-comment event but remains a distinct side effect in #5577.

## Dependency

Depends on #5547 for unique repository-to-installation resolution. After #5547 merges, retarget this PR to `main`.

## What changes

- Aggregate Auto Fix repository configuration across healthy installations with provenance.
- Pin new tickets to the delivering or selected integration.
- Atomically pin legacy retry tickets so sibling events cannot race to claim them.
- Resolve callbacks, configuration, reactions, replies, permissions, and PR-head lookups through the pinned row and app type.
- Admit only secondary `issues.labeled` and created `pull_request_review_comment` events required by Auto Fix.

## What stays unchanged

- Auto Triage event behavior is not changed here.
- Other review-comment side effects remain primary-only until #5577.
- Ambiguous legacy tickets and repositories fail closed.

## Why this touches several Auto Fix modules

A fix ticket crosses webhook processing, retry persistence, configuration lookup, comment replies, and an asynchronous PR callback. Pinning only the first webhook would be lost later, so each existing boundary must consume the same stored integration. The changes remain within Auto Fix plus the narrow webhook admission predicate.

## Review guide

Start with `resolve-integration.ts` and ticket pinning, then follow the integration through review-comment processing and the PR callback. Review `webhook-handler.ts` last to verify only Auto Fix event/action combinations are admitted.

## Validation

- 5 focused suites, 45 tests
- Web typecheck and lint
- Formatting and `git diff --check`

The standard Jest DB harness was unavailable; focused suites used isolated mocks.

### DIFF
--- a/apps/web/src/app/api/internal/auto-fix/pr-callback/route.test.ts
+++ b/apps/web/src/app/api/internal/auto-fix/pr-callback/route.test.ts
@@ -1,6 +1,9 @@
 import { beforeEach, describe, expect, it, jest } from '@jest/globals';
 import type { NextRequest } from 'next/server';
 import type * as fixTicketsModule from '@/lib/auto-fix/db/fix-tickets';
+import type * as resolveIntegrationModule from '@/lib/auto-fix/github/resolve-integration';
+import type * as githubAdapterModule from '@/lib/integrations/platforms/github/adapter';
+import type { PlatformIntegration } from '@kilocode/db/schema';
 import { deriveCallbackToken } from '@kilocode/worker-utils/callback-token';
 
 const mockGetFixTicketBySessionId = jest.fn() as jest.MockedFunction<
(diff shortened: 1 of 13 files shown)


RSO:
Replaced by the smaller webhook-provenance version: https://github.com/Kilo-Org/cloud/pull/5594

PR #5573, opened 2026-08-27
feat(app-builder): keep previews on the GitHub install chosen at migration

RSO:
## Why this PR exists

App Builder migration is not a one-time GitHub operation. The selected repository is used for the initial push, then later preview clones, rebuilds, and restarts inside a Durable Object. If only `owner/repo` is stored, a project can switch to another matching installation after restart or repository movement.

This PR makes installation identity part of App Builder source state and enforces it throughout the preview lifecycle.

## Place in the rollout

This is the App Builder backend-safety layer. #5576 adds the user-facing installation selector only after this PR can persist and honor the choice.

## Dependency

Depends on #5547 for exact and fail-closed legacy repository resolution. Retarget to `main` after #5547 merges.

## What changes

- Add optional expected integration identity to web migration and App Builder Worker contracts.
- Persist it in Preview Durable Object GitHub source state.
- Enforce it for migration validation, initial push, preview clone, fetch/rebuild, and restart.
- Preserve Standard/Lite app type when listing and validating repositories.
- Return a specific stale-installation error when the selected row is no longer usable.
- Keep old preview state readable; unpinned ambiguity fails closed.

## What stays unchanged

- The existing dialog does not gain a full installation selector here; that is #5576.
- Old projects without a stored integration continue through unique legacy resolution.
- This PR does not change deployment push routing, which is #5554.

## Why web and Worker both change

The web app chooses and validates the migration target, but the Worker owns persisted preview state and later Git operations. Both sides must share the optional contract or the choice disappears after the initial request. Most added lines are focused migration and PreviewDO tests.

## Review guide

Trace `expectedPlatformIntegrationId` from router schema through migration service and Worker API into PreviewDO state. Then verify every later Git credential/clone path reads the persisted value.

## Validation

- 44 App Builder tests
- App Builder and web typechecks/lint
- 3 focused web migration tests
- `git diff --check`

### DIFF
--- a/apps/web/src/components/app-builder/MigrateToGitHubDialog.tsx
+++ b/apps/web/src/components/app-builder/MigrateToGitHubDialog.tsx
@@ -51,6 +51,8 @@ type Step = 'create' | 'grant-access' | 'select' | 'success';
 
 const errorMessages: Record<MigrateToGitHubErrorCode, string> = {
   github_app_not_installed: 'GitHub App is not installed. Please install the GitHub App first.',
+  github_integration_unavailable:
+    'The GitHub installation selected for this repository is no longer available.',
   already_migrated: 'This project has already been migrated to GitHub.',
   repo_not_found:
     "Repository not found or not accessible. Make sure you've granted the Kilo GitHub App access to this repository.",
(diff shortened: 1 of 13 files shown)


RSO:
Replaced by the authoritative Git Token Service version: https://github.com/Kilo-Org/cloud/pull/5597

PR #5576, opened 2026-08-27
feat(app-builder): let migrations choose the destination GitHub org

RSO:
## Why this PR exists

#5573 gives App Builder a safe way to persist one GitHub installation, but the migration dialog still projects a single organization installation. Users cannot choose a secondary destination, and duplicate GitHub account/repository names provide no visible installation distinction.

This PR exposes the backend pin as an explicit migration-target choice and refuses to substitute another installation if the selected one becomes unavailable.

## Place in the rollout

This is the App Builder UI child of #5573. The split ensures the Worker can honor the choice before the UI offers it.

## Dependency

Depends on #5573. Merge/deploy App Builder identity first, then retarget this PR to `main`.

## What changes

- Aggregate migration targets from all healthy organization GitHub installations.
- Let the user select the destination GitHub organization/installation.
- Distinguish duplicate account names by Standard/Lite app type.
- Scope repository options to the selected installation.
- Submit its `platformIntegrationId`.
- Keep a stale pinned selection visible with installation-specific recovery instead of falling back.

## What stays unchanged

- Initial push, preview clone, rebuild, and restart enforcement remain in #5573.
- Personal migration behavior remains compatible.
- The selector does not auto-switch when an installation disappears.

## Review guide

Review the pure `migration-target-selector.ts` rules first, then the dialog state and payload. The key invariant is that refresh preserves identity, including unavailable pins, rather than selecting the first current target.

## Validation

- 8 isolated App Builder suites
- Web typecheck, lint, and format check
- `git diff --check`

### DIFF
--- a/apps/web/src/components/app-builder/MigrateToGitHubDialog.tsx
+++ b/apps/web/src/components/app-builder/MigrateToGitHubDialog.tsx
@@ -36,8 +36,21 @@ import {
   DialogTrigger,
   DialogFooter,
 } from '@/components/ui/dialog';
+import { Label } from '@/components/ui/label';
+import {
+  Select,
+  SelectContent,
+  SelectItem,
+  SelectTrigger,
+  SelectValue,
+} from '@/components/ui/select';
 import { RepositoryCombobox, type RepositoryOption } from '@/components/shared/RepositoryCombobox';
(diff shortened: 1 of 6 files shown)


RSO:
Replaced on the authoritative App Builder parent: https://github.com/Kilo-Org/cloud/pull/5600

PR #5577, opened 2026-08-27
feat(review-memory): learn from replies on secondary GitHub orgs

RSO:
## Why this PR exists

#5572 safely admits created pull-request review comments from secondary installations for Auto Fix. The same GitHub event also carries replies to Kilo review comments, which Review Memory uses as feedback. That side effect is still explicitly skipped for secondary installations even though its handler already receives the exact delivering integration and app type.

The result is inconsistent learning: identical reviewer feedback is recorded on the primary GitHub organization but ignored on every secondary one.

## Place in the rollout

This is a small gap found during the final singular-caller/webhook audit. It is separate from Auto Fix because Review Memory is an independent product side effect sharing the same admitted event.

## Dependency

Depends on #5572 because that PR admits the secondary `pull_request_review_comment.created` event. After #5572 merges, retarget this PR to `main`.

## What changes

- Run `handleGitHubReviewCommentReply` for secondary installations.
- Pass the exact delivering `PlatformIntegration`, including Standard/Lite app type.
- Keep Auto Fix and Review Memory as independent same-event handlers.

## What stays unchanged

- Non-replies do not create Review Memory feedback.
- No schema or feedback model changes are introduced.
- Other secondary review-comment actions remain governed by the existing admission predicate.

## Review guide

This is a two-file PR. Review the removal of the secondary-only guard and the four focused webhook cases: reply, non-reply, coexistence with Auto Fix, and Lite app type.

## Validation

- 17 isolated webhook tests
- Focused lint and formatting
- `git diff --check`

### DIFF
--- a/apps/web/src/lib/integrations/platforms/github/webhook-handler.test.ts
+++ b/apps/web/src/lib/integrations/platforms/github/webhook-handler.test.ts
@@ -71,6 +71,10 @@ jest.mock('@/lib/code-reviews/review-memory/github-feedback', () => ({
   handleGitHubReviewCommentReply: (input: unknown) => mockHandleGitHubReviewCommentReply(input),
 }));
 
+jest.mock('@/lib/utils.server', () => ({
+  logExceptInTest: jest.fn(),
+}));
+
 jest.mock('next/server', () => {
   const actual = jest.requireActual('next/server');
   return {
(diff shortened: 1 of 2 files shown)


RSO:
Replaced on the simplified Auto Fix parent: https://github.com/Kilo-Org/cloud/pull/5603

PR #5581, opened 2026-08-27
feat(commerce): add Play Kilo Pass presentation flag

iscekic:
## Summary

No new behavior — this change only prepares the Google Play purchase path for Kilo Pass on Android; a later change turns it on. Every user keeps the purchase options they see today.

---

The `ResolvePurchasePresentationInput` contract gains an optional `supportsNativePlayKiloPass` field. A caller that sends true for an Android Play Kilo Pass now receives `native_iap`, and `isNativeIapMutationAllowed` returns true for the same combination. Old Android callers omit the field or send false and keep the current unavailable or Stripe web-management result, so no stored input breaks.

<details>
<summary>Files</summary>

- `packages/app-shared/src/commerce/purchase-presentation.ts` — adds `supportsNativePlayKiloPass` to `ResolvePurchasePresentationInput`; adds a native_iap branch for Android Play Kilo Pass when the flag is true; extends `isNativeIapMutationAllowed` to admit Android Play Kilo Pass.

</details>

The web `PurchasePresentationInput` gains the same optional field, and `buildPurchasePresentation` forwards it to the shared resolver. A future Android client that calls the web path can signal native Play IAP support without changing the web return shape. Current web callers are unchanged because the field is optional and unset.

<details>
<summary>Files</summary>

- `apps/web/src/lib/kilo-pass/purchase-presentation.ts` — adds `supportsNativePlayKiloPass` to `PurchasePresentationInput` and passes it through in `buildPurchasePresentation`.

</details>

Tests: 2 test files updated.
Generated: none.

---

## Verification

E2E: bot-e2e — runtime verification runs once on the tip PR (level 6) after all levels land. This level is reviewed on its own diff.

## Visual Changes

Visual Changes: N/A

## Reviewer Notes

Level 1 changes no runtime behavior on its own; the flag is inert until level 6 wires the Android client to send true.

Runtime verification (E2E, user advocacy, simplify) runs on the tip PR over every level.
Every level keeps its own checks, its own bot review, and its own threads; each one is answered on its own PR.
Each level is its own deliverable: it builds and passes its own checks alone.

1. `android-iap-e895` — #5581 ← this PR
2. `android-iap-e895-s2` — #5582
3. `android-iap-e895-s3` — #5586
4. `android-iap-e895-s4` — #5598
5. `android-iap-e895-s5` — #5608
6. `android-iap-e895-s6` — #5618 (tip)


apps/web/src/lib/kilo-pass/purchase-presentation.ts
```diff
@@ -35,6 +35,7 @@ export type PurchasePresentationInput = {
   storefront: PurchaseStorefront | null | undefined;
   product: PurchaseProduct;
   program?: string | null;
+  supportsNativePlayKiloPass?: boolean;
```
pandemicsyn, line 38: bot: The new capability flag cannot reach the existing tRPC purchase flows.

Suggested fix: Add `supportsNativePlayKiloPass: z.boolean().optional()` to the request schema(s) intended to support Android native IAP—at least `GetPurchasePresentationInputSchema`, and `PreflightPurchaseInputSchema` if that is the client purchase gate—then forward `input.supportsNativePlayKiloPass` when constructing each `PurchasePresentationInput`. Add router tests showing `true` produces the enabled Android presentation/preflight behavior and omission remains legacy behavior. As written, both schemas strip/reject the field and the router reconstructs inputs without it, so RPC callers always resolve Android Play Kilo Pass as unavailable or Stripe web management.

pandemicsyn (approved):

apps/web/src/lib/kilo-pass/purchase-presentation.ts
iscekic: Not a defect. The flag is wired, but in the later stack levels, not this one.

`GetPurchasePresentationInputSchema` and `PreflightPurchaseInputSchema` both declare `supportsNativePlayKiloPass: z.boolean().optional()` and the router forwards `input.supportsNativePlayKiloPass` into each `PurchasePresentationInput` (`apps/web/src/routers/kilo-pass-router.ts`). The mobile client sends `true` from the subscription card and the subscription screen.

This PR is the bottom of a six-PR stack and adds only the presentation flag. The field is optional, so omitting it keeps the legacy behavior. The stack merges bottom to top, so no intermediate state ships Android Play Kilo Pass as reachable-but-unwired.

PR #5584, opened 2026-08-27
feat(github): choose installations by live repository access

RSO:
## Why this PR exists

Organizations can connect several GitHub App installations, while cached account and selected-repository metadata can become stale. Credential selection therefore needs GitHub to identify the installation that currently serves the exact `owner/repo`, with Kilo still enforcing user or organization authorization and installation health.

The previous proposed foundation in #5547 resolved from database metadata. This replacement moves the authoritative decision into Git Token Service without minting and testing a token for every candidate.

## Place in the rollout

This is the central resolution boundary for Cloud Agent and other server-side GitHub operations that know `owner/repo` but do not already have authoritative installation provenance. Callers that do know the installation, such as webhooks and persisted reviews or findings, continue supplying `expectedIntegrationId` as an exact fence.

The PR is standalone and rollout-compatible with existing retry behavior. Definitive lookup failures stay in the existing result contract, while provider uncertainty follows the existing thrown RPC failure path so callers retry instead of selecting a sibling installation by accident.

## Resolution algorithm

- Load all authorized, active, non-auth-invalid GitHub App rows in the exact personal or organization ownership scope, including current organization membership checks. Cached repository lists are not treated as proof.
- Derive the relevant app types from those rows, bounding provider work to at most one Standard probe and one Lite probe.
- For each relevant app type, app-authenticate and call `GET /repos/{owner}/{repo}/installation` once.
- Match the provider installation ID returned for that app type against the authorized healthy database rows.
- Detect duplicate-row or cross-app ambiguity before issuing credentials.
- After exactly one row is selected, mint one repository-scoped installation token and return its Kilo `platformIntegrationId` with the token and app metadata.
- Preserve exact `expectedIntegrationId` ownership and health checks when a caller supplies a fence.

## Provider safety and failure policy

- Repository-installation probes use a 10-second timeout, `redirect: manual`, fixed GitHub API origin and path construction, strict JSON shape validation, and a 64 KiB bounded response read.
- Only `404` definitively excludes an app type. Authentication, permission, rate-limit, validation, redirect, malformed-response, timeout, network, and 5xx outcomes remain temporary uncertainty.
- A positive match cannot win while another relevant app-type probe is uncertain.
- Internal ambiguity is detected explicitly; the legacy token RPC preserves its existing `no_installation_found` mapping.

## Capability confinement

- Generated session capabilities are pinned to the authoritatively resolved Kilo integration ID even when the caller supplied only `owner/repo`.
[...]

PR #5597, opened 2026-08-27
feat(app-builder): persist the GitHub install resolved at migration

RSO:
## Why

App Builder performs direct Git operations and preview work survives restarts. Migration must resolve one installation authoritatively, persist it, and reuse it for every later clone or rebuild. Cached web metadata is not sufficient.

A migration can also perform an irreversible Git push before the web process receives its result. Concurrent requests therefore need one exclusive database claim before any request reaches the Worker; a lost response must not release that claim and permit a second push.

## Dependency

Depends on #5584.

## What changes

- Atomically acquire an exclusive migration claim only while both `migrated_at` and `git_repo_full_name` are unset.
- Allow only the claim winner to call the migration Worker; concurrent losers fail without pushing.
- Bind finalization and claim release to the exact claim timestamp so one request cannot finalize or release another request claim.
- Replay a completed retry only when repository and optional integration fence match the persisted result.
- Release the claim only for definitive pre-push failures; retain it after ambiguous Worker failures where a push may already have happened.
- Resolve migration `owner/repo` through Git Token Service.
- Treat optional selected `platformIntegrationId` as an exact fence.
- Persist the returned identity on PreviewDO, project, and deployment state.
- Reuse it for preview clone, fetch or rebuild, and restart token requests.
- Keep old unpinned preview state readable with authoritative fail-closed resolution.

## What this removes

No #5547 cached web resolver or resolver tests. Empty-repository validation now occurs in the Worker using the authoritative token.

## Review guide

1. Review the exclusive claim predicate, claim ownership token, and concurrent winner or loser tests in `github-migration-service.ts`.
2. Review which definitive failures release the claim and why ambiguous responses retain it.
3. Trace the Git Token Service result through the migration response into PreviewDO, project, and deployment state.
4. Verify later token calls use the persisted integration as an exact fence.

## Validation

- 46 App Builder tests
- Exclusive concurrent-claim, retry, conflicting-fence, owned-release, and ambiguous-response coverage
- App Builder and web typecheck or lint
- `git diff --check`

### DIFF
--- a/apps/web/src/lib/app-builder/github-migration-service.test.ts
+++ b/apps/web/src/lib/app-builder/github-migration-service.test.ts
@@ -1,0 +1,241 @@
+import { beforeAll, beforeEach, describe, expect, it, jest } from '@jest/globals';
+import type { SQL } from 'drizzle-orm';
+import { PgDialect } from 'drizzle-orm/pg-core';
+import type * as migrationService from './github-migration-service';
+
+type ProjectState = {
+  id: string;
+  title: string;
+  session_id: string | null;
+  deployment_id: string | null;
+  git_repo_full_name: string | null;
+  git_platform_integration_id: string | null;
+  migrated_at: string | null;
+};
+type WorkerMigrationResult =
+  | { success: true; platformIntegrationId: string }
+  | { success: false; error: string };
+
+const projectState: ProjectState = {
+  id: 'project-1',
+  title: 'Test project',
+  session_id: 'session-1',
+  deployment_id: null,
+  git_repo_full_name: null,
+  git_platform_integration_id: null,
+  migrated_at: null,
+};
+const dialect = new PgDialect();
+const migrationPredicates: SQL[] = [];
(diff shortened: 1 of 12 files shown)


RSO:
Fixed the exclusive/idempotent migration claim audit finding in `94b1ecb01`.

- Claim acquisition now requires both `migrated_at IS NULL` and `git_repo_full_name IS NULL`.
- Finalize and release are fenced by the exact claim timestamp, so a loser cannot mutate another operation.
- Completed same-repository retries return the persisted result; conflicting or in-flight migrations return `already_migrated` before calling App Builder.
- Explicit pre-push failures release the claim; ambiguous post-dispatch failures retain it to prevent repeated irreversible Worker pushes.
- Added concurrency, retry, integration-fence, claim-ownership, and ambiguous-failure coverage.

Checks passed: 7 focused web tests, 46 App Builder tests, App Builder typecheck/lint, web typecheck/lint, formatting, and `git diff --check`.

The Kilo Code Review check is currently cancelled because the selected model is unavailable; leaving that status unchanged for rerun after review settings are updated.

PR #5586, opened 2026-08-27
feat(kilo-pass): add Google Play purchase verifier

iscekic:
## Summary

No new behavior — the level adds an internal purchase verification step that no customer-facing path uses yet.

---

## Reviewer Notes

`verifyGooglePlayKiloPassPurchase` is a new entry point that fetches a Google Play subscriptions v2 payload for a purchase token and maps it onto `ValidatedStoreKiloPassPurchase` under the Google Play payment provider. The mapper now rejects a decoded purchase whose `startTimeMs` or `expiryTimeMs` is non-finite, so a pending subscription that omits its start time throws a domain error instead of failing later on an invalid `Date`. The mapping also rejects missing identifiers, expired purchases, and products outside the enabled mobile store catalog, so completion fails before the purchase ledger runs; expired purchases are rejected on this path only because renewals and refunds arrive through the notifications handler.

<details>
<summary>Files</summary>

- `apps/web/src/lib/kilo-pass/google-play-verifier.ts` — adds `verifyGooglePlayKiloPassPurchase`, `decodeGooglePlaySubscriptionPurchase`, and `mapGooglePlayKiloPassPurchase` plus the `GooglePlayEnvironment` and `GooglePlayDecodedPurchase` types; the mapper rejects non-finite timestamps before the expiry check, and `decodeGooglePlaySubscriptionPurchase` falls back to the top-level `latestOrderId` and keeps the raw payload.

</details>

Test: 1 test file added — `apps/web/src/lib/kilo-pass/google-play-verifier.test.ts`, 197 lines.

Generated: none.

---

## Verification

E2E report: none — a lower PR names its own change alone; the tip PR carries the E2E report. No manual test paths were run for this level.

## Visual Changes

Visual Changes: N/A

Human steps: none known.

## Correctness verification, September 7

Purchase verification uses the paid order period for monthly credits. It accepts the observed 232 ms grant timestamp difference. A refund claim blocks delayed completion.

- Current head: `a67c5fede`.
- The stack passes 314 tests across eight scoped suites, web TypeScript, and changed-file lint and formatting.
- Live Google Sandbox purchases pass for all three credit tiers. API-fixture accounting and lifecycle tests remain distinct from live E2E evidence.
- Full runtime evidence remains on #5618. No production deployment occurred.

**Stacked PRs — merge bottom to top.** Each level shows only its own diff.

Runtime verification (E2E, user advocacy, simplify) runs on the tip PR over every level.
[...]

PR #5598, opened 2026-08-27
feat(kilo-pass): handle Play real-time notifications

iscekic:
## Summary

Kilo Pass subscriptions bought on Google Play for Android now follow the store's subscription events.

- A purchase now activates the subscription immediately.
- A renewal, recovery, or restart now extends the subscription automatically.
- A cancellation now keeps the subscription active until the paid period ends.
- A refund or revocation now ends the subscription and removes the credits the purchase granted.

---

A new processor turns Google Play Real-time Developer Notifications into Kilo Pass lifecycle actions. It decodes the base64 Pub/Sub payload, rejects a package-name mismatch, and routes each notification type: purchase, renewal, recovery, and restart complete the store purchase; canceled marks end-of-period cancellation; expired ends the subscription; revoked reverses the granted credits and ends it; unknown types are claimed and ignored. Each event claims a `kilo_pass_store_events` row before it acts, so a delayed or redelivered message cannot apply a purchase or a reversal twice, and the stored payload token is redacted.

<details>
<summary>Files</summary>

- `apps/web/src/lib/kilo-pass/google-play-notifications.ts` — new processor. It base64-decodes the Pub/Sub payload, checks the package name, fetches and decodes the Play subscription purchase, and claims the `kilo_pass_store_events` row with a five-minute stale reclaim. It routes each notification type to its lifecycle action, reverses base, bonus, and promo credits on revocation, skips purchase completion when a processed revoked event already settled, and marks the event processed.

</details>

The notification endpoint authenticates each Pub/Sub push before it reaches the processor. It verifies the OIDC bearer token against the `GOOGLE_PLAY_RTDN_PUSH_AUDIENCE` audience and now also requires the token's verified `email` claim to equal `GOOGLE_PLAY_RTDN_PUSH_SERVICE_ACCOUNT_EMAIL`, so an arbitrary GCP service account cannot invoke the billing webhook. A missing, invalid, or mismatched token returns 401, a malformed body returns 400, a fresh in-flight duplicate returns 503 for Pub/Sub to retry, and a processing error returns 500 without leaking details.

<details>
<summary>Files</summary>

- `apps/web/src/app/api/kilo-pass/play/notifications/route.ts` — new POST route. It parses the body shape, reads both environment values, verifies the bearer with `OAuth2Client`, checks the email and `email_verified` claims against the configured service account, delegates to the processor, and reports failures to Sentry.
- `ENVIRONMENT.md` — catalogs the new `GOOGLE_PLAY_RTDN_PUSH_SERVICE_ACCOUNT_EMAIL` value.

</details>

`trackKiloPassPurchaseCompleted` now accepts a `google_play` channel next to `app_store`, and the property branch now splits on `stripe`. A Google Play completion therefore emits the same `kilo_pass_purchase_completed` event with the provider transaction, product, and environment properties.

<details>
<summary>Files</summary>

- `apps/web/src/lib/kilo-pass/posthog-tracking.ts` — adds `google_play` to the store channel union and changes the property branch condition to compare against `stripe`.

</details>

Tests: 3 files changed — `route.test.ts` and `google-play-notifications.test.ts` added, `posthog-tracking.test.ts` updated.
Generated: none.

---

## Verification

No manual test paths were run. No E2E report is attached for this level — a lower PR names its own change alone, and the tip PR carries the E2E report.

## Visual Changes

Visual Changes: N/A

## Reviewer Notes

- before merge: set `GOOGLE_PLAY_RTDN_PUSH_AUDIENCE` to the public HTTPS URL of the notifications endpoint. Run `pnpm web:env set GOOGLE_PLAY_RTDN_PUSH_AUDIENCE`.
- before merge: set `GOOGLE_PLAY_RTDN_PUSH_SERVICE_ACCOUNT_EMAIL` to the email of the Play service account that the Pub/Sub push uses. The route returns 401 for every notification until both values are set.
- E2E: bot-e2e — runtime verification runs once on the tip PR (level 6) after all levels land. This level is reviewed on its own diff.

## Correctness verification, September 7

Provider state now updates grace, hold, pause, restore, recovery, and cancellation without replaying credits. Refunds reverse only the matching order and prevent later bonuses. Delayed refunded purchase events settle without endless retries.

- Current head: `4d511fb84`.
- The stack passes 314 tests across eight scoped suites, web TypeScript, and changed-file lint and formatting.
- Live Google Sandbox purchases pass for all three credit tiers. API-fixture accounting and lifecycle tests remain distinct from live E2E evidence.
- Full runtime evidence remains on #5618. No production deployment occurred.

Runtime verification (E2E, user advocacy, simplify) runs on the tip PR over every level.
Every level keeps its own checks, its own bot review, and its own threads; each one is answered on its own PR.
Each level is its own deliverable: it builds and passes its own checks alone.

1. `android-iap-e895` — #5581
2. `android-iap-e895-s2` — #5582
3. `android-iap-e895-s3` — #5586
4. `android-iap-e895-s4` — #5598 ← this PR
5. `android-iap-e895-s5` — #5608
6. `android-iap-e895-s6` — #5618 (tip)

## Expanded correctness coverage

- Tested stack head: `16015d47374b30cf28e3b1b9b7f7635e4abcca86`.
- 332 backend tests passed across nine scoped suites. 115 mobile tests passed across thirteen files.
- Web and mobile TypeScript, scoped lint, formatting, and diff checks passed.
- Expired store subscriptions now permit a new purchase before the expiry notification arrives.
- Expiry cleanup locks the subscription before checking receipts. A PostgreSQL concurrency test reproduces the old renewal race.
- Mobile acknowledgement precedes account refresh. Recovery covers slow refresh, failed acknowledgement, backend outage, pending purchases, and mixed ownership results.
- Both Play completion routes acknowledge after committed credits. Notification retries cannot grant credits twice.
- Play center resubscriptions retain the prior account identifier during acknowledgement. Decoder and handler tests cover this temporary provider field.
- An older refunded order reverses only its own credits. The later paid period and its usage bonus remain valid.

### Live evidence

- Stopped the app immediately after Google test payment approval. Relaunch and restore granted credits exactly once.
- A second purchase granted credits and became acknowledged while the app remained stopped. Duplicate notification delivery preserved the balance.
- A final purchase through the updated completion route passed provider acknowledgement and active restore.
- Final artifacts: `parity-server-ack-interrupted`, `parity-server-ack-relaunch`, `parity-server-ack-restored`, `parity-final-server-route-active`, and `parity-final-server-route-restore`.
- All three tiers passed in the earlier correctness round. That round also covers decline, dismissal, ownership rejection, management, cancellation/restart, expiry, refund, and legal links.

### Explicit limits

- Calendar boundaries, concurrency, bonuses, provider failures, grace, hold, pause, deferral, and Play center resubscription use provider fixtures and the real database.
- These fixture cases are integration tests, not live Google end-to-end passes.
- The configured auto-renewing plans have no supported pending-payment test instrument. Pending transaction rejection has automated coverage.
- Google Play tier changes now use deferred replacement at renewal. See the tier-change verification below.
- iOS retains the earlier limited Sandbox smoke coverage.
- A developer refund without revocation does not guarantee a voided event. Exact web parity is not claimed.
- Production migration, deployment, authenticated delivery, and monitoring remain release gates.

### Cleanup

- All test subscriptions ended. The original account retains its initial 38 USD; the second account has zero credits.
- Zero owned events remain unprocessed. Temporary IAM permission, local pull subscription, review databases, and saved purchase payloads are removed.

## Google Play tier changes

- Head: `7b70df79fcaf375e362fce70b5c7f990380d42e0`.
- Upgrades and downgrades take effect at the next renewal, matching monthly web subscriptions.
- Profile Manage opens Kilo Pass plans. A different tier starts Google's deferred replacement flow. The current tier and Manage button open Google management.
- The current paid tier, balance, and subscription history remain intact before renewal.
- The replacement token is acknowledged immediately. A paid renewal updates the tier and uses the existing credit and streak rules.
- Deferred callbacks can name the old product. The app now completes those callbacks.
- The verifier selects the current entitlement from unordered old and future items. It validates the deferred order against the future product.
- Token transfers preserve the subscription ID. Refunds use original receipt tokens despite the database cascade.

### Verification

- 348 backend tests passed across nine suites. 123 mobile tests passed across thirteen files.
- All six transitions between 19, 49, and 199 have request and database coverage.
- Tests cover deferred acknowledgement, duplicate completion, reordered expiry notifications, renewal, retargeting, stale tokens, refunded receipts, account mismatch, and another active provider.
- Web and mobile TypeScript, scoped lint, formatting, and diff checks passed.
- The earlier format-check failure named store-subscription-completion.test.ts. The final commit formats that file; local read-back passes.
- Live downgrade 49 to19 passed approval, acknowledgement, retained current credits, renewal, and restore.
- Live upgrade 19 to199 passed the same checks. Profile Manage opened the plan screen successfully.
- Authenticated Google notifications and duplicate replay passed. Refund cleanup returned both accounts to their initial balances.
- Accelerated Sandbox renewals stay in one calendar month. Calendar-boundary fixtures verify the new month's credit grant and streak.

### Evidence and limits

- Artifacts: tier-change-49-19, tier-change-deferred-restore, tier-change-renewed19-active, tier-change-19-199-active, and tier-change-renewed199.
- The first 19 to49 attempt exposed the deferred-order and callback bugs. That replacement expired/refunded before acknowledgement. Its failure remains recorded.
- The generic high-tier helper expected a fresh-purchase balance during the upgrade. The following active-state assertion and provider read-back passed with the correct unchanged balance.
- Native tests prove one upgrade and one downgrade through renewal. All six combinations have automated fixture coverage, not six live Google passes.
- Current native screenshots remain in the local artifacts. GitHub's browser session is signed out, so this round adds no hosted screenshot.
- No backend deployment or mobile release occurred. Existing production release gates remain.

### Cleanup

- Original test account: 38 USD. Other test account: zero. No active test passes or unprocessed owned events remain.
- Removed temporary IAM permission, local notification subscription, review databases, Jest configuration, and saved purchase payloads.


apps/web/src/app/api/kilo-pass/play/notifications/route.ts
```diff
@@ -0,0 +1,59 @@
...
+  }),
+});
+
+async function verifyGooglePlayRtdnToken(idToken: string): Promise<void> {
+  const audience = getEnvVariable('GOOGLE_PLAY_RTDN_PUSH_AUDIENCE');
+  if (!audience) {
+    throw new Error('GOOGLE_PLAY_RTDN_PUSH_AUDIENCE is not set');
+  }
+  const client = new OAuth2Client();
+  await client.verifyIdToken({ idToken, audience });
```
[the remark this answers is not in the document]
iscekic: (bot) Fixed in 7072951c2 on android-iap-e895-s4: verifyGooglePlayRtdnToken now verifies the email and email_verified claims against GOOGLE_PLAY_RTDN_PUSH_SERVICE_ACCOUNT_EMAIL, mirroring gmail-push. Forwarded up the stack.

apps/web/src/lib/kilo-pass/google-play-notifications.ts
```diff
@@ -0,0 +1,817 @@
...
+    await markGooglePlayStoreEventProcessed(eventId);
+    return { processed: true };
+  }
+
+  if (notificationType === GOOGLE_PLAY_NOTIFICATION_TYPE.SUBSCRIPTION_REVOKED) {
+    await db.transaction(async tx => {
+      let reversal: CreditReversalResult | null = null;
+      try {
+        reversal = await reverseGooglePlayRefundCredits(tx, purchaseToken, decoded.latestOrderId);
+      } catch (error) {
```
pandemicsyn, line 874: bot: A failed refund clawback is acknowledged as a successful revocation.

Suggested fix: Do not catch-and-continue around `reverseGooglePlayRefundCredits`. Let the error escape (or explicitly rethrow after reporting it) so this transaction rolls back and the route returns 500, allowing Pub/Sub to redeliver the event. Only write the successful refund audit and `processed_at` after the reversal completes. Add a test which makes reversal fail and asserts that no processed event/success audit is committed and a retry can perform the clawback.

apps/web/src/lib/kilo-pass/google-play-notifications.ts
```diff
@@ -0,0 +1,817 @@
...
+        messageId: messageId ?? null,
+        providerSubscriptionId: purchaseToken,
+      },
+    });
+    await markGooglePlayStoreEventProcessed(eventId);
+    return { processed: true };
+  }
+
+  if (notificationType === GOOGLE_PLAY_NOTIFICATION_TYPE.SUBSCRIPTION_EXPIRED) {
+    await markGooglePlaySubscriptionEnded(db, purchaseToken);
```
pandemicsyn, line 856: bot: A stale expiry notification can cancel a subscription that has already renewed or recovered.

Suggested fix: Use `eventTimeMillis` to reject terminal events older than the newest applied lifecycle event for this purchase token, or reconcile terminal transitions with the verified current Play subscription state before updating. Apply the same protection to cancellation. Add tests that process a newer active renewal/recovery first and then a delayed cancellation/expiry, asserting the subscription remains active.

apps/web/src/app/api/kilo-pass/play/notifications/route.ts
```diff
@@ -0,0 +1,70 @@
...
+      return Response.json({ error: 'Unauthorized' }, { status: 401 });
+    }
+
+    try {
+      await verifyGooglePlayRtdnToken(bearerToken);
+    } catch {
+      return Response.json({ error: 'Unauthorized' }, { status: 401 });
+    }
+
+    const body = GooglePlayNotificationBodySchema.safeParse(await request.json());
```
pandemicsyn: bot: Malformed request JSON falls through as a 500 instead of the intended 400 client error.

Suggested fix: Wrap `await request.json()` in a small parse-specific `try/catch` and return the same 400 response used for an invalid Pub/Sub envelope; keep processor failures in the outer 500 handler. Add a test using a request with an invalid JSON body.

pandemicsyn (approved):

apps/web/src/lib/kilo-pass/google-play-notifications.ts
iscekic: Valid. Fixed.

The `catch` now rethrows after the Sentry capture, so the transaction rolls back, the success audit and `processed_at` are never written, and Pub/Sub redelivers.

A missing store purchase is not an error path: `reverseGooglePlayRefundCredits` returns `storePurchaseFound: false` for that case, so the rethrow only fires on a genuine failure and cannot cause endless redelivery of a benign revoke. `reversal` is now non-nullable, so the audit payload dropped its `?? ` fallbacks.

No test for the injected failure. Every input to the clawback derives from `storePurchase.kilo_user_id`, which is `ON DELETE CASCADE` / `ON UPDATE CASCADE` against `kilocode_users`, so the failure cannot be reached from test data without adding a seam to production code. The success path stays covered by the existing revoke tests.

Commit `7c9a8c271`.

apps/web/src/lib/kilo-pass/google-play-notifications.ts
iscekic: Valid. Fixed, by reconciling against Play rather than by event timestamps.

The handler already fetches the live subscription for every notification, so Play is available as the source of truth at no extra cost:

- Expiry (type 13) applies only when Play also reports the period as over. If Play still reports a future expiry, the subscription renewed or recovered after the notification was emitted, so the event is stale: mark processed, change nothing.
- Cancellation (type 3) applies only when Play does not report `SUBSCRIPTION_STATE_ACTIVE`. An active state means the buyer restored auto-renew, so the cancellation is stale.

Added `ignores a stale expiry when Play still reports a future expiry` and `ignores a stale cancellation when Play reports the subscription as active`. Two existing tests reused the active fixture for the terminal notification, which cannot happen in production; they now re-mock Play with the expired and canceled state.

Commit `7c9a8c271`.

apps/web/src/app/api/kilo-pass/play/notifications/route.ts
iscekic: Valid. Fixed.

`request.json()` now has its own `try/catch` that returns the same 400 as an invalid Pub/Sub envelope. A body that is not JSON can never become valid, so Pub/Sub must stop redelivering it. Processor failures stay in the outer 500 handler.

Added `returns 400 when the request body is not JSON`, which also asserts the processor is never called.

Commit `7c9a8c271`.

PR #5606, opened 2026-08-27
feat(cli-sessions): refresh PR status with canonical GitHub identity

RSO:
## Why
CLI PR metadata needs both #5565 cache provenance and #5590 Cloud Agent canonical session identity. The previous PR selected legacy installations from cached account/repository metadata; this replacement removes that authorization logic.

## Temporary review base
This PR targets a temporary branch containing patch-equivalent #5565 on top of #5590 so the review diff is one behavior commit. Do not merge the temporary base. After #5565 and #5590 merge, retarget/rebase this PR to main.

## What changes
- Expose the session persisted githubIntegrationId through Cloud Agent getSession.
- Prefer that canonical pin, then exact cache-row provenance.
- Fail legacy unpinned rows closed; web has no Git Token Service binding and does not infer from cache metadata.
- Group review decisions by exact integration/installation/app type.
- Scope secondary PR/review webhook writes to exact cache rows.

## What this removes
No local account-login/repository-access matching resolver.

## Validation
- 153 focused web tests
- 148 focused Cloud Agent tests
- Web/Cloud Agent/DB typechecks and lint
- Formatting and `git diff --check`

PR #5609, opened 2026-08-27
feat(cli-sessions): refresh PR status with canonical GitHub identity

RSO:
Replaces #5606 on a fresh integration base containing the updated #5565 patch-equivalent.\n\n- Persists the canonical GitHub integration on CLI session rows\n- Joins and refreshes PR cache rows only for that exact integration\n- Keeps null legacy cache rows isolated and retry-safe\n- Preserves transient batch claims for retry\n\nMigration note: the stacked #5565 patch still carries the known 0233 collision; unrelated #5536 resolution is intentionally deferred.


RSO:
Superseded by #5612, which replays the same two behavior commits directly on the stable schema child #5611. The replacement stack now includes the latest #5584/#5590 fixes and no longer depends on the temporary hidden integration base.

Issue #4498, opened 2026-07-10
feat: add option to hide forks when selecting repos


shssoichiro:
Reopened from #2599. This issue is still valid.

PR #5770, opened 2026-08-31
feat(web): add a hide-forks option to the repository picker

iscekic:
## What changed

- The repository picker shows a **Hide forks** checkbox when the account has at least one forked repository.
- When the checkbox is on, the picker hides forked repositories from the list.
- The checkbox state persists per browser under the `repo-picker:hide-forks` local-storage key.
- Select All and Deselect All act on the listed repositories only. They keep selections that the fork filter hides.
- The counter reports the listed selection, and adds `N forks hidden` when the filter is on.
- The GitHub repository fetch carries the `fork` flag through to the picker for Code Reviewer, Security Agent, Auto Triage, and Auto Fix.
- A cached repository list that predates the `fork` flag triggers one refetch, so the flag backfills without a migration.

## Why

Issue #4498 asks for a way to hide forked repositories when a user selects repositories to watch. An account with many forks makes the picker hard to read. The GitHub API already returns the `fork` flag, but the picker's data path dropped it.

## Verification

- Driver checks: per-slice checks passed.
- Section e2e `backend` passed: no changed service.

## Notes for the reviewer

- `fork` is optional on the repository types. A repository cached before this change reports `fork: undefined` and stays listed until the backfill refetch replaces it.
- The selection helpers are pure functions in `repository-multi-select-selection.ts` and have unit tests.

### DIFF
--- a/apps/web/src/components/auto-fix/AutoFixConfigForm.tsx
+++ b/apps/web/src/components/auto-fix/AutoFixConfigForm.tsx
@@ -356,6 +356,7 @@ export function AutoFixConfigForm({ organizationId }: AutoFixConfigFormProps) {
                             name: repo.name,
                             full_name: repo.fullName,
                             private: repo.private,
+                            fork: repo.fork,
                           })) as Repository[]
                         }
                         selectedIds={selectedRepositoryIds}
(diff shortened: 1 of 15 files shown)


iscekic:
Opened as part of an internal workflow test, if useful, please take over.

Issue #4268, opened 2026-06-26
nitpick: BYOK Code Reviewer still uses hardcoded Gemma 4 (Kilo credits) alongside the configured BYOK model — please use the BYOK model for everything


baanish:
> **Nitpick:** The amount of Kilo credit at stake is a whopping **$0.07** (seven cents) over the past month, so this is about BYOK expectations, not money.

### Summary

I have my Code Reviewer set to a BYOK model, `Neuralwatt: glm-5.2-short`, and it **does** use GLM 5.2 short for the actual reviews. That part works as expected.

The issue: **in addition to GLM 5.2 short, Code Review also uses Google Gemma 4 (`google/gemma-4-26b-a4b-it`) for an auxiliary step**, and that Gemma call goes through Kilo (not my BYOK provider). So even though I'm on BYOK, my Kilo credit balance keeps ticking down and I have to keep topping it up. I'd expect a BYOK setup to use my configured model for *everything* in the Code Review flow, not fall back to a hardcoded Gemma model that bills Kilo credits.

### What I expect

When a BYOK model is configured for Code Reviewer, the whole Code Review flow (including the auxiliary review-thread classification step) should use that configured BYOK model, so BYOK users don't have to keep topping up Kilo credits for review-related calls.

### What I saw

- Review Configuration shows the selected BYOK model `Neuralwatt: glm-5.2-short`.
- The actual review comment from `kilo-code-bot` is explicitly footed **"Reviewed by glm-5.2-short"** — good, the configured model is honored for the review itself.
- But the Usage dashboard filtered to Code Reviews ($0.07 / 1,032 requests / 32.9M tokens) shows the model breakdown dominated by `google/gemma-4-26b-a4b-it-*`, i.e. a second model running alongside GLM 5.2 short and drawing down Kilo credits.

Configured BYOK Code Reviewer model:

<img width="2200" height="632" alt="Review Configuration AI Model set to Neuralwatt: glm-5.2-short" src="https://github.com/user-attachments/assets/52b996b8-49ce-4dca-b876-3c5de7a2d0cf" />

Actual review, footed "Reviewed by glm-5.2-short":

<img width="1768" height="632" alt="kilo-code-bot Code Review Summary footer: Reviewed by glm-5.2-short" src="https://github.com/user-attachments/assets/73181cf8-ef25-4c89-8b7f-63dc1ec3e83f" />

Usage dashboard (Past Month, Code Reviews) showing Gemma running alongside the configured model:

[...]

PR #5774, opened 2026-08-31
fix(gastown): use BYOK model for review classification

iscekic:
## What changed

- Review-thread classification uses the configured direct BYOK reviewer model.
- Non-BYOK reviews keep the existing Workers AI classifier.
- A failed BYOK call blocks the review without using the Kilo-billed fallback.

## Why

The auxiliary classification call used Kilo credits even when the user configured a BYOK model (#4268).

## Verification

- Unit tests for the changed packages pass (gastown, worker-utils, web drift test).
- Typecheck and lint pass on the changed packages.
- The web Playwright smoke is advisory in the local harness and did not run to completion; CI covers it.


iscekic:
Created as part of my own internal experimentation, closing (unless it's useful, then feel free to take over).

PR #5784, opened 2026-08-31
fix(app-builder): fix user ID validation for App Builder

evanjacobson:
## Summary

- Allow GitHub migration requests to use any non-empty Kilo user ID instead of requiring a UUID.
- Keep optional organization IDs restricted to UUIDs.
- Add schema tests covering non-UUID user IDs, empty user IDs, and organization ID UUID validation.

## Testing

- Added Vitest coverage for `MigrateToGithubRequestSchema`.


services/app-builder/src/api-schemas.ts
```diff
@@ -141,7 +141,7 @@ export type DeleteErrorResponse = z.infer<typeof DeleteErrorResponseSchema>;
 
 export const MigrateToGithubRequestSchema = z.object({
   githubRepo: z.string().regex(/^[^/]+\/[^/]+$/, 'Must be in "owner/repo" format'),
-  userId: z.string().uuid(),
```
evanjacobson, line 144: Users with non-UUID Kilo User IDs fail due to the parsing here

apps/web/src/components/app-builder/ProjectManager.ts
```diff
@@ -123,12 +125,16 @@ export function createProjectManager(config: ProjectManagerConfig): ProjectManag
     const sessionInfos = proj.sessions;
     if (sessionInfos.length === 0) return [];
 
-    const activeInfo =
-      sessionInfos.find(s => s.ended_at === null) ?? sessionInfos[sessionInfos.length - 1];
+    const activeInfo = project.session_id
+      ? sessionInfos.find(s => s.cloud_agent_session_id === project.session_id)
+      : undefined;
 
+    const orderedSessionInfos = activeInfo
```
evanjacobson: `project.session_id` is authoritative, but existing chat and feedback consumers treat the final array entry as active. Moving the canonical session to the end keeps those consumers aligned while older/orphaned sessions remain available as history.

apps/web/src/components/app-builder/ProjectManager.ts
```diff
@@ -339,6 +353,7 @@ export function createProjectManager(config: ProjectManagerConfig): ProjectManag
     }
 
     const effectiveModel = model ?? store.getState().model;
+    const isRecoveringSession = store.getState().isRecoveringSession;
```
evanjacobson: Capture the mode before clearing `pendingNewSession`: a failed mandatory recovery must remain non-cancelable, while a failed user-requested New Chat must remain optional and cancellable.

apps/web/src/components/app-builder/ProjectManager.ts
```diff
@@ -259,15 +266,22 @@ export function createProjectManager(config: ProjectManagerConfig): ProjectManag
...
   // `initiated` lives on ProjectSessionInfo (routing data), not on SessionDisplayInfo.
-  const activeProjectSessionInfo =
-    project.sessions.find(s => s.ended_at === null) ??
-    project.sessions[project.sessions.length - 1];
+  const activeProjectSessionInfo = project.session_id
+    ? project.sessions.find(s => s.cloud_agent_session_id === project.session_id)
+    : undefined;
 
-  if (activeProjectSessionInfo?.initiated === false) {
+  if (activeProjectSessionInfo?.prepared === true && activeProjectSessionInfo.initiated === false) {
```
evanjacobson: This deliberately distinguishes a prepared-but-not-started session from a missing session. Only the former can be started in place; missing sessions must use the replacement-session flow.

apps/web/src/lib/app-builder/app-builder-client.ts
```diff
@@ -294,15 +294,24 @@ export async function migrateToGithub(
     body: JSON.stringify(config),
   });
 
+  const data = await response.json().catch(() => undefined);
+  const parsed = MigrateToGithubResponseSchema.safeParse(data);
+
+  if (parsed.success && !parsed.data.success) {
```
evanjacobson: Worker domain failures intentionally use non-2xx statuses while still returning this typed envelope. Parse that contract before generic HTTP handling so callers retain the actual failure category.

apps/web/src/lib/app-builder/app-builder-service.ts
```diff
@@ -560,7 +573,7 @@ export async function getProject(
           err
         );
         sessionInitiated = null;
-        sessionPrepared = null;
+        sessionPrepared = isDefinitiveSessionNotFoundError(err) ? false : null;
```
evanjacobson: `false` means the backend session is definitively absent and enables replacement recovery. Transient Cloud Agent failures stay `null` so a temporary outage does not abandon an otherwise valid session.

apps/web/src/lib/app-builder/github-migration-service.ts
```diff
@@ -232,11 +232,14 @@ export async function migrateProjectToGitHub(
       });
 
       if (!migrateResult.success) {
-        throw new MigrationError('push_failed', { cause: migrateResult });
+        throw new MigrationError(
+          migrateResult.error === 'push_failed' ? 'push_failed' : 'internal_error',
```
evanjacobson: Only `push_failed` means a Git push was actually attempted and failed. Validation, token, transport, and unexpected Worker failures are surfaced as `internal_error` rather than showing misleading push-failure UI.

apps/web/src/components/app-builder/project-manager/sessions/v2/streaming.ts
```diff
@@ -469,7 +469,6 @@ export function createV2StreamingCoordinator(config: V2StreamingConfig): V2Strea
     }
 
     logger.log('Connecting to existing V2 session', { sessionId });
```
evanjacobson, line 471: A successful transport connection does not mean the agent is working. Replayed `busy` events now establish streaming state, so a zero-event reconnect cannot leave the UI on `Building...` indefinitely.

eshurakov (approved):

evanjacobson:
After a closer look at the issue, I've reduced the scope of this PR to cover only the UUID validation failure. The other fixes require a little more TLC  but we don't have many users asking for the session fixes at this time.

PR #5802, opened 2026-09-01
fix(gateway): decode bounded Zstandard inference requests

marius-kilocode:
## What Problem This Solves

Large inference JSON can exceed the Vercel Function body limit even when the selected model still has context capacity. Images, tool context, and encrypted reasoning contribute to the request. The gateway currently reads compressed bytes as text, so a Zstandard request reaches the handler but fails JSON parsing.

## Why This Change Was Made

Accept Content-Encoding: zstd at the shared OpenRouter gateway handler and its /api/gateway alias, then pass the decoded JSON through the existing validation, authentication, routing, and accounting flow. Ordinary identity requests retain Fetch UTF-8 and BOM handling. No image or reasoning transformation and no provider API schema change is introduced.

Use native Node Zstandard with application-defined 32 MiB input and decoded-output limits and a 32 MiB decoder-window limit. These are resource bounds, not claims about Vercel or provider limits. Unsupported or stacked encodings return 415, malformed frames 400, resource-limit failures 413, and cancellation 499. The upstream request does not inherit the compressed Content-Encoding or Content-Length.

Accept exactly one complete standard Zstandard frame. Node 24 can accept truncated frames or silently discard additional frames, so explicit frame-boundary validation rejects truncated, concatenated, and skippable bodies rather than accepting incomplete content.

## User Impact

Clients can send larger decoded inference payloads within the wire-size limit without reducing image quality or discarding conversation content. This does not remove provider context limits or the Vercel limit on compressed bytes.

Deploy this gateway change before enabling the client counterpart: https://github.com/Kilo-Org/kilocode/pull/13675. The client PR remains in draft for this server-first rollout. The repository pins Node 24; deployed runtimes must provide native Zstandard.

## Evidence

Node 24.14.1 passed 112 focused tests across the decoder, existing gateway route suite, and upstream timeout suite. Web typecheck, changed-file lint, formatting, and whitespace checks pass. Route tests verify validation through both aliases and decoded upstream JSON without compressed transport headers. Independent offline review exercised framing and stalled-stream cleanup on Node 25 without finding an actionable issue.

A local Node 24 HTTP test connected the actual client encoder to this decoder: 4,711,235 identity bytes became 1,123,346 wire bytes, with identical decoded SHA-256 and unchanged image and reasoning fields. The fixture contains repeated images; this ratio is not a general compression guarantee. Additional local checks reject truncation and stacked encodings. No model provider was called, and this PR has not been deployed or validated end to end on hosted Vercel.

### DIFF
--- a/apps/web/src/lib/ai-gateway/request-body.ts
+++ b/apps/web/src/lib/ai-gateway/request-body.ts
@@ -1,0 +1,116 @@
+import { Readable, Transform, Writable } from 'node:stream';
+import { pipeline } from 'node:stream/promises';
+import type { ReadableStream } from 'node:stream/web';
+import { constants, createZstdDecompress } from 'node:zlib';
+
+const MAX_BODY_BYTES = 32 * 1024 * 1024;
+
+function complete(buffer: Buffer): boolean {
+  if (buffer.length < 5 || buffer.readUInt32LE(0) !== 0xfd2fb528) return false;
+  const descriptor = buffer.readUInt8(4);
+  const single = (descriptor & 0x20) !== 0;
+  const dictionary = descriptor & 3;
+  const size = descriptor >>> 6;
+  let offset =
+    5 +
+    (single ? 0 : 1) +
+    (dictionary === 3 ? 4 : dictionary) +
+    (size === 0 ? (single ? 1 : 0) : 2 ** size);
+
+  for (;;) {
+    if (buffer.length - offset < 3) return false;
+    const block = buffer.readUIntLE(offset, 3);
+    const type = (block >>> 1) & 3;
+    if (type === 3) return false;
+    offset += 3 + (type === 1 ? 1 : block >>> 3);
+    if (offset > buffer.length) return false;
+    if (block & 1) break;
+  }
+  return offset + (descriptor & 4 ? 4 : 0) === buffer.length;
+}
+
+export async function readGatewayRequestBody(
+  request: Request,
+  limits: { input?: number; output?: number; window?: number } = {}
+): Promise<{ text: string } | { status: 400 | 413 | 415 | 499; error: string }> {
+  const encoding = request.headers.get('content-encoding')?.trim().toLowerCase() ?? 'identity';
+  if (encoding !== 'identity' && encoding !== 'zstd') {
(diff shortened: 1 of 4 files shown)


apps/web/src/lib/ai-gateway/request-body.ts
```diff
@@ -0,0 +1,116 @@
+import { Readable, Transform, Writable } from 'node:stream';
+import { pipeline } from 'node:stream/promises';
+import type { ReadableStream } from 'node:stream/web';
+import { constants, createZstdDecompress } from 'node:zlib';
+
+const MAX_BODY_BYTES = 32 * 1024 * 1024;
+
+function complete(buffer: Buffer): boolean {
```
chrarnoldus, line 8: bit annoying we need this, but it is what it is I guess

apps/web/src/lib/ai-gateway/request-body.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+  return offset + (descriptor & 4 ? 4 : 0) === buffer.length;
+}
+
+export async function readGatewayRequestBody(
+  request: Request,
+  limits: { input?: number; output?: number; window?: number } = {}
+): Promise<{ text: string } | { status: 400 | 413 | 415 | 499; error: string }> {
+  const encoding = request.headers.get('content-encoding')?.trim().toLowerCase() ?? 'identity';
+  if (encoding !== 'identity' && encoding !== 'zstd') {
```
chrarnoldus, line 37: can we add some logging on what it's doing here?

have you considered just returning text() when no content-encoding is set?

apps/web/src/lib/ai-gateway/request-body.ts
```diff
@@ -0,0 +1,116 @@
...
+    return { status: 415, error: 'Unsupported Content-Encoding. Use identity or zstd.' };
+  }
+
+  const exceeded = new Error('Request body exceeds the gateway resource limit.');
+  const chunks: Buffer[] = [];
+  const utf8 = new TextDecoder();
```
chrarnoldus, line 43: technically we should read the encoding from the request headers I guess, not sure if this is a regression

PR #5804, opened 2026-09-01
fix(mobile): repair session layouts and localized copy

iscekic:
## Summary

- Fix mobile session layouts, target refresh, cancellation notices, suggestion details, and GitHub connection recovery.
- Audit capitalization across all 87 locales, shorten biometric descriptions, and translate the live-session empty state.
- Show the Kilo logo on Home. Remove context selectors from Home and Live Sessions, and remove filter pills from both session lists.
- Use equal search-field margins and preserve filtering through the filter buttons and dialogs.
- Count only unacknowledged questions and permission requests in the Agents tab badge. Keep the header’s all-live count unchanged.

## Verification

- [x] Created a fresh iPhone 17 Pro simulator running iOS 26.5. No existing Kilo E2E simulator was reused.
- [x] Inspected native screenshots for English and Serbian copy, long organization names, and the credits row.
- [x] Inspected the final Home logo, selector-free layouts, centered Live empty state, and one-line session title at the default font size.
- [x] Applied a history filter and verified that the filter count remains visible without a pill strip.
- [x] Verified a fixture with four live sessions and two needs-input sessions; the header showed four and the tab label announced two.
- [x] Opened seeded session details and child transcripts; inspected context removal and the absence of empty message bars.
- [x] Opened the real suggestion and message-details components through temporary fixtures; inspected inline content, read-only details, and centered sheet titles.
- [x] Inspected the actual unlock scene with a temporary forced state; restored normal authentication afterward.
- [ ] Did not exercise live queued cancellation: the local transcripts are read-only and no writable queued session was available.
- [ ] Did not exercise real GitHub reconnection: the local account has fixture integration data, not a real linked authorization.
- [ ] Did not complete native biometric authentication: the simulator requested a passcode.
- [ ] Did not exercise discovery of a newly connected target. Real acknowledgment transitions are covered by automated tests, not native interaction.

## Visual Changes

All screenshots use the default font size. The before images show earlier refinement stages, not a complete baseline for this PR. Unlock screenshots use a temporary forced state.

| Screen | Before | After |
|---|---|---|
| Live Sessions empty state | Sessions with a context selector before the final refinements | Live Sessions without a context selector and with a centered empty state |
| Home | Home with the welcome heading and context selector | Home with the Kilo logo and no context selector |
| All Sessions | All Sessions with a filter pill | All Sessions without filter pills and with an active filter count |
| Unlock alignment | Unlock content near the top of the screen | Vertically centered unlock content |
| Long session title | Long titles could wrap; no matching before screenshot was captured. | Long session title on one line with an ellipsis |

### Needs-input badge

The temporary fixture contains four live sessions: one question, one permission request, one running session, and one idle session.

Four live fixture sessions with two needs-input rows and a tab badge of two

## Reviewer Notes

- Removed all temporary fixture overrides and restored the real shared hook before final validation. The original worktree remains unchanged.
- The badge uses the existing attention and acknowledgment rules, including hydration and repeated requests, independently of mounted list rows.
- Suggestion details are read-only; the composer remains the sole action surface to prevent duplicate submissions.
- GitHub recovery fetches fresh authorization and refreshes active review queries when connected. Transient token failures now return retryable `BAD_GATEWAY`.
- Native testing caught the unlock layout problem: centering now uses the same style object as safe-area padding.
- Final full mobile suite: **625 files and 8,218 tests passed** with `pnpm --dir apps/mobile test --maxWorkers=2`.
- Passed repository lint, affected-package typechecks, unused-code checks, class checks, formatting, and whitespace checks.
- All **87 catalogs passed**, including removal of the two obsolete pill labels.
- Focused backend suite: **11 tests passed**, with database hooks disabled and zero network connection attempts.
- The broader app-wide empty-state layout is a separate proposal; this PR does not implement that migration.


pandemicsyn (approved):

apps/mobile/src/components/agents/use-live-session-query.ts
```diff
@@ -15,16 +15,8 @@ import { LIVE_SESSION_FILTERS_KEY } from '@/lib/storage-keys';
...
-    platformFilter,
-    projectFilter,
-    activeFilterCount,
-    hasLoaded,
-    setFilters,
-    clearFilters,
-    setPlatformFilter,
-    setProjectFilter,
-  } = usePersistedAgentSessionFilters(LIVE_SESSION_FILTERS_KEY);
+  const { platformFilter, projectFilter, activeFilterCount, hasLoaded, setFilters, clearFilters } =
```
pandemicsyn, line 18: bot: Persisted live filters can become invisible and keep the list empty after their matching rows disappear.

Suggested fix: Keep selected platform/project values represented in the modal options (for example, union persisted selections with options derived from live rows), or retain another per-filter removal surface. Add a regression test that restores a filter for a platform/project absent from the current live set, then verifies the user can remove it from the filter UI. The current change removes the individual removal callbacks; options are still built only from current sessions, so the dialog cannot render an absent persisted selection.

pandemicsyn (approved):

apps/mobile/src/components/agents/use-live-session-query.ts
iscekic: Fixed in f5d9bf7fa.

The modal now includes saved platform and project selections, even when their live rows disappear. Users can remove those selections without clearing other filters.

Added two regression cases: other live rows present, and no live rows. Both failed before the fix. All 105 focused tests passed. Mobile typecheck, lint, unused-code, and formatting checks passed.

PR #5821, opened 2026-09-01
fix(mobile): keep dark status bar icons until the splash reveal ends

iscekic:
## Problem

In dark mode the status bar icons (clock, WiFi, battery) turn white at cold start, while the yellow splash overlay is still on screen. White on yellow is unreadable, and the flip happens before the entry animation finishes.

Cause: `<StatusBar style="auto" />` in `src/app/_layout.tsx` resolves to light content as soon as the tree mounts. It does not know the overlay is still painting yellow.

## Fix

`AnimatedSplashOverlay` mounts its own `<StatusBar style="dark" />` while the yellow frame is visible. React Native merges status bar props in mount order, so the overlay's bar wins over the root `auto` bar. The overlay unmounts at the end of the reveal, the override pops, and `auto` takes over again.

## Test

`animated-splash-overlay.mounted.test.tsx` asserts the dark bar is mounted during the splash and gone after dismissal.

## Checks

- `pnpm test` (623 files, 7997 tests) — pass
- `pnpm typecheck` — pass
- `scripts/lint-all.sh` — pass
- `pnpm check:unused` — pass

### DIFF
--- a/apps/mobile/.oxlintrc.json
+++ b/apps/mobile/.oxlintrc.json
@@ -257,7 +257,7 @@
       }
     },
     {
-      "files": ["src/app/_layout.tsx"],
+      "files": ["src/app/_layout.tsx", "src/components/animated-splash-overlay.tsx"],
       "rules": {
         "react/style-prop-object": "off"
       }
--- a/apps/mobile/src/components/animated-splash-overlay-reveal.mounted.test.tsx
+++ b/apps/mobile/src/components/animated-splash-overlay-reveal.mounted.test.tsx
@@ -43,6 +43,7 @@ vi.mock('expo-splash-screen', () => ({
   preventAutoHideAsync: vi.fn(),
 }));
 vi.mock('@sentry/react-native', () => ({ TimeToFullDisplay: () => null }));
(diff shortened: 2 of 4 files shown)


pandemicsyn (approved):

iscekic:
Verified on the iOS simulator in dark mode, 60 fps capture, before and after.

Metric: frames where the status bar sits on the yellow splash **and** the glyphs are lighter than the background.

| build | white-on-yellow frames | duration |
| --- | --- | --- |
| main | 82 | 10.23 s (t=2.38 → 12.60) |
| this PR | 0 | — |

The bar reverts on its own: a screenshot after the reveal shows light glyphs on the dark bar again, so popping the overlay's stack entry hands control back to the root `style="auto"` bar.

Supporting checks:
- `UIViewControllerBasedStatusBarAppearance` is `false` in the built app, so appearance is app-wide via `UIApplication`. `react-native-screens` per-screen `statusBarStyle` cannot preempt the RN status bar module.
- No other `StatusBar` exists in app code, and no navigation library pushes a stack entry.

PR #5864, opened 2026-09-03
feat(admin): add confirmed GitHub App uninstall

pandemicsyn:
## Summary

Follow-up to #5861.

- Add an employee-admin **Uninstall GitHub App** action for verified, locally associated GitHub organization installations. Require the exact installation ID as confirmation and explain repository-access impact and retained Kilo history/settings.
- Recheck current admin access and the selected owner/app/installation/account snapshot under database locks, then verify the upstream installation by ID before deleting it. Only GitHub 204 permits local cleanup; permission errors, missing installations, and timeouts remain unconfirmed.
- Persist owner-specific attempt/outcome audits, distinguish confirmed GitHub deletion from incomplete local cleanup, and prevent immediate retries using stale lookup results.
- Reconcile signed installation-deleted webhooks by exact installation and app scope, including legacy standard-app records with null app type, without touching Lite or sibling installations.
- No schema changes, new dependencies, customer-admin permissions, local-only disconnect, or user OAuth-grant revocation.

## Verification

- [x] Browser-checked with synthetic local identities and mocked destructive responses; no real GitHub installations were removed.
- [x] Checked typed confirmation, safe cancellation, exact target submission, pending-state locks, and refreshing the submitted organization rather than edited input.
- [x] Checked error retry lockout, stale-action removal, and preservation of partial-cleanup warnings when refresh fails.
- [x] Checked confirmation-dialog layouts at 375px, 900px, and 1440px.
- [x] Verified actual HTTP rejection of anonymous callers (401) and non-admin callers (403).

## Visual Changes

New uninstall action on eligible association rows, destructive confirmation dialog, and outcome feedback. Screenshots omitted as requested.

## Reviewer Notes

- Independent static review and re-review completed without remaining feature-specific blockers. Findings drove authorization locking, duplicate detection, retry lockout, and legacy webhook reconciliation fixes.
- Only verified local associations are eligible; orphan installations and uncertain/candidate records do not get destructive controls. Existing agent settings and history are retained.
- GitHub and PostgreSQL are not atomic. A confirmed uninstall with failed local cleanup is reported explicitly; the deletion webhook normally reconciles it, with manual investigation required if the row persists.
- Completed focused automated runs: uninstall/UI/router/adapter tests (5 suites, 50 tests) and lifecycle/database/uninstall tests (3 suites, 50 tests; overlapping coverage). Tests include stale identities, revoked/blocked admins, lock serialization, 401/403/404/timeouts, partial cleanup, audit attribution, legacy records, and sibling preservation.
- Formatting, web lint, and diff whitespace checks passed. The final typecheck found Jest mock-signature errors in the added test; those signatures were corrected afterward. No further check was run after the request to stop testing.
- Broader combined test attempts timed out or were interrupted and are not claimed as passing. All DB-backed checks used the isolated worktree database prepared with `pnpm test:db`.

### DIFF
--- a/apps/web/src/lib/admin/github-installation-uninstall.ts
+++ b/apps/web/src/lib/admin/github-installation-uninstall.ts
@@ -1,0 +1,215 @@
+import 'server-only';
+
+import { and, eq, isNull, or } from 'drizzle-orm';
+import { TRPCError } from '@trpc/server';
+import { db, type DrizzleTransaction } from '@/lib/drizzle';
+import { createAuditLog } from '@/lib/organizations/organization-audit-logs';
+import { verifyAndDeleteGitHubOrganizationInstallation } from '@/lib/integrations/platforms/github/adapter';
+import { kilocode_users, platform_integrations, user_admin_notes } from '@kilocode/db/schema';
+import type { GitHubInstallationUninstallInput } from './github-installation-uninstall-input';
+
+const RETRY_MESSAGE =
+  'GitHub installation uninstall could not be confirmed. Refresh before retrying.';
+
+type Actor = { id: string; email: string; name: string | null };
+type LocalRecord = {
+  id: string;
+  ownerType: 'user' | 'organization';
+  ownerId: string;
+  appType: 'standard' | 'lite';
+  installationId: string;
+  accountId: string;
+};
+
+function uninstallError(code: 'BAD_REQUEST' | 'FORBIDDEN' | 'CONFLICT' = 'CONFLICT') {
+  return new TRPCError({ code, message: RETRY_MESSAGE });
+}
+
+type AuditStatus = 'attempted' | 'confirmed' | 'unconfirmed' | 'confirmed; local cleanup pending';
+
+function auditMessage(record: LocalRecord, status: AuditStatus) {
+  return `GitHub ${record.appType} App integration ${record.id} installation ${record.installationId} (${record.accountId}) uninstall ${status}.`;
+}
+
+async function writeAudit(
+  executor: typeof db | DrizzleTransaction,
+  actor: Actor,
+  record: LocalRecord,
+  status: AuditStatus
+) {
+  if (record.ownerType === 'organization') {
+    const audit = {
+      organization_id: record.ownerId,
+      action: 'organization.settings.change',
+      actor_id: actor.id,
+      actor_email: actor.email,
+      actor_name: actor.name,
+      message: auditMessage(record, status),
+    } as const;
+    if (executor === db) await createAuditLog(audit);
+    else await createAuditLog({ ...audit, tx: executor as DrizzleTransaction });
+    return;
+  }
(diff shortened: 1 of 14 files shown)


iscekic (approved):

apps/web/src/lib/admin/github-installation-uninstall.ts
```diff
@@ -0,0 +1,215 @@
...
+    admin_kilo_user_id: actor.id,
+    note_content: auditMessage(record, status),
+  });
+}
+
+async function tryWriteAudit(params: { actor: Actor; record: LocalRecord; status: AuditStatus }) {
+  try {
+    await writeAudit(db, params.actor, params.record, params.status);
+  } catch {
+    return false;
```
intentionally-left-nil, line 64: If the audit log fails, should we fail out here given this is pretty destructive?

apps/web/src/lib/admin/github-installation-uninstall.ts
```diff
@@ -0,0 +1,215 @@
...
+      platform: platform_integrations.platform,
+      userId: platform_integrations.owned_by_user_id,
+      organizationId: platform_integrations.owned_by_organization_id,
+      appType: platform_integrations.github_app_type,
+      installationId: platform_integrations.platform_installation_id,
+      accountId: platform_integrations.platform_account_id,
+    })
+    .from(platform_integrations)
+    .where(eq(platform_integrations.id, input.integrationId))
+    .for('update');
```
intentionally-left-nil, line 93: As a general pattern do we use row locking a lot? It causes deadlock possibilities, and also you have to remember to unlock on all exception paths

intentionally-left-nil: Instead, can we do a single UPDATE WHERE SELECT sublcause pattern to do this using database primitives?

apps/web/src/lib/admin/github-installation-uninstall.ts
pandemicsyn: Addressed in follow-up PR #5866. The pre-attempt audit still hard-fails before GitHub is called, and the confirmed audit remains atomic with local cleanup. Terminal fallback audit writes stay best-effort because the remote deletion may already be irreversible, but failures are now reported to Sentry with minimal safe context instead of being silently swallowed.

apps/web/src/lib/admin/github-installation-uninstall.ts
pandemicsyn: Addressed in follow-up PR #5866 with rationale and regression coverage. PostgreSQL row locks are transaction-scoped and auto-release on commit or rollback; tests now explicitly prove release on both paths. We retained the locks because a conditional final UPDATE cannot fence actor revocation or target reassignment during the validation-to-irreversible-GitHub-delete gap. A lock-free design would require a durable claimed/deleting state rather than a single UPDATE subquery.

PR #5879, opened 2026-09-04
ci(kilo-app): bump the app version after a store submission

iscekic:
## What

The kilo-app Release workflow patch-bumps the app version after a store submission.

- No bump PR open: build, submit to both stores, then open a PR that bumps `version` in `apps/mobile/app.config.ts`.
- A bump PR already open: build and submit as usual, skip the bump steps with a warning. The open PR already carries the next version.

## Why

A store release needs a version string that was not released before. Today the version is edited by hand, so a release can submit a version that was already released.

The build number is untouched. EAS auto-increments it remotely (`appVersionSource: "remote"`).

## Notes

- The bot cannot push to `main` (ruleset `main-reviews`, no bypass actor), so the bump ships as a PR. The job copies the pattern in `.github/workflows/extension-store-release.yml`.
- `check-changes` skips a run whose only change since the last release tag is the version line. Without it, merging a bump PR starts a release of identical code, which bumps again.
- The bump PR's required checks show as pending: PRs opened with the default `GITHUB_TOKEN` start no workflow runs. Close and reopen the bump PR to run CI.

## Test

`sed` and guard logic checked against the real `app.config.ts`: reads `1.0.7`, writes `1.0.8` on one line only; a version-only diff skips; a mixed diff builds.

### DIFF
--- a/.github/workflows/kilo-app-release.yml
+++ b/.github/workflows/kilo-app-release.yml
@@ -240,8 +240,23 @@ jobs:
     steps:
       - uses: useblacksmith/checkout@41cdeedae8edb2e684ba22896a5fd2a3cb85db6b # v1
 
+      # An open bump PR already carries the next version. A second one would
+      # bump to the same number and conflict with it.
+      - name: Look for an open bump PR
+        id: pending
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          OPEN=$(gh pr list --state open --limit 100 --json headRefName \
+            --jq '[.[] | select(.headRefName | startswith("kilo-app-version-bump-"))] | length')
+          if [ "$OPEN" != "0" ]; then
+            echo "::warning::A kilo-app version bump PR is already open — not opening another. Merge it to release the next version."
+          fi
+          echo "open=$OPEN" >> "$GITHUB_OUTPUT"
+
       - name: Bump patch version
         id: bump
+        if: steps.pending.outputs.open == '0'
         working-directory: apps/mobile
         run: |
           OLD=$(sed -n "s/^  version: '\([0-9][0-9.]*\)',$/\1/p" app.config.ts)
@@ -256,6 +271,7 @@ jobs:
           echo "new=$NEW" >> "$GITHUB_OUTPUT"
 
       - name: Push bump branch
+        if: steps.pending.outputs.open == '0'
         run: |
           NEW="${{ steps.bump.outputs.new }}"
           git config user.name "github-actions[bot]"
@@ -267,6 +283,7 @@ jobs:
           git push -fu origin "kilo-app-version-bump-$NEW"
 
       - name: Open bump PR
+        if: steps.pending.outputs.open == '0'
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
--- a/.github/workflows/kilo-app-release.yml
+++ b/.github/workflows/kilo-app-release.yml
@@ -31,6 +31,9 @@ jobs:
   check-changes:
     runs-on: ${{ vars.RUNNER_DEFAULT_LABEL || 'ubuntu-latest' }}
     timeout-minutes: 5
+    permissions:
+      contents: read # read the release tags
+      pull-requests: read # look for an open bump PR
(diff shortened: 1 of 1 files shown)


.github/workflows/kilo-app-release.yml
```diff
@@ -213,3 +232,92 @@ jobs:
...
+        id: pending
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          OPEN=$(gh pr list --state open --limit 100 --json headRefName \
+            --jq '[.[] | select(.headRefName | startswith("kilo-app-version-bump-"))] | length')
+          if [ "$OPEN" != "0" ]; then
+            echo "::warning::A kilo-app version bump PR is already open — not opening another. Merge it to release the next version."
+          fi
+          echo "open=$OPEN" >> "$GITHUB_OUTPUT"
```
pandemicsyn: bot: An existing bump PR is detected only after the store submissions have completed, so it does not prevent reuse of the current marketing version.

Suggested fix: Move the pending-bump check into `check-changes` (or a job required by `build-and-submit`) and set `should_build=false`/fail before the build when an open `kilo-app-version-bump-*` PR exists. As written, an eligible main push while that PR is open still runs `build-and-submit` against main, whose `app.config.ts` retains the version that was just released; this job merely skips creating another bump PR afterward. Add coverage for a pending bump plus a subsequent eligible code merge.

.github/workflows/kilo-app-release.yml
```diff
@@ -213,3 +232,92 @@ jobs:
...
+             do not trigger workflow runs. Close and reopen this PR (or push an empty
+             commit) to start CI.
+          2. Review and merge. Only the version line changed; no product code.
+          EOF
+          gh pr create --base main \
+            --title "chore(kilo-app): bump version to $NEW" \
+            --body-file /tmp/bump-pr-body.md \
+            --assignee iscekic \
+            --reviewer iscekic \
+            || echo "A bump PR for $NEW is already open"
```
pandemicsyn: bot: The fallback masks every bump-PR creation failure and lets the release succeed without recording the next version.

Suggested fix: Remove `|| echo ...` so `gh pr create` failures fail the job. If duplicate creation must be tolerated for a race, re-query for the expected open bump PR and only treat that verified state as success; propagate permission, API, reviewer/assignee, and other failures. Otherwise a failed PR creation leaves main at the released version and a later release can attempt to submit it again.

pandemicsyn (approved):

.github/workflows/kilo-app-release.yml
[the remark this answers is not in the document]
iscekic: Valid — fixed in 370261503. Dropped the `|| echo`. The pending-bump gate now runs in `check-changes` before the build, so the only case the fallback covered is already handled upstream; any real `gh pr create` failure now fails the job.

iscekic: Valid — fixed in 370261503. Removed `|| echo` so permission, API, and reviewer failures fail the job. No re-query added: the pending-bump gate moved into `check-changes`, so a duplicate creation is no longer a state this job can reach.

.github/workflows/kilo-app-release.yml
iscekic: Valid — fixed in 370261503. The pending-bump check moved into `check-changes` and sets `should_build=false`, so an eligible merge landing while a bump PR is open now skips the build instead of submitting the released version again. It sits ahead of the `workflow_dispatch` override, because a manual run hits the same store rejection. The duplicate check in `bump-version` is deleted; the concurrency group serialises runs, so one gate is enough.

No coverage added: this repo has no test harness for workflow logic, and adding one for this is out of scope for the PR.

iscekic: Reverting my previous fix — the blocking gate is out again as of 479dfb97e. Intended design: the job always tries to submit, and `bump-version` runs in parallel with `build-and-submit`, opening a bump PR when none is open.

So the failure mode you describe stands, deliberately: with a bump PR unmerged, the next eligible merge submits the already-released version and the store rejects it. That is a red job with a clear error, which we prefer over silently skipping a release.

PR #5880, opened 2026-09-04
fix(mobile): translate every catalog against its UI context

iscekic:
## Why

The catalogs were machine-translated key by key with no idea where each string
appears on screen. That produced grammar errors, English title case copied into
languages that do not use it, labels translated as sentences, and English word
order forced around placeholders. `de` shipped `"Teilen {{title}}"` and
`"Kommentar"` on a button that means *comment*, the verb.

Nothing in the repo told a translator, or an agent, where a key renders.
`apps/mobile/AGENTS.md` had no translation guidance at all.

## What changed

**Every key in all 86 non-English catalogs was reviewed against its call site.**
For each key a reviewer saw the surrounding source code, the string's role in the
UI (one of 24: `button`, `alert-action`, `screen-title`, `badge`, `placeholder`,
`empty-body`, …), the English form and capitalization, and the current
translation. Mean 1291 of 2736 strings changed per locale.

Examples from `de`:

| key | before | after | why |
| --- | --- | --- | --- |
| `prReview.composer.comment` | `Kommentar` | `Kommentieren` | the role is `button`, so the action form |
| `prReview.hunkRows.fileLoadedOfTotalCount` | `{{displayCount}} Dateien von {{total}} geladen` | `{{displayCount}} von {{total}} Dateien geladen` | English word order forced onto German |
| `prReview.connectionExpired` | `GitHub-Verbindung abgelaufen.` | `Die GitHub-Verbindung ist abgelaufen.` | the role is `error-body`, needs a full sentence |
| `codeReviewer.title` | `Code Reviewer` | `Code-Prüfer` | never translated |
| `preferences.autoExpandThinking` | `Denken automatisch erweitern` | `Denkprozess automatisch aufklappen` | literal rendering no speaker would use |

**One key per piece of copy.** 206 groups of keys held identical English, which
cannot be translated apart without drifting. Those collapsed onto one key each:
416 keys dropped, 108 new `common.*` keys, 646 call sites repointed. `en.json`
went from 2736 keys to 2428, and stands at 2442 after the merge with main.

Twelve English strings genuinely carry two senses and keep a key each — `Open`
the verb versus the state of a pull request, `Comment` the verb versus the noun,
`High` the severity versus the thinking effort. Each is recorded with its reason.

**Case follows the English lead, where the language has case.** Case carries
meaning a reviewer cannot see one string at a time: a badge is drawn in capitals,
and a dismissal reason is a lowercase fragment that reads inside a sentence
another key starts. Applied mechanically across all 86 catalogs — full capitals
stay capitals, and the first letter follows English. The first-letter rule
touches one character, so it never imposes English title case. A string a
placeholder opens is left alone, because the interpolated value sets the case
there.

Two languages do not take the rule, and the first pass got both wrong:

- Georgian Mkhedruli has no letter case. Uppercasing the initial writes a
  Mtavruli letter, which the language uses only for a full-capitals effect, so
  1980 values had gained a capital the orthography does not use. Georgian is now
  exempt from every case rule, which is what `label-reference.test.ts` already
  assumed.
- Greek keeps the tonos on a capital initial and drops it only in full capitals,
  so `Άνοιγμα` had become `Ανοιγμα` in 137 values. A recase that changes which
  letter it is, and not only its case, is now skipped.

`apps/mobile/AGENTS.md` records both, so the next pass does not repeat them.

## Guardrails, so this does not come back

- `check-catalogs.mjs` now **fails on duplicate copy**, naming both keys and the
  escape hatch. Verified against a deliberately introduced duplicate. It already
  runs in CI via `pnpm check:i18n`.
- `apps/mobile/AGENTS.md` gains a `## Translations` section: edit `en.json` only,
  search for the string before adding a key, put shared copy under `common.`,
  name a label with `$t(key)`, never assume English word order.

Context fit, capitalization and placeholder order need judgement and are **not**
mechanically checkable — a "don't mirror English title case" rule fires on every
German noun. Those are covered by the guide, not a check.

## Three latent bugs this surfaced

`KiloClaw`, `Bitbucket` and `Kilo` are product names that should never be
translated, but a few locales' machine translation had mangled them, so the
"identical to English in every locale" rule never fired. Now that every locale
leaves them as written, it does. Each is allowlisted alongside the other proper
nouns, in its own commit.

## Merged with main

Main added 15 keys under `prReview.overview` for the pull request sidebar
metadata and changed no existing value, so each of the 31 conflicted catalogs
keeps this branch's reviewed copy and takes main's value only for those 15 keys.
Merging the other direction would have resurrected every key the deduplication
removed.

Two follow-ups came out of it:

- `prReview.overview.reviewDismissed` holds the same copy as
  `securityAgent.deadline.dismissed`, and both are a state label with one sense.
  The new duplicate-copy check caught it. Both call sites now read
  `common.dismissed`.
- Those 15 keys arrived after the review pass, so their values in every catalog
  were still the old machine translation. All 86 locales reviewed them against
  the overview screen, like every other key: 549 strings changed.

## Review notes

- 101 commits: one per locale, then the dedup batches, the guardrails, the merge
  with main, and the top-up review. Reviewing per commit is far easier than the
  squashed diff.
- Green: 8937/8937 tests, `check:i18n`, `format:check`, root `lint-all.sh`,
  root `typecheck-all.sh`.
- Lao needed its groups split in half; its catalog is the largest and one
  response kept timing out upstream.


tools/i18n/check-catalogs.mjs
```diff
@@ -382,6 +412,32 @@ for (const catalog of CATALOGS) {
...
   const sharedWordings = [...wordingGroups].filter(([, keys]) => keys.length > 1);
+
+  // One key per piece of copy. Two keys holding the same English string cannot
+  // be translated apart without the catalogs drifting, so the copy belongs to
+  // one key and every call site points at it. A plural family shares copy by
+  // design, so its members are exempt.
+  if (catalog.name === 'mobile') {
+    const byCopy = new Map();
+    for (const [key, value] of english) {
+      if (pluralParts(key)) {
```
pandemicsyn, line 429: bot: The duplicate-copy guardrail skips all plural-suffixed keys, not just duplicates within one plural family.

Suggested fix: Build duplicate groups for all English keys, while exempting only duplicate entries that belong to the same plural stem. Continue comparing plural leaves from distinct families and plural leaves against non-plural keys (or explicitly document and test a narrower intended exclusion). Add regression cases for both cross-family and plural/non-plural duplicates.

pandemicsyn (approved):

tools/i18n/check-catalogs.mjs
iscekic: The observation is correct: the skip covers every plural key, not just one family's members. I fixed the comment, which claimed the narrower scope, in f83e6fc9a. I did not change the logic, for three reasons.

**No instance exists.** I measured the current `en.json`: 0 cross-family duplicate groups and 0 plural-vs-non-plural groups, across 22 plural families.

**Drift is already caught.** The one-wording rule below builds its groups over all English keys, plurals included, and runs per locale. Two plural families holding the same copy still cannot be translated apart in any of the 86 catalogs. What escapes is the duplicate key, not divergent copy.

**Comparing plural leaves against non-plural keys is unsound here.** Every one of the 22 families carries its `_other` copy on the bare stem key by i18next convention — `prReview.overview.file` is `"files"` beside `file_other` `"files"`. Treating those as two keys for one string would fire 21 false failures. Only a stem-aware comparison works, and a non-plural call site cannot repoint at a plural leaf anyway, since i18next resolves `x_one` only through a `count`.

The sound part of the suggestion is narrow: group plural leaves by stem and fail on two distinct stems. Happy to add it if you want the guard now rather than when the first case appears.

PR #5881, opened 2026-09-04
fix(mobile): one iOS Live Activity, an idle timeout, and a tap target

iscekic:
## Problem

The Lock Screen showed three stacked Kilo Live Activity cards at once, each with different counts, one of them in a stale locale.

Three causes:

- `refreshActivity()` adopted one native activity and left every other one on screen. A push-to-start card and an app-started card could both live.
- An idle-only snapshot kept a card alive forever. Nothing retired it.
- `expo-widgets` pins `staleDate: nil`, so a frozen card never dimmed.

Tapping a card, a widget, or an Android ongoing notification did not open the Agents tab.

## Changes

**One card, always.** `refreshActivity()` now ends every native activity it did not adopt. It is the single choke point that `publish`, `startOrUpdate` and `endImmediate` all pass through, so no path can leave a second card behind.

**Idle lifecycle.** A card starts only when `running + needsInput > 0`. It stays for idle work, then ActivityKit dismisses it after 10 minutes (`GLANCEABLE_IDLE_ONLY_MS`). The 8 s terminal dismissal for empty work is unchanged. A JS timer cannot do this — the app is asleep exactly when the window expires — so the dismissal date is handed to `end(after:)`. If work resumes inside the window, the ended card is dismissed before a fresh one starts.

**Server matches the client.** `apnsSendsForTokens` now takes `startable` as well as `eligible`, so an APNs push-to-start cannot raise a card the client would have refused.

**Stale date.** The `expo-widgets` patch adds `staleAfterSeconds` to `start` and `update`, so a card that stops updating dims after 30 minutes instead of showing stale counts as current.

**Deep link.** `OPEN_AGENTS_URL = kiloapp:///cloud/sessions` is persisted at layout registration, not only at `start`, so a push-started card carries a tap destination too. `/cloud/sessions` already maps to the Agents tab in `UNIVERSAL_LINK_ROUTES`.

**Lock Screen tap target.** Device testing found a second bug. Upstream `expo-widgets` attaches `.widgetURL` to the Dynamic Island only, never to the Lock Screen banner, so a banner tap reopened the app on whatever screen it left. The patch adds `.widgetURL` to `LiveActivityBannerView`. The Lock Screen banner is where people actually tap, so requirement 3 was not met without this.

Deleted `openGlanceableAgents()` — nothing called it.

## Tests

`apps/mobile/src/glanceable-ios/ios-sink.native.test.ts` gains three describes: the single-card invariant against a push-to-start orphan, the idle dismissal date, and the resume-inside-the-window replacement. `services/notifications/src/lib/glanceable-delivery.test.ts` covers the idle-only start refusal.

## Lockfile

A cold `pnpm install` in a fresh worktree rewrote `pnpm-lock.yaml` every time. It added the `@swc/helpers@0.5.23` peer key to the `@swc/core@1.15.18` resolutions and dropped an orphan `@swc/helpers@0.5.15` entry. The committed file was not what a cold install produces. This branch commits the converged output; two consecutive cold installs now produce an identical file.

## Verification

Checks:

- 199 mobile glanceable tests and 75 notifications glanceable-delivery tests pass.
- `pnpm typecheck` and `scripts/lint-all.sh` are clean.
- Two cold `pnpm install` runs produce identical lockfiles.

On device, iOS 26.5 simulator, against the local stack with a CLI heartbeat driving the session counts. The final run used the `mobile-native-build.yml` CI artifact for commit `2d0aa5fb3`:

| Behaviour | Result |
| --- | --- |
| One card for one `busy` session | Pass — `0 Needs input / 1 Working / 0 Idle`, a single card |
| One card survives an app reinstall | Pass — the old activity goes `dismissed`, the new one `active` |
| `staleDate` reaches ActivityKit | Pass — "Marking activities stale" logged at +1799.99 s |
| Idle-only card dismissed after 10 min | Pass — "Dismissing activity platters" logged at +599.98 s |
| Work resumes inside the idle window | Pass — the replacement card starts, still one card |
| `kiloapp:///cloud/sessions` deep link | Pass — opens the Agents tab |
| Tapping the Lock Screen card | Pass — opens the Agents tab (failed before `2d0aa5fb3`) |

Push-to-start is **not** device-proven, on either side. What the simulator settled:

- The simulator does vend a push-to-start token, and the app registers it: an `ios_push_to_start` row appears in `user_activity_tokens` after login.
- `xcrun simctl push` cannot raise a Live Activity. It has no way to set `apns-push-type: liveactivity`, so the payload is accepted and no card appears. The app's handling of a push-started card is not reachable this way.
- The worker cannot reach APNs locally either. `readApnsCredentials` returns null without the `.p8`, so `sendIosLiveActivity` returns before any request, and `PUSH_SINK_MODE` does not cover this path — it only intercepts the Expo send in `NotificationChannelDO`. `APNS_BASE_URL` is also pinned to production, while a simulator token is a sandbox token.
- So the worker's decision to send rests on `glanceable-delivery.test.ts`, and the single-card invariant against a push-to-start orphan on `ios-sink.native.test.ts:304`, both unit tests. The device evidence above comes from an app reinstall, a different orphan path.


iscekic:
Both warnings were valid. Reproduced each with a failing test first, then fixed in `afade3d3`.

**`ios-sink.ts:318` — the idle dismissal was never shortened.** Confirmed. Once every agent goes idle, `publish` submits `end(after: +10 min)` and `finishEnd` deletes the map entry as soon as the native call returns. When work then went empty, `endNow(GLANCEABLE_TERMINAL_MS)` found nothing: `refreshActivity` reads `getInstances()`, which excludes an ended card, and the intent-rewrite loop only ran for an immediate dismissal. The card kept the idle counts on the Lock Screen for the rest of the ten minutes.

Fix: `endNow` takes `reachEnded`, which widens the scan to `getInstances(true)`; the terminal caller in `publish` sets it. The intent loop now also shrinks an already-submitted `dismissMs` when the new one is sooner. `finishEnd` already re-reads `ending.intent` at the native boundary, so a shortened intent re-submits on its own. Only a terminal caller may set `reachEnded` — an idle end must never touch a card whose dismissal is already sooner.

**`ios-sink.ts:348` — the replacement started before the dismissal landed.** Confirmed. `endNow(null, null)` submits an asynchronous native end and `start()` ran on the very next line, so both cards were on screen until the dismissal resolved.

Fix: `endNow` returns its pending work, and the replacement waits on it. Two follow-ons that the tests forced out:

- A second `startOrUpdate` arriving in that gap would have started its own card. The waiting start now takes the newer snapshot instead, so it opens on current counts rather than raising a stale card and updating after.
- `scheduleEnd` returned early when a task was already in flight, which handed the caller a promise that resolved before the end. It now awaits the in-flight task.

New tests: `shrinks a submitted idle dismissal to the terminal one once work goes empty` and `holds the replacement card until the idle card is dismissed`. Both fail on the parent commit.

Checks: 8961 mobile tests, `pnpm typecheck`, and `scripts/lint-all.sh` are clean. `origin/main` is merged in.

pandemicsyn (approved):

PR #5897, opened 2026-09-05
fix(cloud-agent): dedupe repositories across GitHub installations

### LINKED ISSUE #5887: Cloud Agent repository picker lists repositories multiple times when an org has several GitHub installations
## Summary

When starting a new Cloud Agent thread in an organization that has more than one connected GitHub installation, the repository picker lists the same repository multiple times — once per installation that can see it.

## Steps to reproduce

1. As an organization with multiple GitHub accounts/orgs connected to the same Kilo organization (each its own GitHub App installation), open Cloud Agent → new thread.
2. Open the repository picker.
3. Observe a repository that is reachable from more than one installation (for example `acme/api` is owned by one connected account and accessible as a collaborator repo on another, or the App was reinstalled leaving two active installation rows for the same account).

## Expected behavior

Each repository appears exactly once in the picker, resolved to the installation that will be used for the session.

## Actual behavior

The same repository appears once per GitHub installation that lists it, so the list shows duplicate entries.

## Cause

`fetchAllGitHubRepositoriesForOrganization` (apps/web/src/lib/cloud-agent/github-integration-helpers.ts) unions the repository lists of every healthy GitHub integration for the organization:

```ts
repositories: results.flatMap(result => result.repositories),
```

No deduplication is performed across installations, so a repo visible through two installations is returned twice. `NewSessionPanel` then renders every entry (keyed by `platformIntegrationId`, not by repo name), producing the duplicate rows.

## Notes

- Applies to the org Cloud Agent new-thread picker and to the org Code Reviews repository list, which share `fetchAllGitHubRepositoriesForOrganization`.
- Introduced together with the "repositories from every connected GitHub org" work (e.g. #5488, #5567, #5591, #5604).

maphew:
## Summary

Fixes #5887 — when an organization has more than one connected GitHub installation, the Cloud Agent new-thread repository picker (and the Code Reviews repository list, which shares the same helper) lists the same repository once per installation.

`fetchRepositoriesForIntegrations` (used by `fetchAllGitHubRepositoriesForOrganization`) unions the repo lists of every healthy GitHub integration with `flatMap` and no dedupe. A repo reachable through two installations — e.g. the app is installed on two accounts that share a repo, or a reinstall left two active installation rows for the same account — is therefore returned multiple times. The web picker renders each row (keyed by `platformIntegrationId`), so the duplicates show up as repeated entries when starting a new thread.

The fix deduplicates by repo `fullName` (case-insensitive) at the aggregation boundary:
- Prefer the installation whose GitHub account owns the repo, so the session runs with the canonical write identity.
- Otherwise keep the first entry (integrations arrive newest-first from `getIntegrationsByOrganization`, `created_at desc`).

Both consumers of `fetchAllGitHubRepositoriesForOrganization` (org Cloud Agent and org Code Reviews) are fixed by this single change.

## Verification

- [ ] Manual: not run locally — sandbox could not complete `pnpm install` (V8 OOM on lockfile resolution) to execute Jest; unit tests below cover the regression.
- [x] Added Jest coverage in `github-integration-helpers.test.ts`:
  - repo reachable through two installations collapses to the owning account's entry
  - one account with two active installations yields one entry per repo
  - existing per-installation provenance test still passes (distinct repos across accounts unaffected)

## Visual Changes

N/A (server-side repo list change; no UI code touched)

## Reviewer Notes

- Dedupe preference is intentionally "owning account first, else newest installation" so a session is prepared against the installation that can always authorize the repo.
- `getIntegrationsByOrganization` returns integrations ordered by `created_at desc`, which makes the first-entry fallback deterministic.


maphew:
Rebased on main and addressed the review warning. Heads-up: main has since merged #5922 for the same issue, which dedupes by keeping the first (oldest) installation because that is the primary installation a session resolves to.

This PR now builds on that behavior instead of contradicting it: the owning account is still preferred when a repo is reachable through several installations, and the fallback keeps the first entry — the comment now correctly says `getIntegrationsByOrganization` returns oldest-first (it is `created_at ASC`), matching the primary-installation rule. The same-account dedupe test now feeds installations in DB order and asserts the primary entry is retained. Merge conflict with the #5922 change is resolved.

Verification: targeted web unit suite passes (18 tests).

eshurakov (approved):

maphew:
Closing as superseded: #5978 changed the contract so repositories are intentionally listed once per granting installation (the picker keys rows by installation + repo, since each installation resolves to different tokens at session start). The cross-installation collapse this PR implemented directly contradicts that, so no rebase can save it. #5887 stays open and needs a fresh fix against the new per-installation listing behavior — likely in the picker UI rather than the aggregation helper.

PR #5898, opened 2026-09-05
fix(cloud-agent-next): retry attachment downloads and hide raw stream errors

### LINKED ISSUE #5883: Cloud Agent: pasted images fail with 'attachment <uuid>.png could not be retrieved (undefined is not a function)'
## Summary

Pasting an image into a Kilo Cloud Agent thread (kilo.ai web app) shows a preview in the composer, but after submitting, the message instead contains:

```
attachment a7393f97-c6bf-4c1f-9fec-2bbb11987962.png could not be retrieved (undefined is not a function)
```

The agent then "carries on" without the image. Reproduced with different UUIDs (e.g. `4024b67d-4294-4493-8884-ce0f9f40b836.png`) in separate sessions — the image is silently dropped and the model never sees it.

## Diagnosis

The string is generated by `services/cloud-agent-next/wrapper/src/session-bootstrap.ts` (`downloadAndMaterializeAttachment`): when the wrapper in the sandbox fetches the presigned R2 attachment URL, any fetch/read error is interpolated verbatim into a synthetic text part:

```ts
text: `attachment ${attachment.filename} could not be retrieved (${message})`
```

So `(undefined is not a function)` is a raw JS `TypeError` message leaking into the user-visible transcript. It matches a known class of Bun Web Streams bugs where reading a `fetch()` response body whose bytes arrive after the response head throws exactly `TypeError: undefined is not a function` (see oven-sh/bun#34302, #28952, #28953). I reproduced the exact message 100/100 times on Bun 1.3.12 by iterating a delayed-body `Response.body`; the failure is intermittent in production and depends on network latency + runtime version.

## Impact

- Pasted images are lost without the user being told anything actionable.
- The raw internal error message (`undefined is not a function`) is meaningless to users and agents.
- No retry exists, so a transient network/stream race permanently drops the attachment.

## Proposed fix

PR (once open): https://github.com/Kilo-Org/cloud/pull/XXXX
- Retry transient attachment-download failures (3 attempts, short backoff).
- On retry, re-fetch and read the body with `response.arrayBuffer()` (the workaround the Bun maintainers verified never hits the stream race) when `content-length` is bounded.
- On exhaustion, surface `attachment <name> could not be retrieved (download failed after N attempts)` and log the raw reason instead of leaking raw JS errors.

maphew:
## Problem

Pasting an image into a Kilo Cloud Agent thread can silently fail: the wrapper's attachment materialization (`services/cloud-agent-next/wrapper/src/session-bootstrap.ts`) catches any error while fetching the presigned R2 attachment URL inside the sandbox and interpolates the raw error into a synthetic text part:

```
attachment a7393f97-c6bf-4c1f-9fec-2bbb11987962.png could not be retrieved (undefined is not a function)
```

The `(undefined is not a function)` text is a raw JS `TypeError` leaking into the user-visible transcript. It matches a known class of Bun Web Streams bugs where reading a `fetch()` `Response.body` whose bytes arrive after the response head throws exactly `TypeError: undefined is not a function` (oven-sh/bun#34302, oven-sh/bun#28952, oven-sh/bun#28953). The failure is intermittent (network-latency dependent), there is **no retry**, and a transient failure permanently drops the image from the prompt. The agent then "carries on" without the image.

Repro evidence: iterating a delayed-body `Response.body` on Bun 1.3.12 fails 100/100 with the identical message; the failure disappears when the body is read via `getReader().read()` or `arrayBuffer()`.

## Fix

- **Retry transient attachment-download failures** — up to `MAX_ATTACHMENT_DOWNLOAD_ATTEMPTS` (3) attempts with short backoff (250ms, then 750ms). Non-2xx `429`/`5xx` responses and fetch/read/timeout errors are retried; `4xx` (except 429) and size-cap overflows are permanent.
- **Use a buffered read on retry** — retries re-fetch and read the body with `response.arrayBuffer()` (the workaround Bun's maintainers verified never hits the stream race) when `content-length` is present and within the 20 MiB cap; otherwise the existing bounded streaming read is used.
- **Stop leaking raw JS error text** — once retries are exhausted the part reads `attachment <name> could not be retrieved (download failed after 3 attempts)` and the underlying reason goes to the wrapper log instead of the transcript. Overflow keeps its typed `Attachment too large: ...` message, and non-retryable HTTP statuses still report `HTTP <status>`.

Each attachment still degrades independently: one failed attachment never blocks the rest of the prompt.

## Tests

- Existing overflow / HTTP-403 / generic-binary / exact-cap cases unchanged (all single-attempt, non-retryable paths).
- Updated the two always-failing transient-error tests to assert the new exhausted-retry message and that 3 attempts were made.
- Added: "retries a transient fetch failure and materializes the attachment on the retry" and "falls back to a buffered read when the streaming read fails mid-transfer".

Fixes https://github.com/Kilo-Org/cloud/issues/5883


maphew:
Addressed both warnings from the automated review:

- Buffered retries no longer throw `unbounded body` when `content-length` is missing/invalid/oversized — `downloadBuffered` now falls back to the bounded streaming read (`downloadBounded`), which enforces the size cap and stays memory-safe. Regression test added.
- `materializeAttachment` now calls `externalSignal?.throwIfAborted()` before the retry backoff, so a caller abort is propagated instead of being treated as a retryable failure (and is never masked as an exhausted-attachment text part on the final attempt). Test added.

Verification: wrapper `bun test src/session-bootstrap.test.ts` passes (87 tests) and `pnpm typecheck` in the wrapper is clean.

eshurakov:
hey @maphew, could you take a look at new kilo bot comments https://github.com/Kilo-Org/cloud/pull/5898#issuecomment-5695928180

thank you!

maphew (commented):
Addressed the review in e594d37 (bun test 91 pass, wrapper tsgo clean): buffered path now documents the re-fetch rationale and the header-as-hint posture with a post-read enforcement check; per-attachment latency is bounded by a 130s overall deadline so retries cannot triple the worst case; added 5xx-then-success and 400-single-attempt tests; fs/empty-body errors propagate to the abort-aware retry loop instead of being labeled retryable up front, and the backoff table is now an ordered array. On the root-cause question:  does use , and the bot is right that the buffered retry only dodges reader-path failures rather than proving the incident stack — the retry still converts any transient failure into a recovered attachment, and every failure now lands in the wrapper log with the attempt history instead of only the sanitized transcript line.

maphew:
Addressed the review in e594d37 (bun test 91 pass, wrapper tsgo clean): buffered path now documents the re-fetch rationale and the header-as-hint posture with a post-read enforcement check; per-attachment latency is bounded by a 130s overall deadline so retries cannot triple the worst case; added 5xx-then-success and 400-single-attempt tests; fs/empty-body errors propagate to the abort-aware retry loop instead of being labeled retryable up front, and the backoff table is now an ordered array. On the root-cause question: the base read path does use getReader, and the buffered retry only dodges reader-path failures rather than proving the incident stack — the retry still converts any transient failure into a recovered attachment, and every failure now lands in the wrapper log with the attempt history.

eshurakov:
@maphew looks good to me, but some checks are failing now. You may need to merge/rebase main as well

maphew:
Fixed the failing lint check: `await expect(result).rejects.toThrow()` in the wrapper's Bun test was the culprit — bun:test's `expect().rejects.toThrow()` isn't typed as thenable, so the type-aware `typescript-eslint(await-thenable)` rule rejected it. Replaced it with a try/catch rejection capture matching the convention already used in that file (c9eeed3fb), and merged current main (e1e6d7965). Verified locally: `scripts/lint-all.sh` clean on the changed files, wrapper suite green (1578 pass), wrapper typecheck clean. The new CI run (c9eeed3f) is pending manual workflow approval (fork gate) — could someone with write access hit 'Approve and run' on the Actions tab.

eshurakov (approved):

PR #5904, opened 2026-09-05
fix(mobile): keep the glanceable surfaces current and one Live Activity card

iscekic:
## Problem

Three reports, one root cause between them:

- The widget, the Live Activity, and the notification showed old values until the app was opened.
- The iOS Lock Screen collected several Live Activity cards.
- A Live Activity sometimes vibrated and expanded on its own.

A push-to-start raises a Live Activity while no JavaScript runs. Nothing read the
native instance, so the server never held an update token for that card. Every
later refresh sent another push-to-start, and each one raised a new card. APNs
requires an `alert` on a push-to-start, and that alert is what lit the screen and
expanded the card. So the stacking and the unprompted expansion are the same bug.

## Changes

**One card, no unprompted expansion**

- The user Durable Object fences a push-to-start token for the maximum life of the
  card it raises. A refresh that finds no update token sends no second start.
- The fence is released as soon as an activity token proves the app adopted the card.
- The iOS sink adopts a card it did not start, at import, inside the background run
  time ActivityKit grants after a push-to-start, and hands the update token to the
  server. The same native read retires any duplicate already on screen.
- An APNs `410` now retires the token it names on every event, not only on an `end`.
  A dead activity row used to absorb every update *and* block the push-to-start that
  would have replaced the card those updates could not reach.

**Honest values instead of stale ones**

- Start and update pushes carry `stale-date`, and the widget timeline carries a
  delayed frame at `updatedAt + 30 min`.
- iOS cannot deliver a background wake to a force-quit app, so a surface can go
  unrefreshed. It now reads as delayed rather than asserting counts the device
  cannot confirm.

**Shared status glyphs**

- `glanceableStatusKind()` in `@kilocode/app-shared` is now the single status map.
  The widget counts and the session list rows both read it, so a row can never
  disagree with the widget beside it.
- The live list and the history list draw the same three glyphs the widget draws:
  needs input, working, idle.

## Verification

- `services/notifications`: 310 tests pass, including four new ones covering the
  fence, the fence release, the `410`-on-update retirement, and the stale date.
  Three of the four fail when the fix is reverted.
- `apps/mobile`: 8968 tests pass. `packages/app-shared`: 370 pass.
- Root `lint-all.sh`, `tsgo` typecheck, `check:unused`, `check:i18n`: clean.

## Known remaining cause of a card replacement

PR #5881 retires a card after a 10 minute idle window. Work that resumes inside
that window dismisses the ended card and starts a replacement, which pops. That
is a trade-off from #5881 and is not changed here.

### DIFF
--- a/apps/mobile/src/components/agents/remote-session-row.tsx
+++ b/apps/mobile/src/components/agents/remote-session-row.tsx
@@ -3,6 +3,8 @@ import { useQueryClient } from '@tanstack/react-query';
 import * as Haptics from 'expo-haptics';
 import { useEffect, useMemo, useRef, useState } from 'react';
 
+import { glanceableStatusKind } from '@kilocode/app-shared/glanceable-agents-snapshot';
+
 import { buildActiveSessionsTrayInput } from '@/lib/active-sessions-live';
 import { currentAuthEpoch, isCurrentAuthEpoch } from '@/lib/auth/auth-epoch';
 import { isSignOutActive } from '@/lib/auth/sign-out-state';
@@ -233,6 +235,7 @@ export function RemoteSessionRow({
             remoteMeta(session)
           )}
           live
+          statusKind={glanceableStatusKind(session.status)}
           needsInput={needsInput}
           metaWhileLive
           platformIcon={platformIcon}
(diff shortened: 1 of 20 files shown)


iscekic:
Reviewed the WARNING on \`apps/mobile/src/glanceable-ios/register.ts:31\`. Valid, and fixed in 3844241.

The root layout imports \`@/glanceable-ios/register\`, which runs \`adoptPushStartedActivity()\`. \`setGlanceableDelivery\` runs from \`@/lib/glanceable/delivery-registration\`, reached only through \`src/app/(app)/_layout.tsx\`, which is evaluated later. So the adoption handed its update token to \`noopDelivery\` and the card it claimed stayed orphaned — the exact failure this PR exists to remove.

Fix: \`adopt-activity.ts\` now imports \`@/lib/glanceable/delivery-registration\` for its side effect, so the delivery is installed before the adoption uses it. No cycle: the chain is register → adopt-activity → delivery-registration → active-agents-live-activity.

Checks after the fix: mobile 8968 tests, typecheck, lint, check:unused all pass.

eshurakov (approved):

marius-kilocode (approved):

PR #5923, opened 2026-09-07
fix(code-review): fail the GitHub check when the model produces no review output

iscekic:
> **kwf trial run.** This PR came from a workflow trial request, not from a tracked work item. Review it as you would any other PR; the label `kwf-trial` marks where it came from.

## Request

> Surface: backend services (services/) — Kilo-Org/cloud issue 5630.
> 
> Code Review reports the GitHub check as successful even when the AI model failed and produced no review.
> 
> Repro from the issue: a review run whose session log ends with "The model hit its output limit while reasoning and produced no actionable output" still shows Complete in Kilo and succeeded in GitHub, with the message "Code review completed successfully."
> 
> Expected: when the model fails, hits an output or token limit, or cannot produce a valid review, the GitHub check is not reported as successful. It is marked failed or incomplete, and the reason reaches the check output.
> 
> Rule: reproduce first. If it does not reproduce on main, answer no_change with the evidence.

## Changelog for users

- A code review whose model hit its output limit and produced no review content no longer reports success: the run settles as failed, the GitHub check is not marked successful, and the reason ("The model output limit was reached") reaches the check output instead of "Code review completed successfully."
- A code review that ends with a real review summary still completes successfully.
- Sessions that are not code reviews are unchanged: they still complete normally when the model reports an output-limit notice.

## Changelog for maintainers

- The failure trigger is a text match on the assistant reply: it fires only when both a "no actionable output" fragment and an "output/token limit" fragment are present, so small copy drift in either half still detects, while a review that merely discusses output limits cannot false-positive without the "no actionable output" half (`assistantReportsNoActionableOutput` in `wrapper-supervisor.ts`).
- Scoped to sessions created on the `code-review` platform; the flag is computed once per settlement loop and threaded into both settlement paths.
- Both settlement paths are covered: sealed-batch completion after a wrapper terminal event, and wrapper-death reconciliation (`projectWrapperDeathReconciliation` gained a `codeReviewSession` parameter; the wrapper-death gate on positive terminal evidence is unchanged).
- The notice is terminalized like a terminal assistant error: `assistant_error` with `assistantFailureReason: 'output_limit'`, `failureStage: 'agent_activity'`, and the shared safe message from `assistantFailureMessage('output_limit')`, so the callback payload projects the reason into the GitHub check output.
- Where to look first: the two call sites in `services/cloud-agent-next/src/session/wrapper-supervisor.ts` and the four new cases in `wrapper-supervisor.test.ts` (terminal-event failure, dead-wrapper reconciliation failure, real review still completes, chat session unaffected).
- Risk: the match is against the CLI's notice copy; if the CLI rewords the notice so neither fragment appears, an empty review would settle as completed again. A false positive would require a real review's text to contain both fragments, e.g. quoting the notice verbatim.

## E2E proof — log excerpts

```
[e1] Rule: reproduce first. If it does not reproduce on main, answer no_change w -> pass :: Reproduced on origin/main via WrapperSupervisor createHarness stub (e1-repro-main.log): `FAIL |unit| src/session/wrapper-supervisor.test.ts > WrapperSupervisor 
```

<details><summary><code>/home/igor_kilocode_ai/.local/share/kwf/sections/surface-backend-services-ser-90c9/e2e-backend/e1-repro-main.log</code></summary>

```
+ Received
  {
-   "assistantFailureReason": "output_limit",
    "assistantMessageId": "ase_dead_wrapper_output_limit",
    "completionSource": "idle_reconciliation",
-   "failureReason": "assistant_error",
-   "status": "failed",
+   "status": "completed",
  }
 ❯ src/session/wrapper-supervisor.test.ts:2605:69
    2603|     await harness.supervisor.runMaintenance(92_000);
    2604|
    2605|     await expect(getSessionMessageState(harness.storage, MESSAGE_ID)).…
       |                                                                     ^
    2606|       status: 'failed',
    2607|       failureReason: 'assistant_error',
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[2/2]⎯
 Test Files  1 failed | 182 passed (183)
      Tests  2 failed | 5421 passed | 3 skipped (5426)
   Start at  01:59:27
   Duration  26.09s (transform 73.79s, setup 0ms, import 263.55s, tests 38.91s, environment 104ms)
/home/igor_kilocode_ai/.local/share/kwf/wt/surface-backend-services-ser-90c9/services/cloud-agent-next:
[ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL] cloud-agent-next@1.0.0 test: `vitest run -- src/session/wrapper-supervisor.test.ts -t 'output.limit|real review summary'`
Exit status 1
```
</details>

### DIFF
--- a/services/cloud-agent-next/src/docker-privileged-proxy-script.test.ts
+++ b/services/cloud-agent-next/src/docker-privileged-proxy-script.test.ts
@@ -95,18 +95,29 @@ async function startProxy(): Promise<{ proxySocket: string; request: Promise<Buf
 }
 
 function sendCreateRequest(proxySocket: string, name: string, image: string): Promise<void> {
-  return new Promise((resolve, reject) => {
-    const body = JSON.stringify({ Image: image, HostConfig: { PidMode: 'host' } });
-    const socket = net.createConnection(proxySocket, () => {
-      socket.write(
-        `POST /v1.48/containers/create?name=${encodeURIComponent(name)} HTTP/1.1\r\n` +
-          `Content-Length: ${Buffer.byteLength(body)}\r\n\r\n${body}`
-      );
+  const body = JSON.stringify({ Image: image, HostConfig: { PidMode: 'host' } });
+  const attempt = (retry: number): Promise<void> =>
+    new Promise((resolve, reject) => {
+      const socket = net.createConnection(proxySocket, () => {
+        socket.write(
+          `POST /v1.48/containers/create?name=${encodeURIComponent(name)} HTTP/1.1\r\n` +
+            `Content-Length: ${Buffer.byteLength(body)}\r\n\r\n${body}`
+        );
+      });
+      socket.once('error', error => {
+        socket.destroy();
+        // The proxy creates the socket file before it starts accepting
+        // connections; under load connect() can land in that window.
+        if (retry > 0 && (error as NodeJS.ErrnoException).code === 'ECONNREFUSED') {
+          setTimeout(() => attempt(retry - 1).then(resolve, reject), 25).unref();
+          return;
+        }
+        reject(error);
+      });
+      socket.once('end', resolve);
+      socket.resume();
     });
-    socket.once('error', reject);
-    socket.once('end', resolve);
-    socket.resume();
-  });
+  return attempt(40);
 }
 
 describe('docker-privileged-proxy.mjs', () => {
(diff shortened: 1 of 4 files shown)


iscekic:

**Status:** 0 Issues

| File | Line | Issue |
| --- | --- | --- |

eshurakov (approved):

marius-kilocode (approved):

PR #5931, opened 2026-09-07
fix(mobile): surface Agents pull-to-refresh failures with Updating and inline Retry

iscekic:
> **kwf trial run.** This PR came from a workflow trial request, not from a tracked work item. Review it as you would any other PR; the label `kwf-trial` marks where it came from.

## Request

> Verify, and fix if real: on the Agents list, a pull-to-refresh while the API (nextjs) is down keeps the rows and shows the native spinner only. The expected unhappy state is an inline "Couldn't refresh." line with a Retry action, and an "Updating" indicator while the pull is in flight.
> Repro (from a verifier run on 2026-09-04 01:34, tree 306a, iOS simulator): sign in, open Agents, stop the nextjs service in the dev stack, pull to refresh, wait 30 s. Observed: rows stay, native spinner, no "Couldn't refresh.", no Retry.
> Rule: reproduce first. If it does not reproduce on main, answer no_change with the evidence.

## Changelog for users

- A pull-to-refresh that fails while the API is down now keeps the cached rows and swaps the native spinner for an inline "Couldn't refresh" line with a working Retry action, on both the live Agents tab and the session history list (previously the spinner spun with no feedback and no next action).
- While a pull or a retry is in flight, a visible "Updating" line shows above the list and is announced to screen readers; the line's height is reserved, so rows never shift when the state swaps.
- A pull whose request never settles stops its spinner after 15 seconds and hands over to the "Couldn't refresh" + Retry line, so a hung fetch can no longer pin the spinner.
- A rejection that lands fast holds the "Updating" feedback through a short beat before the failure line takes over, so the in-flight state is not a missable flash.
- Tapping Retry re-runs the fetch without the native spinner.
- A stale "Couldn't refresh" line retires when a later refresh outside the gesture (tab focus return, app foreground) lands an accepted result; persistent failures keep surfacing through the existing query error state.

## Changelog for maintainers

- Start at `use-pull-refresh.ts`: the shared pull/retry state machine both Agents surfaces now use — a 15 s feedback budget (`PULL_FEEDBACK_BUDGET_MS`) stops the spinner on a hung fetch, an 800 ms minimum beat (`PULL_FEEDBACK_MIN_BEAT_MS`) keeps fast rejections perceivable, token-guarded settlements supersede safely, and `markSettled` only retires standing failures without touching in-flight feedback.
- `SessionListRefreshStatus` is the reserved-height status line above the lists; its full accessibility label must stay exactly the `agents.sessionList.couldNotRefresh` copy — the device verifier asserts the full label, not a substring.
- History-screen contract: its void refetch maps to "accepted", so only the feedback budget can fail a history pull; the query-error state still owns persistent failures and auto-clears on recovery as before.
- Live-tab contract: the wrapped refetch resolves `false` on a rejected refresh; focus-return and app-foreground refreshes that land accepted results call `markSettled` to retire a stale failure line; a rejected context refresh reports through its own error state.
- One line speaks: when the status line owns the rows-state failure (in flight or failed), the duplicate failure label and announcement in `LiveSessionFeedback` are suppressed; a successful pull with cached rows now announces "Updating" exactly once.
- Pure move: `useLiveSessionContext`/`liveSessionContent` moved from the home section into new `home/live-session-state.ts` so the live screen imports them without circulars; home passes no refresh state and keeps its a11y-only announcement.
- Removed the inline error text under the search header and the screen-level `showInlineError` computation and prop plumbing — the reserved status line replaces both.
- New i18n key `agents.sessionList.couldNotRefresh` ("Couldn't refresh") added to every locale; behavior covered by three new mounted suites — the hook state machine, the live tab over the real hook chain including the socket-down outage variant, and the history screen.

## E2E proof

**Recording of the verified flow** (waits trimmed)

https://github.com/user-attachments/assets/54f9b0ec-9db1-4f25-81e4-36ae2d04d9ba

![[e1] Agents list pull-to-refresh while nextjs is down — e2e-mobile-app/e1-could-not-refresh.png](https://github.com/user-attachments/assets/3c0aa71a-3341-40ac-8e1b-201fd2e10127)

![[e1] Agents list pull-to-refresh while nextjs is down — e2e-mobile-app/e1-updating.png](https://github.com/user-attachments/assets/4e9bbe45-229f-4715-9136-1f03b2044a3e)

## Follow-ups (not changed here)
- The 1 LIVE label and the Agents tab badge disappear during the failed refresh while the live session row stays.
- not proved live: Android: not run — the diff forks on no platform, so iOS proves both


iscekic:

**Status:** 2 Issues

| File | Line | Issue |
| --- | --- | --- |
| `apps/mobile/src/components/agents/use-pull-refresh.ts` | 84 | A pull that supersedes an in-flight retry leaves busy=true permanently, pinning the status line on 'Updating' with no Retry and no spinner. |
| `apps/mobile/src/i18n/locales/af.json` | 1881 | A feature PR adds agents.sessionList.couldNotRefresh to all 86 non-en.json locale catalogs; per the mobile i18n rule those catalogs are translated only by the dedicated locales slice. |

iscekic:
bot: Accepted. Pushed 98e12f3 (kwf `kwf-fix-review-1c0d`).

What changed:
', write one bullet per finding that starts with its `file:line`: accepted or rejected, why, what changed. The reply to each review thread is the bullet naming its file.
> 
> # The pull request description as it stands now
> 
> This is the live body; you never need to fetch it, and you must not edit it. The driver merges what your final summary says into it.
> 
> ```
> > **kwf trial run.** This PR came from a workflow trial request, not from a tracked work item. Review it as you would any other PR; the label `kwf-trial` marks where it came from.
> 
> ## Request
> 
> > Surface: the mobile app (apps/mobile).
> > Verify, and fix if real: on the Agents list, a pull-to-refresh while the API (nextjs) is down keeps the rows and shows the native spinner only. The expected unhappy state is an inline "Couldn't refresh." line with a Retry action, and an "Updating" indicator while the pull is in fli

eshurakov (approved):

PR #5943, opened 2026-09-07
fix: remove legacy OpenRouter alias handling

chrarnoldus:
## Summary
- remove the legacy OpenRouter alias allowlist entirely
- validate model IDs against the Redis-backed OpenRouter catalog only
- preserve the existing fallback when no model metadata is available

## Testing
- not run (per request; CI will validate)


chrarnoldus:
Fixed in 7988fd3a5 by removing the obsolete test that asserted legacy aliases bypass Redis. The remaining tests cover Redis catalog acceptance, rejection, and the empty-catalog fallback.

johnnyeric (approved):

PR #5944, opened 2026-09-07
chore(web): remove unavailable OpenRouter models

chrarnoldus:
## Summary
- remove unavailable-model entries that are absent from OpenRouter's live model catalog
- retain the two denylisted Gemma models still exposed by OpenRouter

## Validation
- not run locally per request; CI will validate the change


chrarnoldus:
@kilo-code-bot Addressed in 76e61707f. Updated the stale unavailable-model assertions to cover the two remaining Gemma denylist entries, and changed the retired Tencent/LongCat expectations to confirm they no longer have a local availability restriction.

johnnyeric (approved):

PR #5953, opened 2026-09-08
Add repository_customizations table and routes

intentionally-left-nil:
To continue with the work of allowing per-repository settings (AI model, pr reviews etc) this commit creates a table to store the data. It also adds the routes needed to update the data, and wires it into the GithubRepositoryCustomizationsPreview components.

No callsites consume this data yet, except for the admin page.

## Summary

* Fixed updateIntegrationMetadataForOwner to use proper atomic sql operations
* Added repository_customizations table
* Changed GitHubRepositoryCustomizationsPreview.tsx to be a stateful component, owning its data and connecting to the backend
* Added the react code to the integrations view, under the `PER_REPO_SETTINGS` feature flag

## Verification

Visited the /integrations page, updated the values, saw that it was persisted. The orgs page was harder since I didn't have an integration, but it should still work the same

https://github.com/user-attachments/assets/1a909f10-994e-41d3-a324-8a461e6fd031


apps/web/src/lib/integrations/db/platform-integrations.ts
```diff
@@ -995,40 +999,86 @@ export async function updateIntegrationMetadata(
...
+    .update(platform_integrations)
+    .set({
+      metadata: sql`coalesce(nullif(${platform_integrations.metadata}, 'null'::jsonb), '{}'::jsonb) || ${JSON.stringify(metadataUpdates)}::jsonb`,
+      updated_at: new Date().toISOString(),
+    })
+    .where(
+      and(
+        ownershipCondition,
+        eq(platform_integrations.platform, platform),
+        integrationId ? eq(platform_integrations.id, integrationId) : undefined
```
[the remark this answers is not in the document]
intentionally-left-nil: That is good feedback, thanks bot :)

apps/web/src/routers/gitlab-router.ts
```diff
@@ -247,10 +247,15 @@ export const gitlabRouter = createTRPCRouter({
...
     if (Object.keys(configuredWebhooks).length === 0) {
-      await updateIntegrationMetadataForOwner(owner, 'gitlab', {
-        ...existingMetadata,
-        webhook_secret: newSecret,
-      });
+      await updateIntegrationMetadataForOwner(
+        owner,
+        'gitlab',
+        {
+          ...existingMetadata,
```
pandemicsyn: bot: Do not merge the stale full metadata snapshot into this atomic update.

Suggested fix: Pass only the keys this operation changes: use `{ webhook_secret: newSecret }` in the no-webhook branch, and `{ webhook_secret: newSecret, configured_webhooks: updatedWebhooks }` in the later branch (which has the same spread at line 333). The helper's SQL `metadata || metadataUpdates` can preserve a concurrent change only when `metadataUpdates` is a partial patch; spreading `existingMetadata`, read at line 241, makes its stale value win over concurrent writes to any unrelated key such as a model or review setting.

apps/web/src/components/integrations/GitHubRepositoryCustomizationsPreview.tsx
```diff
@@ -88,93 +92,148 @@ function repositoryCustomizationSummary(
...
+  >({});
+  const [savingRepositoryId, setSavingRepositoryId] = useState<string | null>(null);
   const [search, setSearch] = useState('');
   const [page, setPage] = useState(1);
   const [editingRepository, setEditingRepository] = useState<PreviewRepository | null>(null);
   const [announcement, setAnnouncement] = useState('');
   const editTrigger = useRef<HTMLButtonElement | null>(null);
   const searchInput = useRef<HTMLInputElement | null>(null);
 
+  if (isLoading || !data) {
```
pandemicsyn: bot: A failed customization query is shown as an endless loading state.

Suggested fix: Handle the query error separately from loading (for example destructure `isError`, `error`, and `refetch`, and render an error message with retry). `!data` is also true after React Query finishes retries with an error, so the current condition permanently renders “Loading repository customizations…”. Add a rejected-query component test. The top-level `listIntegrations` query likewise needs an error state: its error result currently renders no installation content.

pandemicsyn (approved):
lgtm, couple of minor things from the bot

PR #5974, opened 2026-09-08
feat(auth): shared token foundation (1/3)

pandemicsyn:
Adds shared authentication support for audience-bound tokens while retaining legacy token acceptance.

Includes:

- Token policy validation and shared issuance/delegation helpers.
- Runtime authorization contracts with bounded lifetimes and current account/organization checks.
- Signed runtime-proxy proof verification in web authentication and Session Ingest.
- Signed organization scope taking precedence over request headers.
- Verified token claims exposed through Worker middleware, with matching Gastown/Wasteland types.
- Default-off master and per-family issuance settings, documented in `ENVIRONMENT.md`.

Merging deploys these readers and helpers. New token issuance remains disabled; producer endpoints, client migrations, and Cloud Agent runtime activation are separate changes.

Validation: 507 focused tests passed. Worker-utils, Session Ingest, web, Gastown, and Wasteland typechecks passed, along with scoped lint, formatting, and diff checks.

PR #5857, opened 2026-09-03
feat(auth): migrate shared resource credentials (phase 5.2)

pandemicsyn:
## Summary

Adds audience-bound tokens behind separate rollout flags. Existing CLI and mobile clients keep legacy issuance by default. Modern credentials retain scoped renewal and recovery support during rollback.

## Rollout settings

**For this merge: enable nothing. Leave all new adoption flags `false` or unset.** Merging automatically deploys the services; wait until all deployments are healthy before changing flags.

### Later: enable one family at a time

First pass that family's end-to-end smoke test. Then, in **both web projects** (`kilocode-app` and `kilocode-global-app`), set `SHARED_RESOURCE_TOKENS_ENABLED=true` and **one** of these flags to `true`. Leave the others off until separately validated.

| Family | Web flag to enable |
| --- | --- |
| Chat, events, notifications | `CHAT_RESOURCE_TOKENS_ENABLED` |
| Explicit API/gateway, attribution, HTML delegation | `DELEGATED_RESOURCE_TOKENS_ENABLED` |
| Workflow gateway credentials | `WORKFLOW_GATEWAY_RESOURCE_TOKENS_ENABLED` |
| Auto-routing benchmarks | `BENCHMARK_RESOURCE_TOKENS_ENABLED` |

The shared flag alone enables none of these families.

### Keep these off for now

| Where | Flag | What must happen before enabling |
| --- | --- | --- |
| Cloud Agent Worker | `RUNTIME_ISOLATION_ENABLED` | Pass Worker/wrapper isolation and startup/recovery smoke tests. Enable this **before** Cloud Agent web issuance. |
| Both web projects | `CLOUD_AGENT_RESOURCE_TOKENS_ENABLED` | Resolve startup/recovery smoke failures and verify isolation above; also requires the shared web flag. |
| Both web projects | `NATIVE_RESOURCE_TOKENS_ENABLED` | Complete CLI/mobile device and downgrade testing. Leave off in this shipping stage. |
| Both web projects | `GASTOWN_RESOURCE_TOKENS_ENABLED` | Fix Session Ingest audience coverage and credential renewal inside running CLI processes. Leave off in this shipping stage. |
| Both web projects | `WASTELAND_RESOURCE_TOKENS_ENABLED` | Verify its complete consumer chain. |
| Security Auto Analysis Worker | Its own `SHARED_RESOURCE_TOKENS_ENABLED` | Validate this Worker's issuance/consumer chain independently. |
[...]

PR #5975, opened 2026-09-08
feat(cloud-agent): scoped runtime credentials (2/3)

pandemicsyn:
## Change

Add Cloud Agent scoped runtime credentials, isolation, renewal/recovery, and web producer wiring. Lost-response create retries retain runtime authorization. Recovery failures emit credential-free diagnostics, including throttled warnings for prolonged locks, while preserving retirement fences.

Stack: [foundation #5974](https://github.com/Kilo-Org/cloud/pull/5974) → **Cloud Agent #5975** → [remaining families #5976](https://github.com/Kilo-Org/cloud/pull/5976). Original #5857 remains unchanged.

## Rollout

**Resolve the outstanding Cloud Agent smoke failures before merging.** Then wait for the foundation deployment to be healthy before merging this PR.

| Setting | Ships as | Next action |
| --- | --- | --- |
| Worker `RUNTIME_ISOLATION_ENABLED` | **true**, production and dev | Verify deployed Worker and wrapper readiness |
| Web `SHARED_RESOURCE_TOKENS_ENABLED` | Off | Enable after this deployment is healthy and verified |
| Web `CLOUD_AGENT_RESOURCE_TOKENS_ENABLED` | Off | Enable with the web master after readiness verification |
| All other producer flags | Off | Activate separately |

Isolation permits modern workspace admission; it does not issue modern tokens or migrate legacy sessions. Cloud Agent activation does not need to wait for #5976. Rollback stops new issuance while retaining readers and runtime support for existing credentials.

## Validation

- Current checks: 5,906 unit tests passed (3 skipped), 651 Workers integration tests passed, and Worker/wrapper typechecks, service lint, formatting, and diff checks passed.
- Earlier validation: 1,251 wrapper tests and 49 focused web tests passed across full runs and isolated timing retries. These suites were not rerun for the registration/diagnostic changes.
- Real smoke remains incomplete: latest original-branch matrix passed 9/13; CLI/container startup, recovery, and SDK cold-wake failures remain.


services/cloud-agent-next/src/kilo/__fixtures__/runtime-url-normalization.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Immutable compatibility fixtures copied from Kilo CLI `route()` behavior.
+ * Sources: Kilo v7.4.20 release artifact and current CLI main (2026-09-04).
+ *
+ * Both versions normalize a configured base then append absolute API paths.
+ * Provider-specific `/api/gateway/v1/...` compatibility aliases are tested by
+ * the Cloud facade resolver, not asserted as behavior of `route()` itself.
+ * Keep this independent of the mutable CLI checkout: older installed CLIs are
+ * part of the Cloud Agent compatibility contract.
+ */
+export const historicalRouteFixtures = [
```
intentionally-left-nil: This file is all dead code, nothing uses either the dict or the function besides the tests. Remove it

services/cloud-agent-next/src/persistence/CloudAgentSession.ts
```diff
@@ -1667,6 +1690,313 @@ export class CloudAgentSession extends DurableObject<WorkerEnv> {
...
+    ownerId: string;
+    expectedOldId: string;
+    recoveryId: string;
+    runtimeAuthorizationSeal: string;
+    runtimeToken: string;
+  }): Promise<{ status: 'recovered' | 'not-needed' | 'denied' | 'busy' | 'retry' }> {
+    const metadata = await this.getMetadata();
+    if (!metadata || metadata.identity.userId !== input.ownerId) return { status: 'denied' };
+    const secret = await resolveSecret(this.env.NEXTAUTH_SECRET);
+    if (!secret) return { status: 'denied' };
```
intentionally-left-nil: Should we log with this happens, because it's our fault?

services/cloud-agent-next/src/persistence/CloudAgentSession.ts
```diff
@@ -1667,6 +1690,313 @@ export class CloudAgentSession extends DurableObject<WorkerEnv> {
...
+            auth: { ...latestMetadata.auth, kilocodeToken: input.runtimeToken },
+          })
+        );
+        await transaction.delete(RUNTIME_PROXY_GRANT_KEY);
+        await transaction.delete(RUNTIME_AUTHORIZATION_RECOVERY_KEY);
+      });
+      await clearWrapperRuntimeIdentity(this.ctx.storage, {}, { incrementGeneration: true });
+      return { status: 'recovered' };
+    } catch (error) {
+      if (error instanceof Error && error.message === 'runtime_authorization_recovery_cas_failed') {
```
intentionally-left-nil: Remove this if statement, you return the same thing anyways

services/cloud-agent-next/src/persistence/CloudAgentSession.ts
```diff
@@ -1667,6 +1690,313 @@ export class CloudAgentSession extends DurableObject<WorkerEnv> {
...
+    }
+    const acquired = await this.ctx.storage.transaction(async transaction => {
+      if (
+        (await countPendingSessionMessages(transaction)) > 0 ||
+        (await hasNonTerminalSessionMessage(transaction))
+      ) {
+        return false;
+      }
+      const existingLock = runtimeAuthorizationRecoveryLockSchema.safeParse(
+        await transaction.get<unknown>(RUNTIME_AUTHORIZATION_RECOVERY_KEY)
```
intentionally-left-nil, line 1825: I'm always going to be squinty eyes about database locks :) I asked the AI what the livelock possibilities are, and it gave me this analysis:

```
What happens if it never clears
The lock schema itself carries no timestamp — runtimeAuthorizationRecoveryLockSchema is just { expectedOldId, recoveryId } (runtime-authorization-persistence.ts:15-17). There's no lock age, no alarm, and no other RPC that unconditionally clears RUNTIME_AUTHORIZATION_RECOVERY_KEY — the only put/delete of that key in CloudAgentSession.ts are the two lines above. So if:
- the sandbox is wedged and observeWrappersWithoutWaking() never reports absent, or
- a terminal client is leaked and listTerminals() never returns empty, or
- metadata/RUNTIME_AUTHORIZATION_KEY permanently diverges from expectedOldId so the final CAS can never succeed,
then every retry (whether it's the client automatically resubmitting, or withDORetry) just re-enters, sees its own recoveryId in the lock, no-ops the write, retries the same doomed step, and returns retry again — forever. There is no self-healing timeout.
```

From that list, leaking a terminal seems to be the most likely future candidate. Perhaps a timeout or alerting would be helpful?

intentionally-left-nil (commented):
too much code to review, here's a first pass :)

services/cloud-agent-next/src/kilo/__fixtures__/runtime-url-normalization.ts
pandemicsyn: Removed in ce43d4345. The compatibility checks now use literal CLI request URLs against the production facade resolver; the unused fixture and emulated CLI helper are gone.

services/cloud-agent-next/src/persistence/CloudAgentSession.ts
pandemicsyn: Added in 8599d647b: missing NEXTAUTH_SECRET now emits a structured error with the session ID and a fixed missing_secret reason before returning denied. Recovery exceptions also log the failed stage without including credentials, seals, or raw exception text. A Workers integration regression verifies the logs and redaction.

services/cloud-agent-next/src/persistence/CloudAgentSession.ts
pandemicsyn: Removed in 8599d647b. The catch now logs the failed recovery stage and returns retry once, without the redundant conditional.

services/cloud-agent-next/src/persistence/CloudAgentSession.ts
pandemicsyn: Confirmed: a leaked terminal or an unconfirmed sandbox stop can leave recovery fenced indefinitely. Automatically expiring that fence would let us replace authority without proving the old runtime is gone.

8599d647b adds durable timing metadata in a separate record, preserving the existing strict two-field lock format for older Workers. Once a lock has been observed for five minutes, continued retries or state inspections emit a warning containing session ID, both recovery IDs, and lockAgeMs, at most once every five minutes. Successful recovery clears that diagnostic record with the lock. Existing locks start timing on their first observation.

This provides an operational warning signal, not a background alarm or configured paging alert. Authenticated terminal closure remains available so a stuck PTY can be closed before retrying. The integration test exercises an aged lock with an active PTY: it warns once, retains the lock/authority, and does not stop the runtime. Unit coverage also verifies throttling and timing reset for a different recovery.

services/cloud-agent-next/src/session/session-registration.ts
```diff
@@ -908,8 +976,11 @@ function buildSessionRegistrationCommand(
     },
     auth: {
       kiloSessionId: allocation.kiloSessionId,
-      kilocodeToken: ctx.authToken,
+      kilocodeToken: allocation.runtimeAuthorization?.token ?? ctx.authToken,
```
eshurakov, line 985: (bot) **P1: Migrate the control-plane ingest bridge when storing a runtime JWT.**

With modern issuance enabled, this replaces the stored bearer with a credential that requires runtime-proxy attestation. `SandboxSession.ts:745-785` still reads that value and calls `publishControlPlaneSessionIngest` through `SESSION_INGEST.fetch`. The bridge (`control-plane-ingest.ts:123-145`) supplies bearer/internal-lineage headers but no attestation. Session Ingest applies `kiloJwtAuthMiddleware` to the internal child routes as well as public routes (`app.ts:64-71`), so its proof check rejects these requests with 401; the internal secret does not bypass it.

Consequently child-session creation and transcript persistence through this bridge fail. The bridge logs and returns on failed child creation, while the event receipt is already committed and duplicate receipts skip publication (`SandboxSession.ts:733-736`). The native proxy only permits the root session ID, so it does not provide a child fallback. This is not a claim that every root transcript is lost.

Please migrate this server-side consumer to an authorized, attested path without weakening direct-runtime-bearer rejection. Verify a real modern credential through the actual bridge and ingest middleware: child creation, persisted lineage, and exported child messages must succeed. Source-traced at this revision; not a live E2E reproduction.

services/cloud-agent-next/src/session/session-registration.ts
```diff
@@ -523,6 +535,55 @@ function effectiveSessionRegistrationInput(
...
+    });
+    runtimeAuthorization = {
+      token: created.token,
+      seal: await sealRuntimeAuthorization(created.authorization, nextAuthSecret),
+    };
+    if (initialTurn?.type === 'prompt') {
+      await assertKiloModelAvailable({
+        env: ctx.env,
+        submittedModel: input.agent.model,
+        originalToken: runtimeAuthorization.token,
```
eshurakov, line 584: (bot) **P1: Preserve authenticated model validation for runtime credentials.**

With modern issuance enabled, this passes the proof-required backing runtime JWT directly to `assertKiloModelAvailable`. Its HTTP requests (`model-validation.ts:59-67`) do not carry runtime-proxy attestation, so web authentication rejects the bearer (`apps/web/src/lib/tokens.ts:254-266`). The personal model validator then has no authenticated user and omits `getDirectByokModelsForUser`; organization validation returns 401 and the caller falls back to the anonymous catalog.

This makes configured private-only BYOK models, for example `kimi-coding/kimi-for-coding`, unavailable for modern session creation. Existing-session preflight uses the same stored backing JWT and has the same problem. Public models can still pass, so the anonymous fallback hides the regression rather than preserving the authenticated catalog.

Please use an authorized, attested validation path. Verify both personal and organization credentials with a configured model absent from the public catalog against the actual web validator: authorized choices must pass as on the legacy path, and unauthorized private models must remain unavailable. Source-traced at this revision; not a live E2E reproduction.

eshurakov (approved):

services/cloud-agent-next/src/session/session-registration.ts
pandemicsyn: Fixed in 505b627e1. The owning DO renews the backing token and supplies its persisted authorization to the bridge. The bridge verifies owner/org, resource, active delegation and exact bearer before issuing Session Ingest attestation; recovery/replacement checks fence publication. Direct runtime bearers still require proof, and legacy publication remains supported.

Added 11 passing PostgreSQL + Durable Object integration cases using the actual bridge, ingest middleware, child-create route and export code. Modern and legacy child creation, persisted lineage and exported messages pass; missing proof and invalid authority/scope are rejected. The old bridge reproduces the 401. This regression suite now runs in CI. Full Cloud Agent validation also passes: 5,929 unit tests and 651 Workers integration tests. These are local receiver integration tests, not a live sandbox/provider E2E claim.

services/cloud-agent-next/src/session/session-registration.ts
pandemicsyn: Fixed in 9c326ff57, with follow-up admission wiring in 505b627e1. Model validation now verifies runtime credentials and attaches bearer-bound proof for the appropriate personal/organization validator. Runtime requests cannot follow redirects or retry anonymously after a 401. Existing/prepared-session and follow-up validation renew the backing JWT through the owning DO within the existing delegation. Legacy catalog behavior is preserved.

Added 13 passing integration cases through the actual web validators/authentication: authorized personal and organization models absent from the public catalog pass, unauthorized private choices fail, legacy access passes, and expired backing tokens renew only within valid authority. The old implementation reproduces the failure. Also passed 56 focused unit tests, the full Cloud Agent unit/Workers suites, and web/Worker/wrapper typechecks.

PR #5978, opened 2026-09-08
feat(github): share installations with web Cloud Agents 

pandemicsyn:
## Outcome
Default-off shared canonical GitHub installations for WEB Cloud Agents with exact tenant-association pinning. Slack shared use remains PR3; Code Reviewer shared routing remains PR4.

Reviewed base: current `main` at `cf59d8662499749228164232615da5af0abe5134`. Reviewed head: `af1734d3ce0b8d11acad16622754be8618603cdb`.

## Exact Association Boundaries
Generic token and Chat SDK webhook paths require canonical `sharing_mode = exclusive`. Managed shared lookup requires an exact active association ID. GitHub bot-link signed state carries exact `platformIntegrationId`; callback validates association, installation, Standard/Lite identity, canonical lifecycle, local health, and current Kilo owner or organization membership.

Predeploy bot-link state without an association ID is accepted only while management and sharing admission remain disabled and only for one healthy canonical-exclusive or true-legacy literal-null app association. Explicit Standard rows require exact signed association identity. Keep features disabled for at least one signed-state TTL after deployment.

## Simple Synchronous Lifecycle
Required canonical/local database cleanup runs first and fails non-2xx without recording completion. Standard bot unlink runs second as best-effort alerting. Completed-success receipts dedupe later deliveries; concurrent and crash-window repeats are idempotent. GitHub does not automatically retry failed webhooks, so manual redelivery after database failure is the accepted recovery. No lifecycle queue, cron, outbox, lease, fencing, Retry-After, or automatic retry worker exists.

Repository refresh follows lifecycle lock order and cannot revive terminal state.

## Provider Reservation Boundary
Slack, Linear, and Discord OAuth attempts are durable, expiring, owner-row locked, purpose-bound, and single-use. PR3 owns provider-installation generation and lazy recovery.

## Migration And Rollout
Generated migration: `0243_exotic_triton.sql`. Replacement indexes build concurrently with required transaction boundaries. Before enabling sharing, stale eligible-unlinked must be zero. PR2 adds no canonical reconciliation DML.

## Verification
- Accepted pre-rebase bot-link/state/DB resolver suite: 99 passed
- Accepted pre-rebase schema/migration suite: 101 passed; fresh bootstrap passed
- Final-rebase Git token service: 633 passed
- Final-rebase focused Cloud Agent suites: 429 passed
- Final-rebase range-diff: all 28 PR2 commits exact
- Changes-only typecheck and web/DB/token/Cloud Agent lint: passed
- Format, whitespace, and conflict-marker checks: passed

The final-rebase PostgreSQL-backed Jest/bootstrap rerun was blocked before test execution because the local OrbStack Docker daemon was unresponsive; `docker info`, Compose health, and `pnpm test:db` timed out. This was not an assertion failure. The accepted pre-rebase head had the schema/bootstrap and focused web suites passing.

No production queries, repairs, associations, settings, flags, backfills, deployments, or external provider operations were performed.


pandemicsyn:
Addressed the review against the current head.

- **Blocking index rebuild:** stale against the prior head. `0243_gray_supernaut.sql` builds the four replacement canonical/pending indexes with `CONCURRENTLY` and the repository-required `COMMIT`/`BEGIN` migrator boundaries.
- **Runtime demotion ambiguity:** valid and fixed in `727326424`. Generic runtime authorization now filters active/local-health/canonical identity and `sharing_mode = exclusive` in SQL before `LIMIT`, so one active exclusive association remains usable when a disconnected sibling is retained; shared mode still fails closed.
- **Exclusive uninstall retry poison:** valid and fixed in `727326424`. `installation.deleted` now uses the attempt-fenced canonical pending/completed receipt for both exclusive and shared installations. Failed unlink/cleanup remains reclaimable, completed receipts suppress redelivery, and concurrent duplicates single-dispatch.

Verification: 69 focused runtime/lifecycle tests, 254 cleanliness behavior tests, 101 schema tests plus fresh bootstrap, 632 token tests, and 419 focused managed Cloud Agent tests passed.

pandemicsyn:
Follow-up on the lifecycle retry disposition: `cd4e7d7ba` closes the remaining quick-redelivery gap. A pending receipt whose lease is still active now returns HTTP 503 with `Retry-After` rather than duplicate 200. Completed receipts alone return duplicate 200; expired pending leases reclaim. Attempt-version fencing protects newer claimants. Tests cover dispatch failure plus receipt-release failure, quick non-2xx retry, post-expiry reclaim/completion, concurrent winner, and completed duplicate.

The pending-index compatibility concern is an explicitly accepted rollout risk: no global bridge index is retained. The migration-to-app promotion window is brief and monitored, and old pending callbacks receive retry guidance after promotion.

pandemicsyn:
Product simplification supersedes the prior receipt-lease disposition. Current head `622241ea3` now uses simple synchronous lifecycle handling:

- Required canonical/local DB cleanup runs first and fails non-2xx without recording completion.
- Standard bot unlink runs second as best-effort alerting and cannot undo or fail successful local cleanup.
- Only completed-success receipts are stored; completed duplicates return 200.
- No pending lease, attempt fencing, Retry-After, queue, cron, outbox, poison state, or automatic retry remains.
- GitHub does not automatically retry failed webhooks; manual redelivery after a DB failure is the explicitly accepted recovery. Concurrent/crash-window repeats are tolerated because cleanup is idempotent.

Verification: 75 focused lifecycle and DB-backed token behavior tests, 101 schema tests plus fresh bootstrap, and all 633 token tests passed.

apps/web/src/lib/integrations/platforms/github/webhook-handler.ts
```diff
@@ -161,6 +167,30 @@ export async function handleGitHubWebhook(
...
+        installationId,
+        appType,
+        deliveryId: eventSignature,
+      });
+      if (receipt === 'completed') {
+        return NextResponse.json({ message: 'Duplicate event' }, { status: 200 });
+      }
+      const response = await dispatch();
+      if (receipt !== 'missing_canonical') {
+        await recordCompletedGitHubInstallationDelivery({
```
iscekic, line 185: bot: `dispatchSharedOnce` records the completion receipt for any resolved `dispatch()` response, regardless of HTTP status. A non-2xx response is stored as completed, so a later GitHub redelivery is short-circuited as `Duplicate event` and manual recovery is blocked. Record only on a success response (for example, `response.ok`), or narrow the receipt to the explicit success paths.

apps/web/src/lib/integrations/provider-oauth-attempts.ts
```diff
@@ -0,0 +1,220 @@
...
+          ownerCondition(input.owner),
+          eq(provider_oauth_attempts.provider, input.provider),
+          or(
+            eq(provider_oauth_attempts.status, 'pending'),
+            eq(provider_oauth_attempts.status, 'consumed')
+          ),
+          lt(provider_oauth_attempts.expires_at, now)
+        )
+      );
+    await tx.insert(provider_oauth_attempts).values({
```
iscekic, line 108: bot: This insert conflicts with `UQ_provider_oauth_attempts_user_pending` / `UQ_provider_oauth_attempts_org_pending` when an unexpired pending attempt already exists for the same owner and provider. `pruneProviderOAuthAttempts` only expires attempts whose `expires_at` has passed, so a user who abandons an OAuth flow cannot start a new one for up to 10 minutes; `/connect` catches the failure and redirects with `oauth_init_failed`. If a new attempt is meant to supersede the old one, expire the existing pending row before insert (or upsert it).

apps/web/src/lib/cloud-agent/github-integration-helpers.ts
```diff
@@ -254,7 +245,7 @@ async function fetchRepositoriesForIntegrations(
     }
     return {
       integrationInstalled: true,
-      repositories: dedupeRepositories(results.flatMap(result => result.repositories)),
+      repositories: results.flatMap(result => result.repositories),
```
iscekic: bot: Removing `dedupeRepositories` means one repository can now appear once per installation that grants access to it. With shared or multiple installations this surfaces duplicate entries for the same repo. Confirm the UI and callers tolerate duplicates; otherwise keep deduping (for example, by `platformIntegrationId` plus repo id).

packages/db/src/migrations/0243_exotic_triton.sql
```diff
@@ -0,0 +1,47 @@
...
+CREATE UNIQUE INDEX "UQ_github_installation_webhook_receipts_delivery" ON "github_installation_webhook_receipts" USING btree ("github_installation_id","delivery_id");--> statement-breakpoint
+CREATE UNIQUE INDEX "UQ_provider_oauth_attempts_user_pending" ON "provider_oauth_attempts" USING btree ("owned_by_user_id","provider") WHERE "provider_oauth_attempts"."status" = 'pending' AND "provider_oauth_attempts"."owned_by_user_id" IS NOT NULL;--> statement-breakpoint
+CREATE UNIQUE INDEX "UQ_provider_oauth_attempts_org_pending" ON "provider_oauth_attempts" USING btree ("owned_by_organization_id","provider") WHERE "provider_oauth_attempts"."status" = 'pending' AND "provider_oauth_attempts"."owned_by_organization_id" IS NOT NULL;--> statement-breakpoint
+CREATE INDEX "IDX_provider_oauth_attempts_expires_at" ON "provider_oauth_attempts" USING btree ("expires_at");--> statement-breakpoint
+ALTER TABLE "platform_integrations" ADD CONSTRAINT "platform_integrations_github_installation_id_github_app_installations_id_fk" FOREIGN KEY ("github_installation_id") REFERENCES "public"."github_app_installations"("id") ON DELETE no action ON UPDATE no action;--> statement-breakpoint
```
iscekic: bot: This FK validates every existing `platform_integrations.github_installation_id` and fails the migration if any row references a missing `github_app_installations.id`. `0239` added the column without an FK and this migration adds no reconciliation DML, so confirm there are no orphan references before deploy. Note also that `ON DELETE no action` makes canonical deletes fail while any association still references them.

iscekic (approved):
Approving to unblock

packages/db/src/migrations/0243_exotic_triton.sql
pandemicsyn: Fixed in ec51eee58.

Split migration 0243 into two: `0243_reconcile_orphaned_github_installation_id.sql` nulls out any `platform_integrations.github_installation_id` that no longer references a live `github_app_installations` row *before* the FK-adding migration runs, so the FK's full-table validation cannot fail on drift (defensive; current app invariants make real orphans very unlikely — the only hard-delete path for `github_app_installations`, `anonymizeCloudUserData`, already guards with a `NOT EXISTS` check).

For ON DELETE: kept it blocking, but made it explicit as `ON DELETE restrict` (was implicit `no action`), matching this repo's convention for other blocking FKs (e.g. `kilocode_users`/`organizations` refs elsewhere in schema.ts). This matches the actual lifecycle: canonical `github_app_installations` rows are soft-deleted (`lifecycle_state`/`deleted_at`) on disconnect/uninstall and are never expected to be hard-deleted while an association still points at them — the code already treats that as an invariant. Documented the reasoning inline in schema.ts next to the FK.

Verified: `drizzle-kit check` passes, migrations apply cleanly against a fresh local Postgres, and `packages/db/src/schema.test.ts` (101 tests) + `github-installations.test.ts` (32 tests) + `lib/user/index.test.ts` (151 tests) all pass.

apps/web/src/lib/integrations/platforms/github/webhook-handler.ts
pandemicsyn: Fixed in 11581780c.

`dispatchSharedOnce` now only records the completion receipt when `response.ok`, leaving non-2xx dispatches unrecorded so GitHub's redelivery can still retry them. The `missing_canonical` short-circuit (skip recording entirely) is unchanged.

Added two tests: one asserting a non-2xx dispatch does not record a receipt and that a manual redelivery still re-invokes the handler, and one asserting a 2xx dispatch does record the receipt. `webhook-handler.test.ts` (30 tests) passes.

apps/web/src/lib/integrations/provider-oauth-attempts.ts
pandemicsyn: Fixed in 731e77549.

`beginProviderOAuthAttempt` now expires any existing **pending** row for that owner+provider unconditionally (dropped the `expires_at < now` filter for the pending branch), so a fresh attempt always supersedes an abandoned-but-unexpired one instead of colliding with `UQ_provider_oauth_attempts_user_pending`/`_org_pending`. Consumed rows are still only expired once actually past their TTL, since they don't hold that unique slot.

I checked `consumeProviderOAuthAttempt`: its pre-step doesn't insert anything — the actual consumption matches the specific row by exact `state_hash` — so it doesn't share this collision pattern and I left it unchanged.

Added a test ("supersedes an unexpired pending attempt when a new attempt starts") asserting a second `beginProviderOAuthAttempt` call succeeds while an unexpired pending attempt exists, that the old attempt can no longer be consumed, and the new one can. `provider-oauth-attempts.test.ts` (9 tests) passes against a real Postgres instance.

apps/web/src/lib/cloud-agent/github-integration-helpers.ts
pandemicsyn: Fixed in 242eb9873, with one deliberate scoping decision worth calling out.

I restored dedup on the flattened `repositories` array, using the composite key you suggested — `(platformIntegrationId, id)` — rather than repo id alone. I checked `mapRepositories`'s output shape and the actual consumers before choosing:

- `NewSessionPanel.tsx` keys and selects repository rows by `platform:platformIntegrationId:id`, not by repo id or `fullName` alone, and requires the exact `platformIntegrationId` to be carried through to session start so the backend uses the right installation's token/permissions.
- There's an existing, still-passing regression test in this file — `preserves both association choices when two installations expose one repository` — added deliberately (alongside `githubAppType`/shared-installation support) specifically to assert that when two *different* installations both grant the same repo, **both** entries must remain selectable, because they can resolve to different tokens/permissions at session start.

Deduping by repo id alone (or by `fullName`, which is what the original `dedupeRepositories` used before it was removed) would silently drop one of those two legitimate, functionally distinct entries and reintroduce the exact regression that test guards against.

So the fix I landed only collapses **exact duplicate entries within a single installation's own repository list** (e.g. a corrupted or duplicated cache) — it does not dedupe across installations. Added a test for that case; the existing cross-installation test is unchanged and still passes (17 tests total in `github-integration-helpers.test.ts`), and I also ran `organization-cloud-agent-next-router.test.ts` (90 tests) to confirm no downstream regression.

Happy to revisit if the intent is actually to collapse cross-installation duplicates in the picker — but that would need a product decision on how to represent the "which installation gets used" choice, since it's tied to session token resolution, not just display.

iscekic (approved):

PR #5980, opened 2026-09-08
fix(ai-gateway): detail efficient routing failures

chrarnoldus:
## Summary
- Return a 503 routing-unavailable response instead of the misleading “organization blocks every model in the auto-routing pool” response when Auto Efficient falls back after decision-making fails and organization policy rejects that fallback.
- Include non-sensitive failure details: a stable reason code, the blocked fallback model ID, and the worker HTTP status when available.
- Log the same message and structured details through `warnExceptInTest`, matching the original error helper’s warning behavior.
- Preserve failure causes for missing worker configuration, request normalization failure, timeout, transport failure, non-2xx responses, invalid worker responses, valid responses with no decision, virtual-model decisions, and missing catalog variants.

## Why
A routing failure is not evidence that the organization blocks every model in the Efficient pool. The prior message conflated two separate facts: decision-making did not produce a usable model, and organization policy rejected the static fallback. The new response and warning log state both facts directly.

Example details:
```json
{
  "reason": "worker_http_error",
  "blocked_fallback_model_ids": [
    "moonshotai/kimi-k3"
  ],
  "worker_status": 502
}
```

Only stable, non-sensitive fields are returned and logged. Exception text, response bodies, worker URLs, credentials, user IDs, and organization IDs are excluded.

## Verification
- Changed-file formatting: passed.
- Changed-file oxlint: passed with 0 warnings and 0 errors.
- `pnpm --filter web exec tsgo --noEmit`: passed (with the local Node 26 versus required Node 24 engine warning).
- `git diff --check`: passed.
- Test suites were left to CI.

### DIFF
--- a/apps/web/src/app/api/openrouter/[...path]/route.test.ts
+++ b/apps/web/src/app/api/openrouter/[...path]/route.test.ts
@@ -1193,7 +1193,8 @@ describe('kilo-auto/efficient classifier billing', () => {
     await Promise.resolve();
 
     expect(mockedFetchEfficientAutoDecision).toHaveBeenCalledWith(
-      expect.objectContaining({ requestedModel: 'kilo-auto/balanced' })
+      expect.objectContaining({ requestedModel: 'kilo-auto/balanced' }),
+      expect.objectContaining({ onFailure: expect.any(Function) })
     );
     expect(mockedLogMicrodollarUsage).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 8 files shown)


chrarnoldus:
Fixed in 08918ddfe. The fallback-specific response now applies only when no abuse model override occurred and the effective model still matches the recorded fallback. A quarantine-3 replacement that is denied by organization policy receives the ordinary model-not-allowed response without stale fallback details or fallback warning logs. Added focused regression coverage for that flow.

PR #6001, opened 2026-09-09
Respect the per-repo override for the model selection

intentionally-left-nil:
If the user has chosen a different model for a github repository, switch out the model once the repository is known

The initial inference to determine the repository and other setup still uses the users default model for their integration

## Summary

Note: I decided explicitly against wiring this into the UI for the New Cloud Agent session. That modal almost always prefers the user's most recent model (unless it's their first experience). I didn't want to create confusion about user expectations (should it always switch away, what happens if the user overrides it etc), so the simplest thing was to leave this alone for now. Instead the scenarios affected are:

1. Starting a bot conversation from a GitHub mention — apps/web/src/lib/bot/agent-runner.ts:211 calls resolveBotModelSlug(integration) to pick the model for the whole conversation, before any specific repo is chosen.
2. Spawning a cloud agent session once a repo is picked within that bot conversation — apps/web/src/lib/bot/tools/spawn-cloud-agent-session.ts:197-206. Here a per-repository override (if configured) takes precedence over this installation-level default.
3. Clicking "fix this review" from a PR review comment — apps/web/src/app/cloud-agent-fork/review/[reviewId]/route.ts:91-101, which resolves the model for the agent session created to address the review, falling back through repo override → installation default → hardcoded DEFAULT_BOT_MODEL.

## Verification

Testing this locally isn't easy, since it's realistically triggered from an `@mention` and not e.g. the new cloud agent page. That tab always has the user's last model as the sticky one, and I don't want to change it. So, unit tests only for now


apps/web/src/lib/bot/tools/spawn-cloud-agent-session.ts
```diff
@@ -184,10 +185,25 @@ export default async function spawnCloudAgentSession(
...
       owner.type === 'org'
-        ? await getGitHubTokenForOrganization(owner.id)
-        : await getGitHubTokenForUser(owner.id);
+        ? getGitHubTokenForOrganization(owner.id)
+        : getGitHubTokenForUser(owner.id),
+      // A per-repository model override (`repository_customizations`) takes
+      // precedence over the installation-default `model` resolved earlier for
+      // this whole bot conversation — the repo is only known now that the LLM
+      // has picked one via this tool call.
+      resolveModelForGitHubRepository(platformIntegration, args.githubRepo),
```
pandemicsyn: bot: A repository-model lookup failure now aborts the GitHub bot tool instead of falling back to the existing model.

Suggested fix: Guard `resolveModelForGitHubRepository` independently (or use `Promise.allSettled`), capture the exception, and retain the incoming `model` as `effectiveModel` when resolution fails. This preserves the prior session-creation behavior during a customization-query failure while still applying overrides when available. Add a test that makes the resolver reject and asserts that `prepareSession` receives the original model.

pandemicsyn (approved):

PR #6003, opened 2026-09-09
fix(cloud-agent-next): isolate control event lanes

eshurakov:
## Summary
- Isolate control event lanes so independent streams cannot block or retire each other.
- Stacked on cleanup.

## Test plan
- [ ] Existing cloud-agent-next control-plane unit tests


iscekic (approved):
Approved.

PR #6005, opened 2026-09-09
fix(cloud-agent-next): isolate scoped Stop across the service boundary

eshurakov:
## Summary
- Isolate scoped Stop across the service boundary.
- Release operation-only root cleanup without fencing sibling chats.
- Keep per-root claims from retiring a successor; reject root+quiescence on the strict abort-result reader.
- Preserve shared-root retirement without a kilo runtime; detach terminals only after successful abort retirement.

## Test plan
- [ ] cloud-agent-next wrapper unit tests for abort, worktree, terminal, operation-registry

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -88,8 +88,8 @@ catalogs:
       specifier: 1.0.0
       version: 1.0.0
     wrangler:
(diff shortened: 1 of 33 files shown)


iscekic (approved):
Approved.

PR #6030, opened 2026-09-09
feat(kilo-mcp): remote MCP worker with hybrid catalog search, tRPC calls, and OAuth 2.1 sign-in (part 1/1)

iscekic:
## Changelog for users

- Agents had no way to discover or invoke the Kilo API over MCP. A new Kilo MCP worker now serves MCP Streamable HTTP with exactly two tools: `search` and `call`.
- `search` turns a plain-language task into matching Kilo endpoints: exact path and keyword matches rank first.
- A Vectorize semantic index also returns the right endpoint for queries with zero keyword overlap. If the index is unavailable, `search` still answers from keyword matching instead of failing.
- `call` executes a Kilo API query as the signed-in user. It accepts only catalog paths and validates arguments against the published schema before sending anything.
- Results larger than 16 KB come back cut at the cap with a `[truncated]` marker.
- Connecting a client runs MCP OAuth 2.1: dynamic client registration, sign-in with the existing Kilo login, an org picker, PKCE-protected token exchange, and refresh-token rotation.
- Access tokens are bound to the user, the chosen org, and this MCP. Requests without a valid token get a 401 challenge pointing clients at the sign-in metadata.
- The catalog publishes all 359 tRPC queries except `admin.*`, `debug.*`, and `test.*`, each with a search-friendly summary authors can hand-edit.

## Changelog for maintainers

- The dump in `apps/web/src/scripts/mcp-catalog` walks the live `rootRouter` (queries only, denylist `admin`/`debug`/`test`) and keeps committed summaries byte-for-byte. Missing summaries are generated per router file via OpenRouter or Anthropic; a corrupt committed catalog fails loudly, and an incomplete catalog is never written.
- CI runs the catalog workflow on every PR and gates the work on a change-detection step, so `catalog (PR)` can be a required status check without leaving unrelated PRs blocked. Same-repo PRs get the refreshed `services/kilo-mcp/catalog.json` committed to the branch, which removes catalog drift. Fork PRs fail with a `catalog.patch` artifact and a comment, including when GitHub withholds the secret. The main-push job re-dumps and upserts Vectorize; PR jobs never touch it. `scripts/kilo-mcp-catalog.test.mjs` asserts this wiring.
- The worker bundles the committed catalog at build time. The Vectorize index pins `@cf/baai/bge-base-en-v1.5` (768 dimensions, cosine) with vector id = procedure path; only the search query is embedded per request. Production deploys from main keep the bundled catalog and the index in lockstep.
- Search ranks exact-path hits above path-sequence hits above token overlap, and every lexical hit above a semantic-only hit. A Vectorize failure degrades to token-only results with a logged warning.
- `call` rejects unknown paths and schema-invalid input before any upstream request. It forwards over apps/web's tRPC GET transport with the verified identity's Kilo credential and the org header from the token claims. `x-kilocode-organizationid` mirrors the apps/web constant — keep them in sync.
- The OAuth 2.1 suite covers well-known metadata, public-client DCR, the `/authorize` pairing with consent page, a single-use status relay, and a membership-validated org picker. `/token` enforces single-use codes, S256 PKCE, and rotating hashed refresh tokens. State lives in one Durable Object (`KiloMcpOAuthStore`, DO SQLite, migration tag v1, 6-hour purge alarm).
- `/mcp` accepts only this worker's HS256 tokens (issuer, audience, expiry, and jti registry checked). A deployment missing `MCP_TOKEN_SECRET` or the Durable Object binding refuses POSTs with 503 instead of forwarding unverified bearers.
- Resolved: commit `97df6a80c` opens the authorize page's Continue with Kilo sign-in link in a new tab (`target="_blank" rel="noopener noreferrer"`), so the pairing-status poll survives sign-in. The authorize screenshot above predates the fix.
- The dump's `kilo run` never uses a maintainer's personal credential. Both jobs exchange `MCP_CATALOG_TOKEN_SECRET` at `POST /api/internal/mcp-catalog/token` for a 1-hour token that belongs to the benchmarking service account.

## Human steps before merge and deploy

**Needs a repo admin.**
- Add `catalog (PR)` to the `main-tests` ruleset required status checks. The automation token lacks repo admin and gets HTTP 404 from the rulesets API. GitHub UI: Settings → Rules → `main-tests` → Require status checks → add `catalog (PR)`.

**Needs the deployer.**
- Run `pnpm web:env set MCP_CATALOG_TOKEN_SECRET --only production --production-file <(op read "op://Kilo Web ENV Production/MCP_CATALOG_TOKEN_SECRET/password")` so Vercel serves the mint route. Agents must not run this command.
- Approve the PR (1 review required).

Already done, no action: `MCP_CATALOG_TOKEN_SECRET` is in 1Password (`Kilo Web ENV Production`) and in GitHub repo secrets; the personal `MCP_CATALOG_KILO_AUTH` secret is deleted from GitHub and 1Password; `MCP_TOKEN_SECRET` is set on both `kilo-mcp` and `kilo-mcp-dev`; the `kilo-mcp-catalog` Vectorize index exists.

## E2E proof (stack-wide)

<details>
<summary>Owner request for the stack</summary>

> 
> # Kilo Remote MCP
> 
> New Cloudflare Worker (`services/kilo-mcp`). Not MCP Gateway. Two tools: `search`, `call`.
> 
> ## Shape
> 
> - Include every tRPC `query` except `admin.*`, `debug.*`, and `test.*`.
> - No mutations in v1. No per-procedure opt-in. Denylist, not allowlist.
> - Worker speaks MCP. It does not run tRPC.
> - `apps/web` stays the authorization server and the tRPC executor.
> - Product auth is MCP OAuth 2.1. The first slice ships without it.
> 
> ## Catalog
> 
> One dump script, in `apps/web` (only place `rootRouter` loads). Two jobs:
> 
> **PR job**
> 
> 1. Keep all queries minus the denylist.
> 2. If a row has no summary, an LLM reads the procedure handler and traces what it does. It writes a search-friendly summary: what the call does, in words an agent would type. Not implementation detail. Not a restatement of the path. If a summary already exists, keep it.
> 3. Emit `catalog.json`: path, kind, summary, JSON Schema, derived tags, search blob.
> 4. Tags and keywords are derived (path, kind, schema keys). Do not author them.
> 5. If `catalog.json` changed, CI commits it to the PR branch. Authors can edit a committed summary; CI will keep the edit.
> 6. Fork PRs cannot be pushed; CI fails and posts the catalog patch.
> 7. Do not generate summaries at MCP runtime. Do not write Vectorize.
> 
> **Merge job** (main only)
> 
> 8. Run the same dump (fill missing summaries if any slipped through).
> 9. Embed every catalog row with one pinned Workers AI model. Upsert Vectorize. Vector id = procedure path.
> 10. PRs do not write Vectorize. Do not embed per `call`.
> 
> Do not copy `trpc-registry.ts`. Do not put hand-written `.meta({ mcp })` on each procedure.
> 
> ## Auth
> 
> - MCP OAuth 2.1: PKCE, DCR, protected-resource metadata, resource indicator.
> - User signs in with existing Kilo login. Picks an org.
> - Token is bound to user + org + this MCP. Worker verifies. Forwards that identity to `/api/trpc/{path}`.
> 
> ## Agent loop
> 
> 1. `search` is hybrid: token overlap on the bundled catalog plus Vectorize kNN on the query embedding. Exact path/token hits rank above semantic hits.
> 2. `call` accepts only catalog paths. Validates input against the published schema.
> 3. Worker POSTs to `/api/trpc/{path}` as that user/org. Truncates the result.
> 
> Production Worker deploys from main, so bundled catalog and Vectorize stay in lockstep.
> 
> ## First slice
> 
> Bootstrap once: dump the query catalog (minus denylist), fill missing summaries, embed, upsert Vectorize. Prove dump → hybrid search → call on the Worker. No OAuth in this slice.

</details>

## Owner manual verification
Owner verification is pending; CI did not prove these paths automatically.
- Real GitHub PR run: same-repo PR gets a bot `catalog.json` commit; fork PR fails with `catalog.patch`. Not run here.
- Open a fork PR touching `apps/web/src/**` without a summary and assert the job fails with the guidance comment. Needs a real fork PR.
- The third original check (merge to main and observe the bot commit before the Vectorize upsert) is superseded: drift is removed because the PR job commits the catalog to the PR branch, and the merge job only fills straggler summaries.

## PR stack (merge bottom to top)
- part 1/1: #6030 (`kwf/kilo-remote-mcp-c204-l1`) ← this PR


services/kilo-mcp/src/auth/token.ts
```diff
@@ -0,0 +1,392 @@
...
+    if (wanted.length === 0 || !wanted.every(token => granted.includes(token))) {
+      return oauthErrorResponse(
+        400,
+        'invalid_scope',
+        `Requested scope exceeds the granted "${MCP_SCOPE}" scope.`
+      );
+    }
+  }
+  const nowIso = now.toISOString();
+  if (record.revokedAt || record.expiresAt <= nowIso) {
```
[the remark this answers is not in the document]
iscekic: bot: Accepted. Pushed bc3aeea (kwf `kwf-fix-review-a613`).

- `services/kilo-mcp/src/auth/token.ts:306` — accepted: replaying a rotated-away refresh token returned `invalid_grant` but left the thief's newer rotation live. `/token` now calls a new `store.revokeGrant` whenever the presented token is already revoked, killing every live token of that (client, user, org, resource) grant per RFC 9700 stolen-grant guidance; a merely expired token is not theft evidence and leaves the grant intact.

services/kilo-mcp/wrangler.jsonc
```diff
@@ -0,0 +1,87 @@
...
+    "WEB_BASE_URL": "https://app.kilo.ai",
+  },
+
+  // ============================================
+  // DEVELOPMENT ENVIRONMENT
+  // Deploy with: wrangler deploy --env dev
+  // This creates a SEPARATE worker: kilo-mcp-dev
+  // ============================================
+  "env": {
+    "dev": {
```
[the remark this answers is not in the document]
iscekic: bot: Accepted. Pushed bc3aeea (kwf `kwf-fix-review-a613`).

- `services/kilo-mcp/wrangler.jsonc:58` — accepted: wrangler environments do not inherit top-level `secrets`, so the dev environment skipped the required-secret enforcement that production has. `env.dev` now declares `secrets.required: ["MCP_TOKEN_SECRET"]`, keeping dev and production deployments on the same check.

services/kilo-mcp/src/index.ts
```diff
@@ -0,0 +1,443 @@
...
+      request,
+      mcpAuth
+        ? {
+            mcpToken: {
+              tokenSecret: mcpAuth.tokenSecret,
+              issuer,
+              resource: `${issuer}/mcp`,
+              isJtiRevoked: jti => mcpAuth.store.isJtiRevoked(jti),
+            },
+            resolveKiloToken: identity =>
```
pandemicsyn, line 306: bot: Credential lookup is not scoped to the verified token's organization or resource.

Suggested fix: Change the store lookup and `resolveKiloToken` contract to accept the verified identity's `organizationId` and `resource`/audience, and filter refresh-token rows on both (including correct NULL-org handling). Pass those claims from this call site. Add a test that creates two live grants for one user/client in different organizations/resources and proves a token for grant A cannot resolve grant B's Kilo credential; cover rotated-token replay revocation as well.

.github/workflows/kilo-mcp-catalog.yml
```diff
@@ -0,0 +1,358 @@
...
+            echo "::error::The catalog token mint returned no token"
+            exit 1
+          fi
+          echo "::add-mask::$token"
+          echo "KILO_API_KEY=$token" >> "$GITHUB_ENV"
+          echo "KILO_ORG_ID=$org" >> "$GITHUB_ENV"
+
+      # Fill any summaries that slipped through before the index is updated
+      # (requirement 8); keep-edit semantics leave author edits alone.
+      - name: Regenerate catalog.json
```
pandemicsyn, line 347: bot: The main job indexes a regenerated catalog but never persists it for the Worker bundle.

Suggested fix: After regeneration, either commit/push `services/kilo-mcp/catalog.json` before the Vectorize upsert and ensure the Worker deployment consumes that commit, or fail the job if regeneration changes the checked-in catalog. Alternatively, explicitly deploy the Worker from the same regenerated artifact. Add a workflow test asserting generated catalog changes cannot be indexed without being persisted/deployed.

services/kilo-mcp/src/call.ts
```diff
@@ -0,0 +1,200 @@
...
+    throw new JsonRpcFailure(
+      INVALID_PARAMS,
+      `Unknown path "${path}". The call tool only accepts paths published in the Kilo catalog — run the search tool first and call one of the paths it returns.`,
+      { path }
+    );
+  }
+
+  const schemaIsEmpty = isNoInputSchema(row.inputSchema);
+  const sendInput = input !== undefined && input !== null;
+  if (sendInput && !schemaIsEmpty) {
```
pandemicsyn, line 131: bot: A no-input catalog procedure accepts and forwards arbitrary supplied input without validation.

Suggested fix: When `isNoInputSchema(row.inputSchema)` is true, reject any non-null/non-undefined `input` with `INVALID_PARAMS` instead of skipping validation and serializing it into the upstream tRPC request. Add a test supplying an object to a no-input row and assert that `fetch` is never called.

pandemicsyn (changes requested):

services/kilo-mcp/src/index.ts
iscekic: bot: Accepted. Pushed c3cf0b81d.

- `services/kilo-mcp/src/index.ts:307` — accepted: the lookup only matched (user, client), so a valid token for one grant could forward a credential from another grant. `getKiloToken` now filters on the verified identity's `organizationId` and `resource` (null-org matches only null rows, same as `revokeGrant`); `VerifiedMcpToken` carries its bound `resource`. Added store tests proving org-A/org-B/other-resource grants resolve only their own credential, and an auth test proving an org-other token resolves nothing.

.github/workflows/kilo-mcp-catalog.yml
iscekic: bot: Accepted. Pushed c3cf0b81d.

- `.github/workflows/kilo-mcp-catalog.yml:347` — accepted: the merge job indexed a regenerated catalog but the Worker bundle still carried the committed one, so the two could diverge. Added a step between the dump and the Vectorize upsert that fails when the dump changes `services/kilo-mcp/catalog.json`, and extended `scripts/kilo-mcp-catalog.test.mjs` to assert the step exists, sits between dump and upsert, and exits non-zero (with a mutation case).

services/kilo-mcp/src/call.ts
iscekic: bot: Accepted. Pushed c3cf0b81d.

- `services/kilo-mcp/src/call.ts:131` — accepted: a no-input row skipped validation and serialized any supplied `input` into the upstream request. `callCatalogEndpoint` now throws `INVALID_PARAMS` when input is non-null for an empty schema, before any request. Test supplies an object to `organizations.list` and asserts `fetch` is never called.

services/kilo-mcp/src/auth/token.ts
```diff
@@ -0,0 +1,409 @@
...
+      resource: record.resource,
+      scope: record.scope,
+      createdAt: nowIso,
+      expiresAt: new Date(
+        now.getTime() + REFRESH_TOKEN_TTL_DAYS * 24 * 60 * 60 * 1000
+      ).toISOString(),
+    },
+    nowIso
+  );
+  if (!rotated) {
```
pandemicsyn: bot: A concurrent refresh replay can leave the newly rotated grant usable instead of revoking it.

Suggested fix: Make refresh rotation and replay handling a single atomic store operation that returns enough state to distinguish a normal rotation from a token that lost a concurrent rotation. When the latter occurs, revoke all live rows for that grant before returning `invalid_grant`. Add a test that pauses two refresh exchanges after both read the old row, lets one rotate, then verifies the losing request revokes the winner's replacement and that it can no longer resolve a Kilo credential.

services/kilo-mcp/src/auth/dcr.ts
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  const parsed = parseRegistration(body);
+  if (!parsed.ok) {
+    return oauthErrorResponse(400, parsed.error, parsed.description);
+  }
+
+  const now = deps.now?.() ?? new Date();
+  const clientId = base64UrlEncode(crypto.getRandomValues(new Uint8Array(32)));
+  await deps.store.registerClient({
```
pandemicsyn: bot: Public dynamic registration can fill the permanent single-instance OAuth client registry without any count or rate limit.

Suggested fix: Enforce a server-side registration quota/rate limit before `registerClient`, and add bounded retention or a managed-registration cleanup strategy. The OAuth client table and `purgeExpired` logic currently have no client expiry path, so per-request body/field caps do not prevent an unauthenticated caller from creating unbounded retained rows. Add a test that reaches the configured limit and receives a deterministic error without inserting another client.

services/kilo-mcp/src/index.ts
```diff
@@ -0,0 +1,447 @@
...
+  auth: ForwardedAuth,
+  deps: McpHandlerDeps
+): Promise<ToolResult> {
+  if (name === 'search') {
+    const query = args['query'];
+    if (typeof query !== 'string' || query.trim().length === 0) {
+      throw new JsonRpcFailure(INVALID_PARAMS, 'search requires a non-empty string "query".');
+    }
+    const limit = args['limit'];
+    if (limit !== undefined && (typeof limit !== 'number' || !Number.isInteger(limit))) {
```
pandemicsyn: bot: `search.limit` declares `1..50`, but the runtime accepts any integer and the search implementation only floors it to at least one.

Suggested fix: Reject values outside 1 through 50 in `runTool` (matching the published tool schema), or clamp them explicitly and document that behavior. Add tests for `0`, negative values, and values above 50; otherwise a raw MCP client can request every matching row in the catalog and receive a response larger than the advertised limit.

services/kilo-mcp/src/auth/dcr.ts
```diff
@@ -0,0 +1,256 @@
...
+      grantTypes: Array.isArray(grantTypes)
+        ? (grantTypes as string[])
+        : ['authorization_code', 'refresh_token'],
+      responseTypes: ['code'],
+      scope: typeof scope === 'string' && scope.length > 0 ? scope : MCP_SCOPE,
+    },
+  };
+}
+
+function validScopeString(scope: string): boolean {
```
pandemicsyn, line 200: bot: nit: Scope validation is duplicated with `validScope` in `authorize.ts`, but the two helpers intentionally differ for an empty string (`validScope` accepts it while this rejects it). Could we share a tokenizer/validator and make the empty-scope/defaulting policy explicit at each caller? That keeps the OAuth input contract from drifting as supported scopes grow.

services/kilo-mcp/src/index.ts
```diff
@@ -0,0 +1,447 @@
...
+      case 'ping':
+        return jsonRpcResult(id ?? 0, {});
+      case 'tools/list':
+        return jsonRpcResult(id ?? 0, { tools: TOOLS });
+      case 'tools/call': {
+        const params = (message.params ?? {}) as { name?: unknown; arguments?: unknown };
+        if (typeof params.name !== 'string') {
+          throw new JsonRpcFailure(INVALID_PARAMS, 'tools/call requires a string "name".');
+        }
+        const args = (params.arguments ?? {}) as Record<string, unknown>;
```
pandemicsyn: bot: nit: Could we validate `params`/`arguments` as records before using them instead of asserting their types? `tools/call` is an external JSON-RPC boundary, and `(params.arguments ?? {}) as Record<string, unknown>` accepts arrays and other non-object JSON values at runtime. A small shared object guard (or MCP envelope schema) would make the contract explicit and remove these unchecked casts.

services/kilo-mcp/src/call.ts
```diff
@@ -0,0 +1,207 @@
...
+    body = await response.json();
+  } catch {
+    body = null;
+  }
+
+  if (!response.ok) {
+    throw toTrpcFailure(response.status, body as TrpcErrorBody, path);
+  }
+
+  const result = (body as { result?: { data?: unknown } } | null)?.result;
```
pandemicsyn: bot: nit: Consider extracting this upstream tRPC response check into a named type guard rather than casting `body` inline. It would document the untrusted-response boundary and keep the shape validation reusable/consistent with the deliberate AJV validation on the request side.

services/kilo-mcp/scripts/embed-catalog.ts
```diff
@@ -0,0 +1,280 @@
...
+    VECTORIZE_INDEX_NAME: read('VECTORIZE_INDEX_NAME'),
+  };
+}
+
+/** Read the committed catalog.json that sits next to this script's package. */
+export function loadCatalog(): Catalog {
+  // import.meta.url is ALREADY a file URL: decode it with fileURLToPath —
+  // pathToFileURL(import.meta.url) double-encodes and yields an ENOENT path.
+  const catalogPath = join(dirname(fileURLToPath(import.meta.url)), '..', 'catalog.json');
+  return JSON.parse(readFileSync(catalogPath, 'utf8')) as Catalog;
```
pandemicsyn: bot: nit: `catalog.json` is parsed and asserted as `Catalog` without any runtime check. If this generated artifact is intentionally trusted, a named loader/guard (or an explicit comment documenting that build-time trust boundary) would make that assumption clearer and avoid a misleading type assertion.

pandemicsyn (approved):
lgmt

services/kilo-mcp/src/auth/token.ts
iscekic: Fixed in beb392ac3. `rotateRefreshToken` now returns `rotated | replayed | missing`; on `replayed` it revokes the whole grant (including the winner replacement) in the same atomic store call. The new test `a refresh that loses a concurrent rotation revokes the winner and its credential` parks two exchanges after both read the old row, lets one rotate, then asserts the loser revokes the winner and `getKiloToken` returns null.

services/kilo-mcp/src/auth/dcr.ts
iscekic: Fixed in beb392ac3. `registerClient` enforces `MAX_REGISTERED_CLIENTS` (10,000) atomically and returns false; the handler answers 429 `temporarily_unavailable`. `purgeExpired` also drops clients older than 90 days that have no live grant, so the registry self-heals. Tests cover reaching the cap without inserting and keeping a client with a live grant.

services/kilo-mcp/src/index.ts
iscekic: Fixed in beb392ac3. `runTool` now rejects any `limit` outside 1..50, and the tool schema and the check share `MAX_SEARCH_LIMIT` so they cannot drift. Tests cover 0, negative, 51, 1.5, NaN, and the 1/50 boundaries.

services/kilo-mcp/src/auth/dcr.ts
iscekic: Fixed in beb392ac3. Added shared `scopeTokens` / `onlyMcpScope` / `normalizeScope` in `auth/http.ts`. DCR keeps its explicit `tokens.length > 0` rejection; authorize passes the tokens straight through, so the empty-scope/defaulting policy stays explicit at each caller.

services/kilo-mcp/src/index.ts
iscekic: Fixed in beb392ac3. Added an `isRecord` guard at the JSON-RPC edge: `tools/call` requires a params object and rejects non-record `arguments` with -32602, and `initialize` tolerates missing params. Tests cover arrays, numbers, strings, and missing params.

services/kilo-mcp/src/call.ts
iscekic: Fixed in beb392ac3. Extracted a named `isTrpcSuccessBody` type guard; the success body is now narrowed instead of cast inline, matching the deliberate AJV validation on the request side.

services/kilo-mcp/scripts/embed-catalog.ts
iscekic: Fixed in beb392ac3. Added a validating `parseCatalog` that checks each row shape and that the key matches `path`; `loadCatalog` uses it and fails loudly on artifact drift. A `parseCatalog` test covers the malformed cases.

PR #6032, opened 2026-09-10
feat(mobile): transcribe voice input through the Kilo gateway

iscekic:
## Changelog for users

- Voice input gains a second engine through the Kilo gateway; a Gateway transcription switch in Preferences is off by default, so voice input behaves exactly as today and never calls the gateway.
- The switch is a two-way choice, not a fallback: off runs the operating system's speech recogniser, on runs the Kilo gateway. The chosen engine owns the whole dictation and the other engine is never called.
- The mode applies to every voice dictation in the app.
- With the switch on, the gateway transcribes with your chosen model; if you have not chosen one, the first model the gateway offers is the default.
- A picker lists the gateway's actual models, and the chosen model persists across launches.
- An unreachable gateway, an unavailable model, no speech, and a timeout each get their own message; an unavailable model and no model open the transcription settings.
- A gateway dictation that times out shows the retryable timeout instead of an unexpected-response error, keeps the sealed recording, and shows the retry message.
- While the gateway transcribes, the composer shows Transcribing... and the microphone button cancels the upload; a transcript that lands right after you tap stop fills the composer.
- Multipart voice transcriptions enforce the organization's provider allow-list and data-collection policy.
- Toasts now appear at the bottom of the screen, and tab-bar navigation away from an unsent draft no longer triggers the discard prompt.

## Changelog for maintainers

- `apps/mobile/src/lib/voice-input/voice-input-engine-select.ts` — the selector composes the OS recogniser and the gateway engine behind the single `VoiceInputNative` the controller consumes. It reads the chosen engine at `start()`, routes start/stop/abort and every capability probe to that one engine, and drops events from the idle engine. No engine ever hands a session to the other. Review this file first.
- `apps/mobile/src/lib/voice-input/voice-input-engine-mode.ts` — `resolveVoiceInputEngineName(enabled)` maps the switch to `os` or `gateway`. The former `gateway-primary`/`device-primary` modes and the `VoiceInputEngineMode` union are gone.
- `apps/mobile/src/lib/voice-input/native-voice-input.ts` — `createSelectingVoiceInputNative` replaces the two-engine dispatcher; the gateway-enabled subscription still refreshes availability so enabling the switch surfaces the mic without a restart.
- `apps/mobile/src/lib/voice-input/gateway/gateway-transcription-preference.ts` — two SecureStore keys remain (enabled default off, chosen model). The `gateway-transcription-primary` key and its hooks are removed.
- `apps/mobile/src/lib/voice-input/gateway/native-gateway-voice-input.ts:22` — `resolveGatewayTranscriptionModelId()` returns the stored model, else the first catalogue entry, else null. An unreachable or empty catalogue reads as no model, which surfaces the actionable picker message instead of an upload that must fail. The engine's `readModelId` dependency is now async.
- `apps/mobile/src/lib/voice-input/use-voice-input-actions.ts` — the start path branches once: switch on starts the gateway engine with no OS consent disclosure (the switch is the consent); switch off runs the existing OS consent flow. The model precondition and the picker alert on start are gone because the engine now resolves the default. `VOICE_INPUT_TOAST_ID` is a module-private constant shared by every voice-input toast, so sonner replaces messages in place; check:unused reports no unused exports, no checker ignore.
- Removed with the fallback: the `engine-fell-back` controller event and its listener, `classifyVoiceInputEngineFallback`, the `voice-engines-failed` error case and its copy, the `tone: 'info'` feedback field, the Preferences "Use as primary" row, and the `voice-input-engine-dispatch*` module plus its five suites. New coverage is `voice-input-engine-select.test.ts`.
- `apps/mobile/src/lib/voice-input/gateway/gateway-transcription-client.ts:133` — a cancelled upload that resolves with no response is classified by checking the timeout signal first: a fired timeout returns the retryable timeout, while a caller abort stays an invalid response.
- `apps/web/src/app/api/openrouter/audio/transcriptions/route.ts:278` — the resolved organization provider policy is computed once and appended to the multipart form as a JSON `provider` field (the JSON branch merges it as before); multipart requests also carry `safety_identifier` and `user`, both the provider-specific hash.
- Wire contract: recordings upload as multipart with a 30 s timeout; responses validate as `{ text: string }`; failures classify into distinct codes that map to one copy each. The gateway route accepts multipart uploads and exempts free models from the zero-balance gate; the catalogue fetch is uncached; the feature value is `mobile-voice-input`. The fake gateway keeps exactly its two committed transcription models and a request counter; no fixture, test hook, or test-only flag is committed. The feature-flags mounted test mocks the gateway transcription preference module, whose import-time Sentry chain cannot resolve under vitest.

## Owner manual verification

- The acoustic microphone-to-transcript checks are skipped, pending owner verification on a real device. No sound was generated or played for microphone pickup. No spoken-utterance transcription through either engine was automated or asserted.

## E2E proof

The device captures for the Preferences rows, persistence across launch, the fake-gateway round trip, and the switch-off no-call check remain valid. The prior "use as primary" and fallback hand-off captures are superseded by this change and are not part of the current proof.

![[e5] gateway off makes no fake-gateway call — e2e-mobile-app/e5-start.png](https://github.com/user-attachments/assets/4b29e97e-1de9-4c56-8e62-016013b9885f)

<details>

> Support transcribing voice input through the Kilo gateway as well as the operating system's speech recognition.
>
> Today voice input is on-device only. The gateway adds a second engine.
>
> Add a setting with two parts:
> 1. A switch that enables gateway transcription. It is OFF by default.
> 2. A model picker for the transcription model.
>
> The switch picks the engine for every dictation in the app: off keeps the operating system's recogniser, on sends the recording to the Kilo gateway. There is no fallback between the two engines — the chosen engine owns the whole dictation.
>
> When gateway transcription is off, behaviour is exactly what it is today: on-device only, and no gateway call is made.
>
> If the user has not chosen a model, the first model the gateway offers is the default.
>
> Cover the failure cases the user can hit: the gateway is unreachable, the request times out, the chosen model is unavailable, and the recording holds no speech. Each gets its own message and its own remedy.
>
> The switch and the chosen model must persist across launches, and the model picker must list the models the gateway actually offers, not a hardcoded list.
>
> Every verification and publication gate must pass. The acoustic microphone-to-transcript conversion — a real utterance becoming a transcript through either engine — awaits owner manual verification by Igor on a real device: automated proof never generates or plays sound for microphone pickup, never depends on the acoustic path, and never uses simulator speech recognition diagnosis that serves only that proof; no acoustic scenario blocks the section and no repair task chases an acoustic failure; the PR body states that the acoustic microphone-to-transcript check awaits owner manual verification and does not claim it passed.
>
> Everything else keeps its automated proof and runs as normal: the recording → upload → response round trip against the fake gateway, whose fixed transcript needs no spoken audio; the failure UX messages; the switch-off path, which starts and ends a device session and makes no gateway call, asserted on the fake server's request counter; and the unit suites covering the engine selector, the request timeout, and persistence.
>
> The no-speech state is proved by the same automated device run, but its empty-transcript engine is temporary harness support that lives outside the product diff, per the owner rule of 2026-09-08: no fixture, E2E-only code, test hook, or test-only runtime flag for it is committed to the product repo, and there is no owner permission for a committed fake-transcribe-silent model. The harness seeds the empty-transcript behaviour at run time in front of the delivered fake gateway, which keeps exactly its two committed transcription models (the fixed transcript and the broken one); ordinary unit tests are unaffected.
>
> The mobile static gate is part of the pass set: check:unused (knip) must report no unused exports, repaired at the source with runtime behavior preserved, never suppressed through a checker ignore comment or config exemption.

</details>

## E2E proof — log excerpts

```
[p1] Dictate a voice message while the gateway stalls past the 30s timeout: mic  -> pass :: owner-manual: acoustic spoken-utterance -> transcript conversion through either engine (not exercised; no sound generated); the automatic timeout classification passed live — p1-timeout.log carries 'fault.sh: stalled pids [59771] for 120s' then 'SCENE p1 OK', i.e. the final assert 'Transcription took too long. Try again.' (voiceInput.gatewayTimeout) rendered, not voiceInput.stopped; supporting focused tests in p1-focused-tests.log (26 passed).
[p2] Enterprise org member multipart transcription through /api/gateway/audio/tr -> pass :: p2-upstream-fields.log carries 'gateway response: HTTP 200 {"text":"Gateway transcription online"}', 'upstream provider = {"only":["openai"],"data_collection":"deny"}' and 'upstream safety_identifier = ...' equal to 'upstream user'; focused route tests 'forwards the organization provider policy on multipart requests' and 'attaches the safety identifier to multipart upstream requests' pass in p2-focused-tests.log.
```

<details><summary><code>.../e2e-mobile-app/p1-timeout.log</code></summary>

```
fault.sh: stalled pids [59771] for 120s
SCENE p1 OK
scenario steps (file p1.json, packed tree HEAD f202d33bc):
  ["tap Start voice input","assert Listening... (10s)","tap Stop voice input","assert Transcribing... (10s)","assert Transcription took too long. Try again. (50s)"]
setup while the composer is open: fault.sh <worktree> stall nextjs --hold 120 (nextjs suspended; port 3300 bound but not answering).
observed: "SCENE p1 OK" -> every assert passed, including the final one, so the mic sheet rendered the timeout copy.
expected state: retryable timeout. voiceInput.gatewayTimeout = "Transcription took too long. Try again."
contrast: gateway-invalid-response maps to voiceInput.stopped = "Voice input stopped. Tap the microphone to try again." (not observed).
owner-manual (not exercised): the acoustic spoken-utterance -> transcript conversion through either engine; no sound was generated or played.
```
</details>

<details><summary><code>.../e2e-mobile-app/p2-upstream-fields.log</code></summary>

```
p2 live check: enterprise org member multipart transcription through /api/gateway/audio/transcriptions
platform: ios (device 2E687210-E477-4D23-94F1-9736C860ACD8); request issued against the local stack (nextjs :3300)
org: 0605a00e-dfb4-4361-946d-b81cca2e02bd plan=enterprise provider_allow_list=["openai"] data_collection=deny
user: fe46cbb5-c197-42c4-ae0f-02a1dd8fc05c (e2e-mobile-kwf-fix-review-ae2c-ios@example.com, role=member)
gateway response: HTTP 200 {"text":"Gateway transcription online"}
upstream request: POST /api/openrouter/audio/transcriptions multipart
upstream model = fake-transcribe
upstream provider = {"only":["openai"],"data_collection":"deny"}
upstream safety_identifier = QJyylBTUbBUKZ2cAKnCpqZAXmtkmZP741isDYxuRLGg=
upstream user = QJyylBTUbBUKZ2cAKnCpqZAXmtkmZP741isDYxuRLGg=
```
</details>


apps/mobile/src/lib/voice-input/gateway/gateway-transcription-client.ts
```diff
@@ -0,0 +1,197 @@
...
+  try {
+    const response = await task.uploadAsync();
+
+    if (!response || response.status < 200 || response.status >= 300) {
+      // A cancelled task resolves without a response; a gateway refusal
+      // carries its status.
+      return {
+        ok: false,
+        status: response?.status,
+        isTimeout: false,
```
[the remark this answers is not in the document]
iscekic: bot: Accepted. Pushed f202d33 (kwf `kwf-fix-review-ae2c`).

- `apps/mobile/src/lib/voice-input/gateway/gateway-transcription-client.ts:133` — accepted. A cancelled upload that resolves with no response was classified as an invalid response. The timeout signal is now checked first: a fired timeout returns the retryable timeout, while a caller abort stays an invalid response. A focused test covers the resolve-without-response cancel.

apps/web/src/app/api/openrouter/audio/transcriptions/route.ts
```diff
@@ -199,12 +275,12 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
         ? currentOnly.filter(route => decision.eligibleProviderRoutes?.has(route))
         : [...decision.eligibleProviderRoutes];
       if (only.length === 0) return modelNotAllowedResponse();
-      body.provider = { ...body.provider, ...providerConfig, only };
+      if (jsonBody) jsonBody.provider = { ...jsonBody.provider, ...providerConfig, only };
```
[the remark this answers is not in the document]
iscekic: bot: Accepted. Pushed f202d33 (kwf `kwf-fix-review-ae2c`).

- `apps/web/src/app/api/openrouter/audio/transcriptions/route.ts:278` — accepted. The resolved organization provider policy was written only into a JSON body, so multipart requests dropped it. The policy is now computed once and appended to the multipart form as a JSON `provider` field; the JSON branch merges it as before. A focused test covers multipart policy forwarding.

apps/web/src/app/api/openrouter/audio/transcriptions/route.ts
```diff
@@ -213,9 +289,19 @@ export async function POST(request: NextRequest): Promise<NextResponseType<unkno
   );
 
   const span = startInactiveSpan({ name: 'transcription-request-start', op: 'http.client' });
-  body.safety_identifier = generateProviderSpecificHash(user.id, provider);
-  body.user = body.safety_identifier;
-  const upstreamBody = buildUpstreamBody(body);
+
+  let upstreamBody: Record<string, unknown> | FormData;
+  if (parsedRequest.kind === 'multipart') {
```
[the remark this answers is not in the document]
iscekic: bot: Accepted. Pushed f202d33 (kwf `kwf-fix-review-ae2c`).

- `apps/web/src/app/api/openrouter/audio/transcriptions/route.ts:294` — accepted. Multipart upstream requests omitted `safety_identifier`. The form now carries `safety_identifier` and `user`, both the provider-specific hash. A focused test covers both fields on multipart.

apps/mobile/src/lib/voice-input/gateway/gateway-voice-input-engine.ts
```diff
@@ -0,0 +1,360 @@
...
+      await recorder.stop();
+    } catch {
+      releaseRecorder(handle);
+      if (!stale(current)) {
+        fail(current, 'client');
+      }
+      return;
+    }
+    const uri = recorder.uri;
+    releaseRecorder(handle);
```
pandemicsyn, line 250: bot: Gateway dictation recordings are released but never deleted from disk.

Suggested fix: Inject a best-effort recording-file deletion dependency (for example, `expo-file-system` `deleteAsync(uri, { idempotent: true })`) and run it once the URI is known on every terminal path: successful upload, classified upload failure, model/auth short-circuit, and cancellation during upload. Keep native-recorder `release()` for the object lifecycle, but do not treat it as file cleanup. Add tests asserting cleanup after success, failure, and abort.

apps/mobile/src/lib/voice-input/gateway/native-gateway-voice-input.ts
```diff
@@ -0,0 +1,107 @@
...
+ * cannot be reached reads as "no model", which surfaces the actionable picker
+ * message on the first dictation instead of an upload that must fail.
+ */
+export async function resolveGatewayTranscriptionModelId(): Promise<GatewayTranscriptionModel | null> {
+  const stored = readGatewayTranscriptionModel();
+  if (stored !== null) {
+    return stored;
+  }
+  try {
+    const models = await fetchTranscriptionModels();
```
pandemicsyn: bot: The default gateway model is fetched without the selected organization scope, so it can choose a model the subsequent scoped upload rejects.

Suggested fix: Read/pass the current organization ID into `fetchTranscriptionModels` when resolving the default model, and also pass that ID from `TranscriptionModelPickerSheet` into `useTranscriptionModels`. The hook already supports an organization argument and emits the corresponding header. Add coverage where the first global catalog entry is unavailable to the selected organization and verify both the picker and automatic default use an allowed model.

apps/web/src/app/api/openrouter/audio/transcriptions/route.ts
```diff
@@ -58,19 +66,54 @@ async function transcriptionProxyRequest(params: {
...
+ * content-type header, before anything reads the body.
+ */
+type ParsedTranscriptionRequest =
+  | { kind: 'json'; body: TranscriptionRequest }
+  | { kind: 'multipart'; file: File; model: string; language: string | null };
 
+async function parseMultipartTranscriptionRequest(
+  request: NextRequest
+): Promise<ParsedTranscriptionRequest | null> {
+  const formData = await request.formData();
```
pandemicsyn: bot: A malformed multipart body can make `request.formData()` reject instead of returning the route's controlled 400 invalid-request response.

Suggested fix: Wrap `await request.formData()` in `try/catch`, optionally capture the parse error without logging audio content, and return `null` so `POST` reaches `invalidRequestResponse()`. Add a test with `Content-Type: multipart/form-data` and an invalid or missing boundary to assert a 400 and no upstream call.

pandemicsyn (approved):

apps/mobile/src/lib/voice-input/gateway/gateway-voice-input-engine.ts
iscekic: bot: Accepted. Fixed in 3074f803b.

`apps/mobile/src/lib/voice-input/gateway/gateway-voice-input-engine.ts` — accepted. Added a required `deleteRecording(uri)` dependency and run it once the URI is known on every terminal path: upload success, classified failure, model/auth short-circuit, stale session, and abort during recording or upload. `release()` still frees the native object; a new `expo-file-system/legacy` `deleteAsync(uri, { idempotent: true })` in the binding deletes the file. Tests cover cleanup after success, failure, model short-circuit, and both aborts.

apps/mobile/src/lib/voice-input/gateway/native-gateway-voice-input.ts
iscekic: bot: Accepted. Fixed in af632cc45.

`apps/mobile/src/lib/voice-input/gateway/native-gateway-voice-input.ts` — accepted. `resolveGatewayTranscriptionModelId` now reads the stored organization id and passes it to `fetchTranscriptionModels(organizationId ?? undefined)`. `apps/mobile/src/components/transcription-model-picker-sheet.tsx` now passes `useOrganization().organizationId` into `useTranscriptionModels`. Tests assert the catalogue read forwards the org id and stays unscoped for a personal account.

apps/web/src/app/api/openrouter/audio/transcriptions/route.ts
iscekic: bot: Accepted. Fixed in 7a60c112f.

`apps/web/src/app/api/openrouter/audio/transcriptions/route.ts` — accepted. `parseMultipartTranscriptionRequest` now wraps `await request.formData()` in `try/catch`, captures the parse error without logging audio content, and returns null so `POST` reaches `invalidRequestResponse()` (400) instead of an unhandled 500. A test sends `Content-Type: multipart/form-data` with an unparseable body and asserts a 400 and no upstream call.

PR #6033, opened 2026-09-10
fix(code-review): pin a same-vendor small model for BYOK reviews

maphew:
Refs #4268.

## Problem

A user set a BYOK model for Code Reviewer. The review itself used that model.
But Kilo credits still went down. The Kilo CLI used a separate small model for
title and auxiliary calls. The CLI fell through to `kilo-auto/small`. That
resolves to `google/gemma-4-26b-a4b-it`. The Gemma call goes through Kilo, not
the user provider.

## Change

The web dispatcher now picks a cheaper model from the same vendor as the
effective review model. It uses the OpenRouter catalog and prompt prices.

- A managed vendor uses the cheapest strictly-cheaper sibling.
- A direct-BYOK vendor uses the primary BYOK model. The primary is used when
  the catalog has no cheaper BYOK sibling. This keeps the call on the user key.
- A managed vendor with no cheaper sibling leaves `small_model` unset. The CLI
  keeps its current default in that case.

The value travels as `smallModel` from `apps/web` to the sandbox session
config. It is stored in the v2 session metadata. It is part of the create
intent fingerprint. The sandbox `KILO_CONFIG_CONTENT` gets `small_model` and
`agent.title.model`. This applies to code-review sessions only. Bitbucket
sessions do not get `agent.title`.

## Relation to other work

- Closes no other item. Refs #4268 only.
- The design follows the maintainer review on PR #4725. That PR was closed for
  inactivity. This PR takes over that design.
- PR #5774 changed gastown's `areThreadsBlocking()` to use a BYOK model. That
  path uses Cloudflare Workers AI, not the Kilo gateway. It does not draw Kilo
  credits. This PR does not change gastown. The hardcoded Gemma id in gastown
  stays as a separate, smaller follow-up.
- No open PR conflicts with this change.

## Adversarial review

An independent review agent examined the diff against main. It found one
critical defect in the first revision. The Durable Object rebuilt
`metadata.agent` from a fixed field list. It dropped `smallModel`, so the fix
was inert. The fix now writes `smallModel` in `registerSession` and in the
admitted-plan agent. A new Durable Object integration test asserts the
persisted value. A second review pass confirmed that all findings are closed.

## Verification

Passed:

- `pnpm run typecheck` in `apps/web`, `services/cloud-agent-next`,
  `services/code-review-infra`, and `packages/worker-utils`.
- `oxlint` and `oxfmt --list-different` on all changed files.
- `vitest run` in `services/cloud-agent-next` for `session-service.test.ts`,
  `session/session-prepare.test.ts`, `persistence/session-metadata.test.ts`,
  and `router/handlers/session-worktree.test.ts`: 453 tests pass.
- `vitest run --config vitest.workers.config.ts` for
  `test/integration/session/code-review-small-model.test.ts` and
  `test/integration/session/start-execution-v2.test.ts`: 41 tests pass.

Not run locally:

- `apps/web` jest tests (`model-selection.test.ts`,
  `prepare-review-payload.test.ts`). The jest global setup needs a live
  PostgreSQL test database. The sandbox has no Docker and no database. CI
  covers these tests.


eshurakov:
Hey @maphew 

I think the small-model choice belongs in the CLI as it already resolves the small model per request in `Provider.getSmallModel` ([provider.ts#L1989](https://github.com/Kilo-Org/kilocode/blob/a6589426458fcbe9f30c2771b36508f3ae3deeed/packages/opencode/src/provider/provider.ts#L1989)): explicit `small_model` config, a plugin hook, `kilo-*` providers, then a family match (`gemini-flash`/`gpt-nano`/`claude-haiku`, [L2113](https://github.com/Kilo-Org/kilocode/blob/a6589426458fcbe9f30c2771b36508f3ae3deeed/packages/opencode/src/provider/provider.ts#L2113)). The last step is an unconditional fallback to `kilo-auto/small` for any non-Azure provider ([L2065](https://github.com/Kilo-Org/kilocode/blob/a6589426458fcbe9f30c2771b36508f3ae3deeed/packages/opencode/src/provider/provider.ts#L2065-L2068)) with no check that the active provider is Kilo-billed. That fallback is what pulls the credits, and titles, prompt enhance, commit messages, branch names and project copy all hit it through `getSmallModel(pid) ?? getModel(pid, modelId)`.

CLI should have access to the model list as well so that we can pick a cheaper same-vendor model by price, could you take a look if we can fix it there for everyone?

maphew:
Good question — I agree the unconditional kilo-auto/small fallback in getSmallModel is the real root cause, and fixing it in the CLI would cover titles, prompt-enhance, commit messages and branch names for everyone, not just code review. This PR stays web-side deliberately as the narrow fix for #4268: the web dispatcher already knows the effective review model and the vendor catalog at dispatch time, so it can pin small_model per review session today with no CLI release coordination and no behavior change to the other aux-call paths. A CLI-side fix (vendor-aware fallback, or at least a Kilo-billed check before falling through to kilo-auto/small) is the right follow-up and would subsume this — happy to open that issue against Kilo-Org/kilocode and link it here, or close this in favor of it if you prefer the CLI fix first. Which do you want?

eshurakov:
@maphew because something similar already exists in the CLI, I think it should be there to avoid spreading similar logic in different places. It may be useful in the end for more than just a code reviewer. 
My preference is to open a PR against the CLI with the minimal fix.
Thanks!

maphew:
@eshurakov Agreed — done: the CLI-side fix is up as Kilo-Org/kilocode#14263. `Provider.getSmallModel` now picks the cheapest chat-capable model from the active provider's own catalog (models.dev prices, ties broken by release date/id) after the family match, so titles, prompt enhance, commit messages, branch names and project copy stay on the user's key; the creds-gated `kilo-auto/small` fallback only remains for providers with no chat-capable model at all. Family matching and explicit `small_model` config behave as before. Happy to close this PR in favor of the CLI one once that approach looks good to you.

PR #6042, opened 2026-09-10
Remove stale Vercel model ID mappings

chrarnoldus:
## Summary
- compare Vercel model mapping keys against the current OpenRouter text and embedding model catalogs
- remove 64 mappings whose keys are not exact OpenRouter model IDs
- retain all 26 exact matches, including the 12 latest-model aliases and 2 embedding models
- align the hardcoded mapping test cases with the retained entries

## Testing
- Not run locally; CI will handle validation as requested.
- `git diff --check`
- formatted the changed files with `oxfmt`


chrarnoldus:
Addressed both review findings in bded5c672:

- Restored `mistralai/mistral-embed-2312` and `mistralai/codestral-embed-2505`; both are exact IDs returned by the embeddings catalog.
- Removed the deleted mappings from `mapModelIdToVercel.test.ts` so its table matches the implementation.

There is no inline review thread to resolve; the findings were posted in the PR-level review summary.

PR #6055, opened 2026-09-10
fix(web): sanitize Gemini JSON ref tool results

chrarnoldus:
## Summary
- extract `sanitizeJsonRefToolResults` into a shared provider utility
- sanitize JSON `$ref` fields for Gemini tool results in provider-specific request handling
- retain opt-in sanitization for direct custom and experiment providers
- cover Gemini and non-Gemini provider-specific behavior

## Testing
- Not run locally; CI will validate the change as requested.
- `oxfmt --list-different` passed for all touched files.
- `git diff --check` passed.

### DIFF
--- a/apps/web/src/lib/ai-gateway/experiments/build-direct-provider.ts
+++ b/apps/web/src/lib/ai-gateway/experiments/build-direct-provider.ts
@@ -1,10 +1,7 @@
 import { addCacheBreakpoints } from '@/lib/ai-gateway/providers/openrouter/request-helpers';
 import type { CustomLlmApiConfig } from '@kilocode/db';
-import {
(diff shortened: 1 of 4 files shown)


chrarnoldus:
Addressed the review warning in 1c93d14e5. The provider-specific tests now use a valid assistant tool-call followed by its matching tool result, and the assertion locates the retained tool message after repair.

Validation:
- `NODE_ENV=test pnpm --filter web exec jest --runInBand src/lib/ai-gateway/providers/apply-provider-specific-logic.test.ts` (38 passed)
- `scripts/typecheck-all.sh` (passed)

jeanduplessis (approved):

PR #6064, opened 2026-09-10
fix(mobile): drop the broken-install DSO crash from native Sentry

iscekic:
## What

Stop the native Sentry SDK reporting the broken-install crash KILO-APP-53.

## Why

KILO-APP-53 reports 145 events from 25 users:

```text
com.facebook.soloader.SoLoaderDSONotFoundError: couldn't find DSO to load: libreactnative.so
```

The crash comes from `MainApplication.onCreate` before JS starts.

The released AAB ships `libreactnative.so` for all four ABIs. I verified this against kilo-app build 34481127753. The crash only happens when the app is side-loaded from split APKs whose ABI split is missing or does not match the device:

- The user installed `base.apk` alone. An AAB keeps native libraries in the ABI split, not in the base.
- The user installed the arm64 split on an x86_64 device.

No code can load a library that is not installed. The crash is an install-integrity failure, not an app defect.

## How

The native SDK captures the crash before JS loads. The JS `beforeSend` never sees it, and the native-init path does not read `ignoreErrors` from `sentry.options.json`. The only hook is the `OptionsConfiguration` passed to `RNSentrySDK.init`.

The new config plugin replaces `RNSentrySDK.init(this)` with a call that adds a narrow ignored-error filter for the `libreactnative.so` DSO message. It registers before `@sentry/react-native/expo`, because config-plugin mods run in reverse registration order, so its mod runs after Sentry has inserted `RNSentrySDK.init`.

The plugin throws if it cannot find the Sentry call, so a Sentry upgrade cannot silently drop the filter.

## Verification

- `expo prebuild --platform android` produced the intended `MainApplication.kt` (filter present, import intact).
- `pnpm --filter kilo-app run test`: 700 files, 9513 tests pass, plus 4 new plugin tests.
- `pnpm --filter kilo-app run typecheck`, `lint`, `check:unused`, `format:check`: pass.

Fixes KILO-APP-53

### DIFF
--- a/apps/mobile/app.config.ts
+++ b/apps/mobile/app.config.ts
@@ -205,6 +205,11 @@ const config: ExpoConfig = {
       },
     ],
     'expo-web-browser',
+    // Registered before @sentry/react-native/expo on purpose: config-plugin
+    // mods run in reverse registration order, so this one runs after Sentry has
+    // inserted RNSentrySDK.init(this) into MainApplication. It then adds the
+    // ignored-error filter for the broken-install DSO crash (KILO-APP-53).
+    './plugins/withAndroidSentryIgnoredInstallErrors',
     [
       '@sentry/react-native/expo',
       {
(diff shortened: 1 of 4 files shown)


iscekic:
Closing in favor of a Sentry ignore. The crash is an install-integrity failure (side-loaded split APKs missing the `libreactnative.so` ABI split), not an app defect. A Sentry ignore stops the noise for every shipped version immediately; the code filter would only apply after the next release. Branch kept for reference.

PR #6076, opened 2026-09-11
Refactor AI gateway provider definitions

chrarnoldus:
## Summary
- move each static AI gateway provider definition into `providers/definitions`
- update consumers to import provider definitions directly without re-exports
- isolate provider lookup in `try-get-provider-by-id.ts`

## Testing
- Not run locally; CI will handle validation as requested.
- `git diff --check`


chrarnoldus:
Updated all stale provider-definition paths in `ENVIRONMENT.md` to their exact leaf modules in 4b8b18cba.

pandemicsyn (approved):

PR #6088, opened 2026-09-11
Remove model experiments from AI gateway and admin UI

chrarnoldus:
## Summary
- remove model-experiment membership, bucketing, direct routing, prompt capture, and usage attribution from the AI Gateway
- stop publishing experiment models and remove their automatic-routing special cases
- remove the Model Experiments and Experiment Requests admin tabs, legacy route, client hooks, prompt download endpoint, and dead R2 prompt-read plumbing
- retain historical database records and admin tRPC APIs while relocating the shared direct-provider helper out of the experiment namespace

## Validation
- `pnpm format:changed`
- `git diff --check`
- local tests intentionally skipped; CI will validate the change


johnnyeric (approved):

johnnyeric:
| Before | After |
| --- | --- |
| Screenshot 2026-09-23 at 18 45 33 |  Screenshot 2026-09-23 at 18 45 45 |

johnnyeric:
Deployed the PR locally and verified. On the UI, the tabs are gone. On the backend, since there is no existent experiments in prod database the removed logic won't be hit and I found no major side effects.
Only one scenario might be worth checking, that a couple models defined with price 0 would show up in a model picker because the check for experiment was removed. I only validated the response locally, I need to check how to verify in prod too.

```
Removing the experiment picker rule also unhides any zero-price model that is not marked free. In the current OpenRouter catalog, that is only google/lyria-3-pro-preview and google/lyria-3-clip-preview. Both output audio as well as text. They can now appear in the Efficient pool picker, where the old rule kept them out. Saving one can still fail if it does not serve every chat API.
```

The picker is defined at:
```
The picker is isEligiblePoolModel in apps/web/src/components/auto-routing/AutoRoutingModeCard.tsx. It is rendered on /profile and /organizations/[id]/providers-and-models. The same filter is also used by the admin benchmarks section.
```

chrarnoldus:
> Removing the experiment picker rule also unhides any zero-price model that is not marked free. In the current OpenRouter catalog, that is only google/lyria-3-pro-preview and google/lyria-3-clip-preview. Both output audio as well as text. They can now appear in the Efficient pool picker, where the old rule kept them out. Saving one can still fail if it does not serve every chat API.

I don't think this is a problem, if it serves text it should work.

PR #6134, opened 2026-09-14
fix(mobile): correct protocol-qualified +new send in react-native patch

iscekic:
## Changelog for users

- The iOS app compiles again: the react-native patch no longer sends `new` to a protocol-qualified class.
- A Fabric component whose handle has no registered view still mounts the unimplemented fallback view instead of crashing.

## Changelog for maintainers

- `RCTComponentViewClassDescriptor.viewClass` is typed `Class<RCTComponentViewProtocol>`, and that protocol declares no `+new`, so the fallback now copies it into a plain `Class` before sending `new`, matching the registered lookup.
- The patch keeps the `_mutex` registration lock, the `_registerComponentViewClassLocked` helper, the missing-handle `RCTUnimplementedViewComponentView` fallback, and the `@synchronized` LegacyViewManagerInterop cache.
- Review first the fallback hunk of `RCTComponentViewFactory.mm` in `patches/react-native@0.86.3.patch`; the only changed statement is the untyped `Class` copy.
- New assertions in `dev/local/react-native-mounting-crash-patch.test.ts` pin the protocol, the descriptor type, the installed copy, and the absence of a protocol-qualified `new` send in both the patch and the installed source.
- `pnpm-lock.yaml` rehashes react-native 0.86.3 to `5b7121a2…` and rewrites every dependent entry; confirm the lockfile diff carries only that hash substitution.
- Run `pnpm install` after pulling so the patched react-native is re-applied and iOS keeps building React Native core from source.

## E2E proof — log excerpts

Proved live on iOS: both must-run scenarios ran on an iOS simulator against the VM dev stack, and both passed.

Ran on ios; the request asked for ios.

```
s1-cold-start-session-list -> pass :: Cold start (stop and relaunch) reached the Agents session list with the Search sessions field and a CLOUD session row (line 68), and crash-sweep.log has no crash line.
s2-remote-session-mount-unmount -> pass :: Remote CLI session ses_f5a736e52ffeQ69FWEO07pL39a streamed a reply (line 11), Go back returned to the list (lines 23, 32), and the second session ses_f5a71732effexcE6HhWxsMattC repeated the cycle (s2b-scene.log line 1).
```


eshurakov (approved):

iscekic:
(bot) ## Live proof (proof-only request; no code change)

Ran on ios; the request asked for ios.

This request asked for live proof of existing behaviour. There is no diff; the verify round ran the must-run scenarios and recorded the evidence below.

- cli gate: not run — the environment could not provide it (resource did not start)
- proved live on iOS through the fleet-ios bridge (device-verify.sh --platform ios on the Mac): # kwf-fix: proof-7b56464-2 # kwf-fix-pr: https://github.com/Kilo-Org/cloud/pull/6134 Prove the behaviour of PR https://github.com/Kilo-Org/cloud/pull/6134 with a live end-to-end run, and make no code change. The pull request description carries no evidence: the '## E2E proof' section says the proof was not captured. Run the pull request's own scenarios on a device or a simulator, for every platfor

PR #6138, opened 2026-09-14
fix(cloud-agent-next): preserve runtime-proxy credentials across warm reuse

eshurakov:
## Problem

Cloud Agent follow-up messages intermittently failed with HTTP 401 (`Assistant request was not authorized`) from the Worker facade on `POST /api/openrouter/chat/completions`.

The legacy runtime-proxy grant was tied to the logical delivery fence. A new delivery could replace the handle while a warm `kilo serve` child still held the previous handle in its inherited environment. Restarting the child for every turn would refresh credentials, but would also close user terminals because `kilo serve` owns their PTYs.

## Fix

- Bind legacy runtime-proxy grants to the physical wrapper instance ID and generation, rather than the per-delivery run and connection IDs. Keep the handle stable while that physical wrapper is retained.
- Preserve strict delivery-fence checks during issuance. Resolution requires the current owned physical lease; physical replacement invalidates the previous grant. Authorization and expiry checks remain enforced, including after backing-token renewal.
- Preserve existing handle claims and expiry when upgrading legacy persisted grants. Control-plane fencing remains unchanged.
- Extract wrapper runtime transitions into `kilo-runtime-lifecycle.ts`. Normal `session/ready` warm reuse keeps the existing child without comparing delivered environment values or maintaining a spawn-time environment baseline. Explicit crash recovery still forces a restart.
- Preserve credential refresh on the devcontainer environment-update path: changed environment values restart the child with the refreshed credentials while retaining the Kilo session ID. Identical updates remain silent. This path uses raw tokens or capabilities rather than the stable runtime-proxy handle.

### Devcontainer trade-off

Restarting on a changed devcontainer environment closes child-owned terminals. This exception is intentional; normal warm runtime-proxy reuse remains restart-free.

## Tests

- Worker grant coverage includes warm retention, stale delivery rejection, physical replacement, legacy grant upgrade, expiry, and authorization changes.
- Wrapper regression tests deliver replacement `KILOCODE_TOKEN` and `GH_TOKEN` values through `updateEnvironment`, check the environment inherited by the replacement child, and verify that the original Kilo session is retained. Both regression cases failed before the devcontainer fix and passed afterward.
- Corrected a warm-start fixture that previously supplied an ignored `deliveredEnv` property. The test now changes the injected parent environment and verifies that normal warm start still reuses the child.

## Verification

Latest checks for the devcontainer review fix:

- Focused lifecycle, start-decision, and server suites: **45 passed, 0 failed**.
- Full wrapper suite: **1,457 passed, 0 failed**. The first run exceeded a 120-second command limit; the retry completed in 148.86 seconds.
- Wrapper `tsgo --noEmit`: passed.
- Targeted `oxlint`, formatting, and `git diff --check`: passed.
- Three-agent static review found one consolidated devcontainer credential-refresh issue, addressed by commit `c113f51c8`; no additional material findings were reported.

Earlier local E2E evidence demonstrated two completed deliveries on a warm wrapper without a new `kilo serve` PID or an observed 401. That run did not exercise the modern runtime-proxy path and is not evidence of end-to-end handle stability.

## Verification Limits

- Stable modern runtime-proxy handles have state-level test coverage, but have not been verified through the real Worker + DO + sandbox + child + upstream flow. The local harness uses a non-policy-bearing token and bypasses that path.
- The devcontainer credential-refresh fix has not been reproduced end to end against a real provider. Its lifecycle regression tests inject child spawning.
- Runtime transitions are serialized, but concurrent different-body `session/ready` requests can still mutate the shared parent environment before a queued spawn. No environment lock was added.

## Deployment

The wrapper is baked into the sandbox image. Shipping this change requires an image rebuild and Worker deployment. No deployment was performed as part of this work.


pandemicsyn (approved):

pandemicsyn (approved):
lgtm

PR #6224, opened 2026-09-16
feat(kilo-mcp): publish allowlisted mutations through search and call

iscekic:
## Changelog for users
- `search` now lists mutation endpoints next to queries, and every hit names its kind and publishes the input schema the call tool validates against.
- `call` can invoke the published mutations; the catalog starts with the agent-profile authoring mutations: create, update, delete, set as default, clear default, set and delete a variable, set commands, and bind or unbind a repository.
- A mutation that fails with an unknown outcome reports that the change may have been applied and tells the agent to check the current state before retrying, never to retry blindly.
- A mutation that returns no value now reports success instead of an error.
- Search results too large for the tool-result cap keep the highest-ranked hits and report how many were dropped.

## Changelog for maintainers
- `apps/web/src/scripts/mcp-catalog/mutations.ts` is the review surface for what an MCP client may change: an exact-path `MCP_MUTATION_ALLOWLIST`. The top-level denylist (`admin`, `debug`, `test`) still applies on top, and subscriptions are never published.
- The dump's rot guard throws when an allowlisted path is not published with kind `mutation` — renamed, deleted, denylisted, or demoted — instead of dropping it silently.
- Look first at the transport split in `call.ts`: queries stay on GET with `?input=`; mutations are POST with a JSON body and never put input on the URL. A no-input mutation posts `{}`.
- Ambiguity rule to check: mutation network failures, gateway non-2xx without a tRPC error message, 5xx-class tRPC errors, and unreadable or non-tRPC 2xx bodies are ambiguous; a 4xx-class tRPC error keeps the ordinary mapping. Queries keep their plain upstream error.
- Catalog rows widen `kind` to `'query' | 'mutation'`; the embed script accepts both kinds and stores the kind in vector metadata.
- New analytics error class `upstream_unreachable_ambiguous` marks unknown-outcome writes, separate from retryable `upstream_unreachable`.
- Search hits now carry the full published input schema, so results can exceed the tool-result cap; over-cap payloads drop the lowest-ranked hits and stay valid JSON with a dropped count.
- Confirm the committed `catalog.json` mutation rows keep hand-written summaries and a `searchBlob` that names each path.

## E2E proof — log excerpts

<details><summary><code>/home/igor_kilocode_ai/.local/share/kwf/sections/owner-kilo-mcp-mutations-cadb/e2e-backend/backend.log</code></summary>

```
 ✓  unit  src/oauth/kilo-pairing.test.ts > fetchOrgOptions > dedupes ids and skips malformed rows 1ms
 ✓  unit  src/oauth/kilo-pairing.test.ts > fetchOrgOptions > throws on a non-OK or unexpected upstream (caller renders a retry) 4ms
 ✓  unit  src/search-knn.test.ts > createSemanticCandidates > embeds the search query with the pinned model and queries the index with that vector 17ms
 ✓  unit  src/search-knn.test.ts > createSemanticCandidates > maps match ids (procedure paths) to candidates with their kNN scores 1ms
 ✓  unit  src/search-knn.test.ts > createSemanticCandidates > requests at least 20 and at most 100 candidates regardless of the caller limit 3ms
 ✓  unit  src/search-knn.test.ts > createSemanticCandidates > raises on an embedding failure so the search degrades to token-only 4ms
 ✓  unit  src/search-knn.test.ts > createSemanticCandidates > raises on an empty embedding vector 1ms
 ✓  unit  src/search-knn.test.ts > createSemanticCandidates > raises on a Vectorize failure so the search degrades to token-only 1ms
 ✓  unit  src/search-knn.test.ts > createSemanticCandidates > logs a non-fatal note and returns no candidates when the index is empty or unpopulated 4ms
 ✓  unit  src/oauth/provider-hooks.test.ts > tokenExchangeCallback > emits oauthSignIn succeeded with the identity from the grant props 7ms
 ✓  unit  src/oauth/provider-hooks.test.ts > tokenExchangeCallback > reports a personal grant with a null organization 1ms
 ✓  unit  src/oauth/provider-hooks.test.ts > tokenExchangeCallback > reports anonymously when the props carry no usable Kilo user id 1ms
 ✓  unit  src/oauth/provider-hooks.test.ts > tokenExchangeCallback > does not emit for a refresh-token exchange 1ms
```
</details>

<details><summary><code>/home/igor_kilocode_ai/.local/share/kwf/sections/owner-kilo-mcp-mutations-cadb/e2e-web/web-e2e.log</code></summary>

```
  1) [chromium] › tests/setup-smoke/profile.spec.ts:10:7 › local setup smoke › signs in with fake auth and renders the profile page
    Test timeout of 30000ms exceeded.
    Error: page.waitForURL: Test timeout of 30000ms exceeded.
    =========================== logs ===========================
    waiting for navigation until "load"
      navigated to "http://localhost:3400/users/sign_in?fakeUser=setup-smoke-15b1d447%2Bstytchpass%40example.com&callbackPath=%2Fprofile"
    ============================================================
      55 |
      56 |     await page.goto(signInUrl);
    > 57 |     await page.waitForURL(url => url.pathname === '/profile', { timeout: 30_000 });
         |                ^
      58 |
      59 |     const profileCard = page.getByRole('region', { name: 'User profile' });
      60 |     await expect(profileCard).toBeVisible({ timeout: 30_000 });
        at /home/igor_kilocode_ai/.local/share/kwf/wt/owner-kilo-mcp-mutations-cadb/apps/web/tests/setup-smoke/profile.spec.ts:57:16
    attachment #1: screenshot (image/png) ──────────────────────────────────────────────────────────
    test-results/setup-smoke/profile-local-setup-smoke--f6e7e-nd-renders-the-profile-page-chromium/test-failed-1.png
    ────────────────────────────────────────────────────────────────────────────────────────────────
    Error Context: test-results/setup-smoke/profile-local-setup-smoke--f6e7e-nd-renders-the-profile-page-chromium/error-context.md
  1 failed
    [chromium] › tests/setup-smoke/profile.spec.ts:10:7 › local setup smoke › signs in with fake auth and renders the profile page
undefined
/home/igor_kilocode_ai/.local/share/kwf/wt/owner-kilo-mcp-mutations-cadb/apps/web:
```
</details>

<details>

> Surface: the Kilo MCP server (services/kilo-mcp).
> 
> Add support for mutations.

</details>

## Owner manual verification
Owner verification is pending; these checks did not pass automatically.
- Real MCP client on the deployed dev worker with a signed-in Kilo account: connect Claude Desktop or Cursor to the dev worker URL (the `dev` env of services/kilo-mcp/wrangler.jsonc, `kilo-mcp-dev`), complete the OAuth sign-in, search "agent profile", call `agentProfiles.create`, confirm the new profile appears in the Kilo app's agent-profile list, then call `agentProfiles.delete` and confirm it disappears. Scope excluded from automation: real Kilo account sign-in and browser pairing cannot be scripted or asserted from the worktree, so the verifier must not run or report this scenario; only the owner runs it. Owner-manual: never scripted, never reported as passed.

## Open findings (not fixed here)
- the '## E2E proof' section shows no screenshot, and this change has a visible surface


services/kilo-mcp/src/call.ts
```diff
@@ -208,16 +276,35 @@ export async function callCatalogEndpoint(options: {
...
 
   if (!response.ok) {
+    // A mutation that fails at the gateway/function layer — app.kilo.ai's
+    // FUNCTION_INVOCATION_TIMEOUT 504, a 502 edge error — comes back non-2xx
+    // with no tRPC error message. The POST reached the platform, so the write
+    // may have landed: report it as ambiguous, like a rejected fetch, instead
+    // of an ordinary upstream error that invites a blind retry. An app-level
+    // tRPC error response answered the request, so its outcome is known and it
+    // keeps the ordinary mapping (queries are never ambiguous).
+    if (row.kind === 'mutation' && !hasTrpcErrorMessage(body as TrpcErrorBody | null)) {
```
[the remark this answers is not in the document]
iscekic: bot: Accepted. Pushed ed983a8 (kwf `kwf-fix-review-caa6`).

- services/kilo-mcp/src/call.ts:286 — accepted. A tRPC 5xx error can follow a committed write (output validation, post-resolver middleware), so a mutation now throws the ambiguous failure with `ambiguous: true` and `httpStatus`, and carries no `retryable` and no `trpcCode`. A 4xx-class error and every query keep the ordinary mapping.

services/kilo-mcp/src/search.ts
```diff
@@ -104,6 +104,7 @@ export async function searchCatalog(
       summary: row.summary,
       tags: row.tags,
       score,
+      inputSchema: row.inputSchema,
```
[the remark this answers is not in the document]
iscekic: bot: Accepted. Pushed ed983a8 (kwf `kwf-fix-review-caa6`).

- services/kilo-mcp/src/search.ts:107 — accepted. Each hit keeps its full `inputSchema`; the search tool result now goes through the call tool's 16 KiB cap, so an over-cap payload is cut at a code-point boundary, marked `truncated: true`, and ends with `[truncated]`. The empty-results payload is capped too.

apps/web/src/scripts/mcp-catalog/catalog.ts
```diff
@@ -182,9 +198,19 @@ export function buildCatalogRows(
       missing.push(leaf);
     }
   }
+  if (mutationPaths.size > 0) {
```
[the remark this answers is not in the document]
iscekic: bot: Accepted. Pushed ed983a8 (kwf `kwf-fix-review-caa6`).

- apps/web/src/scripts/mcp-catalog/catalog.ts:201 — accepted. The rot guard compared the allowlist against mutation paths collected before the denylist, so a denylisted or demoted allowlisted path could disappear. It now compares against paths published with kind `mutation` after the denylist; `catalog-rot-guard.test.ts` covers denylisted, renamed, and demoted entries.

services/kilo-mcp/src/index.ts
```diff
@@ -242,14 +249,17 @@ async function runTool(
...
           results: [],
           message: `No endpoints matched "${query.trim()}". Refine your query: use fewer or different keywords, or describe the task in plain language.`,
         })
       );
     }
-    return textResult(JSON.stringify({ results }));
+    // Every hit carries its published input schema, and a 50-row result can
+    // exceed the tool-result cap: cut and mark the payload like the call tool
+    // instead of handing the client an unbounded body.
+    return toolResult(serializeWithCap({ results }));
```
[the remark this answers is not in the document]
iscekic: bot: Accepted. Pushed ce2075d (kwf `kwf-fix-review-c254`).

- services/kilo-mcp/src/index.ts:262 — accepted: the search branch passed the full hit list to `serializeWithCap`, which cuts JSON at a byte boundary. The payload was unparseable and the dropped hits were never counted. Search now drops the lowest-ranked hits until the payload fits and marks it truncated with a drop count.

services/kilo-mcp/src/call.ts
```diff
@@ -85,17 +85,32 @@ type TrpcErrorBody = {
...
 
 function isTrpcSuccessBody(body: unknown): body is TrpcSuccessBody {
   if (typeof body !== 'object' || body === null || Array.isArray(body)) return false;
   const result = (body as { result?: unknown }).result;
-  return (
-    typeof result === 'object' &&
-    result !== null &&
-    Object.prototype.hasOwnProperty.call(result, 'data')
-  );
+  return typeof result === 'object' && result !== null && !Array.isArray(result);
```
pandemicsyn: bot: Malformed 2xx result objects are reported as successful void mutations.

Suggested fix: Require a recognizable tRPC success envelope before accepting `result` as a void result. For example, distinguish the documented `{ result: {} }` void shape from arbitrary result fields (or validate the complete expected envelope), and route unrecognized 2xx bodies through the existing ambiguous-mutation / query internal-error path. Add tests for `{ result: { nonsense: true } }` for both mutation and query calls.

services/kilo-mcp/src/index.ts
```diff
@@ -242,14 +282,17 @@ async function runTool(
     });
     if (results.length === 0) {
       // Empty state, not an error: tell the agent how to recover.
-      return textResult(
-        JSON.stringify({
+      return toolResult(
+        serializeWithCap({
```
pandemicsyn: bot: An oversized no-results search response is truncated into invalid JSON.

Suggested fix: Do not pass the structured empty-search object to `serializeWithCap`, which cuts serialized JSON. Instead bound/truncate the echoed query before `JSON.stringify`, or return a compact fixed recovery message that always fits. Update the oversized-empty-result test to `JSON.parse` the returned text.

apps/web/src/scripts/mcp-catalog/catalog.ts
```diff
@@ -161,30 +166,60 @@ function shapeRow(leaf: CatalogLeaf, summary: string): CatalogRow {
...
+    // guard, or an allowlisted path could stop being a mutation unnoticed.
+    if (leaf.type === 'mutation') publishedMutationPaths.add(leaf.path);
     const summary = summaries.get(leaf.path);
     if (typeof summary === 'string' && summary !== '') {
       rows.push(shapeRow(leaf, summary));
     } else {
       missing.push(leaf);
     }
   }
+  if (mutationPaths.size > 0) {
```
pandemicsyn: bot: The mutation rot guard is skipped when enumeration contains no mutation leaves.

Suggested fix: Always compare `MCP_MUTATION_ALLOWLIST` with `publishedMutationPaths`; remove the `mutationPaths.size > 0` condition (and the now-unneeded pre-filter tracking if unused). Add a test where a query-only leaf set with a non-empty allowlist throws, so deletion or wholesale router drift cannot silently publish a query-only catalog.

pandemicsyn (approved):

services/kilo-mcp/src/call.ts
iscekic: bot: Accepted. Pushed cd8c552 (kwf `kwf-fix-review-668f`).

- `services/kilo-mcp/src/call.ts:100` — accepted: `isTrpcSuccessBody` now accepts only `{ result: {} }` (void) or `{ result: { data } }`, so `{"result":{"nonsense":true}}` is no longer read as a void success. A mutation routes it to the ambiguous-mutation failure (`ambiguous: true`, "may or may not have been applied"); a query throws the existing "without a tRPC result body" internal error. Tests cover both call kinds.

services/kilo-mcp/src/index.ts
iscekic: bot: Accepted. Pushed cd8c552 (kwf `kwf-fix-review-668f`).

- `services/kilo-mcp/src/index.ts:286` — accepted: the empty-results payload no longer passes through `serializeWithCap`. The echoed query is trimmed and cut to 200 code points with an ellipsis before `JSON.stringify`, so an oversized query yields parseable JSON with `truncated: false`. The oversized-empty-result test now `JSON.parse`s the returned text.

apps/web/src/scripts/mcp-catalog/catalog.ts
iscekic: bot: Accepted. Pushed cd8c552 (kwf `kwf-fix-review-668f`).

- `apps/web/src/scripts/mcp-catalog/catalog.ts:210` — accepted: the rot guard compares `MCP_MUTATION_ALLOWLIST` with `publishedMutationPaths` unconditionally. The `mutationPaths.size > 0` gate and the now-unused `mutationPaths` set are removed, so a query-only enumeration throws instead of publishing a mutation-free catalog. Catalog fixtures that built query-only rows now append the allowlisted mutation leaves.

PR #6232, opened 2026-09-16
test(mobile): pin the iOS Live Activity card count and Approve path

iscekic:
## Changelog for users

No user-visible change.

## Changelog for maintainers

- Adds a unit test that carries one waiting permission from the session row through the glanceable snapshot, the Live Activity content state, the widget interaction event, and the real front-approval service, then asserts the republished count falls to zero.
- The test delivers the native event as the widget process sends it: the source is the per-activity UUID, so the approve handler matches the target string alone.
- Adds a layout-source test that pins the watch card's ranked count row ahead of the Approve gate, so a wait with nothing approvable still reads its count on the wrist.
- The widget layout never runs under vitest, so both card tests read layout source text; an edit to the watch section is a contract change, not an internal refactor.

## E2E proof — log excerpts

```
iOS verification passed. --- shard 1 ---
```

<details>

> 
> Prove PR 6199 live on iOS, on a Mac simulator. The PR's whole proof section is one self-declared line ("-> pass :: owner-manual: the Apple Watch Live Activity card count, the wrist Approve tap ...") and nothing ran on the Mac.
> 
> Prove what the simulator can show, and capture it:
> - the iPhone side: the Live Activity card for a waiting agent, its count, and the approve path the app exposes;
> - the paired Apple Watch simulator: the watch app launched with the Live Activity card visible, captured with `xcrun simctl io <watch-udid> screenshot`.
> 
> A wrist tap that the simulator cannot deliver stays an owner-manual scope and must be reported as `owner_manual`, never as a pass.

</details>

## Owner manual verification
Owner verification is pending; these checks did not pass automatically.
- Apple Watch wrist tap: tapping Approve on the watch wrist itself (the simulator cannot deliver the wrist tap).


pandemicsyn (approved):

iscekic:
(bot) kwf reopened its section for this PR: verification restarted (phase verify) The proof in the body came from an earlier round and can be stale. This PR is human-ready, so nothing here was changed.

PR #6274, opened 2026-09-17
feat(cloud-agent-next): coalesce adjacent part deltas in outbox

eshurakov:
## Summary

The wrapper control-event outbox could only compact by superseding a snapshot entry keyed on an entity id. `message.part.delta` carries an increment, not a snapshot, so it had no key and could never be compacted: a delta burst grew the lane one entry per delta until admission failed and the event was dropped as `queue_overflow` at `MAX_CONTROL_EVENT_OUTBOX_EVENTS`.

- Add an append/merge rule beside the supersede rule. Adjacent same-part deltas for field `text` merge by concatenating `delta`, keeping the tail entry's `receiptId`, `sequence` and `preparedAt`, so the existing receipt identity still describes the published frame.
- Both rules share the in-flight tail guard, so an entry that a pending attempt will remove is never replaced.
- A merged entry takes the newer member's `deadlineAt`, so an actively-growing group is not expired mid-burst; snapshot squash keeps its previous deadline, so supersede semantics are unchanged.
- Merging is declined when the merged payload would exceed `MAX_SANDBOX_CONTROL_FRAME_BYTES`; the delta is appended as its own entry instead. At the entry cap that append can still be dropped as `queue_overflow`, which is correct because an over-frame replacement was never sendable.
- Also attribute non-applied control events: the id-mismatch and child-lineage rejections now return through `result()` and emit `session_event_result`, and a caught batch halt emits one warn-level `session_event_batch_halted` with the halting item's identity and the skipped suffix shape. The halt diagnostic never reads the thrown value.
- Merging is restricted to field `text`; no wire protocol, DO, SDK or schema change is involved.

## Verification

No manual testing performed. Automated checks run on the branch:

- `bun test src/control/control-event-outbox.test.ts` (wrapper): 76 pass, 0 fail
- `pnpm exec vitest run src/sandbox-session/session-message-queue.test.ts src/sandbox-control/diagnostics.test.ts`: 362 pass, 0 fail

## Visual Changes

N/A

## Reviewer Notes

The deadline refresh for merged groups is deliberate: an actively-merged group is bounded by the outbox entry and byte limits instead of the 30s clock.


services/cloud-agent-next/src/sandbox-session/SandboxSession.ts
```diff
@@ -949,12 +950,7 @@ export class SandboxSession extends DurableObject<Env> {
     wrapperInstanceId?: string;
   }): Promise<SandboxEventBatchResult> {
     const outcomes: SandboxEventBatchItemOutcome[] = [];
-    let halted = false;
-    for (const item of input.items) {
-      if (halted) {
-        outcomes.push({ receiptId: item.receiptId, status: 'unattempted' });
-        continue;
-      }
+    for (const [index, item] of input.items.entries()) {
```
pandemicsyn, line 953: bot: Continuing after an `unknown` item can apply same-session events out of order.

Suggested fix: Restore the halt guard after the catch: leave the throwing item as `unknown`, mark every remaining item `unattempted`, and emit one bounded `session_event_batch_halted` diagnostic for the failed item plus suffix shape. This keeps later deltas/snapshots/completions from being applied after a predecessor whose mutation outcome is indeterminate. The changed test at `session-message-queue.test.ts:2226-2275` should again assert that the suffix is unattempted rather than applied.

pandemicsyn (approved):

PR #6286, opened 2026-09-17
Revert per-repo changes

intentionally-left-nil:
Instead of having a single UI for all per-repo changes, we are going to leave the existing code alone for now. There already is the ability to turn code reviewers off per-repo

and a bigger feature ended up having too much scope creep


packages/db/src/migrations/0250_drop_repository_customizations.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE "repository_customizations" CASCADE;
```
intentionally-left-nil: I double checked that this sql table is empty. Let me remove the CASCADE to prevent any potentially scary things (even though that shouldn't be possible)

packages/db/src/migrations/0250_drop_repository_customizations.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE "repository_customizations";
```
pandemicsyn, line 1: bot: Dropping this table before application promotion breaks the currently serving revision during deploy.

Suggested fix: Do not drop the table in this release. First deploy code that no longer reads or writes it while retaining the schema, then remove it in a subsequent migration once old revisions are drained. The production workflow runs `drizzle migrate` before `promote-app`, while the merge-base webhook and accessor code still query this table; those existing instances can receive webhook traffic after this statement succeeds and before promotion.

apps/web/src/lib/integrations/platforms/github/webhook-handlers/pull-request-handler.ts
```diff
@@ -299,35 +297,10 @@ export async function handlePullRequestCodeReview(
...
-        `PR reviews disabled for repository ${repository.full_name} (ID: ${repository.id})`
-      );
-      return NextResponse.json(
-        { message: 'PR reviews disabled for this repository' },
-        { status: 200 }
-      );
-    }
-
-    // 5c. Feature-level guardrail: by default, skip automated reviews of bot-authored PRs
+    // 5b. Feature-level guardrail: by default, skip automated reviews of bot-authored PRs
```
pandemicsyn, line 300: bot: Removing the review-mode guard causes repositories explicitly configured with automated reviews off to receive reviews.

Suggested fix: Preserve the repository/installation review-enable decision in a durable supported configuration source and check it here before duplicate-review creation, or retain the existing settings implementation until an equivalent replacement ships. The removed block fetched `repository_customizations`, resolved the effective `pr_review_mode`, and returned early for `off`; the new flow proceeds directly to bot filtering and review creation. `selected_repository_ids` is not equivalent because it only applies when selection mode is `selected`. Add regression coverage for explicit repository-off and installation-off settings.

pandemicsyn (approved):
pre-approving but might be worth breaking up this change like the bot says.  Since we do canary rollouts the migration will land first and we'll have a window of time where both versions are live.

apps/web/src/lib/integrations/platforms/github/webhook-handlers/pull-request-handler.ts
intentionally-left-nil: Nobody used this feature, we can safely remove the code

packages/db/src/migrations/0250_drop_repository_customizations.sql
intentionally-left-nil: Agreed, fixed by removing the database changes. Will followup in a new PR for this

PR #6289, opened 2026-09-17
Add minimax-token-health cron job

intentionally-left-nil:
Daily job to query the usage URL for each provisioned api key for minimax, to determine whether the API key is healthy or not. Reports the status as a slack message


apps/web/src/lib/coding-plans/minimax-token-health.ts
```diff
@@ -0,0 +1,380 @@
...
+  } catch {
+    return { category: 'unreachable', reason: 'network_error' };
+  }
+
+  if (!response.ok) {
+    await response.body?.cancel().catch(() => undefined);
+    // 429/5xx reflect transient rate limiting or upstream outages (see the
+    // 2026-09-15 Token Plan incident note above), not an actual access
+    // denial, so they must not be bucketed with 401/403 as `denied`.
+    const category = response.status === 429 || response.status >= 500 ? 'unreachable' : 'denied';
```
pandemicsyn: bot: Non-auth HTTP failures are reported as access denials.

Suggested fix: Classify only statuses that actually indicate failed authentication/authorization (normally 401 and 403) as `denied`. Route transient statuses such as 408/429/5xx to `unreachable`, and classify unexpected client/protocol statuses (for example 400, 404, or 422) as `bad_response` or a separate controlled category. Add mocked-fetch cases so the alert labels cannot regress. The current expression maps every non-OK status other than 429 and 5xx to `denied`, so a changed endpoint, malformed request, or timeout-like 408 is surfaced in Slack as a bad API key; all non-healthy live subscriptions are escalated for immediate follow-up.

pandemicsyn (approved):
lgmt , looks like robot just flagged a minor one

PR #6313, opened 2026-09-18
fix(admin): bound credit category aggregation

jrf0110:
## Summary
- scope the admin credit-categories list aggregation to configured promo categories and the paid-credit bucket
- avoid grouping and serializing dynamically generated per-payment and per-instance categories
- preserve exact-key detail queries and the dedicated credit-campaign admin flow
- add regression coverage proving a dynamic KiloClaw settlement category is excluded while a configured promo remains visible

## Root cause
`GET /admin/api/credit-categories` grouped the entire `credit_transactions` table by `credit_category`. Dynamic categories can be unique per transaction, so the endpoint returned an unbounded response and the client mounted every row and tooltip, locking the browser tab.

## Validation
- `pnpm --filter web test -- --runInBand src/app/admin/api/credit-categories/route.test.ts`
- `pnpm exec oxlint --config .oxlintrc.json apps/web/src/app/admin/api/credit-categories/route.ts apps/web/src/app/admin/api/credit-categories/route.test.ts`
- `git diff --check`
- authenticated Chromium stress test with 5,000 unique dynamic categories: API returned 92 configured categories, rendered 46 rows under default filters, rendered 0 dynamic rows, and completed in approximately 1.05 seconds


apps/web/src/app/admin/api/credit-categories/route.ts
```diff
@@ -26,7 +26,13 @@ export async function GET(
   oneWeekAgo.setDate(oneWeekAgo.getDate() - 7);
 
   const whereCondition = !key
-    ? undefined
+    ? or(
```
pandemicsyn, line 29: bot: The new allowlist hides fixed operational ledger categories from the category catalog.

Suggested fix: If the catalog is expected to remain an admin audit surface, retain a bounded operational-category allowlist alongside promo keys and `<null:paid>` (at least `accounting_adjustment` and `stripe-dispute-enforcement`), and add regression coverage. Otherwise, explicitly confirm that reconciliation and dispute-enforcement records are intentionally available only through per-user/per-organization transaction history, because the category-detail route rejects unconfigured keys.

pandemicsyn (approved):
lgtm

PR #6315, opened 2026-09-18
feat(kilo-mcp): publish all non-admin/debug mutations

iscekic:
## Problem

PR #6224 added mutation support to the Kilo MCP, but the catalog only published the 10 paths in `MCP_MUTATION_ALLOWLIST` (agentProfiles). Every other mutation was invisible to `search` and rejected by `call`.

## Root cause

`buildCatalogRows` gated mutations on an exact-path allowlist:

```ts
const kept = leaf.type === 'query' || (leaf.type === 'mutation' && isAllowedMutation(leaf.path));
```

The `rootRouter` exposes 671 mutations. 517 are non-denylisted (top-level segment not `admin`, `debug`, or `test`); only 10 were published. This is a catalog-scope bug, not a deployment failure: the catalog-merge job for #6224 succeeded and its `Upsert Vectorize index` step passed (run [35289442683](https://github.com/Kilo-Org/cloud/actions/runs/35289442683)).

## Fix

- Remove `mutations.ts` (`MCP_MUTATION_ALLOWLIST`, `isAllowedMutation`) and its exact-path gate. `buildCatalogRows` now keeps every query and mutation that survives the existing top-level denylist, and still drops subscriptions.
- Replace the allowlist rot guard with a mutation drift guard: an enumeration that publishes no mutation row throws instead of emitting a query-only catalog.
- Regenerate `services/kilo-mcp/catalog.json`: **897 rows (380 queries, 517 mutations)**, every row with a summary. `catalog.test.ts` asserts the committed catalog regenerates byte-for-byte and that every non-denylisted router mutation is present.

## Verification

- `apps/web` jest `src/scripts/mcp-catalog`: 31 passed.
- `services/kilo-mcp` vitest: 282 passed.
- `node --test scripts/kilo-mcp-catalog.test.mjs`: 29 passed.
- `oxlint` and `oxfmt --list-different` clean on the changed files.
- `tsgo --noEmit` reports no error in `mcp-catalog` (remaining output is pre-existing, in unrelated packages).

## Note

The denylist is top-level-segment based, as before. Nested paths whose segment name merely contains `admin` (for example `codingPlans.adminMarkRevocationFailed`) stay published, consistent with how queries are already filtered.


apps/web/src/scripts/mcp-catalog/catalog.ts
```diff
@@ -166,38 +164,28 @@ function shapeRow(leaf: CatalogLeaf, summary: string): CatalogRow {
...
-  /** Every allowlisted path the catalog will publish with kind `mutation`. */
-  const publishedMutationPaths = new Set<string>();
+  let publishedMutations = 0;
   const rows: CatalogRow[] = [];
   const missing: CatalogLeaf[] = [];
   for (const leaf of leaves) {
-    const kept =
-      leaf.type === 'query' || (leaf.type === 'mutation' && isAllowedMutation(leaf.path));
-    if (!kept) continue;
+    if (leaf.type !== 'query' && leaf.type !== 'mutation') continue;
```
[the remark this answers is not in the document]
iscekic: Valid. Fixed in 994fed8e0: `isDenylistedPath` now matches `admin`/`debug`/`dev`/`test` as segments anywhere in the path, plus the `adminX`/`devX` prefixes. This drops 21 nested admin mutations, 5 dev-only mutations, and 18 nested admin queries. `test` stays exact-only, so user-facing calls such as `slack.testConnection` remain published. `catalog.json` regenerated: 853 rows (491 mutations, 362 queries).

apps/web/src/scripts/mcp-catalog/catalog.ts
```diff
@@ -166,38 +164,28 @@ function shapeRow(leaf: CatalogLeaf, summary: string): CatalogRow {
...
     if (denylisted.has(leaf.path.split('.')[0] ?? '')) continue;
-    // Only a mutation leaf counts: a demoted query leaf must not satisfy the
-    // guard, or an allowlisted path could stop being a mutation unnoticed.
-    if (leaf.type === 'mutation') publishedMutationPaths.add(leaf.path);
+    if (leaf.type === 'mutation') publishedMutations += 1;
```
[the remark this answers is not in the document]
iscekic: Valid as a wording mismatch. The guard intentionally counts exposed mutation leaves, not emitted rows: a mutation without a committed summary is a legitimate `missing` entry that the dump generates before writing, so requiring an emitted mutation row would break first generation. Renamed to `exposedMutations`, the error now says "exposed zero mutation procedures", and a comment explains the distinction. Added a test for the query-only enumeration.

apps/web/src/scripts/mcp-catalog/catalog.ts
```diff
@@ -166,38 +178,30 @@ function shapeRow(leaf: CatalogLeaf, summary: string): CatalogRow {
...
   const missing: CatalogLeaf[] = [];
   for (const leaf of leaves) {
-    const kept =
-      leaf.type === 'query' || (leaf.type === 'mutation' && isAllowedMutation(leaf.path));
-    if (!kept) continue;
-    if (denylisted.has(leaf.path.split('.')[0] ?? '')) continue;
-    // Only a mutation leaf counts: a demoted query leaf must not satisfy the
-    // guard, or an allowlisted path could stop being a mutation unnoticed.
-    if (leaf.type === 'mutation') publishedMutationPaths.add(leaf.path);
+    if (leaf.type !== 'query' && leaf.type !== 'mutation') continue;
```
pandemicsyn, line 289: bot: Publishing every non-name-denylisted mutation makes high-impact writes callable by MCP agents without a capability policy.

Suggested fix: Do not use the absence of an `admin`/`debug`/`dev`/`test` name as the sole eligibility rule for MCP writes. Keep a reviewed opt-in capability list (or add an equivalent explicit classification) for sensitive mutation classes, and exclude or require interactive confirmation for credential/token issuance and rotation, secrets, organization membership/authorization, deployments, billing/purchase, and destructive actions. Add regression tests that these representative paths cannot enter the runtime catalog without that explicit decision.

pandemicsyn, line 289: bot: This newly broadens mutation publication, but the catalog generator represents a chained tRPC procedure using only `_def.inputs[0]`. For example, the newly exposed `workspaceFolders.create` entry advertises only `organizationId`; its router adds a second `.input()` that requires `name` and `color`. An MCP request can therefore pass the published schema and then fail upstream tRPC validation.

Compose every chained input schema before JSON-Schema conversion, regenerate `catalog.json`, and add regression coverage that a chained-input mutation includes all required fields. Check the newly published folder update/delete/move/reorder operations as well, since they follow the same pattern.

pandemicsyn (approved):
I'd probably have a preferred an opt-in/default-deny approach (like the orig) , let make sure we at least have a brief mention in REVIEWS.md or something to try and catch us leaking admin/sensitive routes.

apps/web/src/scripts/mcp-catalog/catalog.ts
iscekic: Rejected, with the documentation you asked for added.

The default-deny opt-in capability list is out of scope for this PR: it reverses the catalog policy already agreed in #6224 and this PR (publish every query and mutation that is not internal), and server-side tRPC authorization still gates every call. The concrete leak this concern names (nested `admin`/`dev` procedures) was fixed in 994fed8e0.

I added the brief review guidance from your summary: `REVIEW.md` now flags a newly published MCP mutation that issues or rotates credentials or tokens, touches secrets, changes organization membership or authorization, deploys, charges or purchases, or deletes data. Commit a544a6f47. I did not add an opt-in list or exclude those mutation classes.

iscekic: Valid. Fixed in 7da532c87.

`collectCatalogLeaves` keeps every `.input()` schema, and `toInputSchema` composes them: object schemas merge their properties and required keys, a key declared by two schemas keeps both constraints through `allOf`, and a non-object chain falls back to a top-level `allOf`. `workspaceFolders.create` now publishes `organizationId`, `name`, and `color`, all required; `update`/`delete`/`moveWorktree`/`reorder` are fixed the same way. Added unit tests, a regression test on the committed catalog, and a check that tags cover every chained field. `catalog.json` regenerated: 853 rows, 207 with expanded input schemas.

apps/web/src/scripts/mcp-catalog/catalog.ts
```diff
@@ -119,22 +134,89 @@ export function collectCatalogLeaves(router: { _def?: unknown }): CatalogLeaf[]
...
+          properties[key] = { allOf: [properties[key], value] };
+        }
+      }
+      if (Array.isArray(schema.required)) {
+        for (const key of schema.required) {
+          if (typeof key === 'string' && !required.includes(key)) required.push(key);
+        }
+      }
+    }
+    const composed: Record<string, unknown> = {};
```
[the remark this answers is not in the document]
iscekic: Valid. Fixed in ee6b15bfc.

The object-merge branch now runs only when every chained schema carries no keyword beyond `$schema`, `type`, `properties`, and `required`. A chain with `additionalProperties` (`.strict()`) or any other keyword falls back to a top-level `allOf`, which preserves each schema's own keywords. Copying `additionalProperties` into the flat union would not be faithful: a strict subschema rejects a key that another input contributes, so the union would still over-advertise.

`organizations.cloudAgentNext.listRepositoryBranches` and `organizations.groups.getPolicyEditorData` now publish `additionalProperties: false`; a request with an unknown top-level key is rejected by the published schema. 32 rows use the intersection. Added a unit test for the strict fallback; `catalog.json` regenerated (853 rows).

PR #6354, opened 2026-09-19
fix(mobile): keep disabled primary button labels legible

iscekic:
## Changelog for users

- A disabled primary button keeps a legible label in both light and dark themes instead of fading into its fill.
- The disabled Start session button on the new-session form now reads as inactive but stays readable.
- A primary button that is busy keeps its brand fill and shows a spinner, so it still reads as working.
- Disabled buttons that are not primary keep their existing dimmed appearance.

## Changelog for maintainers

- Finding accepted: the disabled primary dimmed via `opacity-50`, compositing the brand fill and ink into olive-on-olive; the disabled primary no longer dims.
- Added a `--primary-disabled` token (light `#6F6A61`, dark `#8A8680`) in `apps/mobile/src/global.css`, generated into `theme-colors.generated.ts` by `generate-theme-colors.mjs`.
- Disabled, non-busy default-variant buttons paint `bg-primary-disabled` and skip `opacity-50`; non-default variants keep `opacity-50`.
- A loading primary keeps `bg-primary` and the inline spinner; `new-session-start-button.tsx` passes `loading` instead of rendering its own `ActivityIndicator`.
- Contrast tests assert `primaryForeground` on `primaryDisabled` clears 4.5:1 in both themes; mounted Button and bootstrap-error-screen tests cover the class precedence.
- Review hint: confirm `bg-primary-disabled` merges after the variant fill so it wins over `bg-primary`, and that the busy path never takes the muted fill.

## E2E proof — log excerpts

![[e1] composer-repo-chosen: primary 'Start session' label contrast in the dark theme — e2e-mobile-app/e1-composer-repo-chosen-dark.png](https://github.com/user-attachments/assets/73d0c1d1-65ea-42d1-acf0-e0797552bf36)

```
[e1] composer-repo-chosen: primary 'Start session' label contrast in the dark th -> pass :: Android emulator-5554, app Appearance=Dark (e1x-theme-dark-scene.xml: content-desc="Dark" checkable="true" checked="true"); with the repository chosen (e1-composer-repo-chosen-dark-scene.xml: content-desc="Repository: GitHub · iscekic/panon-deking") and the prompt empty (e1x-composer-top-dark-scene.xml: showing-hint="true") the primary control renders disabled (e1-composer-repo-chosen-dark-scene.xml: content-desc="Start session" checkable="false" checked="false" clickable="true" enabled="false"); the disabled-label contrast judgement is the visual reviewer's on capture e1-composer-repo-chosen-dark.png.
```

## Open findings (not fixed here)
- e1: Observed: The primary 'Start session' label is low-contrast olive-on-olive in the dark theme even though the form is complete, so the main action reads as disabled and is hard to read. Expected: the screen renders without this defect
- [e1] Observed: The primary 'Start session' label is low-contrast olive-on-olive in the dark theme even though the form is complete, so the main action reads as disabled and is hard to read. Expected: the screen renders without this defect
- [e1] composer-repo-chosen: primary 'Start session' label low-contrast in the dark theme: android emulator-5554: the pack carries no change for this finding (files/change.diff is empty, worktree $WT clean at the reported revision fd04ca1ef), so the expected fixed rendering is not delivered; on the de

e1-composer-repo-chosen-dark


eshurakov (approved):
Approved after a shallow triage pass. No blocking findings.

PR #6374, opened 2026-09-19
fix(test): stop the ChatGPT routing tests depending on today's catalog

iscekic:
## What

Stubs `findKiloExclusiveModel` in `routing.test.ts` so two cases stop using a real catalog model id as their fixture.

## Why

`main` is red. Two cases in `apps/web/src/lib/ai-gateway/openai-chatgpt/routing.test.ts` fail:

- `isOpenAiChatGptEligible › is not eligible for a Kilo-exclusive OpenAI model`
- `getOpenAiChatGptByokModelIds › never tags a gpt-oss or Kilo-exclusive model`

Both used a real Kilo-exclusive model id as the "exclusive" fixture. `findKiloExclusiveModel` returns only models that are not disabled (`apps/web/src/lib/ai-gateway/models.ts:161`), and once the last live exclusive OpenAI promotion was disabled the fixture stopped being exclusive. The cases then asserted the opposite of what they were written to check.

No live Kilo-exclusive OpenAI model remains, so swapping in another catalog id would not fix it and would break again on the next promotion change.

## How

The lookup is stubbed for a sentinel id. The cases now assert the routing rule rather than which promotions happen to be running.

## Test

`npx jest src/lib/ai-gateway/openai-chatgpt/routing.test.ts` — 35 passed, was 2 failed / 33 passed.

## Note

This blocks #6372, which is otherwise green.


iscekic:
Folded into #6372 as c54cb4ad0 — branch protection needs a human approval here, and #6372 is already approved, so carrying the fix there is the shorter path to a green main. Same change, unmodified.

PR #6383, opened 2026-09-19
fix(mobile): keep question card action labels readable

iscekic:
## Changelog for users

- The question card's actions now stack as full-width buttons instead of sharing one row.
- "Send answers", "Skip", "Retry", "Retry skip", and "Submitting…" each keep their whole label on one readable line.
- Action labels are centered and never shortened with an ellipsis or smaller text.
- The order stays Skip first, then any retry, then the primary Send answers button.

## Changelog for maintainers

- `apps/mobile/src/lib/auth/passkey-client.test.ts:250` — rejected (no_change): deleting the `'lib/auth/passkey-client.ts'` entry at `apps/mobile/src/lib/case-guard.test.ts:43` makes the case guard fail (`expected [ 'lib/auth/passkey-client.ts' ] to deeply equal []`), and `classifyPasskeyError` only folds native error strings that are never displayed (`apps/mobile/src/lib/auth/passkey-client.ts:133`). No question-card code changed.
- The question-card action footer is now a vertical (`flex-col`) stack instead of a horizontal row, so actions no longer share one row's width.
- Every action button is `w-full` instead of `flex-1`, and its label uses `shrink text-center` instead of a bare `text-sm`.
- The conditional `ml-2` label offset used while submitting is gone; the shared Button already applies `gap-2` between the spinner and the label.
- Labels keep full text and automatic font scaling: no `numberOfLines` and no `adjustsFontSizeToFit`.
- A new mounted test `question-card-actions.mounted.test.tsx` covers the label layout before and after selecting an answer, while submitting, after a retryable answer or skip failure, on a non-retryable terminal card, and the existing Skip confirmation.
- Review hint: the primary action now sits below Skip rather than beside it; check the stacked order and height on narrow and wide screens. The repeated `prior/` capture blocks were removed as duplicates, and the current android captures come from emulator-5604 and emulator-5554; the needs-input card was reached through the `__fake__:question` directive because the scenario trigger only sets the session status.
- No copy, catalog, or API change ships with this fix.

## E2E proof

![[e1] After a failed Skip request, restore connectivity and press Retry skip; confirm the readable action retries skipping rather than submitting answers. — e2e-mobile-app/e1-failed.png](https://github.com/user-attachments/assets/2c4e65b7-eacd-4b5e-80b5-779aa9480fbd)

![[e1] needs-input question card — Send answers readable before/after selecting — prior/e1-before-select.png](https://github.com/user-attachments/assets/22e7e20f-fba5-415c-afc4-75799026860b)

![[e1] needs-input question card — Send answers readable before/after selecting — prior/e1-after-select.png](https://github.com/user-attachments/assets/d03e54aa-033a-4323-8a26-2ae8b5dcee3f)

<details>

> Explorer finding: session-question: The primary "Send answers" button breaks its label mid-word across three lines ("Send / answ / ers"), so the main action is hard to read.
> 
> The user-agent explorer found this while using the app like a user.
> One finding per item; the explorer never edits product code.
> 
> Flow: session-question
> Found on revision: fd04ca1ef
> 
> Repro:
> 1. set this state first: account e2e-mobile-cloud-android@example.com signed in via emailed code (~$258.98 credits), Code Reviewer seeded with 3 pull requests (the `checks` state is blocked by a harness preflight: tracked edits present), a second organization to switch to, one needs-input session; system night mode, display size 480x1040, portrait.
> 2. open the app on emulator-5554
> 3. reach the screen the capture names (session-question)
> 4. the capture shows the defect named below
> 
> Observed: The primary "Send answers" button breaks its label mid-word across three lines ("Send / answ / ers"), so the main action is hard to read.
> Expected: the screen renders without this defect
> 
> Evidence (from the device run):

</details>

## E2E proof — log excerpts
```
[e1] After a failed Skip request, restore connectivity and press Retry skip; con -> pass :: android emulator-5554; with nextjs down the skip failed and the card showed text="Failed to skip question. Please try again." plus content-desc="Retry skip" ... clickable="true" enabled="true" (e1-failed.log), and after <internal>, leaving content-desc="Question dismissed tool, completed" (e1-retried.log) — the retry re-ran the skip, no answer submit.
```


iscekic:

**Status:** 1 Issues

| File | Line | Issue |
| --- | --- | --- |
| `apps/mobile/src/lib/auth/passkey-client.test.ts` | 250 | The patch includes unrelated passkey-error tests and a case-guard allowlist change in a question-card-only fix. |

iscekic:
bot: Rejected, no code change (kwf `kwf-fix-review-1569`).

Why: (already implemented, verified live: no change needed: Rejected removal of the passkey validation changes because the documented exception is necessary for the required case guard to pass. Reproduction: temporarily removed only apps/mobile/src/lib/case-guard.test.ts:43 and ran pnpm -C apps/mobile exec vitest related --run src/lib/case-

eshurakov (approved):
Approved after a shallow triage pass. No blocking findings.

